Issue #15, opened 2018-04-20
Logger


SteffenSeckler:
we need a proper way to do logging

PR #16, opened 2018-04-20
Feature/spdlog

SteffenSeckler:
spdlog, ref #15 
actually (almost) complete ;)

done:
* create the logger using `autopas::logger::create()` to print to `std::out` OR `autopas::logger::create(std::ostream& oss)` OR `create(std::string& filename)`
* logger will be created with the name `AutoPasLog`
* log via AutoPasLogger->debug("this is an output")
* logger uses fmt syntax: 
   * AutoPasLogger->info("Hello, {}!", "world");  // uses Python-like format string syntax
   * AutoPasLogger->warn("Hello, %s!", "world"); // uses printf format string syntax
* log levels
beware:
* the logger has to be created before you can log sth. otherwise errors occurr - we might want to add a base class for tests, that always does this. (same for exceptions, maybe?)

missing:
* the only thing not yet here is an interface of autopas to always enable the logger.

open question:
* either we add libraries as is, i.e. copy them into libs
* or we just add the includes (for header only libs, such as the logger)

log levels:
* setting via: `AutoPasLogger->set_level(level);`
* the levels (most verbose to least verbose):
  1.  #`AutoPasLogger->trace("trace");`
  2. `AutoPasLogger->debug("debug");`
  3. `AutoPasLogger->info("info");`
  4. `AutoPasLogger->warn("warn");`
  5. `AutoPasLogger->error("error");`
  6. `AutoPasLogger->critical("critical");`
  7. `set_level(spdlog::level::off);`


FG-TUM:
regarding the open question:
xSDK M12 applies here as far as I see it.

> M12. If a package imports software that is externally developed and maintained, 7 then it must allow installing, building, and linking against an outside copy of that software. Acceptable ways to accomplish this include (1) forsaking the internal copied version and using an externally provided implementation or (2) changing the file names and namespaces of all global symbols to allow the internal copy and the external copy to coexist in the same downstream libraries and programs.

SteffenSeckler:
I decided to actually just import this using the add_target_includedirectories routine and delete all other unimportant things.

PR #18, opened 2018-04-23
Added Jenkinsfile, fix 13

### LINKED ISSUE #13: Jenkins status
https://github.com/AutoPas/AutoPas/commits/master

commits do not retain their build status once a new build is started:

* all old builds show the state "pending"
* only the newest build shows the actual state
* we might want to use some github plugin for that instead of using curl + xpost ;)

NeoTheThird:
This should fix #13. The job was moved to a multi-branch project, so we now have the opportunity to store the pipeline directly in the source tree. There are various advantages to this approach, because it allows us to use some smart new features.

I also removed the e-mail notification that went out to the core development team on every build to reduce the verbosity. I would propose to replace that with either email that is only sent to the last committer. Another option would be to have Jenkins simply post a comment on the commit/Pull-Request, that way every user can specify whether they want to receive an e-mail in their GitHub settings.

The latter option could be a good idea either way, because we could also display useful additional information like test coverage in a comment.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,71 @@
+pipeline{
+    agent { label 'atsccs11' }
+    stages{
+        stage('setup'){
+            steps{
+                echo 'Starting AutoPas Pipeline'
+            }
+        }
+        stage("build") {
+            steps{
+                dir("build"){
+                    sh "cmake .."
+                    sh "make"
+                }
+            }
+        }
+        stage("test") {
+            steps{
+                dir("build"){
+                    //sh "env CTEST_OUTPUT_ON_FAILURE=1 make test"
+                    sh 'env GTEST_OUTPUT="xml:$(pwd)/test.xml" ctest'
+                }
+                dir("build-addresssanitizer"){
+                    sh "cmake -DCodeCoverage=OFF -DCMAKE_BUILD_TYPE=Debug -DENABLE_ADDRESS_SANITIZER=ON .."
+                    sh "make -j 4"
+                    sh "make test"
+                }
+                dir("build-threadsanitizer"){
+                    sh "cmake -DCodeCoverage=OFF -DCMAKE_BUILD_TYPE=Debug -DENABLE_THREAD_SANITIZER=ON .."
+                    sh "make -j 4"
+                    sh "make test"
+                }
+            }
+        }
+        stage("documentation and coverage"){
+            steps{
+                dir("build"){
+                    sh 'make doc_doxygen 2>DoxygenWarningLog.txt'
+                    sh 'touch /import/www/wwwsccs/html/AutoPas/doxygen_doc/master || echo 0'
+                    sh 'rm -rf /import/www/wwwsccs/html/AutoPas/doxygen_doc/master || echo 0'
+                    sh 'cp -r doc_doxygen/html /import/www/wwwsccs/html/AutoPas/doxygen_doc/master'
+                }
+                junit 'build/test.xml'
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,71 @@
...
+                    sh "make -j 4"
+                    sh "make test"
+                }
+            }
+        }
+        stage("documentation and coverage"){
+            steps{
+                dir("build"){
+                    sh 'make doc_doxygen 2>DoxygenWarningLog.txt'
+                    sh 'touch /import/www/wwwsccs/html/AutoPas/doxygen_doc/master || echo 0'
```
SteffenSeckler: do this and the following two lines only for the master

SteffenSeckler: do this and the following two lines only for the master

SteffenSeckler:
oh and regarding your comment.
Yes a message to the commit/PR would be good. (unless you get an email for PR that fail tests anyways. -- no idea how github handles that.)

SteffenSeckler:
* the links going out from the check visible in this PR are dead - you might want to check that ;)

NeoTheThird:
> oh and regarding your comment.
Yes a message to the commit/PR would be good. (unless you get an email for PR that fail tests anyways. -- no idea how github handles that.)

You don't always get an email. Some CI services (like for example Travis) send you one, though. It comes down to personal preference, i guess.

NeoTheThird:
> the links going out from the check visible in this PR are dead - you might want to check that ;)

Sorry about that, let me move to a different plugin real quick. Hang on.

NeoTheThird:
Ok, autostatus was still active, let me go shout at Jenkins and try that again...

SteffenSeckler:
minor remark:
maybe updating the githubautostatus plugin works. At least an update is available...

NeoTheThird:
Yeah, i saw that. But the changes don't address this issue, and it doesn't seem that useful to have Jenkins report the status on *every* stage. For now, i changed it to only report for build and test, i will add more later. But i want to be able to set customized messages, the generic messages autostatus spits out aren't that useful.

NeoTheThird:
Ok, there seems to be something wrong with the default url that's provided by the multibranch api, for *build* and *test* i added an override now, for the third one i first have to find out what component is generating the report... I'm ok with merging this for now if you are, will come back to it later.

SteffenSeckler:
ok then just merge it.
one remark: maybe it's better to actually use the autostatus plugin. it 
doesn't change anything regarding the url problem and the jenkins file 
becomes less crowded.


-- Steffen Seckler, M.Sc. with honors
Chair of Scientific Computing
Technische Universität München, Fakultät für Informatik
Boltzmannstr. 3, 85748 Garching, Germany
Phone: +49-89-289-18616
Fax: +49-89-289-18607
Office: MI 02.05.036

Am 24. April 2018 5:43:31 nachm. schrieb Jan Sprinz <notifications@github.com>:
Ok, there seems to be something wrong with the default url that's provided 
by the multibranch api, for build and test i added an override now, for the 
third one i first have to find out what component is generating the 
report... I'm ok with merging this for now if you are, will come back to it 
later.

NeoTheThird:
Ok, merged. Well, it does make a difference, because with the override I added, the links are correct. We also have some extra flexibility, for example i'd like the URL for *test* to point to [here](http://vmbungartz10.informatik.tu-muenchen.de/mardyn/blue/organizations/jenkins/AutoPas-Multibranch/detail/master/1/tests) in the future. I can simply assemble the url from the other parameters, i will send a separate PR for that later. Unless you want me to put the autostatus plugin back in, i'd prefer that approach.

NeoTheThird:
Btw, you'll have to rebase #20 and #21 for jenkins to recognize them.

SteffenSeckler:
ok done. 
I also changed the discovery option of branches in jenkins a bit, so pull requests are now correctly recognized.

Issue #19, opened 2018-04-24
ExceptionHandler


SteffenSeckler:
add a thread safe way to handle exceptions and define how they are supposed to be handled (throw, abort, mpi_abort, ...)

PR #20, opened 2018-04-24
thread safe exceptions and openmp compilation

SteffenSeckler:
ExceptionHandling is now thread safe and uses static functions and one static std::mutex.
usage:
```C++
ExceptionHandler::setBehavior(ExceptionBehavior::ignore);
// other possible values for the bahavior: 
enum ExceptionBehavior {
  ignore,                    /// ignore all exceptions, t
  throwException,            /// throw the exception
  printAbort,                /// print the exception and
  printCustomAbortFunction,  /// print the exception and call a custom abort
  /// function
};
ExceptionHandler::exception("testignore");
// or:
ExceptionHandler::exception(std::exception);

// defaults:
ExceptionBehavior::throwException;
// standard custom abort function is abort
```

* I also added a test to check that it actually is thread safe. This test uses openmp, so it seems that that combination is fine. - this test requires OPENMP to be set to ON
* Removal of the locks actually makes the test fail (when using the threadsanitizer errors are spawned)
* regarding the use of static functions and simple locking: I thought that this only happens once in a while and should not be too much overhead. In most cases this should end up as failure anyways.

Defaults:
* throw exception
* if printCustomAbortFunction is selected, but no custom abortfunction is specified, the standard abort() will be called.

ref #19

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,112 @@
+pipeline{
+    agent { label 'atsccs11' }
+    stages{
+        stage('setup'){
+            steps{
+                echo 'Starting AutoPas Pipeline'
+                githubNotify context: 'build', description: 'build pending...',  status: 'PENDING', targetUrl: currentBuild.absoluteUrl
+                githubNotify context: 'test', description: 'test pending...',  status: 'PENDING', targetUrl: currentBuild.absoluteUrl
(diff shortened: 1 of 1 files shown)


SteffenSeckler:
todo: add openmp target to jenkins

PR #21, opened 2018-04-24
adds AutoPasTestBase class, that ensures that a logger is setup.

SteffenSeckler:
* adds the class (implements constructor, that initializes logger)
* makes all test classes inherit from that class

pull request, as I want your feedback and want to let you know of the change, should you want to add new tests ;)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,7 @@ include(version.cmake)
 include(clang-tidy)
 include(clang-format)
 include(doxygen)
-
+include(openmp)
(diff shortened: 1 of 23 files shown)


FG-TUM:
We could use this base class for typical functions that are reused often like a function to fill a domain with particles.

SteffenSeckler:
hmm possible, but I would probably put that in some extra file to have it somewhere, where it belongs and not clutter everything in some akward class that just initializes the logger and stuff like that ;)

PR #24, opened 2018-04-25
corrects exception throwing of ExceptionHandler

### LINKED ISSUE #19: ExceptionHandler

SteffenSeckler:
we assume that ExceptionHandler::_behavior is set to throw throughout the explanations of this PR.

adds tests to properly check for the thrown error/exception when using the ExceptionHandler

if behavior is set to throw:
* fixes: throwing exceptions always produced a std::exception with meaningless what() methods
* feature: throwing a caught exception using the ExceptionHandler looses the type of the exception
e.g.
```C++
try{
 // sth. that produces an error
} catch (std::exception& e){
 ExceptionHandler::exception(e);  // produced wrong result
}
```
calling `ExceptionHandler::rethrow();` inside a catch block will now rethrow the caught exception. You can still reuse the normal `exception()` method, to provide additional information.

```C++
try{
 // sth. that produces an error
} catch (std::logic_error&) {
 ExceptionHandler::exception("logic error caught you should do input parameters properly");  // will throw an AutoPasException
}
catch (...) {
 ExceptionHandler::rethrow();  // will rethrow the current error
}
```

closes #19

### DIFF
--- a/src/utils/ExceptionHandler.cpp
+++ b/src/utils/ExceptionHandler.cpp
@@ -9,4 +9,19 @@
 std::mutex autopas::utils::ExceptionHandler::exceptionMutex;
 autopas::utils::ExceptionBehavior autopas::utils::ExceptionHandler::_behavior =
     ExceptionBehavior::throwException;
-std::function<void()> autopas::utils::ExceptionHandler::_customAbortFunction = abort;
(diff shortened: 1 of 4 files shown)


SteffenSeckler:
@FG-TUM review please :)

Issue #10, opened 2018-04-17
VerletList Container


SteffenSeckler:
Implement the VerletListContainer
* should reuse functionality from LinkedCells

deletion of particles should set VerletLists::_neighborListIsValid to false
benchmark for rebuild + md or so -- how much speedup is theoretically possible
add real life example and functionality needed
updating halo particles without change of data structure - i.e. get particle by id or position and modify it that way
particle iterators: should they include the halo particles?

Todos from #25 
ToDo:
1. deletion of particles should set VerletLists::_neighborListIsValid to false
2. <del> benchmark for rebuild + md or so -- how much speedup is theoretically possible</del>
3. <del>add real life example and functionality needed</del>
   * <del>updating halo particles without change of data structure - i.e. get particle by id or position and modify it that way</del>
   * <del>particle iterators: should they include the halo particles?</del>
4. Vectorization (new issue #35)

SteffenSeckler:
@AutoPas/design 
any ideas for proper implementations of the SoALoader and SoAExtractor functions?
currently they are cell based. Do probably have to change that...

SteffenSeckler:
see also https://github.com/orgs/AutoPas/teams/design/discussions/6

SteffenSeckler:
Todo
2. benchmark for rebuild + md or so -- how much speedup is theoretically possible
is done (see https://github.com/AutoPas/AutoPas/commit/56911aea93cd016c59be713159a6056beef52500 and https://github.com/AutoPas/AutoPas/commit/c884235435a468687010117f51695296b25e04a7)

tchipev:
very cool :)
how does it look like with vectorization? This paper discusses simple LJ Verlet list implementation with intrinsics, maybe you can look at it:
https://software.intel.com/sites/default/files/article/478279/mxhmd.final.pdf

Also, in LAMMPS, the Verlet lists are vectorized with compiler pragmas only (I think) and the code is more or less readable. You can look at it in void PairLJCutIntel::eval(...) here:
https://github.com/lammps/lammps/blob/master/src/USER-INTEL/pair_lj_cut_intel.cpp

Best,
Nikola

SteffenSeckler:
see https://github.com/AutoPas/AutoPas/wiki/Benchmarks

SteffenSeckler:
@FG-TUM @tchipev 
currently thinking about how to make an interface to exchange particles for verlet lists.
even if the verlet list is reused, we still need to update the halo particles.
[...]

PR #25, opened 2018-04-26
Feature/reusable verlet lists

SteffenSeckler:
adds the possibility to reuse verlet lists over multiple steps.
#10 
Done:
* adds fixed rebuild frequency
* rebuild of the neighbor lists triggered after frequency reached
* rebuild of list forced if:
   * particle/haloparticle is added
   * container is updated
* unit tests:
   * check rebuild frequency
   * check forced rebuild
* checkNeighborListsAreValid(): method to check whether the particles have not moved to much. this is more or less a function for the user to check whether his used rebuild frequency was small enough. (particles don't move too much)

ToDo:
1. deletion of particles should set VerletLists::_neighborListIsValid to false
4. benchmark for rebuild + md or so -- how much speedup is theoretically possible
5. add real life example and functionality needed
   * updating halo particles without change of data structure - i.e. get particle by id or position and modify it that way
   * particle iterators: should they include the halo particles?

### DIFF
--- a/src/containers/CellBlock3D.h
+++ b/src/containers/CellBlock3D.h
@@ -84,6 +84,25 @@ class CellBlock3D {
   ParticleCell &getContainingCell(const std::array<double, 3> &pos) const;
 
   /**
+   * Get the lower and upper corner of the cell at the 1d index index1d
+   * @param index1d the 1d index
+   * @param boxmin the lower corner (out)
+   * @param boxmax the upper corner (out)
+   */
+  void getCellBoundingBox(const index_t index1d, std::array<double, 3> &boxmin,
+                          std::array<double, 3> &boxmax);
+
+  /**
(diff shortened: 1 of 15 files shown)


SteffenSeckler:
@FG-TUM 
ad ToDo 1.:
deletion of particles should set VerletLists::_neighborListIsValid to false

what do you think is the best option to do that? I have more or less 2-3:
1. do it using the particle iterators -> the iterator will need a reference/pointer to the particlecontainer
2. do it using the cells -> the cell will need a reference/pointer to the particlecontainer
3. (don't do it let the user do it (potentially dangerous))

SteffenSeckler:
@FG-TUM should you have time, a review would be nice :)
the todos from the top are still open, but I don't want this PR to get too big, so I will do that with the next.

P.S: we probably want to add the cellboarder-and-flag manager soonishly

PR #27, opened 2018-04-29
omp sliced

FG-TUM:
OpenMP parallelization for the sliced traversal.

ToDo:
* currently if the traversal is not applicable (e.g. domain too small) a stupid, not parallel, fallback version is used. This needs to be exchanged by e.g. c08 and decided/handled by a traversal tuner (=autopas class?)
* ~manually merge OpenMP cmake modules~
* maybe ~some doc and~ formatting since this was done in a rush _sorry_ :)
* ~maybe clean up some attributes... I'll look into that tomorrow.~


SteffenSeckler:
regarding the Todo's:
* I would actually go for some kind of strategy pattern to manage the selection of different things (traversals, ...) and put that outside of the AutoPas class. This makes the implementation of strategies, where one specific combination is specified really straight forward, as we simply have another strategy for that. Possible strategies are (names are by no means final):
   * auto-tuning strategy
   * fixed strategy
   * ...

CMakeLists.txt
```diff
@@ -28,10 +28,11 @@ include(version.cmake)
 include(clang-tidy)
 include(clang-format)
 include(doxygen)
+include(OpenMP)
```
SteffenSeckler: merge openmp cmake module files (they seem to do exactly the same)

src/containers/cellPairTraversals/CellPairTraversal.h
```diff
@@ -32,13 +32,22 @@ class CellPairTraversals {
...
    */
   virtual ~CellPairTraversals() = default;
 
+  /**
+   * checks if the traversal is applicable to the current state of the domain.
+   * @return true iff the traversal can be applied.
+   */
+  virtual bool isApplicable() {
+      // return false by default to force thinking about this
+      return false;
```
SteffenSeckler: we could also make it pure virtual

src/utils/WrapOpenMP.h
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+#include <cassert>
+#if defined(_OPENMP)
+#include <omp.h>
+
+inline int autopas_get_thread_num() { return omp_get_thread_num(); }
+inline int autopas_get_num_threads() { return omp_get_num_threads(); }
+inline int autopas_get_max_threads() { return omp_get_max_threads(); }
+typedef omp_lock_t autopas_lock_t;
```
SteffenSeckler: I'm using c++11's std::thread's locks for the ExceptionHandler and it is also used for the logging library. should not pose any problems, right?

src/containers/cellPairTraversals/SlicedTraversal.h
```diff
@@ -5,9 +5,11 @@
  *      Author: tchipevn
```
SteffenSeckler: add authors

src/containers/cellPairTraversals/SlicedTraversal.h
```diff
@@ -5,9 +5,11 @@
...
 
-#ifndef SRC_CONTAINERS_CELLPAIRTRAVERSALS_SLICEDTRAVERSAL_H_
-#define SRC_CONTAINERS_CELLPAIRTRAVERSALS_SLICEDTRAVERSAL_H_
+#pragma once
 
+#include <utils/WrapOpenMP.h>
+#include <algorithm>
+#include <iomanip>
 #include "CellPairTraversal.h"
 #include "utils/ThreeDimensionalMapping.h"
```
SteffenSeckler: add documentation (todo: enhance documenation)

src/utils/WrapOpenMP.h
FG-TUM: don't think so. But I do not really want want to mix omp threads with std::thread locks.

SteffenSeckler:
TODO: remove one of openmp.cmake / OpenMP.cmake file

FG-TUM:
> is this the sliced traversal variant from mardyn, or did you make any changes? If yes, which?

Conceptually it is the same. I just tried a different way to implement it without the need to copy the `Permutations` class and with the use of the `static` scheduler of OpenMP (instead of manually reimplementig it).

SteffenSeckler:
one remark:
I'm currently trying to get a proper data race detector running. So this will be added in another pull request (today)

PR #29, opened 2018-05-02
See if automated status can be overridden

NeoTheThird:
Please ignore this for now.


NeoTheThird:
Okay, never mind. It was worth a try.

PR #38, opened 2018-05-18
pipeline changes for docker and openshift cloud

SteffenSeckler:
this moves the pipeline almost entirely to the [openshift cloud](https://en.wikipedia.org/wiki/OpenShift)

**Features:**
* better parallelizability (12 cpus per node, (partially) with hyperthreading
* everything in doxygen containers -- easier testing of different compilers, ... (errors fixed in https://github.com/AutoPas/AutoPas/commit/d59d48950a6c56fb01b0dc3f083b56e09a34a123, https://github.com/AutoPas/AutoPas/commit/b10382677a0279a615bdc4a4bee8ed1b4ced0986 could be identified that way)
* no dependency on installed software, easy "software" installation
* automatic scaling of executors

the only thing that is not in the cloud is the update of the doxygen documentation. As long, as we remain hosting this, this will probably not be changed. (In theory we could move the documentation to github pages)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -70,37 +70,49 @@ pipeline{
                 parallel(
                     normal: {
                         githubNotify context: 'test', description: 'test in progress...',  status: 'PENDING', targetUrl: currentBuild.absoluteUrl
-                        dir("build"){
-                            //sh "env CTEST_OUTPUT_ON_FAILURE=1 make test"
-                            sh 'env GTEST_OUTPUT="xml:$(pwd)/test.xml" ./tests/testAutopas/runTests'
+                        container('autopas-gcc7-cmake-make') {
+                            dir("build"){
+                                //sh "env CTEST_OUTPUT_ON_FAILURE=1 make test"
+                                sh 'env GTEST_OUTPUT="xml:$(pwd)/test.xml" ./tests/testAutopas/runTests'
+                            }
                         }
                     },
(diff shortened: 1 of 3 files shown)


SteffenSeckler:
@NeoTheThird for future reference ;)

FG-TUM:
Regarding Doc on github.
I stumbled upon [a gist](https://gist.github.com/vidavidorra/548ffbcdae99d752da02) where someone explained how to deploy a doxygen doc gh-pages. I have neither tested nor fully read it, just leaving it here...

Jenkinsfile
```diff
@@ -10,36 +15,62 @@ pipeline{
...
+                    },
+                    addresssanitizer: {
+                        container('autopas-gcc7-cmake-make') {
+                            dir("build-addresssanitizer"){
+                                sh "cmake -DCMAKE_BUILD_TYPE=Debug -DENABLE_ADDRESS_SANITIZER=ON .."
+                                sh "make -j 4"
+                            }
+                        }
+                    },
+                    addresssanitizerrelease: {
```
FG-TUM: Don't want to be nitpicky but camelCase would improve readability here or does this kill jenkis somehow?

SteffenSeckler:
yup sounds good

Jenkinsfile
SteffenSeckler: yup sounds good

Issue #42, opened 2018-05-21
CellFunctor loads and extracts SoA after every call


SteffenSeckler:
**Describe the bug**
The processCell(Pair)SoA(No)N3 functions currently all call SoALoaders and Extractors. Due to the nature of the traversals, this means the soa are loaded and extracted many times during one traversal, e.g:
```C++
void processCellPairSoAN3(ParticleCell &cell1, ParticleCell &cell2) {
    _functor->SoALoader(cell1, &cell1._particleSoABuffer);
    _functor->SoALoader(cell2, &cell2._particleSoABuffer);

    _functor->SoAFunctor(cell1._particleSoABuffer, cell2._particleSoABuffer,
                         true);

    _functor->SoAExtractor(&cell1, &cell1._particleSoABuffer);
    _functor->SoAExtractor(&cell2, &cell2._particleSoABuffer);
  }
```

**Expected behavior**
The SoA should only be loaded and extracted once per traversal, i.e. loaded at the beginning and extracted at the end.
This should be done in within ParticleContainer::iteratePairwiseSoA2(), but outside of the actual traversal, e.g:
```C++
template <class ParticleFunctor>
  void iteratePairwiseSoA2(ParticleFunctor *f, bool useNewton3 = true) {
    if (useNewton3) {
       // new:
       loadSoA();

       // old:
       CellFunctor<Particle, ParticleCell, ParticleFunctor, true, true>
          cellFunctor(f);
      SlicedTraversal<ParticleCell, CellFunctor<Particle, ParticleCell,
                                                ParticleFunctor, true, true>>
          traversal(this->_data, _cellBlock.getCellsPerDimensionWithHalo(),
                    &cellFunctor);
      traversal.traverseCellPairs();

      // new:
      extractSoA();
    } else {
      ...
    }
  }
```
**Additional context**
Note, that some tests probably have to be rewritten to acomodate that behavior.

SteffenSeckler:
fixed with #48

PR #48, opened 2018-05-27
Fewer SoALoad/Extract calls

FG-TUM:
Referencing #42 

This removes the `SoALoad/Extract` calls from the cell functor since it called `Load/Extract` for each pairwise interaction. The `Load/Extract` is now done in separate iterations in the `iteratePairwiseSoA` methods of the containers.

Speed improvement on a grid example:

Container | Iterations | \#Particles | grid spacing | cutoff | 
--------------|--------------|--------------|---------------|-----------
Linked-Cells | 10 | 60^3 | 1 | 2.5 | 

Time for one iteratePairwise call in s:

Traversal | Old | New | Speedup
-------------|------|--------|--------------
c08 | .96 | .84 | 1.14
sli | .93 | .82 | 1.13

### DIFF
--- a/src/containers/LinkedCells.h
+++ b/src/containers/LinkedCells.h
@@ -228,7 +228,7 @@ class LinkedCells : public ParticleContainer<Particle, ParticleCell> {
    * @param functor
    */
   template <class ParticleFunctor>
-  void loadSoAs(ParticleFunctor &functor) {
+  void loadSoAs(ParticleFunctor *functor) {
 #ifdef AUTOPAS_OPENMP
     //TODO find a condition on when to use omp or when it is just overhead
 #pragma omp parallel for
@@ -244,7 +244,7 @@ class LinkedCells : public ParticleContainer<Particle, ParticleCell> {
    * @param functor
    */
   template <class ParticleFunctor>
-  void extractSoAs(ParticleFunctor &functor) {
+  void extractSoAs(ParticleFunctor *functor) {
 #ifdef AUTOPAS_OPENMP
     //TODO find a condition on when to use omp or when it is just overhead
 #pragma omp parallel for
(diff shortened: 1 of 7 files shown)


FG-TUM:
I think there is still room for improvement.
It would be better (cache efficient) if we would load one SoA directly before its first usage and unload after the last. However, then this needs to be done by the traversal either through some counter or smart reasoning about the traversal order.

src/containers/LinkedCells.h
```diff
@@ -216,6 +221,38 @@ class LinkedCells : public ParticleContainer<Particle, ParticleCell> {
...
   CellBlock3D<ParticleCell> _cellBlock;
   // ThreeDimensionalCellHandler
+
+  /**
+   * Iterate over all cells and load the data in the SoAs
+   * @tparam ParticleFunctor
+   * @param functor
+   */
+  template <class ParticleFunctor>
+  void fillSoAs(ParticleFunctor &functor) {
```
SteffenSeckler: 1. seems like you have chosen a different name for these functions compared to the verlet-list-vectorization branch.
I had: loadSoA and extractSoA
to make it more consistent with the function names of the Functor. about the `s` I dont really care, so you can leave it as is and I will change it in the Verletlists.
2. the `&` should probably be a `*`, otherwise the `functor->` looks awkward. you are also calling this function with a `Functor *`. so in the end, you have here: `Functor*&` so a functor pointer by reference

src/containers/LinkedCells.h
```diff
@@ -216,6 +221,38 @@ class LinkedCells : public ParticleContainer<Particle, ParticleCell> {
...
+    }
+  }
+
+  /**
+   * Iterate over all cells and fetch the data from the SoAs
+   * @tparam ParticleFunctor
+   * @param functor
+   */
+  template <class ParticleFunctor>
+  void extractSoAs(ParticleFunctor &functor) {
```
SteffenSeckler: probably `*` instead of `&`

src/containers/LinkedCells.h
FG-TUM: Let's stick to the plural since we treat all cells ;)

PR #50, opened 2018-05-31
particle iterator wrapper

SteffenSeckler:
for now just to track the changes.

this introduces an interface and a wrapper for the particleiterator class, to get rid of some templates and to make everything better designable, i.e. we can now have different iterators for different containers, which is needed for things like a tree container where the normal RegionParticleIterator can no longer be used.

also introduces a ParticleContainerInterface, where no ParticleCell template is needed.
The ParticleCell template should later on also be removable from the AutoPas class (most likely)


SteffenSeckler:
@FG-TUM 
you may have a first look at this and leave some comments.
I am not yet fully convinced that we need all of it.

FG-TUM:
I can't make a proper review at the moment (also today is a holiday in Germany :P ) but for a first glance, the combination of interfaced AND abstract classes seems a bit overly complicated. Maybe a look at a class diagram might help to throw some stuff out.

SteffenSeckler:
well the abstract class does implement a few things that most, but not all ParticleContainers should have in common, i.e. sth like a vector of `ParticleCell`s, as well as other things (iteratePairwise, ...), so it needs the ParticleCell template. At least for now.
The interface does NOT have that template, instead it only has `Particle` as a template parameter. Which is the only thing a ParticleContainer should have.

The interface does also NOT include the `iteratePairwise*` functions, as they need the ParticleCell template parameter (see #52)

SteffenSeckler:
@FG-TUM proper review please :)

src/iterators/ParticleIteratorWrapper.h
```diff
@@ -0,0 +1,65 @@
...
+ *      Author: seckler
+ */
+#pragma once
+
+#include "ParticleIteratorInterface.h"
+
+namespace autopas {
+
+/**
+ * ParticleIteratorInterface class to iterate over particles.
```
FG-TUM: Wrong doc. Here also the necessity for a wrapper should be explained in a way someone new to the code understands it.

src/containers/ParticleContainerInterface.h
```diff
@@ -0,0 +1,135 @@
...
+  /**
+   * iterate over all particles in a specified region
+   * for(auto iter = container.getRegionIterator(lowCorner,
+   * highCorner);iter.isValid();++iter)
+   * @param lowerCorner lower corner of the region
+   * @param higherCorner higher corner of the region
+   * @return iterator to iterate over all particles in a specific region
+   */
+  virtual ParticleIteratorWrapper<Particle> getRegionIterator(std::array<double, 3> lowerCorner,
+                                                              std::array<double, 3> higherCorner) = 0;
```
FG-TUM: this one does not care weather the given region is in the halo, owned or both areas, correct?

src/containers/ParticleContainer.h
```diff
@@ -5,26 +5,25 @@
...
+// consider multiple inheritance or delegation to avoid virtual call to Functor
 /**
  * The ParticleContainer class stores particles in some object and provides
  * methods to iterate over its particles.
  * @tparam Particle Class for particles
  * @tparam ParticleCell Class for the particle cells
  */
 template <class Particle, class ParticleCell>
-class ParticleContainer {
+class ParticleContainer : public ParticleContainerInterface<Particle> {
```
FG-TUM: when compiling with g++7 I'm still getting lots of warinings that methods can be marked as override here (mostly for getters & setters). Apparently you need to put both `final` and `override` (order irrelevant but plz consistent ;) ).

src/containers/DirectSum.h
```diff
@@ -144,8 +144,10 @@ class DirectSum : public ParticleContainer<Particle, ParticleCell> {
     return false;
   }
 
-  ParticleIterator<Particle, ParticleCell> begin(IteratorBehavior behavior = IteratorBehavior::haloAndOwned) override {
-    return ParticleIterator<Particle, ParticleCell>(&this->_data, &_cellBorderFlagManager, behavior);
+  ParticleIteratorWrapper<Particle> begin(IteratorBehavior behavior = IteratorBehavior::haloAndOwned) override {
+    return ParticleIteratorWrapper<Particle>(
+        (ParticleIteratorInterface<Particle> *)new internal::ParticleIterator<Particle, ParticleCell>(
```
FG-TUM: is this cast necessary? In the respective LinkedCell version, there is none and g++ compiles this without a cast.

src/iterators/RegionParticleIterator.h
```diff
@@ -10,6 +10,7 @@
 #pragma once
 
 namespace autopas {
+namespace internal {
```
FG-TUM: i like the internal namespace but we should specify somewhere how to use them / whats in what. Can this be done via doxygen or should we do this via the github wiki (or both)?

src/iterators/SingleCellIteratorInterface.h
```diff
@@ -0,0 +1,72 @@
...
+
+namespace internal {
+
+/**
+ * All implementations of the interface should inherit from this class. It extends the interface just by the clone
+ * method, which is needed by the Wrapper.
+ * @tparam Particle
+ */
+template <class Particle>
+class SingleCellIteratorInterfaceImpl : public SingleCellIteratorInterface<Particle> {
```
FG-TUM: Naive question: Why do we need this? Can the `clone()` method not go into the `SingleCellIteratorInterface`? And if not please explain in the doc the why.

src/iterators/SingleCellIteratorWrapper.h
```diff
@@ -0,0 +1,78 @@
...
+   * copy constructor
+   * @param otherParticleIteratorWrapper the other ParticleIteratorWrapper
+   */
+  SingleCellIteratorWrapper(const SingleCellIteratorWrapper& otherParticleIteratorWrapper) {
+    _particleIterator = std::unique_ptr<internal::SingleCellIteratorInterfaceImpl<Particle>>(
+        otherParticleIteratorWrapper._particleIterator->clone());
+  }
+
+  /**
+   * copy constructor
```
FG-TUM: istn't this called copy assignment operator?

tests/testAutopas/AutoPasTest.cpp
```diff
@@ -0,0 +1,16 @@
...
+
+#include "AutoPasTest.h"
+#include "AutoPas.h"
+
+TEST_F(AutoPasTest, getRegionParticleIterator) {
+  auto iter = autoPas.getRegionIterator({0., 0., 0.}, {4., 4., 4.});
+
+  for (; iter.isValid(); ++iter) {
+    iter->setR(iter->getR());
+  }
```
FG-TUM: Isn't there an assert missing here? So far this test seems to just test for crashes or am I missing s.th.? :D

tests/testAutopas/RegionParticleIteratorTest.cpp
```diff
@@ -55,3 +55,63 @@ TEST_F(RegionParticleIteratorTest, testLinkedCellsRegionParticleIteratorEmpty) {
...
+  }
+
+  // check the touch using the normal iterator
+  for (auto iterator = lcContainer.begin(); iterator.isValid(); ++iterator) {
+    //  std::cout << "id: " << iterator->getID() << " at [" <<
+    //  iterator->getR()[0]
+    //         << ", " << iterator->getR()[1] << ", " << iterator->getR()[2]
+    //              << "] touched:" << iterator->getNumTouched() << std::endl;
+
+    ASSERT_EQ(iterator->inBox(_regionMin, _regionMax) ? 1 : 0, iterator->getNumTouched());
```
FG-TUM: is there a test for `inBox` and `isValid`? Many tests seem to rely on those.

src/iterators/SingleCellIteratorWrapper.h
```diff
@@ -0,0 +1,78 @@
...
+ */
+#pragma once
+
+#include <utils/ExceptionHandler.h>
+#include "SingleCellIteratorInterface.h"
+
+namespace autopas {
+
+/**
+ * ParticleIteratorInterface class to iterate over particles.
```
FG-TUM: Wrong doc. Here also the necessity for a wrapper should be explained in a way someone new to the code understands it.

src/containers/DirectSum.h
SteffenSeckler: not necessary, since I have added that in the constructor

src/iterators/SingleCellIteratorInterface.h
SteffenSeckler: the reason:
the wrapper should not have that method. It is also completely undefined how the wrapper would handle that. I also don't want it to be visible for the wrapper to a user.

tests/testAutopas/AutoPasTest.cpp
SteffenSeckler: yes and it makes certain, that the method is actually called, so that potential template errors are found.
I have managed to compile the code, while this block was not there and adding that test let the compilation fail...

src/containers/ParticleContainerInterface.h
SteffenSeckler: correct, probably todo

src/containers/ParticleContainer.h
SteffenSeckler: sounds like a bug in g++7
final implicitly always means override

src/containers/ParticleContainerInterface.h
SteffenSeckler: done

src/iterators/SingleCellIteratorWrapper.h
SteffenSeckler: yep, copy past error

src/iterators/SingleCellIteratorWrapper.h
SteffenSeckler: fixed

src/containers/ParticleContainer.h
SteffenSeckler: ok, added the overrides, even though clang-tidy complains about this.
(we should probably add a .clang-tidy file + add this to jenkins...)

src/iterators/RegionParticleIterator.h
FG-TUM: This is not an general info for the README. I think the best way would be Doxygen so to keep all documentation in one place.

FG-TUM: I just tested putting a documentation block on top of one namespace line:
```
/**
 * doc for the namespace
 */
namespace autopas { ... }
```

I think what doxygen turns this into is quite useful.

SteffenSeckler: ok, somewhere in an extra file? namespaces.h or so?

FG-TUM: probably a good idea yes.

tests/testAutopas/AutoPasTest.cpp
FG-TUM: Then at least document this here. The way it is now it looks a bit strange and superfluous.

src/iterators/RegionParticleIterator.h
SteffenSeckler: ok, I will add that

tests/testAutopas/AutoPasTest.cpp
FG-TUM: Then at least document this here. The way it is now it looks strange and superfluous.

src/iterators/RegionParticleIterator.h
SteffenSeckler: hmm there are a total of three namespaces that only span one class:
```
namespace arrayMath {
namespace logger {
namespace ThreeDimensionalMapping {
```

at least the logger I would like to convert to a class. What about the others? They are sometimes used via:
```C++
  using ThreeDimensionalMapping::threeToOneD;
```
do we want them to stay namespaces or should we convert them to classes, too?

FG-TUM: Turning `ThreeDimensionalMapping` into a class `ThreeDimensionalMapper` with static methods sounds the most intuitive to me.

SteffenSeckler: ok done (without the renaming though)

tests/testAutopas/AutoPasTest.cpp
SteffenSeckler: done

SteffenSeckler:
I have added a few more things, mainly regarding formatting and so:
* always use `#pragma once` instead of header guards using ifndef
* end namespaces with `}  // namespacename` and not with `} /* namespacenanme */`

Issue #53, opened 2018-06-01
missing SoALoader and SoAExtractor in FlopCounter


SteffenSeckler:
**Describe the bug**
The FlopCounter does not implement the SoALoader/SoAExctractor functions.
Executing iteratePairwiseSoA with the FlopCounter will lead to:
```
terminate called after throwing an instance of 'autopas::utils::ExceptionHandler::AutoPasException'
  what():  Functor::SoALoader: not yet implemented
```
**To Reproduce**
1. build AutoPas normally
2. execute e.g. `md-flexible --container linked --cutoff 1. --data-layout soa --functor lj --iterations 10 --particles-per-dimension 10 --particle-spacing 0.4`
3. See error

**Expected behavior**
FlopCounter sshould implement the SoALoader and SoAExtractor, where the latter is empty

**gdb-trace**
```
Starting program: 

Program received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) where
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff6cb9801 in __GI_abort () at abort.c:79
#2  0x00007ffff76ac8fb in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007ffff76b2d3a in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007ffff76b2d95 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007ffff76b2fe8 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x000000000049326a in autopas::utils::ExceptionHandler::exception<autopas::utils::ExceptionHandler::AutoPasException> (e=...)
    at ../../src/utils/ExceptionHandler.h:58
#7  0x000000000049295b in autopas::utils::ExceptionHandler::exception<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >
    (exceptionString=<incomplete type>) at ../../src/utils/ExceptionHandler.cpp:17
[...]

PR #55, opened 2018-06-01
Check Examples

### LINKED ISSUE #54: examples are not tested in Jenkins
**Describe the Bug**
The examples are currently not tested in Jenkins

**Expected Behavior**
Do we want it to be tested? If we don't do it, we run into the risk of having non-running examples

**Possible Solutions**
* Add a cmake target to run all examples. This would require all of them to be relatively short.
* Maybe even add some method to test for correctness? e.g. correct output.

@FG-TUM: Opinions?

SteffenSeckler:
fixes #54 
* `make checkExamples` or `ninja checkExamples` will check the examples. The checks are disabled for the normal `make test` (`ninja test`) or `ctest`.
* adds the check to jenkins using the release target. (everything else is too time-costly)
* the checks use ctest's addTest() feature.

To add new tests just copy one of the other tests.

build failing because of #53

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,4 +1,5 @@
-ADD_CUSTOM_TARGET(checkExamples)
+ADD_CUSTOM_TARGET(checkExamples
+        COMMAND ctest -C checkExamples -j 4)
 
 ADD_SUBDIRECTORY(basic)
 ADD_SUBDIRECTORY(md-flexible)
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,3 +1,5 @@
+ADD_CUSTOM_TARGET(checkExamples)
+
 ADD_SUBDIRECTORY(basic)
 ADD_SUBDIRECTORY(md-flexible)
 ADD_SUBDIRECTORY(md)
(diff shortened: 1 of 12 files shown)


SteffenSeckler:
@FG-TUM @NeoTheThird review please ;)
there shouldn't be too much to complain in this case :)

examples/CMakeLists.txt
```diff
@@ -1,7 +1,11 @@
+ADD_CUSTOM_TARGET(checkExamples
+        COMMAND ctest -C checkExamples -j 4)
```
NeoTheThird: Any reason to not make this the same as https://github.com/AutoPas/AutoPas/pull/55/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3R203?

examples/md-flexible/CMakeLists.txt
```diff
@@ -18,3 +18,15 @@ ADD_EXECUTABLE(md-flexible
 #set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -qopt-report-phase=vec")
 TARGET_LINK_LIBRARIES(md-flexible
         autopas)
+
+#-----------------test-----------------
+# add check for current target
+add_test(
+        NAME md-flexible.test
```
NeoTheThird: Fails with `Exception: Child aborted`. Fyi, this is not an error message you'll want to google.

SteffenSeckler:
well the failing test is because there is a bug (#53)

examples/CMakeLists.txt
SteffenSeckler: jup, on normal pc's you don't have 8 procs

PR #56, opened 2018-06-01
SoA stuff for flop counter

FG-TUM:
- FlopCounter now supports SoA
- Restores Examle(s) -> #53
- Simple tests for FlopCounter -> Any suggestions for more?

### DIFF
--- a/src/pairwiseFunctors/FlopCounterFunctor.h
+++ b/src/pairwiseFunctors/FlopCounterFunctor.h
@@ -183,11 +183,10 @@ class FlopCounterFunctor : public Functor<Particle, ParticleCell> {
   unsigned long getKernelCalls() const { return _kernelCalls; }
 
  private:
-
   double _cutoffSquare;
   unsigned long _distanceCalculations, _kernelCalls;
   // 3 sub + 3 square + 2 add
-  static constexpr double  numFlopsPerDistanceCalculation = 8.0;
+  static constexpr double numFlopsPerDistanceCalculation = 8.0;
 };
 
 } /* namespace autopas */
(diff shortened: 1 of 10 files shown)


FG-TUM:
I did

```
mkdir build && cd build
cmake ..
make clangformat
git commit -a
git push
```

why do I still get rejected by jenkins with the message that format errors were detected?

```
make clangformat
git status
```
at least for me tells me that there are no changes.

SteffenSeckler:
hmm, Jenkins uses clang-format-6.0 maybe you are using an older version.
It's a bit problematic that the behavior is not the same over the different versions.
I have formatted it (https://github.com/AutoPas/AutoPas/pull/56/commits/e3aa6111435f022398c0f06c179229fa4085f409). The main differences are indentation of comments and reordering of includes

src/pairwiseFunctors/FlopCounterFunctor.h
```diff
@@ -165,12 +183,10 @@ class FlopCounterFunctor : public Functor<Particle, ParticleCell> {
...
   double _cutoffSquare;
   unsigned long _distanceCalculations, _kernelCalls;
+  // 3 sub + 3 square + 2 add
+  static constexpr double numFlopsPerDistanceCalculation = 8.0;
 };
 
-} /* namespace autopas */
-
-#endif /* SRC_PAIRWISEFUNCTORS_FLOPCOUNTERFUNCTOR_H_ */
+} /* namespace autopas */
```
SteffenSeckler: I would prefer
``` C++
}  // namspace autopas
```
in general. What is you opinion? I think that is also what google wants

FG-TUM: I actually have no preference here

FG-TUM:
I have `clang-format-5.0`
The fact that different versions format different is a problem since clang is a bit of a hassle to install from source. But I guess we need to agree on one version and document it in the [README](https://github.com/AutoPas/AutoPas/blob/master/README.md). Which version I do not mind but I think it makes more sense to use the more (or most?) recent.

SteffenSeckler:
@FG-TUM llvm provides ppa's. also for clang 6.0:
https://apt.llvm.org/

FG-TUM:
Neat. I'll update the Readme

PR #57, opened 2018-06-05
verlet list rebuild optimizations

SteffenSeckler:
the verlet-lists are now (re)build using SoA's.
This saves a lot of lookups (hash generations) and some other things.

Speedup is clearly visible, especially as soon as there are more than one particle per cell:
Verlet Lists (md-main, 1000 cells):

type | particle count | rebuild time before | rebuild time now | kernel time (LJ)
------ | ----------------- | ------------------------ | ---------------------- | -----------------
soa | 10.000 | 0.12 | 0.07 | 0.010
aos | 10.000 | 0.10 | 0.05 | 0.018
soa | 1.000 | 1.8e-3 | 1.4e-3 | 1.9e-4
aos | 1.000 | 1.6e-3 | 1.2e-3 | 1.8e-4

Linked Cells (md-main, 1000 cells):

type | particle count | kernel time (LJ)
------ | --------------- | ----
soa | 10.000 | 0.032
aos | 10.000 | 0.07
soa | 1.000 | 6.8e-4
aos | 1.000 | 1.0e-3

meaning, that in this case verlet lists are faster than linked cells, as soon as the rebuild happens at most every 4 time steps.

remaining optimization possibilities:
* from what I see, vectorization is not really possible here, as there are not really any flops done. the main time is spent in storing the lists
* upper optimization limit for the rebuild time should be roughly that of the kernel time of the aos LJ  linked-cells iteration. so we are already close to optimal.
* put neighborlists into particle class to prevent lookup alltogether -- involves lots of template magic and might not pay off

### DIFF
--- a/src/containers/VerletListHelpers.h
+++ b/src/containers/VerletListHelpers.h
@@ -51,17 +51,16 @@ class VerletListHelpers {
     virtual void SoAFunctor(SoA &soa, bool /*newton3*/ = true) override {
       if (soa.getNumParticles() == 0) return;
 
-      Particle **const __restrict__ idptr = reinterpret_cast<Particle * *const>(soa.begin(Particle::AttributeNames::id));
+      Particle **const __restrict__ idptr = reinterpret_cast<Particle **const>(soa.begin(Particle::AttributeNames::id));
       double *const __restrict__ xptr = soa.begin(Particle::AttributeNames::posX);
       double *const __restrict__ yptr = soa.begin(Particle::AttributeNames::posY);
       double *const __restrict__ zptr = soa.begin(Particle::AttributeNames::posZ);
 
-      for (unsigned int i = 0; i < soa.getNumParticles(); ++i) {
+      size_t numPart = soa.getNumParticles();
+      for (unsigned int i = 0; i < numPart; ++i) {
         auto &currentList = _verletListsAoS.at(idptr[i]);
 
-        for (unsigned int j = i + 1; j < soa.getNumParticles(); ++j) {
-          if (i == j) continue;
-
+        for (unsigned int j = i + 1; j < numPart; ++j) {
           const double drx = xptr[i] - xptr[j];
           const double dry = yptr[i] - yptr[j];
           const double drz = zptr[i] - zptr[j];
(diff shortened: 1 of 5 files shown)


SteffenSeckler:
@FG-TUM review here, too :)

src/containers/VerletListHelpers.h
```diff
@@ -45,7 +45,102 @@ class VerletListHelpers {
...
+      Particle **const __restrict__ idptr = reinterpret_cast<Particle **const>(soa.begin(Particle::AttributeNames::id));
+      double *const __restrict__ xptr = soa.begin(Particle::AttributeNames::posX);
+      double *const __restrict__ yptr = soa.begin(Particle::AttributeNames::posY);
+      double *const __restrict__ zptr = soa.begin(Particle::AttributeNames::posZ);
+
+      size_t numPart = soa.getNumParticles();
+      for (unsigned int i = 0; i < numPart; ++i) {
+        auto &currentList = _verletListsAoS.at(idptr[i]);
+
+        for (unsigned int j = i + 1; j < numPart; ++j) {
```
FG-TUM: we could check if it is worthwhile to vectorize this with `omp simd`

src/containers/VerletListHelpers.h
```diff
@@ -45,7 +45,102 @@ class VerletListHelpers {
...
+      double *const __restrict__ y2ptr = soa2.begin(Particle::AttributeNames::posY);
+      double *const __restrict__ z2ptr = soa2.begin(Particle::AttributeNames::posZ);
+
+      size_t numPart1 = soa1.getNumParticles();
+      for (unsigned int i = 0; i < numPart1; ++i) {
+        auto &currentList = _verletListsAoS.at(id1ptr[i]);
+
+        size_t numPart2 = soa2.getNumParticles();
+
+        for (unsigned int j = 0; j < numPart2; ++j) {
```
FG-TUM: same as above

src/containers/VerletLists.h
```diff
@@ -262,7 +273,17 @@ class VerletLists : public LinkedCells<Particle, ParticleCell> {
...
-    LinkedCells<Particle, ParticleCell>::iteratePairwiseAoS2(&f, useNewton3);
+    switch (_buildVerletListType) {
+      case BuildVerletListType::VerletAoS:
+        LinkedCells<Particle, ParticleCell>::iteratePairwiseAoS2(&f, useNewton3);
+        break;
+      case BuildVerletListType::VerletSoA:
+        LinkedCells<Particle, ParticleCell>::iteratePairwiseSoA2(&f, useNewton3);
+        break;
+      default:
+        utils::ExceptionHandler::exception("VerletLists::updateVerletListsAoS(): unsupported BuildVerletListType");
```
FG-TUM: add actual _buildVerletListType to the message

src/containers/VerletListHelpers.h
SteffenSeckler: there is a function call inside here: `currentList.push_back(...)`, so a vectorization is not possible immediately.
I have actually tried to do a vectorization here by allocating an additional array and just marking the respective fields. This was vectorizable, but slower than the method in here.

src/containers/VerletLists.h
SteffenSeckler: fixed with 
https://github.com/AutoPas/AutoPas/pull/57/commits/93b6908b38d3a6372b47530d61a95358a3f6dd2c and
https://github.com/AutoPas/AutoPas/pull/57/commits/064f22266ba915738221a6509910591a0eaf3abc:
adds possibility to pass fmt format like strings to the exception handler and uses it here

src/containers/VerletListHelpers.h
SteffenSeckler: see below

PR #66, opened 2018-06-26
sph soa

### LINKED ISSUE #58: vectorization - sph
vectorization of sph kernels

- [x] blocked by #59 
- sph SoA particle type
   - support for calcdensity functor
   - support for calchydroforce functor
- sph SoA vectorization
   - support for calcdensity functor
   - support for calchydroforce functor

SteffenSeckler:
fixes #58 by adding (partially) vectorized soa's for sph functors:

- sph SoA particle type
   - support for calcdensity functor
   - support for calchydroforce functor
- sph SoA vectorization (+ verlet)
   - support for calcdensity functor
   - support for calchydroforce functor
- sph SoA vectorization (+ verlet) tests
   - tests for calcdensity functor
   - tests for calchydroforce functor

## performance measurements:

numparticles | numsteps | container | functor | MFUPS aos | MFUPS soa | note 
----------------- | ------------- | ------------ | ---------- | ---------------- | ---------------- | -----
10000           | 30             | LC           | density | 0.075            | 0.30             | 
10000           | 30             | LC           | density | 0.075            | 0.19             |  without pragma omp simd
10000           | 30             | direct       | density | 0.011            | 0.044           |
10000           | 30             | verlet       | density | 0.32              | 0.48             | zero skin, rebuild every 10
10000           | 30             | LC           | hydro    | 0.059            | 0.122           | 
10000           | 30             | LC           | hydro    | 0.059            | 0.10             | forced vectorization (pragma omp simd+native)
10000           | 30             | direct       | hydro    | 0.011            | 0.047           |
10000           | 30             | verlet       | hydro    | 0.15              | 0.17             | zero skin, rebuild every 10

some of the performance improvement does not stem from vectorization. 
* in fact vectorization was not beneficial for the calchydroforcefunctor at all. clang only vectorizes it with -march=native, even with -mavx2 it does not vectorize the call, as there is a function call to sqrt inside the functor.
* density functor is vectorizable without much problems.
* soa verlet functor for density is vectorized (according to clang) and not vectorizable for hydro

## graphics:
https://github.com/AutoPas/AutoPas/wiki/Benchmarks

### DIFF
--- a/examples/sphDiagramGeneration/sph-diagram-generation.cpp
+++ b/examples/sphDiagramGeneration/sph-diagram-generation.cpp
@@ -96,7 +96,7 @@ int main(int argc, char *argv[]) {
     exit(1);
   }
 
-  autopas::VerletLists<autopas::sph::SPHParticle> verletCont(boxMin, boxMax, cutoff, skin, rebuildFrequency);
+  autopas::VerletLists<autopas::sph::SPHParticle> verletCont(boxMin, boxMax, cutoff, skin*cutoff, rebuildFrequency);
 
   addParticles(lcCont, numParticles);
 
--- a/examples/sphDiagramGeneration/sph-diagram-generation.cpp
+++ b/examples/sphDiagramGeneration/sph-diagram-generation.cpp
@@ -8,8 +8,8 @@
 #include "sph/autopassph.h"
 #include "utils/Timer.h"
 
-template <class Container>
-void measureContainer(Container *cont, int numParticles, int numIterations);
+template <class Container, class Functor>
+void measureContainer(Container *cont, Functor *func, int numParticles, int numIterations);
 
 double fRand(double fMin, double fMax) {
   double f = static_cast<double>(rand()) / RAND_MAX;
@@ -66,15 +66,23 @@ int main(int argc, char *argv[]) {
   autopas::DirectSum<autopas::sph::SPHParticle, autopas::FullParticleCell<autopas::sph::SPHParticle>> dirCont(
       boxMin, boxMax, cutoff);
 
+  autopas::sph::SPHCalcDensityFunctor densfunc;
+
+  autopas::sph::SPHCalcHydroForceFunctor hydrofunc;
+
   int numParticles = 16;
   int numIterations = 100000;
   int whichContainer = 0;
-  // int whichFunctor = 0;
-  if (argc == 5 or argc == 4) {
+  int whichFunctor = 0;
+  if (argc == 5) {
     numParticles = atoi(argv[1]);
     numIterations = atoi(argv[2]);
     whichContainer = atoi(argv[3]);
-    //  whichFunctor = atoi(argv[4]);
+    whichFunctor = atoi(argv[4]);
+  } else if (argc == 4) {
+    numParticles = atoi(argv[1]);
+    numIterations = atoi(argv[2]);
+    whichContainer = atoi(argv[3]);
   } else {
     exit(1);
   }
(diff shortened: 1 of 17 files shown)


SteffenSeckler:
@FG-TUM next review please :)

examples/sphDiagramGeneration/sph-diagram-generation.cpp
```diff
@@ -66,57 +66,115 @@ int main(int argc, char *argv[]) {
...
+    numParticles = atoi(argv[1]);
+    numIterations = atoi(argv[2]);
+    whichContainer = atoi(argv[3]);
+    whichFunctor = atoi(argv[4]);
+  } else if (argc == 4) {
     numParticles = atoi(argv[1]);
     numIterations = atoi(argv[2]);
     whichContainer = atoi(argv[3]);
-    //  whichFunctor = atoi(argv[4]);
   } else {
```
FG-TUM: a help text on how to use this would be useful but I don't know how user-friendly you want this example to be ;)

examples/sphDiagramGeneration/sph-diagram-generation.cpp
```diff
@@ -66,57 +66,115 @@ int main(int argc, char *argv[]) {
...
+  autopas::VerletLists<autopas::sph::SPHParticle> verletCont(boxMin, boxMax, cutoff, skin * cutoff, rebuildFrequency);
+
   addParticles(lcCont, numParticles);
 
   for (auto it = lcCont.begin(); it.isValid(); ++it) {
     dirCont.addParticle(*it);
+    verletCont.addParticle(*it);
   }
 
   if (whichContainer == 0) {
```
FG-TUM: enums would drastically increase readability here

examples/sphDiagramGeneration/sph-diagram-generation.cpp
```diff
@@ -66,57 +66,115 @@ int main(int argc, char *argv[]) {
...
   }
 }
 
-template <class Container>
-void measureContainer(Container *cont, int numParticles, int numIterations) {
-  autopas::sph::SPHCalcDensityFunctor func;
-  autopas::FlopCounterFunctor<autopas::sph::SPHParticle, autopas::FullParticleCell<autopas::sph::SPHParticle>>
-      flopFunctor(cont->getCutoff());
+template <class Container, class Functor>
+void measureContainer(Container *cont, Functor *func, int numParticles, int numIterations) {
```
FG-TUM: add `TODO: fix flopFunctor for verlet` or we might delete this by accident
(at least this was the problem if I remember corretly?)

SteffenSeckler: hmm I'm not a huge fan of the flopcounters :p 
they are only working for fixed cutoff radii and don't really show what is happening.
MMUPS, etc. are way more meaningful and they sort of force you to compare against other codes...

the flops are only a theoretical value and hold no real comparability against other codes (each code counts flops differently). e.g.
* ls1: only flops that actually need to be computed are counted (kernel calls only counted if `distance < cutoff` ) -> less flops are counted than actually calculated, i.e. you don't count unnecessary flop calculations that didn't need to be calculated
* other codes: count all the flops actually performed: even if they result from vectorization and masking, i.e. independent of whether the distance is smaller than the cutoff. -> makes everything appear way (~8x) faster.

SteffenSeckler:
@FG-TUM 
merged the soaiterator things, which breaks this PR. The problem is that the RMM cells are not compatible with the sphparticles (they don't have the fields `forceX, forceY, forceZ`). So we have two ways to fix this:

a) fix RMM by introducing an RMMSoAType -- needs to be done for every particle
b) remove the RMMParticleCells -- they are not really flexible for now, require different functors, ...
c) remove the SingleCellIteratorWrapper -- might give performance boost, the wrapper is also not really necessary, ASSUMING: the cell type is defined from the outside, i.e. the user.
d) just state the type of SingleCellIterator in the cell classes -- **i will do this for now**

which do you think is the best? som

SteffenSeckler:
compiles again with: 
* https://github.com/AutoPas/AutoPas/pull/66/commits/e71230504932b38772a00c62acd31257f55c368a
* the change is to get the iterator type at compile time instead of determining this dynamically.

examples/sphDiagramGeneration/sph-diagram-generation.cpp
SteffenSeckler: done

examples/sphDiagramGeneration/sph-diagram-generation.cpp
SteffenSeckler: done

SteffenSeckler:
updated graphics in benchmark:
https://github.com/AutoPas/AutoPas/wiki/Benchmarks
mainly aos is now a lot faster than previously (due to the better iterators)

SteffenSeckler:
new performance values of aos (lc + direct), verlet does not change much:

performance gain is here ~2x for aos (compared to before aos improvements)

numparticles | numsteps | container | functor | MFUPS aos | MFUPS aos old
----------------- | ------------- | ------------ | ---------- | ---------------- | ---------------------
10000           | 30             | LC           | density | 0.13              | 0.075             
10000           | 30             | direct       | density | 0.020            | 0.011           
10000           | 30             | LC           | hydro    | 0.094            | 0.059             
10000           | 30             | direct       | hydro    | 0.021            | 0.011

PR #70, opened 2018-06-28
AoS performance improvements

SteffenSeckler:
Problem:
ParticleIteratorWrappers were decreasing performance of AoS functors.
See these plots (orange and blue solid lines):
### Before Wrappers:
image

### After Wrappers:
image

### With the improvements made here:
image

### Speedup now vs. before wrappers:
image
notes:
* (without the linked-cells soa version, as this is sort of totally different and the previous performance was very bad)
* verlet lists: for very few particles (less than 0.1 particles per cell), they are slower than before. not entirely sure why this happens.
* linked cells: slower than before up to 0.5 particles per cell. this most likely stems from the overhead of the wrappers, i.e. having additional `new` calls, etc.. This is probably also the reason for the decreased performance of the verlet lists.
* some of the speedup comes from new return statements inside the CellFunctors.

@FG-TUM In theory we can remove the wrapper for the SingleCellIterators, as they are anyways internal only, so not visible to the outside. What is your opinion?

### DIFF
--- a/src/iterators/SingleCellIteratorInterface.h
+++ b/src/iterators/SingleCellIteratorInterface.h
@@ -71,7 +71,6 @@ class SingleCellIteratorInterfaceImpl : public SingleCellIteratorInterface<Parti
 }  // namespace internal
 }  // namespace autopas
 
-
 /**
  * gets a static cell iterator from an iteratorwrapper of a cell.
  * @param iter the iterator to be defined
--- a/src/iterators/SingleCellIteratorInterface.h
+++ b/src/iterators/SingleCellIteratorInterface.h
@@ -70,3 +70,26 @@ class SingleCellIteratorInterfaceImpl : public SingleCellIteratorInterface<Parti
 };
 }  // namespace internal
 }  // namespace autopas
+
+
+/**
+ * gets a static cell iterator from an iteratorwrapper of a cell.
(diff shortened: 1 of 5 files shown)


src/iterators/SingleCellIteratorInterface.h
```diff
@@ -70,3 +70,25 @@ class SingleCellIteratorInterfaceImpl : public SingleCellIteratorInterface<Parti
...
+  auto __ptr = __wrapper.get();                                                                                       \
+  {                                                                                                                   \
+    if (auto __##iter##ptr = dynamic_cast<                                                                            \
+            internal::SingleCellIterator<Particle, typename std::remove_reference<decltype(cell)>::type> *>(__ptr)) { \
+      auto iter = *__##iter##ptr;                                                                                     \
+      body                                                                                                            \
+    } else if (auto __##iter##ptr = dynamic_cast<RMMParticleCellIterator<Particle> *>(__ptr)) {                       \
+      auto iter = *__##iter##ptr;                                                                                     \
+      body                                                                                                            \
```
FG-TUM: since both branches seem to be identical can they be merged?

SteffenSeckler: no idea what you mean, but will merge anyways :)

PR #86, opened 2018-08-13
test examples with sanitizers

### LINKED ISSUE #76: Jenkins: test examples with sanitizers
**Describe the bug**
Examples are currently not tested with sanitizers. This should probably be done, as there is some functionality not in the tests!

SteffenSeckler:
closes #76:
examples are run with sanitizers.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -21,7 +21,7 @@ pipeline{
                         container('autopas-gcc7-cmake-make') {
                             dir("build"){
                                 sh "cmake .."
-                                sh "make -j 4 2>&1 > buildlog.txt || (cat buildlog.txt && exit 1)"
+                                sh "make -j 4 &> buildlog.txt || (cat buildlog.txt && exit 1)"
                             }
                         }
                     },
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -229,28 +225,126 @@ pipeline{
...
                         container('autopas-gcc7-cmake-make') {
                             dir("build/examples") {
                                 sh 'ctest -C checkExamples -j8'
                             }
                         }
+                    },
+                    "gcc openmp": {
+                        container('autopas-gcc7-cmake-make') {
+                            dir("build-openmp/examples") {
+                                sh 'ctest -C checkExamples -j8'
```
FG-TUM: why do you use -j 4 for building but -j 8 for testing?

SteffenSeckler: * for building: all builds are done on the same machine, thus -j4 is the maximum one should do.
* for testing: here the -j doesn't have much influence, it just says how many tests can be run in parallel. We could probably reduce this number, to reduce the number of parallel things a bit, but it shouldn't change much.

PR #92, opened 2018-09-20
Faster update container

### LINKED ISSUE #26: LinkedCells::updateContainer() can duplicate or remove particles if halo particles are present
LinkedCells::updateContainer() should only be called if all halo particles are DELETED.
There should be a run-time check for that, as it is not to expensive to check (still on the high-level).

problem:
if we don't do that particles can be duplicated, or lost, as they can no longer be differentiated:
```
addParticle(p1);
addHaloParticle(p2);
p1->setR(); // set the position of the normal particle to somewhere outside of the domain
updateContainer();
// now both particle and haloParticle are outside of the domain
```

possible implementation:
* bool that indicates whether halos are present
* if haloparticle is added, this bool is set to true
* if all haloparticles are deleted, this bool is set to false

other possible solution (more complicated!):
* don't store halo and leaving particles together

### LINKED ISSUE #72: updateContainer
`LinkedCells::updateContainer()` currently deletes all particles and adds them again.

Do not delete particles, only move those which crossed cell borders.

Use Functors?!

FG-TUM:
Faster and more parallel implementation of `LinkedCells::updateContainer()`

- [x] locking mechanism for `FullParticleCell::addParticle()`
- [x] full parallel implementation of `LinkedCells::updateContainer()`
- [x] unit tests for `LinkedCells::updateContainer()`
- [x] resolves #72 
- [x] resolves #26  

Tested on my machine (i7-7700 8x3.6GHz)
- 1,000,000 Particles (uniform random distribution)
- 52x 52 x 52 Cells
- random = particle jumps to a complete random location in the domain
- realistic = particle jumps to a random neighbor cell
- omitting locks (= allowing data races) has no significant impact

Particle relocations [ms] |  no | random 50%  | random 100% | realistic 5% | realistic 10%
------------ | ------------- | ------------- | ------------- | ------------- | -------------
old                         | 180 | 285 | 285 | 185 | 190
lock everything      |   30 | 240 | 250 |   42 | 50
add-barrier-delete |   35 | 250 | 288 |   47 | 58

Versions can be compared in 2330c573c0dde6b5da3b1a1eec670edf915ccf0f

### DIFF
--- a/src/autopas/containers/LinkedCells.h
+++ b/src/autopas/containers/LinkedCells.h
@@ -113,89 +113,6 @@ class LinkedCells : public ParticleContainer<Particle, ParticleCell, SoAArraysTy
     extractSoAs(f);
   }
 
-/**
- * selector for implementation of updateContainer.
- * 1 = old version       : delete and add all particles
- * 2 = lock everything   : fully parallel check all particles and delete/add all with locking cells
- * 3 = add-barrier-delete: first check all particles fully parallel and delete invalid then after a barrier add with
- * locks
- */
-#define variant 2
-
-#if variant == 1
-  // old variant
-  void updateContainer() override {
-    auto haloIter = this->begin(IteratorBehavior::haloOnly);
-    if (haloIter.isValid()) {
-      utils::ExceptionHandler::exception(
-          "Linked Cells: Halo particles still present when updateContainer was called. First particle found:\n" +
-          haloIter->toString());
-    }
-    std::vector<Particle> invalidParticles;
-// custom reduction with templates not supported
-//#pragma omp parallel reduction(vecMerge: invalidParticles)
-#ifdef AUTOPAS_OPENMP
-#pragma omp parallel
-#endif  // AUTOPAS_OPENMP
-    {
-      std::vector<Particle> myInvalidParticles;
-      for (auto iter = this->begin(); iter.isValid(); ++iter) {
-        myInvalidParticles.push_back(*iter);
-      }
-#ifdef AUTOPAS_OPENMP
-#pragma omp critical
-#endif  // AUTOPAS_OPENMP
-      invalidParticles.insert(invalidParticles.end(), myInvalidParticles.begin(), myInvalidParticles.end());
-    }
-    for (auto &cell : this->_cells) {
-      cell.clear();
-    }
-    for (auto &particle : invalidParticles) {
-      if (inBox(particle.getR(), this->getBoxMin(), this->getBoxMax())) {
-        addParticle(particle);
-      } else {
-        addHaloParticle(particle);
-      }
-    }
-  }
-
-#elif variant == 2
-
-  // new with locking
-  void updateContainer() override {
-    auto haloIter = this->begin(IteratorBehavior::haloOnly);
-    if (haloIter.isValid()) {
-      utils::ExceptionHandler::exception(
-          "Linked Cells: Halo particles still present when updateContainer was called. First particle found:\n" +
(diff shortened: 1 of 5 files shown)


src/autopas/containers/LinkedCells.h
```diff
@@ -112,31 +112,33 @@ class LinkedCells : public ParticleContainer<Particle, ParticleCell, SoAArraysTy
...
-    for (auto &particle : invalidParticles) {
-      if (inBox(particle.getR(), this->getBoxMin(), this->getBoxMax())) {
-        addParticle(particle);
-      } else {
-        addHaloParticle(particle);
+    for (size_t cellId = 0; cellId < this->getCells().size(); ++cellId) {
+      // if empty
+      if (not this->getCells()[cellId].isNotEmpty()) continue;
+
+      std::array<double, 3> cellLoweCorner, cellUpperCorner;
```
SteffenSeckler: LoweR

src/autopas/containers/LinkedCells.h
```diff
@@ -112,31 +112,33 @@ class LinkedCells : public ParticleContainer<Particle, ParticleCell, SoAArraysTy
...
-//#pragma omp parallel reduction(vecMerge: invalidParticles)
-#ifdef AUTOPAS_OPENMP
-#pragma omp parallel
-#endif  // AUTOPAS_OPENMP
-    {
-      std::vector<Particle> myInvalidParticles;
-      for (auto iter = this->begin(); iter.isValid(); ++iter) {
-        myInvalidParticles.push_back(*iter);
-      }
+    auto haloIter = this->getRegionIterator(this->_cellBlock.getHaloBoxMin(), this->_cellBlock.getHaloBoxMax(),
```
SteffenSeckler: should be done with the normal iterator (+ haloCellsOnly) and not with the RegionIterator.
you will anyways iterate through all halo Cells, which is exactly what the normal iterator will do.

src/autopas/containers/LinkedCells.h
```diff
@@ -112,31 +112,33 @@ class LinkedCells : public ParticleContainer<Particle, ParticleCell, SoAArraysTy
...
+
+      for (auto &&pIter = this->getCells()[cellId].begin(); pIter.isValid(); ++pIter) {
+        // if not in cell
+        if (notInBox(pIter->getR(), cellLoweCorner, cellUpperCorner)) {
+          // if not in halo
+          if (inBox(pIter->getR(), this->getBoxMin(), this->getBoxMax()))
+            addParticle(*pIter);
+          else
+            addHaloParticle(*pIter);
+          pIter.deleteCurrentParticle();
```
SteffenSeckler: nope :P
deleteCurrentParticle, pIter.isValid() (needs Cell::getNumParticles() ) etc. are not thread safe.

SteffenSeckler: alternatives:
1. make everything thread safe
2. safe to temporary vector -> add afterwards (with or without parallelization)

FG-TUM: both alternatives implemented and compared. See text at the top.

src/autopas/containers/LinkedCells.h
FG-TUM: fixed 1084179

src/autopas/containers/LinkedCells.h
FG-TUM: fixed d7c84c0f2d066bc59baaa1a2f87ec78a83c140bb

Issue #95, opened 2018-09-25
LinkedCells::updateContainer() not reordering correctly


SteffenSeckler:
**Describe the bug**
updateContainer() does not properly reorder the particles!
See branch feature/updateContainerIssue

**To Reproduce**
1. run the new unit test "LinkedCellsTest.testUpdateContainerCloseToBoundary"
2. it crashes

**Expected behavior**
the unit test does not crash

SteffenSeckler:
testUpdateContainerCloseToBoundary

PR #96, opened 2018-09-25
adds unit test for LinkedCells' and DirectSum's UpdateContainer  very close to boundary

SteffenSeckler:
ref #95 
ref #97


SteffenSeckler:
fixes #97

SteffenSeckler:
fixes #95

PR #151, opened 2019-01-11
Make autopas use a proper lock class

### LINKED ISSUE #132: encapsulate omp_lock_t in c++ lock style class
omp_lock_t cannot be copied, so extra copy operators etc. are needed wherever omp_lock_t is used. See #131

Encapsulate it in some proper class. I.e. copyable, ...

Leave as is. Meaning: wherever omp_lock_t is used, the class using it has to have custom copy operators, etc. So lots of code duplication.

SteffenSeckler:
# Description

Replaces autopas_lock_t (wrapper for omp_lock_t if AutoPas is compiled with OpenMP // to some dummy lock if AutoPas is compiled without OpenMP) with a proper lock class.
This makes copying of objects that have a lock straightforward and is no longer a possible problem.

## Resolved Issues # (issue)

- [x] fixes #132

### DIFF
--- a/src/autopas/utils/WrapOpenMP.h
+++ b/src/autopas/utils/WrapOpenMP.h
@@ -116,24 +116,24 @@ inline int autopas_get_num_threads() { return 1; }
 inline int autopas_get_max_threads() { return 1; }
 
 /**
- * AutoPasLock for the sequential case, that uses an enum for checking the base functionality.
+ * AutoPasLock for the sequential case.
  */
 class AutoPasLock {
  public:
   /**
    * Default constructor
    */
-  AutoPasLock() { _lock = unlocked; }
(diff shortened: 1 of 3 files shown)


src/autopas/utils/WrapOpenMP.h
```diff
@@ -100,44 +116,56 @@ inline int autopas_get_num_threads() { return 1; }
...
+   * Release the lock.
+   */
+  void unlock() {
+    assert(_lock == locked);
+    _lock = unlocked;
+  }
+
+ private:
+  // lock: 0 means unlocked, 1 locked.
+  enum { unlocked, locked } _lock;
```
FG-TUM: do you even bool?

PR #127, opened 2018-11-28
adds calculation of global values to ljfunctor's soa functor

SteffenSeckler:
# Description

adds calculation of global values to ljfunctor's soa functor

## Related Pull Requests

- #102 
- #113 

## Resolved Issues # (issue)

- [x] #100 (SoA part, not yet verlet)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes.

- [x] Unit tests
- [x] Using ls1

PR #160, opened 2019-01-18
Global value calculation for verlet lists SoA functor

### LINKED ISSUE #100: Provide possibility to calculate global values in force functors
The calculation of the potential energy or the virial (needed to calculate the pressure) is a pairwise interaction and is often done in the main kernel call, i.e. within the pairwise functors.
For the correct calculation of these global values, you have to make certain that contributions are not summed up multiple times. This is currently not easily possible.

Provide an easy way to calculate these values correctly.
Multiple options exist:
1. Multiply the global results by 1/2. This is possible only if no newton3 optimization is performed. (also only for full-shell if boundary evaluations are not calculated more than 2 times)
2. Provide a boolean to the functor calls that specifies whether the current interaction should contribute to the global values. This can be done 
   1. cell-pair wise:
   Problems for: direct-sum + verlet-lists
   2. particle-pair wise:
   Problem: relatively expensive
3. Use the eighth shell traversal + decomposition scheme.
   Problems for: direct-sum.
   Problem: a force-exchange needs to be implemented by the user.

**The problems**
- Direct-Sum(2.i. + 3.): Here the problem arises because only 1 halo cell exists. For the correct handling of the outer particle pairs (cell-wise), 26 halo cells are needed.
- Verlet-Lists(2.ii.): Once the verlet-lists are generated, it becomes ambiguous to differentiate particle-pair-interactions that contribute to the global values and those that don't. A speedy implementation probably needs a separate verlet list storage for interactions that don't contribute to the global values.

**Edit**:
**Current Solution**
Do it functor based.

**Implementation State for the LJFunctor**
- [x] LJFunctor: AoS noomp -- #102 
- [x] LJFunctor: AoS omp -- #113 
- [x] LJFunctor: SoA noomp -- #127
- [x] LJFunctor: SoA omp -- #127
- [ ] LJ Functor: Verlet-Lists SoA

SteffenSeckler:
# Description

introduces global value calculation for verlet lists SoA functor.

## Related Pull Requests

- #102 
- #113 
- #127

## Resolved Issues # (issue)

- [x] fixes #100

# How Has This Been Tested?

- [x] testSoAFunctorGlobalsVerlet
- [x] tested using ls1's autopas integration


tests/testAutopas/tests/pairwiseFunctors/LJFunctorTest.cpp
```diff
@@ -114,15 +124,16 @@ void LJFunctorTest::testSoANoGlobals(bool newton3, CellInteractionType interacti
...
       break;
-    case CellInteractionType::pair:
+    case InteractionType::pair:
       f2 = cell2.begin()->getF();
       break;
   }
   // if the interactiontype is own, then the forces of the second particle should always be calculated!
-  if (newton3 or interactionType == CellInteractionType::own) {
+  if (newton3 or interactionType != InteractionType::pair) {
     EXPECT_NEAR(f2[0], -expectedForce[0], absDelta);
```
FG-TUM: why not EXPECT_DOUBLE_EQ ?

src/autopas/pairwiseFunctors/LJFunctor.h
```diff
@@ -653,12 +513,265 @@ class LJFunctor : public Functor<Particle, ParticleCell, typename Particle::SoAA
...
+      // this is a magic number, that should correspond to at least
+      // vectorization width*N have testet multiple sizes:
+      // 4: does not give a speedup, slower than original AoSFunctor
+      // 8: small speedup compared to AoS
+      // 12: highest speedup compared to Aos
+      // 16: smaller speedup
+      // in theory this is a variable, we could auto-tune over...
+#ifdef __AVX512F__
+      // use a multiple of 8 for avx
+      const size_t vecsize = 16;
```
FG-TUM: this should probably a constexpr for the whole class

tests/testAutopas/tests/pairwiseFunctors/LJFunctorTest.cpp
SteffenSeckler: 4ULP's might not be enough

src/autopas/pairwiseFunctors/LJFunctor.h
SteffenSeckler: changed to constexpr, but left where it is.

Issue #144, opened 2019-01-07
md-flex allow selection of newton3


SteffenSeckler:
**Is your feature request related to a problem? Please describe.**
Currently md-flex is not able to benchmark non-newton3 runs.
Especially needed to compare different things of new verlet-list implementations.

**Describe the solution you'd like**
Make it able to benchmark non-newton3 runs.

**Describe alternatives you've considered**
Benchmark / Check with sph examples

FG-TUM:
this was resolved by #166

PR #166, opened 2019-01-28
All verlet options for md flex

FG-TUM:
# Description

Adapted md-flex and the measurePerformance script to benchmark all containers.

- [x] md-flex supports newton 3 on/off
- [x] md-flex default behaviour is now to use all containers and traversals

~## Related Pull Requests~

## Resolved Issues

#144 

# How Has This Been Tested?

- [x] ran the script and it worked....
- [x] added the script to tests


examples/md-flexible/main.cpp
```diff
@@ -285,6 +293,8 @@ int main(int argc, char **argv) {
   if (numIterations > 0)
```
SteffenSeckler: { }

FG-TUM: 0eb1489

Issue #149, opened 2019-01-10
Update Timer from ctime to chrono


SteffenSeckler:
**Describe the bug**
The timer class is currently using ctime.

**To Reproduce**
1. Look into Timer.cpp

**Expected behavior**
Use something more appropriate. Best: std::chrono

PR #181, opened 2019-03-29
Update Timer from ctime to std::chrono

Garfield96:
# Description

Update Timer from ctime to std::chrono

# Resolved Issues: #149 

# How Has This Been Tested?
tested with TimerTest

### DIFF
--- a/src/autopas/utils/Timer.cpp
+++ b/src/autopas/utils/Timer.cpp
@@ -7,31 +7,18 @@
 #include "utils/Timer.h"
 #include <iostream>
 
-using namespace std;
+using namespace std::chrono;
 
-autopas::utils::Timer::Timer() : _startTime{} {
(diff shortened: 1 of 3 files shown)


src/autopas/utils/Timer.cpp
```diff
@@ -7,31 +7,18 @@
...
-using namespace std;
+using namespace std::chrono;
 
-autopas::utils::Timer::Timer() : _startTime{} {
-  struct timespec info {};
-  if (clock_getres(CLOCK_REALTIME, &info)) {
-    std::cout << "Could not retrieve time resolution!" << endl;
-  }
-}
+autopas::utils::Timer::Timer() : _startTime{} {}
```
SteffenSeckler:
```suggestion
autopas::utils::Timer::Timer() = default;
```

PR #194, opened 2019-05-02
Cpp17

FG-TUM:
# Description

bumped c++ version from 14 to 17

## ~Resolved Issues # (issue)~

# How Has This Been Tested?

see Jenkins results
- [ ] tested mardyn integration


FG-TUM:
@SteffenSeckler The CUDA compiler Jenkins uses does not support cuda + c++17

SteffenSeckler:
well as clang is currently not properly supported in cmake as a cuda compiler, we will probably have to hold off with c++17.
See also: https://gitlab.kitware.com/cmake/cmake/issues/16586

src/autopas/CMakeLists.txt
```diff
@@ -13,6 +13,8 @@ ADD_LIBRARY(autopas STATIC
         $<$<BOOL:${ENABLE_CUDA}>:${CU_SRC}>
         )
 
+target_compile_features(autopas PRIVATE cxx_std_17)
```
SteffenSeckler: should definitely NOT be private!

SteffenSeckler: but public

FG-TUM:
Closed because currently not fixable

PR #201, opened 2019-05-04
Allow cell sizes < 1.0 for linked cells, refactoring

Garfield96:
# Description

1. support for cell size factors < 1.0 for all traversals of LinkedCells. 
2. The existing tests for the LinkedCells Traversals were replaced by new automatically generated tests (TraversalTest).
3. refactoring: Basic traversal iteration for traversals based on domain coloring moved to own class (CBasedTraversal),  replaced `#define AUTOPAS_WITH_STATIC_CELL_ITER` by template function

## Related Pull Requests

- 

# How Has This Been Tested?

- [x] Jenkins
- [x] TestTraversal tests all LinkedCells Traversals with Newton 3 enabled / disabled


FG-TUM:
Does this stuff work if particles travel further than one cell per time step? :)

Garfield96:
> Does this stuff work if particles travel further than one cell per time step? :)

The only critical step I can think of is the update process, which put the particles into their new cells. But since the current method doesn't take neighbourhood relationships into account, it doesn't matter whether a particles travel 1, 2 or more cells. Other than that, there shouldn't be problem.

src/autopas/containers/linkedCells/traversals/C18Traversal.h
```diff
@@ -87,26 +92,48 @@ class C18Traversal : public C18BasedTraversal<ParticleCell, PairwiseFunctor, Dat
...
+   * Pairs for processBaseCell(). overlap[0] x overlap[1] Array for each special case in x and y direction.
    */
-  std::array<std::array<std::vector<unsigned long>, 3>, 3> _cellOffsets;
+  std::vector<std::vector<std::vector<unsigned long>>> _cellOffsets;
+
+  /**
+   * returns the array index for current position pos.
+   * @param pos current position in dimension dim
+   * @param dim current dimension
+   * @return
```
FG-TUM:
```suggestion
   * @return Index for the _cellOffsets Array.
```

FG-TUM:
ready to merge

Issue #214, opened 2019-05-12
Global values for verlet lists probably broken


SteffenSeckler:
**Describe the bug**
Assume two particles close to the inner boundary. 
If both particles move over the domain boundaries while the lists are not rebuild, the interaction between them will most likely not be counted for the global values, even though it should. 

**To Reproduce**
1.Think
2. Write test

PR #209, opened 2019-05-10
Test and Fix Container Conversion

SteffenSeckler:
# Description

This PR introduces tests and fixes for the conversion between containers

- [x] Test for the conversion between all containers
- [x] Fixes container conversion
- [x] Linked Cells will now simply ignore particles that should be added in the halo if they are outside of the halo (fixes conversion of verlet type container to LC type container.

## Resolved Issues # (issue)

- [x] #208 (might still be broken for VerletClusterLists)

## TODO:
- [ ] #155 (will come later)

## Additional context

- ref #155 

# How Has This Been Tested?

- [x] New test for container conversion

Issue #208, opened 2019-05-10
Conversion to VerletClusterLists can fail


denizcandas:
**Describe the bug**
When you call md-flexible with --particles-per-dimension=x, where x is not a multiple of 4, the experiment throws the following exception:
terminate called after throwing an instance of 'autopas::utils::ExceptionHandler::AutoPasException'
  what():  LinkedCells: trying to add halo particle that is not in the halo box.

**To Reproduce**
1. compile the program in release mode with openmp enabled
2. call md-flexible with --particles-per-dimension=x, where x!= 0 (mod 4) (edit: also x >4)
3. See error

**Expected behavior**
The experiment finishes without an exception

**Screenshots**
![image](https://user-images.githubusercontent.com/16650999/57516263-b08bc180-7314-11e9-929a-87d86fa65270.png)

**Build environment (please complete the following information):**
 - OS: Linux
 - Compiler: gcc 7.3.0
 - Build System: cmake 3.10.2
- Mpp2 Cluster

SteffenSeckler:
thanks for reporting this.
We are currently working on a proper test to check this behavior. (See branch https://github.com/AutoPas/AutoPas/tree/feature/containerConversion)

For now you can disable the verlet list containers, especially the VerletClusterLists container

SteffenSeckler:
fixed in #209 (except for

SteffenSeckler:
fixed as of #430

Issue #207, opened 2019-05-09
Test all containers against reference container


SteffenSeckler:
**Is your feature request related to a problem? Please describe.**
Currently, tests are duplicated a lot and some tests are missing for some containers. 

**Describe the solution you'd like**
There should be a test that tests all possible containers with all possible options for correctness.

SteffenSeckler:
verlet cluster lists is missing

FG-TUM:
@SteffenSeckler I think this can be closed since we test the interface functionality of all containers in `AllContainersTests.cpp` and compare the results of`iteratePairwise()` for all containers in `TraversalComparison.cpp`

SteffenSeckler:
I think so, too.

PR #215, opened 2019-05-13
[WIP] AutoPas interface refactoring

SteffenSeckler:
# Description

Introduce a proper interface for AutoPas to enable one common interface for all containers.

Change list:

- [x] positions of particles no longer completely define the owned state
- [x] LinkedCells: deleteHaloParticles can also delete particles from cells inside the domain
- [x] updateContainer does not necessarily update the container, e.g., for verlet lists it can ignore this to not destroy the container. 
   - [x] there exists a method to force the rebuild
- [x] updateContainer now checks for invalid particles, i.e. particles which do no longer belong into the container (this property depends on the container type), invalid particles will be returned as a vector. #217  
- [x] On a proper rebuild of a container the halo particles are removed first.
- ~Verlet-Type container: addParticle is actually an addOrUpdateParticle - if it knows the particle (and it is a halo particle), it will change it to an owned particle. If it is already owned it throws an exception. if it is not known (e.g. for real particle insertion) it will invalidate the lists.~
  -  ~document this properly~
- [x] Verlet-Type container (maybe all containers): addHaloParticle is actually an addOrUpdateHaloParticle - if it knows the halo particle, it will update its state (overwrite it). if it is not known (e.g. for real particle insertion) it will invalidate the lists and add it (if the neighbor lists are invalid) or ignore the halo particle if the neighbor lists are valid and it is not too close (i.e. within a distance of `cutoff radius + skin/2` from the domain). 
  - [x] document this properly
- [x] Rewrite sph examples for new interface + AutoPas classes
- [x] Potentially let non-verlet type containers know about skin to be able to check halo particle insertion.
- [x] Fix functors to use bool isOwned instead of check for position.

## Related Pull Requests + Issues

- #206 
- #207 
- #208 
- #209
- #214 

## subtasks
- #217 
- #220 
- #223 
- #224 

## Resolved Issues # (issue)

- [ ] #Issue

## Todo
- [x] New return type of updateContainer has to be implemented for all containers
- [x] Enable halo addition for LinkedCell container (ContainerSelectorTest.cpp)
- [x] Test with two autopas containers
  - [x] Tests of interfaces need to check all options (newton3, ...)
  - [ ] Let tests run longer to check conversion.
- [ ] Think about renaming `updateContainer()`
- [ ] Think about renaming halo particles to sth like not-owned particles
- [ ] update documentation
- [ ] ~parallelize VerletClusterLists::updateContainer()~ -- postponed issue: 
- [ ] check ///@ todo
- [ ] ~optimize deleteHaloParticles() funtions, or rather the iterators.~ -- postponed issue:
- [ ] ~Fix VerletClusterLists -- will probably go to different PR~ -- postponed issue: 
- [x] remove deleteHaloParticles calls (only possible once addHaloParticle can update already existing particles) from sph examples 
   - [ ] and check results again
- [x] Bug still existing as of 25.6.2019, 13:00: enforcing a rebuild of the container + lists (after enough samples have been taken) also has to enforce an updateContainer call before this.
- [x] adapt gpu version
- [ ] Document that using two functors one supporting a N3 option the other does not is not possible.

# How Has This Been Tested?

- [x] Test for particle addition and deletion
- [x] Test for simulation loop with one single container
- [x] Test for simulation loop with two containers


tests/testAutopas/tests/autopasInterface/AutoPasInterfaceTest.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+void addHaloParticles(autopas::AutoPas<Molecule, FMCell>& autoPas, std::vector<Molecule> haloParticles) {
+  for (auto& p : haloParticles) {
+    autoPas.addHaloParticle(p);
+  }
+}
+
+template <typename Functor>
+void doSimulationLoop(autopas::AutoPas<Molecule, FMCell>& autoPas, Functor* functor) {
+  // 1. update Container; return value is vector of invalid = leaving particles!
```
FG-TUM:
```suggestion
  // 1. update Container; return value is vector of invalid == leaving particles!
```

tests/testAutopas/tests/autopasInterface/AutoPasInterfaceTest.cpp
```diff
@@ -0,0 +1,154 @@
...
+void defaultInit(AutoPasT& autoPas) {
+  autoPas.setBoxMin(boxMin);
+  autoPas.setBoxMax(boxMax);
+  autoPas.setCutoff(cutoff);
+  autoPas.setVerletSkin(skin);
+  // init autopas
+  autoPas.init();
+}
+
+std::vector<Molecule> convertToEnteringParticles(const std::vector<Molecule>& leavingParticles) {
```
FG-TUM: document this

tests/testAutopas/tests/autopasInterface/AutoPasInterfaceTest.cpp
```diff
@@ -0,0 +1,154 @@
...
+        // should at least be boxMin
+        pos[dim] = std::max(boxMin[dim], pos[dim] - (boxMax[dim] - boxMin[dim]));
+      }
+    }
+    p.setR(pos);
+  }
+  return enteringParticles;
+}
+
+auto identifyAndSendHaloParticles(autopas::AutoPas<Molecule, FMCell>& autoPas) {
```
FG-TUM: document this

FG-TUM:
Before the final review, we should think about renaming `updateContainer()`

PR #186, opened 2019-04-16
Allow cell sizes > 1.0 for linked cells, argument validation, refactoring

Garfield96:
# Description

1. Allow bigger cell sizes (>1.0) for linked cells container. Cell size option was added to md-flexible as well.
2. Cutoff and cell size are now checked in setters of autopas.h to avoid invalid values.
3. CMake: Copy script files to build dir to avoid manual copying.
4. Refactoring of testing helpers.

# How Has This Been Tested?

1. Tests were added to CellBlock3dTest to test new cell sizes.
3. AutoPasTest.cpp tests whether invalid arguments causes an exception.

Issue #90, opened 2018-09-06
Further (possible) AutoTuning Parameters


SteffenSeckler:
This is a collection of further possible auto-tuning parameters:
- [ ] Traversals: chunk size of OpenMP schedule dynamic. This is currently set to 1, but with many threads, this can be problematic (especially if the particle density is low)
- [ ] Cell Size: Especially if vectorization is enabled and the kernel is relatively inexpensive, it could be advantageous to increase the Cell size to get better performance (since we have fewer cells). This mainly affects the linked-cells, but can also be used for the Verlet-list generation.

SteffenSeckler:
some done via #170

FG-TUM:
Verlet Cluster size

SteffenSeckler:
@FG-TUM can you assign tuning of the verletclustersize to someone?

FG-TUM:
Semi-likely. @HobbyProgrammer might be a sensible follow-up investigation after the investigation into the cell stuff?

PR #218, opened 2019-05-14
Cell size factor for cell based verlet lists

Garfield96:
# Description

Introduce cell size factor for cell based verlet lists.

## Related Pull Requests

- #201, #186 

## Resolved Issues

- #90 

# How Has This Been Tested?

- [x] new test cases for bigger cell sizes


src/autopas/containers/verletListsCellBased/VerletListsLinkedBase.h
```diff
@@ -38,17 +38,22 @@ class VerletListsLinkedBase : public ParticleContainer<Particle, FullParticleCel
    * neighbor lists are to be rebuild. A frequency of 1 means that they are
    * always rebuild, 10 means they are rebuild after 10 traversals
    * @param applicableTraversals all applicable traversals
+   * @param cellSizeFactor cell size factor ralative to cutoff
```
FG-TUM: Document here that 1 is an enforced lower boundary

PR #227, opened 2019-05-28
Refactored TraversalSelector to remove a lot of unnecessary dependencies

HobbyProgrammer:
# Description
Refactored TraversalSelector: Extracted TraversalSelectorInfo, which represents the container specific data, and made generateTraversal() method static.

Before, all containers included TraversalSelector which itself included all of the traversals. Now, these dependencies are reduced to the necessary ones.

This refactoring was probably the simplest one as other ideas would have been problematic because of the template parameters of generateTraversal().

# How Has This Been Tested?

With all the tests from before.

### DIFF
--- a/src/autopas/selectors/TraversalSelector.h
+++ b/src/autopas/selectors/TraversalSelector.h
@@ -39,8 +39,6 @@ namespace autopas {
  */
 template <class ParticleCell>
 class TraversalSelector {
-  friend TraversalSelectorInfo<ParticleCell>;
-
  public:
   /**
    * Generates a given Traversal for the given properties.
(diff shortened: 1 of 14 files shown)


HobbyProgrammer:
This refactoring was easy, so just 90 minutes of work. I think it is flexible enough and reduces the dependencies, although it might not be pretty from a software design point of view. The template parameters from `generateTraversal()` make nice object oriented design a lot harder, though.

Since `TraversalSelector` now only contains one static method that "generates", not "selects" a traversal, it might make sense to rename it. Please share your suggestions, if you think that would be good.

src/autopas/selectors/TraversalSelector.h
```diff
@@ -38,21 +39,9 @@ namespace autopas {
...
-  /**
-   * Constructor of the TraversalSelector class.
-   * @param dims Array with the dimension lengths of the domain
-   * @param cutoff Cutoff radius
-   * @param cellLength cell length.
-   */
-  TraversalSelector(const std::array<unsigned long, 3> &dims, const double cutoff = 1.0,
-                    const std::array<double, 3> &cellLength = {1.0, 1.0, 1.0})
-      : _dims(dims), _cutoff(cutoff), _cellLength(cellLength) {}
+  friend TraversalSelectorInfo<ParticleCell>;
```
FG-TUM: TraversalSelectorInfo params public. We dont linke friends.
-> remove underscores

HobbyProgrammer:
The GPU cloud tests got a timeout error after only adding some comments. The tests were successful before. What could be the reason for this? Can I restart the tests somehow?

FG-TUM:
You can trigger the tests by updating the branch with the master. Other than that only Steffen and I can do that but we'll do this if we see problems like these.

PR #230, opened 2019-05-30
CCache

NeoTheThird:
# Description

This looks for ccache in `$PATH` and uses it, unless `CCACHE_NO` is specified. Obviously this is not yet used in CI, because the images don't have ccache installed. I filed a PR for that [here](https://github.com/AutoPas/AutoPas-Dockerfiles/pull/1).

@SteffenSeckler I'm not sure how this will work with the Kubernetes setup we have. Ccache uses `$HOME/.ccache` by default, i guess that is volatile, right? What's the best way to do this? Volumes? One shared volume? Something else?

Are there cases where it's definitely a bad idea to use ccache?

## Related Pull Requests

- https://github.com/AutoPas/AutoPas-Dockerfiles/pull/1

# How Has This Been Tested?

- GCC build
- Clang build


SteffenSeckler:
so the default way without using volumes would be to use artifacts. for volumes you would probably need to specify things like locks, ...

CMakeLists.txt
```diff
@@ -14,6 +14,20 @@ if (NOT CMAKE_CXX_COMPILER_ID IN_LIST SUPPORTED_COMPILERS)
     message(WARNING "Compiler " ${CMAKE_CXX_COMPILER_ID} " not supported!")
 endif ()
 
+# Use ccache if available, unless CCACHE_NO is specified
+if (CCACHE_NO)
```
SteffenSeckler: needs an Option to enable, disable ccache. this should also go into a module file

NeoTheThird: Okay, so you don't want it to be enabled by default?

NeoTheThird: AHHH, you mean a cmake option, gotcha. Not an env var. Sure thing.

SteffenSeckler: that as well. but in general it should have an option tag to show up in ccmake

SteffenSeckler: yup cmake option :)

NeoTheThird:
> artifacts

Ok, on it

NeoTheThird:
I'm not so sure if artifacts are really the cleanest way to do it, as the problem with race conditions only gets bigger here (that can be avoided by name-spacing as well, but it reduces cache hits on the first build, uses a lot of space and needs a lot of extra code in the pipeline script). From what i gather from [the docs](https://ccache.dev/manual/3.4.1.html#_sharing_a_cache), ccache should be smart enough to account for different branches, compilers and flags, even if multiple people write to the same dir simultaneously.

> "Another reason to use ccache is that the same cache is used for builds in different directories. If you have several versions or branches of a software stored in different directories, many of the object files in a build directory can probably be taken from the cache even if they were compiled for another version or branch.
> A third scenario is using ccache to speed up clean builds performed by servers or build farms that regularly check that the code is buildable.
> You can also share the cache between users, which can be very useful on shared compilation servers." (https://ccache.dev/)

Or do you think that's too risky/error prone?

SteffenSeckler:
ok, then you would like to use a shared volume, correct?

cmake/modules/ccache.cmake
```diff
@@ -0,0 +1,13 @@
+option(CCACHE "Use ccache for accellerated compilation through caching" ON)
```
SteffenSeckler: probs disabled by default would be good :)

SteffenSeckler:
i'll probably be able to look into storages in ~3 weeks

NeoTheThird:
> i'll probably be able to look into storages in ~3 weeks

Ok, should it wait for that or continue the artifact approach?

NeoTheThird:
If you could excuse me one naive question, just clicking "Add volume" in the kubernetes config section of jenkins would not do the trick? Is space very limited on our openshift?

SteffenSeckler:
hmm that all works a bit differently. For our kubernetes cloud we did not yet configure a proper network storage that all nodes can access. And without that the add volumes won't do anything :)

NeoTheThird:
Ok, should it wait for that or continue the artifact approach?

SteffenSeckler:
waiting seems more reasonable for now

cmake/modules/ccache.cmake
```diff
@@ -0,0 +1,13 @@
+option(CCACHE "Use ccache for accellerated compilation through caching" OFF)
+if (CCACHE)
+    find_program(CCACHE_FOUND ccache)
+    if (CCACHE_FOUND)
+        message(STATUS "ccache found, building with ccache")
+        set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
+        set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK ccache)
```
SteffenSeckler:
```suggestion
```

cmake/modules/ccache.cmake
```diff
@@ -0,0 +1,13 @@
+option(CCACHE "Use ccache for accellerated compilation through caching" OFF)
+if (CCACHE)
+    find_program(CCACHE_FOUND ccache)
+    if (CCACHE_FOUND)
+        message(STATUS "ccache found, building with ccache")
+        set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
```
SteffenSeckler:
```suggestion
        set(CMAKE_CXX_COMPILER_LAUNCHER ${CCACHE_FOUND} CACHE STRING "ccache executable")
```

FG-TUM:
Is this already merge-able? (apart from updating with master)

SteffenSeckler:
yup

SteffenSeckler:
also merged clang and cmake format, as they both modifyfiles and git diff could be influenced by each other.

PR #242, opened 2019-06-13
Lock resource to avoid weird kubernetes errors

NeoTheThird:
# Description

This limits the number of parallel builds jenkins performs by acquiring a lock on the `openshift_cluster` resource, of which there's currently two. This is to avoid weird circular wait conditions when many builds are triggered at the same time. It happened a lot when master was updated, because all PRs would then be triggered at the same time and request a lot of default containers where they would build and then sit idly for hours waiting for the GPU container of which we can only have one at a time, resulting in timeouts and weird behavior.

# How Has This Been Tested?

- [ ] Jenkins build

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,7 @@
 pipeline{
+    options {
+        lock(quantity: 1, resource: 'openshift_cluster')
+    }
     agent {
         node {
             //cloud 'kubernetes'


NeoTheThird:
Okay, that's actually not how this works, so as of now this would limit it to only one parallel build. Have to enquire if that's even possible with this approach.

SteffenSeckler:
should rather be done using the jenkins config.

SteffenSeckler:
limiting it to 3 for now.

PR #244, opened 2019-06-14
MemProfiler Cpp style

FG-TUM:
# Description

Rewrote Memory profiler with Cpp style file access.

## Resolved Issues # (issue)

The previous implementation leaked file descriptors which could result in segfaults.

# How Has This Been Tested?

- [x] md-flexible with > 2000 Iterations

### DIFF
--- a/src/autopas/utils/MemoryProfiler.cpp
+++ b/src/autopas/utils/MemoryProfiler.cpp
@@ -6,7 +6,7 @@
 
 #include "MemoryProfiler.h"
 #include <sys/stat.h>
-#include <fstream>
+#include <ifstream>
 #include "Logger.h"
 
 size_t autopas::memoryProfiler::currentMemoryUsage() {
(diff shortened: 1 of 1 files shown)


src/autopas/utils/MemoryProfiler.cpp
```diff
@@ -5,38 +5,28 @@
  */
 
 #include "MemoryProfiler.h"
-#include "Logger.h"
-
-#include <fcntl.h>
 #include <sys/stat.h>
-#include <unistd.h>
-#include <cstring>
+#include <fstream>
```
SteffenSeckler:
```suggestion
#include <ifstream>
```
you are only using ifstream, so only include that.

src/autopas/utils/MemoryProfiler.cpp
```diff
@@ -5,38 +5,28 @@
  */
 
 #include "MemoryProfiler.h"
-#include "Logger.h"
```
SteffenSeckler: should still be there

src/autopas/utils/MemoryProfiler.cpp
FG-TUM: there is no `ifstream` header ^^

PR #245, opened 2019-06-14
make internal namespace inline to prevent usage from outside of autopas

SteffenSeckler:
# Description

Makes the namespace autopas::internal inline to prevent its usage from outside of autopas.


SteffenSeckler:
@FG-TUM maybe we should add more things to the internal namespace?

FG-TUM:
Ok. Like what?

PR #246, opened 2019-06-14
Bump CMake version

### LINKED ISSUE #184: Remove workaround in cmake scripts related to CUDA
CMake aborts when COMPILE_LANGUAGE:xx is used without enabling the language. This is 
circumvented by checking if Cuda is enabled before thus requiring code duplication  

Require CMake 3.12 or higher (Problem is fixed there)
Remove Cuda enabled check from other-compileroptions.cmake and vectorization.cmake

### LINKED ISSUE #231: Bump CMake Version
Higher CMake Version -> More features like target_compile_definitions for imported targets which is interesting for the mardyn integration.

Bump to 3.12

* Bump to 3.13 because of: 
https://github.com/AutoPas/AutoPas/blob/98bf2b17c8966a4bf34823da2f2cb0326d91f5c5/src/autopas/CMakeLists.txt#L24-L30

* Bump to latest because why not. Installing CMake is trivial.

**Additional context**
Smuc NG has 3.12
Hazel Hen has 3.13

Garfield96:
# Description

* bump CMake version to 3.13
* add `formatall` target to execute `cmakeformat` and `clangformat` with a single command
* require OpenMP version >= 4.5

## Resolved Issues

- [x] closes #231 
- [x] closes #184 

# How Has This Been Tested?

- [x] Jenkins


Garfield96:
@SteffenSeckler Can you please upgrade all compilers used in Jenkins to a version supporting at least OpenMP 4.5. OpenMP 5.0 would be truly awesome, especially for using tasks.

Garfield96:
Where should we place the OpenMP stuff? (policy already enforced with 29d2663) 
https://github.com/AutoPas/AutoPas/blob/98bf2b17c8966a4bf34823da2f2cb0326d91f5c5/src/autopas/CMakeLists.txt#L24-L30

SteffenSeckler:
the openmp cmake check is broken, see:
https://gitlab.kitware.com/cmake/cmake/issues/19150
and
https://bugs.llvm.org/show_bug.cgi?id=42298

eventually, one would also want to have cmake adapt to the bug described in https://bugs.llvm.org/show_bug.cgi?id=42298, but for me, it looks to be on the clang side.

SteffenSeckler:
lgtm file is certainly needed, as lgtm only uses cmake 3.12 by default.

SteffenSeckler:
i'm currently checking for the lgtm config

Garfield96:
@SteffenSeckler 
> finally got lgtm running :)

Thanks for that! Nice job.
> is there any need to change cmake to 3.14?

We need at least CMake 3.13 to enforce policy `CMP0079`. I still choose 3.14 since I expected some problems, like Jenkins and lgtm, even with 3.13. By using 3.14, we don't need to think about updating CMake again in the near future.
> as far as i remember there were still some todos regarding upgrading cmake to sth like 3.12.

There are no open todos in the code, but there is an issue (#184) regarding CMake 3.12 and CUDA. I'm working on that.

SteffenSeckler:
@ openmp detection: for clang we would need to specify `-fopenmp-version=45`
for it to correctly show in cmake, see:

https://bugs.llvm.org/show_bug.cgi?id=42298#c1

SteffenSeckler:
from what i see from the cmake changelog, 3.13 is definitely useful as it contains a couple of nice features (e. g. target link libraries from module files). but i don't see any changes in 3.14 or 3.15 that force us to use these versions. So I would suggest that we go with cmake 3.13 for now. We will anyways have to update the required cmake versions again at latest when for cuda+clang or c++20 modules is added

SteffenSeckler:
changed the lgtm s.t. only cmake v3.13.3 is used. to reenable 3.14, the following workaround might be needed:
```yml
extraction:
  cpp:
    prepare:    # Customizable step used by all languages.
      packages:
        - python-pip
    after_prepare:    # Customizable step used by all languages.
      - wget https://files.pythonhosted.org/packages/5c/e0/be401c003291b56efc55aeba6a80ab790d3d4cece2778288d65323009420/pip-19.1.1-py2.py3-none-any.whl
      - wget https://files.pythonhosted.org/packages/bb/10/44230dd6bf3563b8f227dbf344c908d412ad2ff48066476672f3a72e174e/wheel-0.33.4-py2.py3-none-any.whl
      - wget https://files.pythonhosted.org/packages/ec/51/f45cea425fd5cb0b0380f5b0f048ebc1da5b417e48d304838c02d6288a1e/setuptools-41.0.1-py2.py3-none-any.whl
      - pip install --user --upgrade *.whl
      - export PATH=~/.local/bin:$PATH
      - pip --version
      - wget https://files.pythonhosted.org/packages/09/bb/0121c0ae40144b99420a1a0ea081d48ac4d77332600896d7e04bf23aa957/cmake-3.14.4-py2-none-manylinux2010_x86_64.whl
      - pip install --user --upgrade cmake*.whl
      - ls ~/.local/bin
      - cmake --version
```

PR #259, opened 2019-06-19
Removed newton3 parameter from all iteratePairwise methods in the containers

HobbyProgrammer:
# Description
Removed newton 3 parameter from all iteratePairwise methods in the containers.

## Related Pull Requests

None

## Resolved Issues

None

# How Has This Been Tested?
Jenkins

PR #260, opened 2019-06-20
Removed all templates from iteratePairwise() and made it virtual

### LINKED ISSUE #229: TraversalSelector::generateTraversal() has CellPairTraversal as fixed return type, but not all traversals will be CellPairTraversals
`TraversalSelector::generateTraversal()` has `CellPairTraversal` as fixed return type, but not all traversals will be `CellPairTraversal`s. Especially the Verlet List traversals will work on neighbor lists, not cells.

`TraversalSelector::generateTraversal()` should not be limited to `CellPairTraversal`s, but allow all traversals deriving from `TraversalInterface`. This is not straightforward to implement though, as a lot of code depends on it. There are only CellPairTraverals at the moment.

Have all traversals derive from `CellPairTraversal`.

HobbyProgrammer:
# Description

Refactoring! While removing the newton3 parameter from `iteratePairwise`, I noticed that it might be possible to remove all template parameters. I consider this a lot better than now so just as Fabio told me, I tried it out. Here, I removed all templates from `iteratePairwise()`. It is now pure virtual in `ParticleContainerInterface`! 

The code now uses `TraversalInterface` instead of `CellPairTraversal` in places where all traversals are wanted. `TraversalInterface` now has virtual `initTraversal()`, `traverseParticlePairs()`, and `endTraversal()` methods. This solves #229 !

Removed data layout and newton 3 templates from CellPairTraversal during this refactoring.

VerletClusterLists don't support Traversals yet, so they just throw an exception in iteratePairwise() at the moment. I will refactor them to use traversals anyway, so as soon as that is there, I can solve this.

The container traversal interfaces like `LinkedCellTraversalInterface` are now mostly marker interfaces to mark which traversals are applicable, and sometimes to specify which information the traversals get (like the setters in the verlet traversal interfaces). The traversals themselves implement the `TraversalInterface`. 
I chose to do it like that (and not have the interfaces inherit from `TraversalInterface`) to avoid deadly diamonds of death that would occur for all `CellPairTraversal`s like the `C08Traversal`, because it inherits from both `CellPairTraversal` and `LinkedCellTraversalInterface`, and `CellPairTraversal` inherits from `TraversalInterface`.

Furthermore, there might be some problems in code inside `if defined (AUTOPAS_CUDA)` since I don't have that installed and can't try it out.

## Changelist:
- [x] removed templates `dataLayout` and `useNewton3` from `CellPairTraversal`.
gain -> ?
- [x] promoted `initTraversal()` `endTraversal()` to `TravesalInterface`
Removed their arguments. CellPairTraversal gets new `setCellsToTraverse()` to store a pointer to the cells.

## Related Pull Request

- #259 
- #262 

## Resolved Issues

- closes #229 

# How Has This Been Tested?

- Al the Unit Tests
- No tests for CUDA!! (Since I don't have it installed.)


src/autopas/containers/directSum/DirectSumTraversalInterface.h
```diff
@@ -8,18 +8,13 @@
...
-  /**
-   * Traverse all pairs of cells.
-   * This function needs to be implemented by derived classes and handles to
-   * order in which the cells are traversed.
-   * @param cells Vector of cells to traverse
-   */
-  virtual void traverseCellPairs(std::vector<ParticleCell> &cells) = 0;
-};
\ No newline at end of file
+class DirectSumTraversalInterface {};
```
FG-TUM: If there is nothing left in the Interface we might as well delete it

src/autopas/containers/linkedCells/traversals/LinkedCellTraversalInterface.h
```diff
@@ -15,15 +16,6 @@ namespace autopas {
...
- public:
-  /**
-   * Traverse all pairs of cells.
-   * This function needs to be implemented by derived classes and handles to
-   * order in which the cells are traversed.
-   * @param cells Vector of cells to traverse
-   */
-  virtual void traverseCellPairs(std::vector<ParticleCell> &cells) = 0;
-};
+class LinkedCellTraversalInterface {};
```
FG-TUM: If there is nothing left in the Interface we might as well delete it

src/autopas/containers/verletClusterLists/VerletClusterLists.h
```diff
@@ -63,17 +63,9 @@ class VerletClusterLists : public ParticleContainer<Particle, FullParticleCell<P
...
-   * @tparam the type of ParticleFunctor
-   * @tparam Traversal
-   * @param f functor that describes the pair-potential
-   * @param traversal not used
-   * @param useNewton3 whether newton 3 optimization should be used
-   */
-  template <class ParticleFunctor, class Traversal>
-  void iteratePairwise(ParticleFunctor *f, Traversal *traversal) {
+  void iteratePairwise(TraversalInterface *traversal) override {
+    autopas::utils::ExceptionHandler::exception("VerletClusterLists have no traversals yet here, so it does not work.");
```
FG-TUM: yea with this exception we cannot merge it into the master. So we probably want to merge #262 first and then remove this exception ;)

src/autopas/selectors/AutoTuner.h
```diff
@@ -320,10 +320,9 @@ void AutoTuner<Particle, ParticleCell>::iteratePairwiseTemplateHelper(PairwiseFu
   // if tuning execute with time measurements
   if (inTuningPhase) {
     auto start = std::chrono::high_resolution_clock::now();
-    // @todo remove useNewton3 in iteratePairwise by introducing traversals for DS and VL
 
     f->initTraversal();
-    withStaticContainerType(container, [&](auto container) { container->iteratePairwise(f, traversal.get()); });
+    container->iteratePairwise(traversal.get());
```
FG-TUM: I think if we do it this way `withStaticContainerType()` wouldn't be needed anymore correct?

src/autopas/containers/linkedCells/traversals/LinkedCellTraversalInterface.h
HobbyProgrammer: The linked cells container really only requires a `CellPairTraversal`. That is why this interface is empty. We could definitely delete it. 
The reason why I left it was that I think it makes a lot of sense to have the information which traversals are allowed for this container somehow reflected in the class hierarchy. The direct sum traversal is also a `CellPairTraversal` at the moment, for example. This can be used to check whether a traversal fits to the container, like we would otherwise do purely with the `TraversalOption` and `ContainerOption`, but we also have a place to add new methods that all linked cells traversals should have in the future.

Just tell me your decision on this, and I will realize it. (I am assuming that this decision also holds for the `DirectSumTraversalInterface`.

src/autopas/containers/verletClusterLists/VerletClusterLists.h
HobbyProgrammer: Definitely!

src/autopas/selectors/AutoTuner.h
HobbyProgrammer: Correct! Do you want me to delete it?

HobbyProgrammer:
The first reason why I removed `useNewton3` and `dataLayout` as template parameters from `CellPairTraversal` was that I had a the DirectSumTraversalInterface inherit from it temporarily (until I decided that it having it extra makes more sense).

It might make still sense to have them removed, though, as they are not needed in the implementation, and I would say that as few template parameters as possible are the way to go.

Furthermore, if you take a look at `LinkedCells::iteratePairwise` for example: Currently, I cast the traversal parameter to `CellPairTraversal` to have access to the `CellPairTraversal::setCellsToTraverse()` method. This would not be possible with the template parameters, because this information is not there statically. If we had those template parameters, we would need to add the `setCellsToTraverse()` method to the `LinkedCellsTraversalInterface`. The same method we would have to the `DirectSumTraversalInterface`, too, though, because it also needs cells. So I think having this method in `CellPairTraversal` makes the most sense, which implies that we can't have these template parameters here.

src/autopas/containers/linkedCells/traversals/LinkedCellTraversalInterface.h
FG-TUM: document use for casting

src/autopas/selectors/AutoTuner.h
FG-TUM: yes

HobbyProgrammer:
@FG-TUM Would it maybe make sense to not have `CellPairTraversal` implement the `TraversalInterface`, but all of the traversal interfaces for the specific containers (`LinkedCellsTraversalInterface` etc.)? Actually, `CellPairTraversal` is not implementing a single of the methods specified in `TraversalInterface`.
What is does is, it creates a implementation inheritance hierarchy. `CBasedTraversal`, `SlicedBasedTraversal` etc. inherit from it. By removing this inheritance, we would be able to separate the implementation inheritance and specification (or interface) inheritance. This is good practice: [C++ Core Guidelines: C.129](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c129-when-designing-a-class-hierarchy-distinguish-between-implementation-inheritance-and-interface-inheritance)

FG-TUM:
@HobbyProgrammer besides the random jenkins crashes this still seems to fail the address sanitizer test

HobbyProgrammer:
Jenkins needs to be restarted, again. 
Since only the gpu cloud had problems before, it will probably run through, and the PR is ready for review, then @FG-TUM.

src/autopas/containers/linkedCells/traversals/C01CudaTraversal.h
```diff
@@ -145,7 +145,6 @@ inline void C01CudaTraversal<ParticleCell, PairwiseFunctor, dataLayout, useNewto
 
 template <class ParticleCell, class PairwiseFunctor, DataLayoutOption dataLayout, bool useNewton3>
 inline void C01CudaTraversal<ParticleCell, PairwiseFunctor, dataLayout, useNewton3>::traverseParticlePairs() {
-  auto &cells = *(this->_cells);
```
SteffenSeckler: should not be removed, still needed by cuda!

SteffenSeckler: so move it further down

HobbyProgrammer: Oh yes, sorry! Thanks!

PR #265, opened 2019-06-21
Autotune bayesian

Njan0:
# Description

AutoTuning using bayesian statistics.
 - Added GaussianProcess: Stochastical model for blackbox-function approximation.
 - Added TuningStrategy BayesianSearch: Predict next Configuration to test using GaussianProcess.
 - Added Library: Eigen: Library for linear algebra.

# How Has This Been Tested?

In addition to following tests, I ran some simulations with md-flex.

- [x] GaussianProcessTest
- [x] BayesianSearchTest


src/autopas/selectors/tuningStrategy/GaussianProcess.h
```diff
@@ -0,0 +1,265 @@
...
+ * TODO: maybe offer some options.
+ * @tparam Vector
+ */
+template <class Vector>
+class GaussianProcess {
+ public:
+  /**
+   * Constructor
+   * @param theta prior variance
+   * @param dimScale scale each dimension before applying kernel
```
FG-TUM:
```suggestion
   * @param dimScale scale of each dimension before applying kernel
```
?

tests/testAutopas/tests/selectors/tuningStrategy/BayesianSearchTest.cpp
```diff
@@ -0,0 +1,57 @@
+/**
+ * @file BayesianSearchTest.cpp
+ * @author Jan Nguyen
+ * @date 12.06.19
+ */
+
+#include "BayesianSearchTest.h"
+#include <gmock/gmock-matchers.h>
+#include <gmock/gmock-more-matchers.h>
+
```
FG-TUM: would a unit test make sense where you add some evidence and then expect a certain outcome?

src/autopas/CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ target_compile_definitions(
     PUBLIC
     $<$<BOOL:${AUTOPAS_OPENMP}>:AUTOPAS_OPENMP>
     $<$<BOOL:${AUTOPAS_ENABLE_CUDA}>:AUTOPAS_CUDA>
+    $<$<NOT:$<BOOL:${AUTOPAS_OPENMP}>>:EIGEN_DONT_PARALLELIZE>
```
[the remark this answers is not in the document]
Njan0: I think so too but without setting this flag the test on Jenkings failed. If I remember correctly "clang ninja address sanitizer" and "clang ninja address sanitizer release" could not compile because of missing OpenMP functions.

tests/testAutopas/tests/selectors/tuningStrategy/BayesianSearchTest.cpp
Njan0: I added a test, which assigns a dummy time for each configuration. The BayesianSearch should find the constructed optimum via tuning. I also fixed the seed of the RNG to avoid random fails.

https://github.com/AutoPas/AutoPas/blob/691599b696e17a59afa7e650ad40d260e6ee08ee/tests/testAutopas/tests/selectors/tuningStrategy/BayesianSearchTest.cpp#L77-L103

src/autopas/selectors/tuningStrategy/BayesianSearch.h
```diff
@@ -0,0 +1,181 @@
...
+        _cellSizeFactors(allowedCellSizeFactors.clone()),
+        _currentConfig(),
+        _gp(0., std::vector<double>(featureSpaceDims, 1.), 0.001),
+        _predAcqFunction(predAcqFunction),
+        _predNumSamples(predNumSamples),
+        _maxEvidences(maxEvidences),
+        _lastAcqFunction(lastAcqFunction),
+        _lastNumSamples(lastNumSamples),
+        _rng() {
+    /// @todo setting hyperparameters
```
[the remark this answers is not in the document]
FG-TUM: -> later

FG-TUM:
TODO:
- [ ] remember invalid configurations

src/autopas/selectors/tuningStrategy/BayesianSearch.h
```diff
@@ -163,16 +175,39 @@ bool BayesianSearch::tune() {
...
+    auto best = std::find_if(samples.begin(), samples.end(), [this](const FeatureVector &fv) {
+      return _invalidConfigs.find(fv) == _invalidConfigs.end();
+    });
+
+    if (best != samples.end()) {
+      // valid config found!
+      return *best;
+    } else {
+      // No valid configuration. This should rarely happen.
+      AutoPasLog(debug, "BayesianSearch: Tuning could not generate a valid configuration.");
```
FG-TUM:
```suggestion
      AutoPasLog(critical, "Tuning could not generate a valid configuration.");
```
sounds more like a critical (error) instead of debug message. Or is this a recoverable error?
Also the prefix can be dropped as it will be inserted if the user wants it (verbose logging = on).

src/autopas/selectors/tuningStrategy/TuningStrategyInterface.h
```diff
@@ -36,9 +36,10 @@ class TuningStrategyInterface {
    * The new configuration can be obtained by getCurrentConfiguration. It is the configuration which is either the next
    * configuration to test (=true) or the optimum (=false).
    *
+   * @param currentInvalid set to true to mark current configuration as invalid
```
FG-TUM:
```suggestion
   * @param currentInvalid Tells the tune() function that the currently selected configuration is invalid. This can be used to avoid getting stuck in an invalid optimum.
```

src/autopas/selectors/tuningStrategy/BayesianSearch.h
Njan0: This could happen because there is actually no valid configuration or by chance. To prevent the simulation from crashing in the second case, I allow the function to sample again up to 'maxAttempts = 10' (maybe find a sensible value for this) times. But it if keeps failing, an exception is thrown.

PR #274, opened 2019-06-26
Particle refactoring

Garfield96:
# Description

* Move `ParticleBase` to own file
* Make particle id type template parameter

# How Has This Been Tested?

- [x] Jenkins


src/autopas/particles/Particle.h
```diff
@@ -7,222 +7,18 @@
...
-  /**
-   * Defines whether the particle is owned by the current AutoPas object (aka (MPI-)process)
-   */
-  bool _isOwned;
-};
 /**
  * Particle with all variables in 32 bit precision
  */
-typedef ParticleBase<float> ParticleFP32;
+typedef ParticleBase<float, unsigned int> ParticleFP32;
```
SteffenSeckler: is there any reason to make this an `unsigned int`? it might introduce errors if too many particles are used (more than 4 billion).
Maybe introduce a `ParticleFP32UI` or sth with a better name?

Garfield96: The idea was to make the id as wide as the floating point numbers, but I'll change that. Actually, the reason to set the id type via a template parameter was, that in huge simulations the difference between 64bit and 32bit has a significant impact on memory consumption. However, since the floating point precision doesn't depend on the size of the simulation, it might be better to follow your suggestion.

PR #224, opened 2019-05-24
Add update possibility for addPartice, addHaloParticle + adapt functors + add LogicHandler (ad #215)

SteffenSeckler:
# Description

- [x] correct LJFunctor.h to use isOwned
- [x] add owned to SoA
-  ~~add possibility to update particles using addParticle~~
- [x] add possibility to update particles using addHaloParticle
- [x] bump to cmake 3.14, as of error in lgtm

## Todo:
- [x] uncomment updateHaloParticle VerletListsTest
- [x] updateHaloParticle() for DS
- [x] updateHaloParticle() for LC
- [x] make LJFunctor decide on Halo particles cell-wise again
- [x] write test for addHaloParticle / updateHaloParticle
- [x] rename addHaloParticle to addOrUpdateHaloParticle
- [x] enforcing a rebuild of the container + lists (after enough samples have been taken) also has to enforce an updateContainer call before this.

## Related Pull Requests

- #215 

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes.

- [x] updates LJFunctorTest
- [x] adds a shitload of new tests (>300) in AutoPasInterfaceTest

PR #223, opened 2019-05-23
Adapt sph example to #215

SteffenSeckler:
Adapts the sph examples to the interface change.
* Also makes allowsNewton3 and allowsNonNewton3 to be specified mandatorily
* minor fix for verlet list generation: unused AoS mode was not proper, this is fixed now.

## Description
- [ ] Sph single node example
  - [x] Change from LinkedCells to AutoPas
  - [x] use updateContainers result
  - [x] take care of iterators (they should mostly use ownedOnly)
     - [x] .begin
     - [x] .getRegionIterator
  - [ ] remove deleteHaloParticles calls (only possible once addHaloParticle can update already existing particles)
  - [x] Adapt bigger halo region
- [ ] sph MPI example
  - [x] Change from LinkedCells to AutoPas
  - [x] use updateContainers result
  - [x] take care of iterators (they should mostly use ownedOnly)
     - [x] .begin
     - [x] .getRegionIterator
  - [ ] remove deleteHaloParticles calls (only possible once addHaloParticle can update already existing particles)
  - [x] Adapt bigger halo region

## Related Pull Requests

- #215 
- #220 

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes.

- [x] single node: Comparison against the old results.
- [x] MPI: Comparison against the old results.

PR #217, opened 2019-05-14
adapt updateContainer interface change

SteffenSeckler:
# Description

Adapts change of updateContainer function, i.e., they now return a vector of invalid particles.

## Related Pull Requests

- part of #215

Issue #155, opened 2019-01-14
Finish VerletClusterLists


SteffenSeckler:
VerletClusterLists is not yet finished. 
**Todo's**:
- [x] iteratePairwiseAoS with newton3 enabled. -- here an openmp scheme has to be found. -> #295 
- [ ] halo particle handling:
  - [ ] updateContainer needs to be parallelized
  - [x] halo particles have to be marked, otherwise they can not be differentiated from leaving particles. (as long as they are saved over multiple steps).
  - [ ] addParticle might need to be changed.
  - [ ] addHaloParticle needs to be implemented.
    - [ ] reenable TraversalComparison for halo particles.
  - [ ] deleteHaloParticles needs to be implemented.
  - [ ] updateHaloParticle needs to be implemented.
  - [x] The RegionParticleIterator needs to be implemented. Might make use of: [std::upper_bound](https://en.cppreference.com/w/cpp/algorithm/upper_bound) and [std::lower_bound](https://en.cppreference.com/w/cpp/algorithm/lower_bound) or alternatively [std::equal_range](https://en.cppreference.com/w/cpp/algorithm/equal_range)
  - [x] Implement Traversal
- [ ] enable ContainerConversionTest for VerletClusterLists
- [ ] Deletion of particles is not implemented
- [ ] Particles passed to 'addParticle()' are only really added in the next rebuild. Up until then, also the iterators don't iterate over them. (This breaks e.g. the container conversion test. It does not fail, but silently loses all particles.)
- [ ] OpenMP chunk sizes aren't optimized.
- [ ] iterators are ignoring the iterator behavior -> VERY WRONG!

HobbyProgrammer:
Notes for after #319 is merged besides halo stuff that is still not implemented:

- Deletion of particles is not implemented

- Particles passed to 'addParticle()' are only really added in the next rebuild. Up until then, also the iterators don't iterate over them. (This breaks e.g. the container conversion test. It does not fail, but silently loses all particles.)
- OpenMP chunk sizes aren't optimized.

PR #278, opened 2019-06-26
AutoPas Interface Refactoring

### LINKED ISSUE #214: Global values for verlet lists probably broken

SteffenSeckler:
# Description

Introduce a proper interface for AutoPas to enable one common interface for all containers.

Change list:

- [x] positions of particles no longer completely define the owned state
- [x] LinkedCells: deleteHaloParticles can also delete particles from cells inside the domain
- [x] updateContainer does not necessarily update the container, e.g., for verlet lists it can ignore this to not destroy the container. 
   - [x] there exists a method to force the rebuild
- [x] updateContainer now checks for invalid particles, i.e. particles which do no longer belong into the container (this property depends on the container type), invalid particles will be returned as a vector. #217  
- [x] On a proper rebuild of a container the halo particles are removed first.
- ~Verlet-Type container: addParticle is actually an addOrUpdateParticle - if it knows the particle (and it is a halo particle), it will change it to an owned particle. If it is already owned it throws an exception. if it is not known (e.g. for real particle insertion) it will invalidate the lists.~
  -  ~document this properly~
- [x] Verlet-Type container (maybe all containers): addHaloParticle is actually an addOrUpdateHaloParticle - if it knows the halo particle, it will update its state (overwrite it). if it is not known (e.g. for real particle insertion) it will invalidate the lists and add it (if the neighbor lists are invalid) or ignore the halo particle if the neighbor lists are valid and it is not too close (i.e. within a distance of `cutoff radius + skin/2` from the domain). 
  - [x] document this properly
- [x] Rewrite sph examples for new interface + AutoPas classes
- [x] Potentially let non-verlet type containers know about skin to be able to check halo particle insertion.
- [x] Fix functors to use bool isOwned instead of check for position.
- [x] Also enforces cmake 3.14

## Related Pull Requests + Issues

- #207 
- #208 
- #209
- #214 

## subtasks
- #217 
- #220 
- #223 
- #224 

## Resolved Issues # (issue)

- [x] fixes #214 
- adds tests for #208 , but not yet all enabled.
- parts of #207  

## Todo
- [x] New return type of updateContainer has to be implemented for all containers
- [x] Enable halo addition for LinkedCell container (ContainerSelectorTest.cpp)
- [x] Test with two autopas containers
  - [x] Tests of interfaces need to check all options (newton3, ...)
  - [x] Let tests run longer to check rebuild.
- ~Think about renaming `updateContainer()`~ -- will not be renamed, it is documented instead
- ~Think about renaming halo particles to sth like not-owned particles~ -- will not be renamed (too many merge conflicts projected)
- [x] update documentation
- ~parallelize VerletClusterLists::updateContainer()~ -- postponed issue: #155 
- [x] check ///@ todo
- ~optimize deleteHaloParticles() funtions, or rather the iterators.~ -- postponed issue: #280
- ~Fix VerletClusterLists -- will probably go to different PR~ -- postponed issue: #155
- [x] remove deleteHaloParticles calls (only possible once addHaloParticle can update already existing particles) from sph examples 
   - [x] and check results again
- [x] Bug still existing as of 25.6.2019, 13:00: enforcing a rebuild of the container + lists (after enough samples have been taken) also has to enforce an updateContainer call before this.
- [x] adapt gpu version
- [x] Document that using two functors one supporting a N3 option the other does not is not possible.
- [x] revert some changes to cellborderandflagmanager's (CellBlock3D and DirectSum)

# How Has This Been Tested?

- [x] Test for particle addition and deletion
- [x] Test for simulation loop with one single container
- [x] Test for simulation loop with two containers


README.md
```diff
@@ -122,42 +122,62 @@ The particle can be accesses using `iter->` (`*iter` is also possible).
...
-You have to update the container when the two conditions are fullfilled:
-* If you moved particles
-* You want to use `iteratePairwise()` or a RegionParticleIterator
-
-#### When it is not enough
-If you moved particles by more than one interaction length.
-If you are planning to move particles by a long distance,
-e.g. because of boundary conditions please delete the particles and add them again:
+### Simulation Loop
+One simulation loop should always consist of the following phases:
```
FG-TUM: would make sense to distinguish the MPI / no-MPI cases

README.md
```diff
@@ -122,42 +122,62 @@ The particle can be accesses using `iter->` (`*iter` is also possible).
...
+   ```
+
+2. Handling the leaving particles
+   * Apply boundary conditions on them
+   
+   * Potentially send them to other mpi-processes
+   
+   * Add them to the containers using
+      ```C++
+      addParticle(particle)
```
FG-TUM: autopas.addParticle(particle)

(also for next code blocks)

README.md
```diff
@@ -122,42 +122,62 @@ The particle can be accesses using `iter->` (`*iter` is also possible).
...
+4. Perform an iteratePairwise step.
+   ```C++
+   autoPas.iteratePairwise(functor);
+   ```
+
+In some iterations step 1. will return an empty list of invalid particles to benefit of not rebuilding the containers and the associated neighbor lists.
+
+### Using multiple functors
+
+AutoPas is able to work with simulation setups using multiple functors.
```
FG-TUM: multiple forces described by multiple functors?

README.md
```diff
@@ -122,42 +122,62 @@ The particle can be accesses using `iter->` (`*iter` is also possible).
...
+A good example for that is the sph example found under examples/sph or examples/sph-mpi.
```
FG-TUM: A demonstration for that

README.md
```diff
@@ -122,42 +122,62 @@ The particle can be accesses using `iter->` (`*iter` is also possible).
...
+There exist some things you have to be careful about when using multiple functors:
+* If you use multiple functors it is necessary that all functors support the same newton3 options. If there is one functor not supporting newton3, you have to disable newton3 support for AutoPas by calling
+  ```C++
+  autoPas.setAllowedNewton3Options({false});
+  ```
+
+* If you have `n` functors within one iteration and update the particle position only at the end or start of the iteration, the rebuildFrequency and the samplingRate have to be a multiple of `n`.   
+
+### Inserting additional particles
+When inserting additional particles, you always have to enforce a containerUpdate before doing that on ALL AutoPas instances, i.e., on all mpi processes, by calling
```
FG-TUM: before doing *that* on ALL AutoPas instance
-> whats "that"? Sentence unclear

README.md
```diff
@@ -122,42 +122,62 @@ The particle can be accesses using `iter->` (`*iter` is also possible).
...
-for (auto p: invalidParticles) {
-  sphSystem.addParticle(p);
-}
-```
-
-#### Special exceptions
-* Verlet-Lists, here it is safe to not update the container as long as particles move not more than a skin radius.
+autoPas.updateContainerForced();
+``` 
+This will invalidate the internal neighbor lists and containers.
```
FG-TUM: Maybe we should describe this ownership model for particles in a dedicated section (can happen in a separate PR)

src/autopas/containers/verletListsCellBased/verletListsCells/VerletListsCells.h
```diff
@@ -116,38 +116,38 @@ class VerletListsCells
     switch (_buildTraversal) {
       case c08: {
         if (useNewton3) {
-          auto traversal = C08Traversal<LinkedParticleCell, typename verlet_internal::VerletListGeneratorFunctor,
+          auto buildTraversal = C08Traversal<LinkedParticleCell, typename verlet_internal::VerletListGeneratorFunctor,
```
FG-TUM: can't you remove / simplify this switch tree using 
```
std::unique_ptr<TraversalInterface> TraversalSelector<ParticleCell>::generateTraversal()?
```
(the new one with less templates)

README.md
SteffenSeckler: there really isn't much different there...

README.md
SteffenSeckler: ok

SteffenSeckler: #281

src/autopas/containers/verletListsCellBased/verletListsCells/VerletListsCells.h
SteffenSeckler: can probably be replace be sth. like
```
// first check whether _buildTraversal is a valid option.
    std::set<TraversalOption> validTraversalOptions{c08, c01, c18};
    if (std::find(validTraversalOptions.begin(), validTraversalOptions.end(), _buildTraversal) ==
        validTraversalOptions.end()) {
      utils::ExceptionHandler::exception("VerletListsCells::updateVerletLists(): unsupported Traversal: {}",
                                         _buildTraversal);
    }
    if (_buildTraversal == TraversalOption::c01 and useNewton3) {
      utils::ExceptionHandler::exception("VerletListsCells::updateVerletLists(): c01 does not support newton3");
    }

    // generate buildTraversal and iterate
    auto buildTraversal = TraversalSelector<LinkedParticleCell>::generateTraversal(
        _buildTraversal, f, this->getTraversalSelectorInfo(), DataLayoutOption::aos,
        useNewton3 ? Newton3Option::enabled : Newton3Option::disabled);
    this->_linkedCells.iteratePairwise(&f, &buildTraversal);
```
but will not do that for now

SteffenSeckler: #283

PR #290, opened 2019-06-28
Cutoff sanity check

FG-TUM:
# Description

Previously it was possible to use a functor with a larger cutoff than the containers'. This sanity check prevents this.

## ~Resolved Issues~

# How Has This Been Tested?

- [x] md-flexible


src/autopas/sph/SPHCalcHydroForceFunctor.h
```diff
@@ -25,6 +25,10 @@ class SPHCalcHydroForceFunctor : public Functor<SPHParticle, FullParticleCell<SP
   /// particle cell type
   typedef FullParticleCell<Particle> ParticleCell;
 
+  SPHCalcHydroForceFunctor()
+      // for the cutoff sanity check this works only if smoothing length is never > 1. (currently hard coded to 0.012)
+      : autopas::Functor<Particle, ParticleCell>(autopas::sph::SPHKernels::getKernelSupportRadius()){};
```
SteffenSeckler: doesn't make too much sense here, as the smoothing length is normally not 1

PR #292, opened 2019-06-28
Adds more c++17 features

SteffenSeckler:
# Description

- nested namespaced
- using constexpr if instead of enable_if in SoAStorage


SteffenSeckler:
cuda exception handler cannot have nested namespaces as it is compiled with the cuda compiler

PR #299, opened 2019-07-07
Add first version of SoAView

### LINKED ISSUE #300: Give Range to SoAFunctor() Methods
When developing the SoA version of the VerletClusterLists Coloring traversal, I came across the problem again that `Functor` only provides methods for calculating forces between all particles in a SoA, or the forces between all particles from two SoAs. Especially when working with clusters from the VerletClusterLists, this is very tedious. 

In the current implementation of `SoA`, there is a view start that can be defined, but it is not enough, because there is no view end and for each `SoA`, only one view start can be defined, so it does not support multiple independent views.

I would like to specify for each SoA in these functions the range of particles in the SoA the function works on, so e.g. calculate the forces between all particles in the SoA from index 12 to index 16. 

- **Idea 1:**
#298: Add start index and number of particles to work on to the signature of the two `Functor::SoAFunctor()` methods. All functor implementations only need minimal changes for that, but there are many changes. Old signatures can stay and call the new ones.

- **Idea 2:**
#299: Add a `SoAView` that defines a range on a `SoA` and allows to iterate over that. Then replace the `SoA` parameters of the `SoAFunctor()` with `SoAViews`. This changes the signatures, but not the implementations.

If this is interface is not provided, for every cluster, a new SoA could be created, like it did in #269. I didn't measure, but I think it is slower because of the whole copying. (What exactly is described there).

_Number 300 :smirk:_

HobbyProgrammer:
# Description

Adds views on `SoA`s. The view contains only the methods that are needed by the functors at the moment, `begin()` and `getNumParticles()`. It can be used to iterate over only a small part of a `SoA`, making it possible to e.g. use the same SoA for multiple clusters of the VerletClusterLists.

They are meant to replace the `SoA` parameters in the methods of `Functor`. This would only change the signatures, not the implementations.
- [x] Replace the `SoA` parameters in the methods of `Functor`.

## Related Pull Requests

- Views could replace functor refactoring of #298 

## Resolved Issues

- [x] closes #300

# How Has This Been Tested?

- Unit Tests


Garfield96:
Additional tasks:
- [ ] replace occurences of `viewStart` in `C01Traversal` and `C04SoATraversal` (not yet in master) by `SoAView`
- [ ] remove `viewStart` from class `SoA`

HobbyProgrammer:
I now changed SoAFunctor() signatures to use SoAView instead of SoA&.

It was kind of hard to decide whether to use SoAView, const SoAView&, or SoAView&& as type in the signatures. SoAView& fell out pretty early, because it cannot accept rvalues, which is needed by nearly every call to SoAFunctor const SoAView& would only be possible if SoAView::begin was const, too. It would be possible to implement that, but this prevents having "real" const SoAViews which can only be read and not written to. SoAView&& is the most similar to SoAView and would also be a good option, I think. The problem here is, that if a SoAView was an lvalue, it would need to be moved to pass it to SoAFunctor() which seems unnecessary. The option that I chose, SoAView, is the most broadly applicable option and , I think, compilers don't even copy rvalue SoAViews that are passed, but instead construct them directly inplace, so I don't expect much performance impact.

If you think differently, feel free to share your opinion.

HobbyProgrammer:
I now removed view start from `SoA`. @Garfield96 Do you have a better idea for replacing view start in `C01Traversal` than I had? Just calling the pairwise functor directly works, but it is not nice to just bypass the cell functor. All the other ideas I came up with involved a lot of otherwise unnecessary refactoring of `CellFunctor` or `ParticleCell`, though.

Garfield96:
> I now removed view start from `SoA`. @Garfield96 Do you have a better idea for replacing view start in `C01Traversal` than I had? Just calling the pairwise functor directly works, but it is not nice to just bypass the cell functor. All the other ideas I came up with involved a lot of otherwise unnecessary refactoring of `CellFunctor` or `ParticleCell`, though.

I agree with you that bypassing the cell functor is not optimal, especially because it is also necessary in `C04SoATraversal`, which is now in the master branch. The only idea I came up with so far is to replace the SoA attribute inside the cells by an SoAView. The SoAViews could use `std::shared_ptr` to manage the lifetime of the real SoA datastructure. In this case bypassing wouldn't be necessary. But I admit that this refactoring requires a lot of time.

HobbyProgrammer:
https://github.com/AutoPas/AutoPas/blob/043c38a20ed883aedc2c0b9321d177395c070405/src/autopas/utils/SoAView.h#L80-L84
Do you have any idea what clangformat is doing here?

Otherwise, asserts are removed, so besides the C04SoATraversal not working, everything is ready for review!

Garfield96:
@HobbyProgrammer I fixed the C04SoATraversal in branch container-tests-refactoring.

Garfield96:
> https://github.com/AutoPas/AutoPas/blob/043c38a20ed883aedc2c0b9321d177395c070405/src/autopas/utils/SoAView.h#L80-L84
> 
> Do you have any idea what clangformat is doing here?
> 
> Otherwise, asserts are removed, so besides the C04SoATraversal not working, everything is ready for review!

This is a bug. Clang-format breaks, because of the `[[nodiscard]]` attributes in the method declarations before `private`. If the you remove the `[[nodiscard]]` attributes, the failure disappears. This failure is fixed in clang-format-7, so there is no need to open a bug report.

HobbyProgrammer:
@FG-TUM Ready for review! (For some reason, I can't rerequest a review from you. Probably because of that bug that you already are in the reviewer list twice...)

src/autopas/containers/linkedCells/traversals/C04SoACellHandler.h
```diff
@@ -206,57 +214,54 @@ inline void C04SoACellHandler<ParticleCell, PairwiseFunctor, DataLayout, useNewt
...
-
-          cell1->_particleSoABuffer.setViewStart(0);
-          cell1->_particleSoABuffer.resizeArrays(numParticlesBaseCell);
-          this->_cellFunctor.processCellPair(*cell1, temp);
-
-          cell1->_particleSoABuffer.append(temp._particleSoABuffer);
+          auto restView = cell1->_particleSoABuffer.constructView(numParticlesBaseCell,
+                                                                  cell1->_particleSoABuffer.getNumParticles());
+          _pairwiseFunctor->SoAFunctor(stripeView, restView, useNewton3);
+          if (not useNewton3) {
```
FG-TUM: @Garfield96  looking at this traversal in this way it seems Newton3 off does not make too much sense, does it?

src/autopas/containers/verletListsCellBased/verletLists/VerletListHelpers.h
```diff
@@ -84,7 +84,7 @@ class VerletListHelpers {
      * @param soa the soa
      * @param newton3 whether to use newton 3
      */
-    void SoAFunctor(SoA<SoAArraysType> &soa, bool newton3) override {
+    void SoAFunctor(SoAView<SoAArraysType> soa, bool newton3) override {
```
FG-TUM: have you tested if this is faster than passing references? Is move used here?

HobbyProgrammer: I have talked about this here: https://github.com/AutoPas/AutoPas/pull/299#issuecomment-509968207

Summed up, I think it doesn't make any difference for nearly all use cases.

Do you have another opinion?

src/autopas/containers/linkedCells/traversals/C04SoACellHandler.h
```diff
@@ -206,57 +214,54 @@ inline void C04SoACellHandler<ParticleCell, PairwiseFunctor, DataLayout, useNewt
...
         // two cases intervall in current  stripe
         cell1 = &combinationSlice[currentSlice];
-        cell1->_particleSoABuffer.setViewStart(0);
+        auto stripeView = cell1->_particleSoABuffer.constructView(0, numParticlesBaseCell);
+        cell1ViewStart = 0;
         if (slice == currentSlice) {
           // process stripe with itself
-          cell1->_particleSoABuffer.setViewLength(numParticlesBaseCell);
-
-          this->_cellFunctor.processCell(*cell1);
```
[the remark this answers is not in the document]
HobbyProgrammer: True!

src/autopas/containers/verletListsCellBased/verletLists/VerletListHelpers.h
HobbyProgrammer: I haven't measured what is faster, but I haven't measured any performance difference at least on my machine with gcc between the old implementation without SoAViews and with SoAViews, so I think regardless what method to pass the SoAView is used, measuring the time it needs is hard, because it is such a small impact.

FG-TUM: Ah I forgot I already read that comment^^
In theory, I agree. Would be interesting to verify though.

FG-TUM: But then again GCC doesn't manage to vectorize the functor (at least with N3). I think this should be fine as it is now.

PR #301, opened 2019-07-09
Container Tests Refactoring

### LINKED ISSUE #309: C04SoA Traversal is not compared with others in tests
I tried to replace the old view stuff of `SoA` in `C04SoATraversal` with my new `SoAView` in #299 , and it worked instantly, at least the tests didn't fail. That is impossible, so I checked the tests.

**Describe the bug**
The tests don't check if the `C04SoATraversal` gives correct results.

**To Reproduce**
1. Go to `C04SoACellHandler.h`
2. Remove line 251: `this->_cellFunctor.processCellPair(*cell1, currentCS);` (it is pretty important, I think)
3. See that all tests pass.

**Expected behavior**
Some test should fail if I destroy the `C04SoATraversal` implementation.

HobbyProgrammer:
# Description

Refactors the tests of the containers to remove duplicate code.

**TODO:**
- [x] Pull together common tests for all containers
- [x] Write generic ContainerVersusContainer test and use it for all versus tests.
- [x] Check globals in TraversalComparison
- [x] Check halo particles
- [x] Check slight moves
- [x] Cell size factors

**not done**
- <strike>Pull together common tests for all Verlet lists containers</strike>
- <strike>Check that Verlet-like containers need at most as many flops as others</strike> these tests have now actually been removed.

## Fixes 
- fixes incorrect calculation of globals when using the ClusterFunctor.
- fixes incorrect calculation of globals when using the LJFunctorCudaGlobals and more than 64 particles per cell.
- fixes C01CudaTraversal claiming to be applicable for cellLengths < interactionLength, see also #417 
- fixes incorrect calculation of energy in LJFunctorCudaGlobals. only bodyBodyFN3, i.e., cell pairs and newton3 and interactions with halo particles.
- fixes misplaced `__syncthreads()` in `SoAFunctorN3Pair()` of LJFunctorCuda(Globals).cu

## hotfixes
- disables sorting, see #418 
- does not test cuda c01 for very high densities, see #419
- disables C01CudaTraversal for enabled newton3, see #420
- disables C04SoATraversal completely, see #422, #423

## Related Pull Requests
None

## Resolved Issues
- closes #309 (8d3f289 implements base step comparison with C08, a9403fa compares C04SoA with other traversals using many particles (100+) )

# How Has This Been Tested?

This adds lots of tests.

# Notes
The gtest_discover_tests does now take more than a minute. I have submitted a merge request to fix this:
https://gitlab.kitware.com/cmake/cmake/merge_requests/4371


HobbyProgrammer:
For the generic ContainerVersusContainer test:
- Should this be run for all traversals or only for the containers? 
- How should the kernel call and distance calculation count checks be build into these generic tests, as there are several exceptions depending on the traversal or container?

HobbyProgrammer:
For the common tests for all containers:
- How much can `isContainerUpdateNeeded()` be tested generically? Is there something else than testing than a container update is needed if a particle moves to halo?

Garfield96:
As far as I know, @SteffenSeckler wants to extend the AutoPasInterfaceTest to test 2 containers against each other. A basic version is already there.

HobbyProgrammer:
> 
> 
> As far as I know, @SteffenSeckler wants to extend the AutoPasInterfaceTest to test 2 containers against each other. A basic version is already there.

I think these tests will only test the containers with very few particles against each other. Correct me if I am wrong @SteffenSeckler.

tests/testAutopas/tests/containers/AllContainersTests.cpp
```diff
@@ -0,0 +1,149 @@
+/**
```
SteffenSeckler: I think it would make more sense to test this using the AutoPas interface instead of the actual containers.
@FG-TUM what do you think?
the only problem with that is that these here, look more like unit tests and not tests of the entire thing...

this also means that, e.g., that tests of the actual containers are more error-prone (as these do no longer detect these exceptions)

tests/testAutopas/tests/containers/AllContainersTests.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * Checks if addParticle() only accepts particles in the domain and throws otherwise, and if addHaloParticle() never
+ * throws.
+ *
+ * @Reviewer: Meant to replace
+ *      LinkedCellsTest::testParticleAdding
+ *      DirectSumContainerTest::testParticleAdding
+ */
+TEST_P(AllContainersTests, testParticleAdding) {
```
SteffenSeckler: this test partially fails in the moment for some tests.
it's also no longer correct (see below)

tests/testAutopas/tests/containers/AllContainersTests.cpp
```diff
@@ -0,0 +1,149 @@
...
+        if (x == -1.5 or y == -1.5 or z == -1.5 or x == 11.5 or y == 11.5 or z == 11.5) {
+          EXPECT_ANY_THROW(_container->addParticle(p));     // outside, therefore not ok!
+          EXPECT_NO_THROW(_container->addHaloParticle(p));  // much outside, still ok because it is ignored!
+        } else if (x == 10. or y == 10. or z == 10. or x == -.5 or y == -.5 or z == -.5 or x == 10.5 or y == 10.5 or
+                   z == 10.5) {
+          EXPECT_ANY_THROW(_container->addParticle(p));     // outside, therefore not ok!
+          EXPECT_NO_THROW(_container->addHaloParticle(p));  // outside, therefore ok!
+        } else {
+          EXPECT_NO_THROW(_container->addParticle(p));  // inside, therefore ok!
+          EXPECT_NO_THROW(
```
SteffenSeckler: this should actually be an EXPECT_ANY_THROW(

tests/testAutopas/tests/containers/AllContainersTests.cpp
```diff
@@ -0,0 +1,149 @@
...
+              _container->addHaloParticle(p));  // inside, but still ok, as halo particle can be inside of the domain!
```
SteffenSeckler:
```suggestion
              _container->addHaloParticle(p));  // inside, and not ok, as halo particle cannot be added inside of the domain!
```

SteffenSeckler: it might also make sense to remove the exceptions from the containers.

SteffenSeckler:
@HobbyProgrammer I have extended these tests a bit. somehow, the verletClusterLists tests are failing to calculate correct global values, do you have an idea, why this happens?

SteffenSeckler:
@HobbyProgrammer ok found the bug (https://github.com/AutoPas/AutoPas/pull/301/commits/57a637fe94a08da641aaa80f19cb3145d933964f)

SteffenSeckler:
@FG-TUM this is more or less complete. So review, please :)

tests/testAutopas/tests/containers/TraversalComparison.cpp
```diff
@@ -0,0 +1,236 @@
...
+ * generation.
+ * @return Tuple of forces for all particles, ordered by particle id, and array of the globals values (upot, double).
+ */
+std::tuple<std::vector<std::array<double, 3>>, std::array<double, 2>> TraversalComparison::calculateForces(
+    autopas::ContainerOption containerOption, autopas::TraversalOption traversalOption,
+    autopas::DataLayoutOption dataLayoutOption, autopas::Newton3Option newton3Option, unsigned long numMolecules,
+    unsigned long numHaloMolecules, std::array<double, 3> boxMax, double cellSizeFactor, bool doSlightShift) {
+  // Construct container
+  autopas::ContainerSelector<Molecule, FMCell> selector{_boxMin, boxMax, _cutoff};
+  double skin = _cutoff * 0.1;
```
FG-TUM:
```suggestion
  constexpr double skin = _cutoff * 0.1;
```
since `_cutoff` is also `constexpr

tests/testAutopas/tests/containers/TraversalComparison.cpp
```diff
@@ -0,0 +1,236 @@
...
+                                                               container->getBoxMin(), container->getBoxMax(),
+                                                               numMolecules);
+
+  autopasTools::generators::RandomGenerator::fillWithHaloParticles(
+      *container, Molecule({0., 0., 0.}, {0., 0., 0.}, numMolecules), container->getCutoff(), numHaloMolecules);
+
+  container->rebuildNeighborLists(traversal.get());
+
+  if (doSlightShift) {
+    executeSlightShift(container, skin / 2, numMolecules + numHaloMolecules);
```
FG-TUM: `skin / 2` could also be replaced by a `constexpr` but ok...

tests/testAutopas/tests/containers/TraversalComparison.cpp
```diff
@@ -0,0 +1,236 @@
...
+  if (_forcesReference.count(key) == 0) {
+    generateReference(key);
+  }
+
+  for (size_t i = 0; i < numParticles; ++i) {
+    for (unsigned int d = 0; d < 3; ++d) {
+      double calculatedForce = calculatedForces[i][d];
+      double referenceForce = _forcesReference[key][i][d];
+      EXPECT_NEAR(calculatedForce, referenceForce, std::fabs(calculatedForce * rel_err_tolerance))
+          << "Particle id: " << i;
```
FG-TUM: why not print the whole particle

tests/testAutopas/tests/containers/TraversalComparison.cpp
```diff
@@ -0,0 +1,236 @@
...
+    }
+  }
+
+  auto &globalValuesReferenceRef = _globalValuesReference[key];
+  for (size_t index : {0, 1}) {
+    EXPECT_NE(calculatedGlobals[index], 0) << "index:" << index;
```
FG-TUM: for readability, calcuclatedGlobals might be better as map<string, double>

src/autopas/containers/directSum/DirectSum.h
```diff
@@ -54,21 +54,14 @@ class DirectSum : public ParticleContainer<ParticleCell> {
   ContainerOption getContainerType() const override { return ContainerOption::directSum; }
 
   /**
-   * @copydoc ParticleContainerInterface::addParticle()
+   * @copydoc ParticleContainerInterface::addParticleImpl()
```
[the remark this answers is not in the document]
SteffenSeckler: yes

src/autopas/containers/linkedCells/traversals/C04SoACellHandler.h
```diff
@@ -236,6 +236,14 @@ inline void C04SoACellHandler<ParticleCell, PairwiseFunctor, dataLayout, useNewt
         }
         cell1 = &combinationSlice[index];
         cell1ViewEnd = combinationSlicesOffsets[index][1];
+      } else if (offset1 == _baseOffsets.back().front()) {
```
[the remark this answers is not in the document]
SteffenSeckler: no purpose wrong merge ;)

tests/testAutopas/tests/containers/AllContainersTests.cpp
```diff
@@ -0,0 +1,107 @@
...
+                           auto allOptions =
+                               std::set<autopas::ContainerOption>{autopas::ContainerOption::getAllOptions()};
+                           allOptions.erase(autopas::ContainerOption::verletClusterLists);
+                           return allOptions;
+                         }()),
+                         AllContainersTests::getParamToStringFunction());
+
+INSTANTIATE_TEST_SUITE_P(DISABLED_Generated, AllContainersTests,
+                         testing::Values(autopas::ContainerOption::verletClusterLists),
+                         AllContainersTests::getParamToStringFunction());
```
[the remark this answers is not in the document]
SteffenSeckler: added an additional @todo instead.

tests/testAutopas/tests/containers/AllContainersTests.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Checks if ParticleContainer::getNumParticle() returns the correct number of particles.
+ *
+ */
+TEST_P(AllContainersTests, testGetNumParticles) {
+  EXPECT_EQ(_container->getNumParticles(), 0);
+
+  std::array<double, 3> r = {2, 2, 2};
+  Particle p(r, {0., 0., 0.}, 0);
```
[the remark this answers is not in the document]
SteffenSeckler: better for readability.

tests/testAutopas/tests/containers/TraversalComparison.cpp
SteffenSeckler: ignored

tests/testAutopas/tests/containers/TraversalComparison.cpp
SteffenSeckler: because I don't have the entire particle here.

tests/testAutopas/tests/containers/TraversalComparison.cpp
SteffenSeckler: chosen to use a struct.

src/autopas/LogicHandler.h
```diff
@@ -76,7 +76,7 @@ class LogicHandler {
     if (not isContainerValid()) {
       if (not utils::inBox(haloParticle.getR(), _autoTuner.getContainer()->getBoxMin(),
                            _autoTuner.getContainer()->getBoxMax())) {
-        container->addHaloParticle(haloParticle);
+        container->template addHaloParticle<false>(haloParticle);
```
FG-TUM:
```suggestion
        container->template addHaloParticle</* checkInBox */ false>(haloParticle);
```
:innocent:

PR #304, opened 2019-07-09
remove dummy traversal

SteffenSeckler:
# Description

Removes dummy traversal, as it is no longer needed


SteffenSeckler:
@HobbyProgrammer you no longer need to do this :)

PR #307, opened 2019-07-10
Add generic to_string function, Bug fix NumberSet

Garfield96:
# Description

This PR adds a generic `to_string` function to `ArrayMath`, which can convert all containers (`cbegin` + `cend` are required) to a string representation (really helpful for debugging). 
It also fixes a bug in `NumberSet`. If the set was empty, the string conversion method never terminated.

# How Has This Been Tested?

- [x] New test cases
- [x] Jenkins

### DIFF
--- a/src/autopas/utils/ArrayMath.h
+++ b/src/autopas/utils/ArrayMath.h
@@ -23,7 +23,7 @@ namespace autopas::ArrayMath {
  * @return a + b
  */
 template <class T, std::size_t SIZE>
-constexpr std::array<T, SIZE> add(const std::array<T, SIZE> &a, const std::array<T, SIZE> &b) {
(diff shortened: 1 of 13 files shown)


src/autopas/utils/ArrayMath.h
```diff
@@ -163,4 +164,27 @@ constexpr std::array<T, SIZE> normalize(const std::array<T, SIZE> &a) {
...
 
+/**
+ * Generates a string representation of a container which fulfills the Container requirement (provide cbegin and cend).
+ * @tparam T Type of Container.
+ * @param a Container.
+ * @param delimiter String delimiter.
+ * @return String representation of a.
+ */
+template <class T>
+std::string to_string(const T &a, const std::string &delimiter = ", ") {
```
FG-TUM: move to new namespace ArrayUtils

PR #317, opened 2019-07-18
C04SoA: Bug fix

Garfield96:
# Description

This PR fixes a bug in C04SoA Traversal.

## Related Pull Requests

- #301 

# How Has This Been Tested?

- [x] test case C04SoATraversalTest (in branch container-tests-refactoring)
- [x] Jenkins

### DIFF
--- a/src/autopas/containers/linkedCells/traversals/C04SoACellHandler.h
+++ b/src/autopas/containers/linkedCells/traversals/C04SoACellHandler.h
@@ -232,8 +232,16 @@ inline void C04SoACellHandler<ParticleCell, PairwiseFunctor, DataLayout, useNewt
       } else if (offset1 == _baseOffsets.front().back()) {
         cell1 = &combinationSlice[currentSlice];
         cell1->_particleSoABuffer.setViewStart(
-            combinationSlicesOffsets[slice][combinationSlicesOffsets[slice].size() - 2]);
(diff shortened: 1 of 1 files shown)


HobbyProgrammer:
I just locally merged this into soa-views and tried to port it to them, but I encountered the following:
When adding the following code to `SoA::getNumParticles()`:
```
if(viewLength != -1l) {
  auto realSoASize = soaStorage.template get<0>().size();
  if(viewStart + viewLength > realSoASize) {
#pragma omp critical
    std::cerr << "Error: " << "ViewStart: " << viewStart << " ViewLength: " << viewLength << " Size: " << realSoASize << std::endl;
  }
}
```
Quite a few errors are reported when executing the `Newton3OnOffTest` for the `C04SoATraversal`.

Is this intended behaviour? It seems to me that `viewStart + viewLength` should never be greater than the real size of the `SoA`, right?
I have no idea what the effects of this are, maybe it does not change the behaviour of the traversal, but for my `SoAView`, I implemented the checks that this can never happen, and they fail, so I need to know what I should do with this.

FG-TUM:
@HobbyProgrammer I cannot reproduce this on this branch. We should look into this on your branch. Please reference it here.

HobbyProgrammer:
> 
> 
> @HobbyProgrammer I cannot reproduce this on this branch. We should look into this on your branch. Please reference it here.

Okay, this is odd, as I tested it on this branch locally. We can look at this together tomorrow, on my laptop.

PR #319, opened 2019-07-26
VerletClusters refactoring

### LINKED ISSUE #271: Iterator of VerletClusterLists iterates over dummy particles
**Describe the bug**
When iterating over a `VerletClusterLists` container using `begin()`, the iterator also iterates over all dummy particles that have been used to fill up the clusters. These are constructed using the default constructor of the `Particle` class template parameter and are not distinguishable from regular particles after they have been sorted into the clusters.

**To Reproduce**
1. Construct a `VerletClusterLists` container with one particle and cluster size > 1.
2. Iterate over it.
3. See that you iterate over cluster size particles.

**Expected behavior**
The iterator should only iterate over the particles that are added to the container and relevant for the user.

**Additional context**
In `LinkedCellsVersusVerletClusterListsTest.cpp`, where the container is tested against `LinkedCells`, the test just skips all particles for the comparison that have id 0 to avoid dummy particles. This is documented there.

Related issue: #155

HobbyProgrammer:
# Description
Almost completely refactors the VerletClusterLists container. Goals for the refactoring were:
 - better readability of the code
 - better extensability of the code
 - Remove the need for any `std::unordered_map`s (previously needed especially needed for SoA)
 - not much worse performance

## What has been done?
 - [x] constexpr cluster size
 - [x] Overall refactoring
    - Pull out classes `Cluster` and `ClusterTower`
    - Save neighbors and SoAView in Cluster to avoid all `std::unordered_map`s
    - Use extract method object refactoring resulting in `VerletClusterListsRebuilder`
 - [x] Parallelize rebuild and neighbor list generation
 - [x] Parallelize SoA loading and extracting
 - [x] Have iterator only iterate over the real particles, not the dummies.

## What still has to be done?
- [x] Discuss if it is okay that added particles are only really added once rebuildVerletLists() is called. (This impacts the iterator, too).
- [x] Find some heuristic for all of the chunksizes for the parallelizations.
- [x] Deletion in the iterators. Same problem as for adding particles.
- [x] Add more tests

Everything that worked before should also work here already.

## Benchmark Comparison
I also did some performance comparison on the mpp2 cluster between before and after, and I am pretty happy with the result.

_Worst change: Strong Scaling with 2^19 particles and box length 25_
VerletClusterLists_verlet-clusters-coloring_AoS_Newton3disabled_RebuildFreq100000_gaussian_strongScaling_MFUPs

_Best Change: Normal Benchmark with box length 25_
VerletClusterLists_verlet-clusters_SoA_Newton3disabled_RebuildFreq100000_uniform_numParticles_MFUPs

## Related Pull Requests
- #298 is already contained here.

## Related Issues
- ref #155 
- closes #271

## How Has This Been Tested?
Jenkins


HobbyProgrammer:
If you have any suggestions what should be documented more or to improve code quality, tell me!

HobbyProgrammer:
@FG-TUM This should be ready for review!

src/autopas/iterators/SingleCellIterator.h
```diff
@@ -11,11 +11,15 @@
 #include "autopas/utils/ExceptionHandler.h"
 
 namespace autopas::internal {
+
+//@TODO Besides the methods of ParticleCell, a method Particle& getParticle(size_t index) and Particle&
+//    * getParticle(size_t index) const is needed for the template type ParticleCell. It might be possible to add them
+//    to the ParticleCell class.
```
SteffenSeckler: remove todo, as it is already done.

src/autopas/cells/FullParticleCell.h
```diff
@@ -90,6 +90,20 @@ class FullParticleCell : public ParticleCell<Particle> {
 
   std::array<double, 3> getCellLength() const override { return _cellLength; }
 
+  /**
+   * Returns the particle at position index. Needed by SingleCellIterator.
+   * @param index the position of the particle to return.
+   * @return the particle at position index.
+   */
+  decltype(auto) getParticle(size_t index) { return _particles.at(index); }
```
SteffenSeckler: I would prefer it if the function signature is written out, renamed to `at` and moved close to the `operator []` i.e.,
```suggestion
  Particle& at(size_t index) { return _particles.at(index); }
```
as this makes it easier to understand, especially for less experienced students.

src/autopas/cells/FullParticleCell.h
```diff
@@ -90,6 +90,20 @@ class FullParticleCell : public ParticleCell<Particle> {
...
+
+  /**
+   * Returns the const particle at position index. Needed by SingleCellIterator.
+   * @param index the position of the particle to return.
+   * @return the particle at position index.
+   */
+  decltype(auto) getParticle(size_t index) const { return _particles.at(index); }
```
SteffenSeckler: and:
```suggestion
  const Particle& at(size_t index) const { return _particles.at(index); }
```

src/autopas/cells/SortedCellView.h
```diff
@@ -74,6 +74,20 @@ class SortedCellView : public ParticleCell<Particle> {
 
   std::array<double, 3> getCellLength() const override { return _cell->getCellLength(); }
 
+  /**
+   * Returns the particle at position index. Needed by SingleCellIterator.
+   * @param index the position of the particle to return.
+   * @return the particle at position index.
+   */
+  decltype(auto) getParticle(size_t index) { return _particles.at(index); }
```
SteffenSeckler: same as above

src/autopas/cells/SortedCellView.h
```diff
@@ -74,6 +74,20 @@ class SortedCellView : public ParticleCell<Particle> {
...
+
+  /**
+   * Returns the const particle at position index. Needed by SingleCellIterator.
+   * @param index the position of the particle to return.
+   * @return the particle at position index.
+   */
+  decltype(auto) getParticle(size_t index) const { return _particles.at(index); }
```
SteffenSeckler: same as above

src/autopas/containers/verletClusterLists/ClusterTower.h
```diff
@@ -0,0 +1,261 @@
...
+    return SingleCellIteratorWrapper<Particle, false>{
+        new SingleCellIterator<Particle, ClusterTower<Particle, clusterSize>, false>(this)};
+  }
+
+  /**
+   * Returns the particle at position index. Needed by SingleCellIterator.
+   * @param index the position of the particle to return.
+   * @return the particle at position index.
+   */
+  decltype(auto) getParticle(size_t index) { return _particles._particles.at(index); }
```
SteffenSeckler: same as above

src/autopas/containers/verletClusterLists/ClusterTower.h
```diff
@@ -0,0 +1,261 @@
...
+
+  /**
+   * Returns the const particle at position index. Needed by SingleCellIterator.
+   * @param index the position of the particle to return.
+   * @return the particle at position index.
+   */
+  decltype(auto) getParticle(size_t index) const { return _particles._particles.at(index); }
```
SteffenSeckler: same as above

src/autopas/containers/verletClusterLists/VerletClusterLists.h
```diff
@@ -79,13 +73,13 @@ class VerletClusterLists : public ParticleContainer<FullParticleCell<Particle>>
     traversal->endTraversal();
   }
 
+  // @TODO: Somehow make the iterator also iterating over the _particlesToAdd. Otherwise, e.g. ContainerSelectorTest
```
SteffenSeckler: link issue

src/autopas/containers/verletClusterLists/VerletClusterLists.h
```diff
@@ -117,15 +114,16 @@ class VerletClusterLists : public ParticleContainer<FullParticleCell<Particle>>
...
   std::vector<Particle> updateContainer() override {
+    // @TODO What happens when some particles are just deleted here?
     AutoPasLog(debug, "updating container");
     // first delete all particles
     this->deleteHaloParticles();
 
     // next find invalid particles
     std::vector<Particle> invalidParticles;
-    /// @todo: parallelize
+    // @todo: parallelize
```
SteffenSeckler: these todos all need the `///` to make them appear in the doxygen doc

tests/testAutopas/tests/containers/LinkedCellsVersusVerletClusterListsTest.cpp
```diff
@@ -41,19 +41,17 @@ void LinkedCellsVersusVerletClusterListsTest::test(unsigned long numMolecules, d
   _verletLists.iteratePairwise(verletTraversal.get());
   _linkedCells.iteratePairwise(&traversalLinkedLJ);
 
-  std::vector<std::array<double, 3>> forcesVerlet(numMolecules), forcesLinked(numMolecules);
+  std::vector<std::array<double, 3>> forcesVerlet{numMolecules}, forcesLinked{numMolecules};
   // get and sort by id, skip id=0 to avoid dummy particles
```
SteffenSeckler: remove comment

src/autopas/containers/verletClusterLists/VerletClusterLists.h
```diff
@@ -140,45 +139,54 @@ class VerletClusterLists : public ParticleContainer<FullParticleCell<Particle>>
...
     return ParticleIteratorWrapper<Particle, false>();
   }
 
-  void rebuildNeighborLists(TraversalInterface *traversal) override { rebuild(traversal->getUseNewton3()); }
+  void rebuildNeighborLists(TraversalInterface *traversal) override {
+    internal::VerletClusterListsRebuilder<Particle> builder{*this, _particlesToAdd, traversal->getUseNewton3()};
+    auto res = builder.rebuild();
+    std::tie(_towerSideLength, _numTowersPerInteractionLength, _towersPerDim, _numClusters, _neighborListIsNewton3) = {
+        res._towerSideLength, res._interactionLengthInTowers, res._towersPerDim, res._numClusters,
+        res._neighborListIsNewton3};
```
FG-TUM: probably better to return a tuple from `builder.rebuild()` and assign them directly

src/autopas/containers/verletClusterLists/VerletClusterLists.h
```diff
@@ -189,576 +197,197 @@ class VerletClusterLists : public ParticleContainer<FullParticleCell<Particle>>
...
    */
-  auto getCellsPerDimension() const { return _cellsPerDim; }
+  auto getTowersPerDimension() const { return _towersPerDim; }
 
   /**
    * Returns the 2D grid for the XY-plane of this container that defines the cluster towers.
    * @return the grids of this container for usage in traversals.
    */
-  auto &getGrids() { return this->_cells; }
+  auto &getTowers() { return _towers; }
```
FG-TUM: Mention in the doc that this returns a ref! (also applies to other getters)

src/autopas/containers/verletClusterLists/VerletClusterLists.h
```diff
@@ -117,6 +115,7 @@ class VerletClusterLists : public ParticleContainer<FullParticleCell<Particle>>
    */
   AUTOPAS_WARN_UNUSED_RESULT
   std::vector<Particle> updateContainer() override {
+    /// @todo What happens when some particles are just deleted here?
     AutoPasLog(debug, "updating container");
```
FG-TUM: remove this log message

src/autopas/containers/verletClusterLists/VerletClusterListsRebuilder.h
```diff
@@ -0,0 +1,413 @@
...
+     */
+    bool _neighborListIsNewton3;
+  };
+  /**
+   * Rebuilds the towers, clusters, and neighbor lists. Clusters are filled with dummies as described in
+   * ClusterTower::fillUpWithDummyParticles.
+   *
+   * @return all new values for VerletClusterLists member variables.
+   */
+  BuildingReturnValues rebuild() {
```
FG-TUM: see above

src/autopas/containers/verletClusterLists/VerletClusterListsRebuilder.h
```diff
@@ -0,0 +1,413 @@
...
+   */
+  void calculateNeighborsForTowerPair(internal::ClusterTower<Particle, clusterSize> &tower,
+                                      internal::ClusterTower<Particle, clusterSize> &neighborTower,
+                                      double distBetweenTowersXYsqr) {
+    const bool isSameTower = &tower == &neighborTower;
+    for (size_t towerIndex = 0; towerIndex < tower.getNumClusters(); towerIndex++) {
+      auto startIndexNeighbor = _neighborListIsNewton3 and isSameTower ? towerIndex + 1 : 0;
+      auto &towerCluster = tower.getCluster(towerIndex);
+      double towerClusterBoxBottom = towerCluster.at(0).getR()[2];
+      double towerClusterBoxTop = towerCluster.at(clusterSize - 1).getR()[2];
```
FG-TUM: are you using `at()` here on purpose?

FG-TUM: same further below

src/autopas/containers/verletClusterLists/traversals/ClusterFunctor.h
```diff
@@ -0,0 +1,72 @@
...
+  /**
+   * Traverses pairs of all particles in the given cluster. Always uses newton 3 in the AoS data layout.
+   * @param cluster The cluster to traverse.
+   */
+  void traverseCluster(internal::Cluster<Particle, clusterSize> &cluster) {
+    if constexpr (dataLayout == DataLayoutOption::aos) {
+      for (size_t i = 0; i < clusterSize; i++) {
+        // Always use newton 3 for interactions within one cluster.
+        for (size_t j = i + 1; j < clusterSize; j++) {
+          _functor->AoSFunctor(cluster.at(i), cluster.at(j), true);
```
FG-TUM: especially in functions that are incredibly often called we really should consider using `[]` over `at()`. Is there a reason why `at()` should stay here? If no change it please.

src/autopas/containers/verletClusterLists/traversals/ClusterFunctor.h
```diff
@@ -0,0 +1,72 @@
...
+   * Traverses all pairs of particles between two clusters.
+   * @param cluster The first cluster.
+   * @param neighborCluster The second cluster.
+   */
+  void traverseClusterPair(internal::Cluster<Particle, clusterSize> &cluster,
+                           internal::Cluster<Particle, clusterSize> &neighborCluster) {
+    if constexpr (dataLayout == DataLayoutOption::aos) {
+      for (size_t i = 0; i < clusterSize; i++) {
+        for (size_t j = 0; j < clusterSize; j++) {
+          _functor->AoSFunctor(cluster.at(i), neighborCluster.at(j), useNewton3);
```
FG-TUM: see above

src/autopas/containers/verletClusterLists/traversals/VerletClustersTraversal.h
```diff
@@ -44,139 +47,31 @@ class VerletClustersTraversal : public TraversalInterface,
...
-      for (int i = 0; i < clusterSize; i++) {
-        cell.addParticle(*(clusterStart + i));
-      }
-      SoA<typename Particle::SoAArraysType> &soa = _clusterSoAs[currentClusterIndex];
-      soa.resizeArrays(clusterSize);
-      _functor->SoALoader(cell, soa);
-    };
-
-    clusterList.template traverseClusters<true>(_clusterTraverseFunctor);
+    clusterList.loadParticlesIntoSoAs(_functor);
```
FG-TUM: why not just:

```
if (dataLayout = DataLayoutOption::soa) {
   VerletClustersTraversalInterface<Particle>::_verletClusterLists->loadParticlesIntoSoAs(_functor);
}
```
same for `endTraversal()`

src/autopas/containers/verletClusterLists/VerletClusterLists.h
HobbyProgrammer: Hm, I am not sure about this. I also thought about doing this, but since it has 5 return values, I thought having an extra struct that contains all these values with names and also with documentation increases readability.

But I can of course change it, if you want!

src/autopas/containers/verletClusterLists/traversals/VerletClustersTraversal.h
HobbyProgrammer: Good point! I think I changed that part too often and this resulted in this code.

HobbyProgrammer:
@FG-TUM Regarding the calls to the `at()` function: I think you mixed up this `at()` function that is implemented in the `Cluster` class with the `at()` function of `std::vector` or something like that. The function called there actually does not include a check for the boundaries of the cluster.

Before, the method was called `getParticle()`. @SteffenSeckler changed it to `at()` earlier. For this particular method that does not include a check of the boundaries, it might make sense to reverse it or implement the `[]`-operator to avoid this confusion. Just tell me what you want to have!

SteffenSeckler:
yep, the at() function of the cluster should probably be an `operator []`

FG-TUM:
True, I have misinterpreted the  `Cluster.at()` calls since, I thought that your `at()` functions would call `vector.at()` at some point, as is the case with `ClusterTower.at()`. Therefore, I would suggest implementing a `[]` operator for `Cluster` and `ClusterTower` that then also only uses the `vector[]`.

src/autopas/containers/verletClusterLists/Cluster.h
```diff
@@ -0,0 +1,83 @@
...
+  /**
+   * @copydoc operator[](size_t)
+   */
+  const Particle &operator[](size_t index) const { return *(_firstParticle + index); }
+
+  /**
+   * Returns the SoAView for this cluster.
+   * @return the SoAView for this cluster.
+   */
+  auto &getSoAView() { return _soaView; }
```
SteffenSeckler: getSoAView should actually return an object and not a reference (Views should always be lightweight and be copied)

src/autopas/containers/verletClusterLists/ClusterTower.h
```diff
@@ -0,0 +1,261 @@
...
+   * Loads the particles into the SoA stored in this tower and generates the SoAView for each cluster.
+   * @tparam Functor The type of the functor to use.
+   * @param functor The functor to use for loading the particles into the SoA.
+   */
+  template <class Functor>
+  void loadSoA(Functor *functor) {
+    functor->SoALoader(_particles, _particles._particleSoABuffer);
+    for (size_t index = 0; index < getNumClusters(); index++) {
+      auto &cluster = getCluster(index);
+      cluster.getSoAView() = {&(_particles._particleSoABuffer), index * clusterSize, (index + 1) * clusterSize};
```
SteffenSeckler: don't do this please.
Write a proper setter and use it instead!

PR #328, opened 2019-08-05
AVX Functor shortcut

FG-TUM:
# Description

In the SoAKernel if all four force calculations are masked away skip the calculation. 

## ~Resolved Issues~

# How Has This Been Tested?

- [x] existing unit tests


FG-TUM:
@LG-TUM this should interest you

PR #331, opened 2019-08-12
VerletClusterLists Static Scheduling Traversal

HobbyProgrammer:
# Description

Adds a new traversal for the VerletClusterLists container.

It works pretty similar to the standard verlet-clusters traversal, but partitions all cluster pairs as equally as possible without dividing the neighbors for one cluster on rebuilding the neighbor lists.
This partitioning is then used for traversing with static scheduling.

The new traversal supports only NoNewton3, but AoS and SoA.

## Related Pull Requests

- Based on #319

## Performance Results
 - Strong Scaling is almost equal to verlet-clusters
 - SoA varies
 - With AoS and gaussian distribution (mean=10.5, stddev=5.5), it is better with less particles and equal with many particles

_Domain Size is 25 with cutoff 1 and skin 0.2._
**On mpp2:**
VERLET~2
VE2525~1
VERLET~4
VE75C8~1

**On mpp3:**
VERLET~1
VE2525~1
VERLET~3
VE75C8~1

# How Has This Been Tested?

- New unit tests in `LinkedCellsVersusVerletClusterListsTest.cpp` that compare the results of the new traversal with LinkedCells.


SteffenSeckler:
I have merged the master into this

HobbyProgrammer:
Ah, thanks. I also did this locally, but I struggled with the maxEvidence, so the tests wouldn't run through.

src/autopas/containers/verletClusterLists/VerletClusterLists.h
```diff
@@ -348,6 +383,69 @@ class VerletClusterLists : public ParticleContainer<FullParticleCell<Particle>>
...
+    constexpr int minNumClusterPairsPerThread = 1000;
+    auto numThreads =
+        std::clamp(static_cast<int>(numClusterPairs / minNumClusterPairsPerThread), 1, autopas_get_max_threads());
+
+    size_t numClusterPairsPerThread = numClusterPairs / numThreads;
+    fillClusterRanges(numClusterPairsPerThread, numThreads);
+  }
+
+  /**
+   * Fills in the cluster ranges of the cluster thread partition.
```
SteffenSeckler: Extend this description a bit. What does it actually do?
sth. like "gives each thread an appropriate range of clusters, s.t., the number of cluster pairs is similar."

SteffenSeckler:
seems like there is an error when using at least 23 threads.

HobbyProgrammer:
I already know about this. This is the last thing I have to do, fix this bug. As far as I know, it is in the static cluster-to-thread distribution, when too many threads are used.

SteffenSeckler:
I fixed that error:
* Added sanity check if the last threads don't get any work (this happened before)
* Added better checking using `numClusterPairsTotal` instead of `numClusterPairsThisThread` which should give the last thread more clusters on average.

tests/testAutopas/tests/containers/LinkedCellsVersusVerletClusterListsTest.cpp
```diff
@@ -71,6 +71,7 @@ void LinkedCellsVersusVerletClusterListsTest::test(unsigned long numMolecules, d
       traversalOption, flopsVerlet, _verletLists.getTraversalSelectorInfo(), dataLayout,
       useNewton3 ? autopas::Newton3Option::enabled : autopas::Newton3Option::disabled);
 
+  _verletLists.rebuildNeighborLists(&*traversalFLOPSVerlet);
```
FG-TUM:
```suggestion
  _verletLists.rebuildNeighborLists(traversalFLOPSVerlet.get());
```

tests/testAutopas/tests/containers/LinkedCellsVersusVerletClusterListsTest.cpp
```diff
@@ -71,6 +71,7 @@ void LinkedCellsVersusVerletClusterListsTest::test(unsigned long numMolecules, d
...
   _verletLists.iteratePairwise(&*traversalFLOPSVerlet);
```
FG-TUM:
```suggestion
  _verletLists.iteratePairwise(traversalFLOPSVerlet.get());
```

tests/testAutopas/tests/selectors/AutoTunerTest.cpp
```diff
@@ -83,14 +83,14 @@ TEST_F(AutoTunerTest, testAllConfigurations) {
     ++iterations;
     ++collectedSamples;
     auto currentConfig = autoTuner.getCurrentConfig();
-
+    std::cout << "currentConfig:" << currentConfig.toString() << std::endl;
```
FG-TUM:
```suggestion

```

SteffenSeckler:
@HobbyProgrammer can you resolve the remaining review comments?
mainly this includes the `getTowers()` function. it should return a `const &` instead of a `&`. additionally, there should be sth. like a setter that the VerletClusterListsRebuilder can use.
Alternatively, store the towers somewhere else.

HobbyProgrammer:
@SteffenSeckler Yes!

src/autopas/containers/verletClusterLists/VerletClusterLists.h
```diff
@@ -348,6 +383,71 @@ class VerletClusterLists : public ParticleContainer<FullParticleCell<Particle>>
...
+    size_t numClusterPairs = 0;
+    this->template traverseClusters<false>(
+        [&numClusterPairs](auto &cluster) { numClusterPairs += cluster.getNeighbors().size(); });
+
+    constexpr int minNumClusterPairsPerThread = 1000;
+    auto numThreads =
+        std::clamp(static_cast<int>(numClusterPairs / minNumClusterPairsPerThread), 1, autopas_get_max_threads());
+
+    size_t numClusterPairsPerThread = numClusterPairs / numThreads;
+    fillClusterRanges(numClusterPairsPerThread, numThreads);
```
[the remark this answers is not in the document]
HobbyProgrammer: @FG-TUM Something else I noticed while thinking about this:
At the moment, the side length of a cluster only depends on the domain size and the number of particles, not at all on the interaction length. This seems wrong!

HobbyProgrammer:
Regarding the `getTowers()` method, I opted for removing it completely, as a const version alone wasn't really necessary. Before, this method was the only way for traversals to iterate over the clusters themselves. Alternatively, they could use the traverse methods provided by the `VerletClusterLists` class. This was sufficient for all traversals, but not for the verlet-clusters-static traversal.

Now, I solved this analog to the Linked Cells. I just pass the towers to the `VerletClustersTraversalInterface` next to the container instance, similar to how linked cells traversals access the cells from the Linked Cells container.

PR #335, opened 2019-08-22
Add missing namespaces to namespaces.h

Garfield96:
# Description

See title.

# How Has This Been Tested?

nothing to test

### DIFF
--- a/src/autopas/namespaces.h
+++ b/src/autopas/namespaces.h
@@ -15,7 +15,7 @@
 namespace autopas {
 
 /**
- * Namespace to handle mathematical operations of floating point std::array's.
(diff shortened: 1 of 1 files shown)


src/autopas/namespaces.h
```diff
@@ -19,12 +19,22 @@ namespace autopas {
  */
 namespace ArrayMath {}
 
+/**
+ * In this namespace some helper functions for floating point std::array's can be found.
```
FG-TUM:
```suggestion
 * In this namespace some helper functions for std::array can be found.
```

No (any more) restricted to floating point! :) please also adapt ArrayMath description.

PR #336, opened 2019-08-22
Documentation Linked-Cells

Garfield96:
# Description

This PR introduces various graphics created in the context of my Bachelor's Thesis.

# How Has This Been Tested?

nothing to test


Garfield96:
> 
> 
> if doxygen only needs the pngs is there a reason to save the pdfs?

There is no reason which is related to Doxygen. I thought it might be helpful to also include a PDF version since this can be used for integration into (Latex) documents.  But I can remove all PDFs if you want.

FG-TUM:
well we can always create them from the tex files so I think we should remove the pdfs.

PR #348, opened 2019-09-06
Simpler Option Handling

### LINKED ISSUE #368: Newton3OnOff Test does not test DataLayout=Cuda
**Describe the bug**
Newton3OnOff Test does not test DataLayout=Cuda. Instead, it tests two times for SoA.

**To Reproduce**
1. compile with cuda
2. run `ctest -R 'Newton3.*Direct.*Cuda.*' --verbose` to check which test is with cuda... (PR #348 will make this more verbose). Lets assume /2 is DirectSum + Cuda.
3. `gdb --args tests/testAutopas/runTests --gtest_filter='*Newton3OnOff*/2'`
4. `break Newton3OnOffTest.cpp:25`
5. 
```
p dataLayoutStr
$1 = "structure-of-arrays-on-cuda-capable-device"
``` 
6.
```
p dataLayoutOption
$2 = autopas::soa
``` 

**Expected behavior**
```
p dataLayoutOption
$2 = autopas::cuda
``` 

~**Screenshots**~

* OS: Linux
* Compiler: g++-9 + nvcc 10.1
* Build System: make

**Additional context**
* PR #348 will change the parsing quite a bit...
* When choosing the correct DataLayout the tests crash here: https://github.com/AutoPas/AutoPas/blob/5890c412a88aa5d601044c227ec2f183458022a2/src/autopas/pairwiseFunctors/Functor.h#L219

FG-TUM:
# Description

At the moment if a new option is added many things need to be done: Adding an enum, adding a case in the respective to_string(), and defining a new case in the ever-growing if tree of the parse function.

What this PR does:

- [x] replace if trees in parse functions by Needleman-Wunsch algorithm (scores string alignment).
- [x] move string representations to option files
- [ ] use `string_view`s instead of `string`s (this might introduce problems with the intel compiler)
- [x] update Contribution guidelines

## Related Pull Requests

- expects #360 to be merged

## Resolved Issues

- [ ] #193 <- this cannot be fixed by the listed options, because the intel compiler does not support anything.
- [x] closes #368 <- now strings are properly parsed and the correct enum is selected for cuda.

# How Has This Been Tested?

- [x] old unit tests
- [x] new unit tests


src/autopas/options/Option.h
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  /**
+   * Converts a string of options to a set of enums. For best results, the options are expected to be lower case.
+   *
+   * Allowed delimiters can be found in autopas::utils::StringUtils::delimiters.
+   * Possible options can be found in getAllOptions().
+   *
+   * This function uses the Needleman-Wunsch algorithm to find the closest matching options.
+   * If an option is ambiguous an execption is thrown.
```
SteffenSeckler:
```suggestion
   * If an option is ambiguous an exception is thrown.
```

src/autopas/options/Option.h
```diff
@@ -0,0 +1,133 @@
...
+    auto match = mapOptNames.find(actualThis);
+    if (match == mapOptNames.end()) {
+      return "Unknown Option (" + std::to_string(actualThis) + ")";
+    } else {
+      return match->second;
+    }
+  }
+
+  /**
+   * Converts a string of options to a set of enums. For best results, the options are expected to be lower case.
```
SteffenSeckler:
```
For best results, the options are expected to be lower case.
```
As far as I see this, a better description would be sth. like:
```
The algorithm is case-insensitive.
```
Or sth. like this, as you are anyways transforming all options to lowercase.

tests/testAutopas/tests/autopasInterface/Newton3OnOffTest.cpp
```diff
@@ -78,17 +77,18 @@ INSTANTIATE_TEST_SUITE_P(
             }()),
             ValuesIn([]() {
               std::vector<std::string> ret;
-              std::transform(
-                  autopas::allDataLayoutOptions.begin(), autopas::allDataLayoutOptions.end(), std::back_inserter(ret),
-                  [](autopas::DataLayoutOption d) -> std::string { return autopas::utils::StringUtils::to_string(d); });
+              auto allDataLayoutOptions = autopas::DataLayoutOption::getAllOptions();
```
SteffenSeckler: here, you should use the actual options for the parametrization and not the string representations of them.

tests/testAutopas/tests/autopasInterface/Newton3OnOffTest.cpp
```diff
@@ -17,11 +17,10 @@ TEST_P(Newton3OnOffTest, countFunctorCallsTest) {
   auto contTravStr = std::get<0>(GetParam());
   auto dataLayoutStr = std::get<1>(GetParam());
 
-  transform(contTravStr.begin(), contTravStr.end(), contTravStr.begin(), ::tolower);
-  transform(dataLayoutStr.begin(), dataLayoutStr.end(), dataLayoutStr.begin(), ::tolower);
-  auto containerOption = autopas::utils::StringUtils::parseContainerOptions(contTravStr).begin().operator*();
-  auto traversalOption = autopas::utils::StringUtils::parseTraversalOptions(contTravStr).begin().operator*();
-  auto dataLayoutOption = autopas::utils::StringUtils::parseDataLayout(dataLayoutStr).begin().operator*();
+  auto contTravStrVector = autopas::utils::StringUtils::tokenize(contTravStr, "+");
```
SteffenSeckler: make this unneeded, by using the actual options and not the string representations inside of the parametrization

SteffenSeckler:
```
::testing::WithParamInterface<std::tuple<std::string, std::string>>
```
will become sth. like:
```
::testing::WithParamInterface<std::tuple<ContainerOption, TraversalOption>>
```

tests/testAutopas/tests/autopasInterface/Newton3OnOffTest.h
```diff
@@ -56,4 +56,17 @@ class Newton3OnOffTest : public AutoPasTestBase,
    */
   template <bool useNewton3, class Container, class Traversal>
   std::pair<size_t, size_t> eval(autopas::DataLayoutOption dataLayout, Container &container, Traversal traversalOption);
+
+  struct PrintToStringParamName {
+    template <class ParamType>
+    std::string operator()(const testing::TestParamInfo<ParamType> &info) const {
```
SteffenSeckler: only in this function you will need the toString() functions of the option objects.

tests/testAutopas/tests/options/OptionTest.cpp
```diff
@@ -0,0 +1,108 @@
...
+#include "autopas/options/DataLayoutOption.h"
+#include "autopas/options/SelectorStrategyOption.h"
+#include "autopas/options/TraversalOption.h"
+#include "autopas/options/TuningStrategyOption.h"
+#include "tests/utils/StringUtilsTest.h"
+
+// parseOptions tests
+// these tests shall not (yet :) be generated since we here want to pass strings that do not match exactly.
+
+TEST(OptionTest, parseTraversalOptionsTest) {
```
SteffenSeckler: maybe comment what these tests do?

tests/testAutopas/tests/options/OptionTest.cpp
```diff
@@ -0,0 +1,108 @@
+/**
```
SteffenSeckler: OptionTest: maybe add a test:
```
parseOptions(option.toString()) == option
```

src/autopas/containers/verletClusterLists/VerletClusterCellsParticleIterator.h
```diff
@@ -86,7 +86,8 @@ class VerletClusterCellsParticleIterator : public ParticleIteratorInterfaceImpl<
   bool isValid() const override { return _cellId < _vectorOfCells->size(); }
 
   /**
-   * @copydoc ParticleIteratorInterface::deleteCurrentParticle()
+   * Deletes the current particle.
+   * @note This function is disabled for const iterators.
```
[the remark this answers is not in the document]
FG-TUM: revertred back to copydoc. Seems to be working again....

src/autopas/options/Option.h
```diff
@@ -0,0 +1,133 @@
+/**
+ * @file Option.h
+ * @author F. Gratl
+ * @date 10/14/19
```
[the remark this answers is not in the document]
FG-TUM: true but not here. This really annoys me...

src/autopas/options/Option.h
FG-TUM: The algorithm actually is case sensitive. However, since the idea for the new parsing process is that the user does not need to know exactly how anything is spelled we compare only to lowercase. The argument `optionsString` however, is not transformed to lowercase since this would be redundant most of the time. Therefore the algorithm will probably work when given a mixed-case string but not as good as when presented with a lower case one.
Example: when given `AoS` it is unclear if `soa` or `aos` is meant.

SteffenSeckler: oh right, this looked different to me, as you were doing that lowercase transformation twice. so i thought you did this also for the input...

PR #360, opened 2019-10-15
Bayesian Search: Change FeatureVector diff

FG-TUM:
# Description

Changes the return type of the - operator of FeatureVector to Eigen::VectorXd. This seems reasonable since the previously returned feature vector was only used to be cast to a Eigen::VectorXd. Also with #348 the construction of invalid options will not be possible anymore.

## Related Pull Requests

- #348

## ~Resolved Issues~

# How Has This Been Tested?

- [ ] Tested by Jan :)

### DIFF
--- a/src/autopas/selectors/FeatureVector.h
+++ b/src/autopas/selectors/FeatureVector.h
@@ -48,7 +48,10 @@ class FeatureVector : public Configuration {
   FeatureVector(Configuration conf) : Configuration(conf) {}
 
   /**
-   * Distance between two FeatureVectors
+   * Distance between two FeatureVectors.
+   * Since there is no real ordering all discrete options are assumed to have a distance 
+   * of one to each other.
+   * This function ignores the container dimension since it is encoded in the traversal.
    * @param other
    * @return
    */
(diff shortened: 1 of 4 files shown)


src/autopas/selectors/FeatureVector.h
```diff
@@ -52,13 +52,12 @@ class FeatureVector : public Configuration {
...
    */
-  FeatureVector operator-(const FeatureVector &other) const {
-    ContainerOption co = ContainerOption((traversal == other.traversal) ? 0 : 1);
-    double cfs = (cellSizeFactor - other.cellSizeFactor);
-    TraversalOption to = TraversalOption((traversal == other.traversal) ? 0 : 1);
-    DataLayoutOption dlo = DataLayoutOption((dataLayout == other.dataLayout) ? 0 : 1);
-    Newton3Option n3o = Newton3Option((newton3 == other.newton3) ? 0 : 1);
-    return FeatureVector(co, cfs, to, dlo, n3o);
+  Eigen::VectorXd operator-(const FeatureVector &other) const {
+    Eigen::VectorXd result(featureSpaceDims);
```
SteffenSeckler: add a comment why we don't care about the container option and why the difference is always 1

FG-TUM: done

PR #370, opened 2019-11-04
Fmm example

JoachimMarin:
# Description

Implements the fast multipole method for 1/r for a number of randomly generated particles in a certain domain.
The fast multipole method operates on an adaptive octree.

## ~~Resolved Issues~~

# How Has This Been Tested?

- [ ] The results of the fast multipole method are compared to the exact results for each particle


examples/fmm/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+        "*.h"
+        "../../tests/testAutopas/testingHelpers/RandomGenerator.cpp"
+)
+
+add_executable(fmm ${MY_SRC})
+
+target_link_libraries(fmm autopas)
+
+# add the test to checkExamples
+add_dependencies(checkExamples fmm)
```
SteffenSeckler: you might actually want to add a test, which might also be a run of the fmm executable.

You might orient yourself on the other tests, e.g.,
The test will only pass if 0 is returned from COMMAND, so you might also use grep or sth. else to add some more functionality.
```
# -----------------test-----------------
# add check for current target
# cmake-format: off
add_test(
    NAME fmm.test
    COMMAND fmm
    CONFIGURATIONS checkExamples
)
# cmake-format: on
```

examples/fmm/FmmParticle.h
```diff
@@ -0,0 +1,24 @@
...
+
+class FmmParticle : public autopas::Particle {
+ public:
+  double charge;
+  double resultFMM = 0;
+  double resultExact = 0;
+  double longRange = 0;
+  double shortRange = 0;
+
+  FmmParticle(std::array<double, 3> r, std::array<double, 3> v, int id, double charge) : ParticleBase(r, v, id) {
```
SteffenSeckler:
```suggestion
  FmmParticle(std::array<double, 3> r, std::array<double, 3> v, int id, double charge) : ParticleBase(r, v, id), charge(charge) {
```

examples/fmm/FmmParticle.h
```diff
@@ -0,0 +1,24 @@
...
+    this->charge = charge;
```
SteffenSeckler:
```suggestion
```

examples/fmm/Math3D.cpp
```diff
@@ -0,0 +1,179 @@
...
+  }
+  if (n == m + 1) {
+    return x * (2 * m + 1) * associatedLegendrePolynomialRec(m, m, x);
+  }
+
+  ret = x * (2 * n - 1) * associatedLegendrePolynomialRec(m, n - 1, x) -
+        (n + m - 1) * associatedLegendrePolynomialRec(m, n - 2, x);
+  ret /= (n - m);
+  return ret;
+}
```
Garfield96: C++17 implements this in `cmath`:
[https://en.cppreference.com/w/cpp/numeric/special_math/assoc_legendre](https://en.cppreference.com/w/cpp/numeric/special_math/assoc_legendre)
I don't know if the current version of the standard library used in Jenkins already implements this.

SteffenSeckler: according to:
https://en.cppreference.com/w/cpp/compiler_support#cpp17
(Mathematical special functions)
it should be there since libstdc++7 and thus also on the jenkins nodes.

FG-TUM:
Closed because FMM integration is not easily doable and maintaining this example seems not beneficial. See #4 for discussion.

PR #380, opened 2019-11-19
Look for system libs

FG-TUM:
# Description

For all dependencies:
- [x] Look for an already installed version
- [x] If nothing is found or version is too old use the bundled zip
- [x] Switch for forcing the usage of the bundled version

## ~Resolved Issues~

# How Has This Been Tested?

- [ ] Jenkins


examples/md-flexible/tests/TimeDiscretizationTest.cpp
```diff
@@ -126,8 +125,12 @@ void TimeDiscretizationTest::Pos_and_Velo_Test(
...
       EXPECT_EQ(iter->getID(), i);
-      EXPECT_EQ(iter->getV(), nextVelocity(oldVelocityValues.at(i), velForces.at(i), forces.at(i), particleD));
-      EXPECT_EQ(iter->getR(), nextPosition(oldPositionValues.at(i), forces.at(i), oldVelocityValues.at(i), particleD));
+      auto expectedV = nextVelocity(oldVelocityValues.at(i), velForces.at(i), forces.at(i), deltaT);
+      auto expectedR = nextPosition(oldPositionValues.at(i), forces.at(i), oldVelocityValues.at(i), deltaT);
+      for (int dim = 0; dim < 3; ++dim) {
+        EXPECT_NEAR(iter->getV()[dim], expectedV[dim], 1e-13);
+        EXPECT_NEAR(iter->getR()[dim], expectedR[dim], 1e-13);
+      }
       i++;
```
FG-TUM: TBH this test looks sketchy. I will rework it in a different branch but this version here at least passes...

SteffenSeckler: did this not pass before this change?

cmake/modules/autopas_yaml-cpp.cmake
```diff
@@ -1,19 +1,40 @@
-message(STATUS "yaml-cpp - using bundled version")
+option(yaml-cpp_ForceBundled "Do not look for an installed version, always use bundled." OFF)
+
+if (NOT ${yaml-cpp_ForceBundled})
+    set(expectedVersion 0.5.2)
+    # first try: check if we find any installed version
+    find_package(yaml-cpp ${expectedVersion} QUIET)
+    if (yaml-cpp_FOUND)
+        message(STATUS "yaml-cpp - using installed system version ${yaml-cpp_VERSION}")
+        return()
```
SteffenSeckler:
```suggestion
        // return here, as we have now found and imported the target.
        return()
```

cmake/modules/autopas_eigen.cmake
```diff
@@ -1,17 +1,41 @@
...
+
+if (NOT ${Eigen3_ForceBundled})
+    set(expectedVersion 3.3.4)
+    # capital E actually required...
+    find_package(Eigen3 ${expectedVersion} QUIET)
+    # actually I don't know our minimal supported version but this is the one I tested.
+    if (Eigen3_FOUND)
+        message(STATUS "Eigen3 - using installed system version ${Eigen3_VERSION}")
+        # to later add the alias Eigen3::Eigen needs to be promoted to global visibility
+        set_target_properties(Eigen3::Eigen PROPERTIES "IMPORTED_GLOBAL" "TRUE")
```
SteffenSeckler: this is needed because we include this cmake file in the `lib` subdirectory, correct?

cmake/modules/autopas_eigen.cmake
```diff
@@ -1,17 +1,41 @@
...
+        # we need to alias this because aparently make sometimes breaks on '::'
+        add_library(Eigen3 ALIAS Eigen3::Eigen)
+        return()
```
SteffenSeckler:
```suggestion
        // return, as we have found the target
        return()
```

examples/md-flexible/tests/TimeDiscretizationTest.cpp
FG-TUM: failed for me once while testing because it compared doubles via ==

cmake/modules/autopas_eigen.cmake
FG-TUM: True. Targets are by default only visible in the directory they are created/imported.

PR #386, opened 2019-12-02
Jenkins: enable ccache

SteffenSeckler:
# Description

- enable ccache for all Jenkins builds.


SteffenSeckler:
1:40 -> 1:11 if mostly cache hits (intel and gpu versions were not yet cached)

PR #394, opened 2020-01-09
ContainerSelector use owned state instead of try-catch

SteffenSeckler:
# Description

ContainerSelector: when changing a container use the owned state to specify whether to add a particle as owned or halo particle instead of using try-catch exception workflow.


SteffenSeckler:
closed in favor of including this to #395

PR #395, opened 2020-01-09
fixes #393: handling of boundary conditions for md-flexible

### LINKED ISSUE #393: VerletClusterCells::rebuild() creates segfault when particles are too far outside the domain
**Describe the bug**
In VerletClusterCells::rebuild() when particles are too far outside the domain, an invalid cell index is calculated resulting in a segmentation fault.

**To Reproduce**
Execute md-flexible with VerletClusterCells as only container option.
`./md-flexible --container VerletClusterCells`

SteffenSeckler:
# Description

- enables periodic boundary conditions by default for MDFlexConfig.
- fixes: calculate position update first, then boundary update.
- enables an error message if no boundary condition is set.
- ContainerSelector: changing a container now uses the owned state to specify whether to add a particle as owned or halo particle instead of using try-catch exception workflow.
- introduce isAppropriateClusterSize() and use them in VerletClusterCells.

## Resolved Issues

- [x] fixes #393 

# How Has This Been Tested?

- [x] Adds an additional test for md-flex to checkExamples for VerletClusterCells.
- [x] Adds new unit test in IteratorTest.cpp that compares the iterators with and without openmp for all containers.

### DIFF
--- a/src/autopas/containers/verletListsCellBased/verletLists/VerletListHelpers.h
+++ b/src/autopas/containers/verletListsCellBased/verletLists/VerletListHelpers.h
@@ -65,6 +65,10 @@ class VerletListHelpers {
       return true;
     }
 
+    bool isAppropriateClusterSize(unsigned int clusterSize, DataLayoutOption::Value dataLayout) const override {
+      return false;  // this functor shouldn't be called with clusters!
+    }
+
     void AoSFunctor(Particle &i, Particle &j, bool /*newton3*/) override {
       auto dist = utils::ArrayMath::sub(i.getR(), j.getR());
 
@@ -249,6 +253,11 @@ class VerletListHelpers {
       utils::ExceptionHandler::exception(
           "VerletListGeneratorFunctor::allowsNonNewton3() is not implemented, because it should not be called.");
       return true;
+    }
+
+    bool isAppropriateClusterSize(unsigned int clusterSize,
+                                  autopas::DataLayoutOption::Value dataLayout) const override {
+      return false;  // this functor shouldn't be used with clusters!
     }
 
     void AoSFunctor(Particle &i, Particle &j, bool newton3) override {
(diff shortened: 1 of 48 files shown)


src/autopas/containers/verletListsCellBased/verletLists/VerletListHelpers.h
```diff
@@ -251,6 +255,11 @@ class VerletListHelpers {
       return true;
     }
 
+    bool isAppropriateClusterSize(unsigned int clusterSize,
+                                  autopas::DataLayoutOption::Value dataLayout) const override {
```
FG-TUM:
```suggestion
                                  DataLayoutOption::Value dataLayout) const override {
```

FG-TUM: will need reformat after applying this

tests/testAutopas/tests/autopasInterface/Newton3OnOffTest.cpp
```diff
@@ -32,7 +32,7 @@ INSTANTIATE_TEST_SUITE_P(
 
               // container factory
               autopas::ContainerSelector<Particle, FPCell> containerSelector({0., 0., 0.}, {10., 10., 10.}, 1.);
-              autopas::ContainerSelectorInfo containerInfo(1., 0.);
+              autopas::ContainerSelectorInfo containerInfo(1., 0., 64);
```
FG-TUM: is 64 really what we want here?

SteffenSeckler: it's sth. that works with cuda, so yes !

SteffenSeckler: and it was the previous default value.

PR #401, opened 2020-01-15
Header cleanup

### LINKED ISSUE #324: Remove unnecessary includes
A lot of files include headers which are not used. This leads to unnecessary dependencies and increases the build time.

Remove all unnecessary includes (especially remove `autopasIncludes.h`)

**Additional context**
Finding and resolving unused includes is a tedious task. To find unused includes, the IWYU (include-what-you-use) tool is handy:
[IWYU Website](https://include-what-you-use.org/)
[IWYU Github repo](https://github.com/include-what-you-use/include-what-you-use)

There is also a [python script ](https://github.com/include-what-you-use/include-what-you-use/blob/master/README.md#applying-fixes) which automates the cleanup (not working completely correct and some manual work is needed afterwards, but should be still faster than doing anything by hand).

SteffenSeckler:
# Description

Removes "autopasIncludes.h" and tries to remove AutoPas.h from includes, wherever possible.

In the process of this, the following was done:
- add templates for Particle and ParticleCell to both SPHFunctors
- change Timer to return nanoseconds instead of µs, the Timer is now also used by the AutoTuner.

## Resolved Issues

- [x] fixes #324

# How Has This Been Tested?

- [x] Adds an additional test for checking whether <> are used when including files from AutoPas.


src/autopas/molecularDynamics/autopasmd.h
```diff
@@ -0,0 +1,14 @@
+/**
+ * @file autopasmd.h
+ * main include file for all sph functionality
```
FG-TUM:
```suggestion
 * main include file for all molecular dynamics functionality
```

Jenkinsfile
```diff
@@ -353,5 +353,27 @@ void checkCustom() {
...
+            // if .cpp or .h files contain #include <autopas, return 2
+            sh "grep -qlrE '#include <autopas' . && exit 2 || exit 0"
+        } catch (Exception e) {
+            // change detected
+            echo 'Usage of #include <autopas...> is discouraged, please use "". Affected files:'
+            sh "grep -lrE '#include <autopas' ."
+            sh "exit 1"
+        }
+
+        // check that no file contains an #include <autopas/...>
```
FG-TUM:
```suggestion
        // check that no file contains typedef
```

Jenkinsfile
```diff
@@ -353,5 +353,27 @@ void checkCustom() {
             sh "grep -lrE '(NULL|[^_]assert)' . | grep '\\.cpp\\|\\.h'"
             sh "exit 1"
         }
+
+        // check that no file contains an #include <autopas/...>
+        try{
+            // if .cpp or .h files contain #include <autopas, return 2
```
FG-TUM:
```suggestion
            // if any files contain #include <autopas, return 2
```
We probably should restrict this to source files only

Jenkinsfile
```diff
@@ -353,5 +353,27 @@ void checkCustom() {
...
+        } catch (Exception e) {
+            // change detected
+            echo 'Usage of #include <autopas...> is discouraged, please use "". Affected files:'
+            sh "grep -lrE '#include <autopas' ."
+            sh "exit 1"
+        }
+
+        // check that no file contains an #include <autopas/...>
+        try{
+            // if .cpp or .h files contain #include <autopas, return 2
```
FG-TUM: same as above. Also: Sure we want the same exit code?

tests/testAutopas/tests/utils/SoATest.cpp
```diff
@@ -18,7 +18,7 @@ TEST_F(SoATest, SoATypeTest) {
 }
 
 TEST_F(SoATest, SoAStorageTestGet) {
-  typedef autopas::utils::SoAType<size_t, double, double, double>::Type soatype;
+  using soatype = autopas::utils::SoAType<size_t, double, double, double>::Type;
```
FG-TUM: looks like we can also move this to higher scope

PR #409, opened 2020-02-04
Cmake fetchcontent

ssauermann:
# Description

Dependencies are now included by using the cmake FetchContent module instead of the ExternalProject module. Thus targets / libraries are already known at configure time instead of only at compile time. This allows using AutoPas in other projects without workarounds simply like this:

````cmake
    include(FetchContent)

    set(autopasRepoPath https://github.com/AutoPas/AutoPas.git)
    if (GIT_SUBMODULES_SSH)
        set(autopasRepoPath git@github.com:AutoPas/AutoPas.git)
    endif ()

    FetchContent_Declare(
            autopas
            GIT_REPOSITORY ${autopasRepoPath}
            GIT_TAG origin/cmake-fetchcontent
    )

    option(AUTOPAS_BUILD_TESTS "" OFF)
    option(AUTOPAS_BUILD_EXAMPLES "" OFF)
    option(AUTOPAS_BUILD_TARGET_DOC "" OFF)
    option(AUTOPAS_ENABLE_ADDRESS_SANITIZER "" OFF)
    option(AUTOPAS_OPENMP "" ${BUILD_OMP})
    option(spdlog_ForceBundled "" ON)
    option(Eigen3_ForceBundled "" ON)
    option(yaml-cpp_ForceBundled "" ON)

    FetchContent_MakeAvailable(autopas)
````

## Related Pull Requests

None

## Resolved Issues

None

# How Has This Been Tested?

- `make all && make test && make doc_doxygen` complete successfully (forcing bundled dependencies)
- Including AutoPas via FetchContent in lammps compiles

### CODE REVIEW

**Comment by SteffenSeckler** [comment]
for now looks good, but we should take care, that ALL options are now prepended by `AUTOPAS_`.

The one thing I don't like about this solution is that you more or less have to specify all options from added libraries as `advanced`, or otherwise they will show up.

**Comment by SteffenSeckler** [comment]
will not merge as long as this does not suppress the warnings from the libraries included via fetch_content

File: cmake/modules/autopas_spdlog.cmake [hunk 1]
```diff
@@ -21,53 +21,50 @@ endif ()
...
-# make
-file(MAKE_DIRECTORY "${install_dir}/src/spdlog_bundled/include")
+FetchContent_MakeAvailable(spdlog)
+
+if(IS_DIRECTORY "${spdlog_SOURCE_DIR}")
+    set_property(DIRECTORY ${spdlog_SOURCE_DIR} PROPERTY EXCLUDE_FROM_ALL YES)
+endif()
+
+# Disable warnings
+target_compile_options(spdlog PRIVATE -w)
```
FG-TUM: this is well and fine but won't help for warnings from AutoPas files that include spdlog. We also do not want it as `PUBLIC` since this would mute all AutoPas warnings.

PR #421, opened 2020-02-14
some VerletCluster fixes

SteffenSeckler:
# Description

- verlet cluster cells iteratePairwise shouldn't rebuild the neighbor lists, 
- track the validity state of the verletclusters properly.

# How Has This Been Tested?

- [x] new test to check for an exception


tests/testAutopas/tests/containers/verletClusterCells/VerletClusterCellsTest.cpp
```diff
@@ -300,6 +300,26 @@ TEST_F(VerletClusterCellsTest, testVerletParticleLoss) {
...
+  std::array<double, 3> min = {1, 1, 1};
+  std::array<double, 3> max = {3, 3, 3};
+  double cutoff = 1.;
+  double skin = 0.2;
+  int clusterSize = 32;
+  autopas::VerletClusterCells<Particle> verletLists(min, max, cutoff, skin, clusterSize);
+
+  autopasTools::generators::RandomGenerator::fillWithParticles(verletLists, Particle(), verletLists.getBoxMin(),
+                                                               verletLists.getBoxMax(), 500);
+  autopasTools::generators::RandomGenerator::fillWithHaloParticles(verletLists, Particle(), cutoff, 50);
```
FG-TUM, line 313: I don't think we need that many particles here

src/autopas/containers/verletClusterLists/VerletClusterLists.h
```diff
@@ -716,10 +717,19 @@ class VerletClusterLists : public ParticleContainerInterface<FullParticleCell<Pa
...
   double _skin{};
 
+  /**
+   * Enum to specify the validity of this container.
+   */
+  enum class ValidityState : unsigned char {
+    invalid = 0,        // nothing is valid.
+    cellsAreValid = 1,  // only the cell structure is valid, but the lists are not.
+    listsAreValid = 2   // the lists are valid
+  };
```
FG-TUM, line 727: Just an option to think / talk about:
```
bool cellsAreValid;
bool listsAreValid;

bool containerValid() { return cellsAreValid and listsAreValid;}
```
Might be more clear? Not sure though...

SteffenSeckler: I thought about this, too.
But it makes everything less clear, as you (almost always) have to set both of these values at the same time and listsAreValid implies that cellsAreValid, so an enum ist more logical...

FG-TUM: I think my main problem with this is that `cellsAreValid` implies that the lists are not and that this fact only becomes clear if you look at the comment in line 725 but not right away when reading the code.
```suggestion
  enum class ValidityState : unsigned char {
    invalid = 0,        // nothing is valid.
    cellsAreValid = 1,  // only the cell structure is valid, but the lists are not.
    listsAreValid = 2   // the lists and cells are valid
  };
```

FG-TUM: Possible rename to clear things up:
```suggestion
  enum class ValidityState : unsigned char {
    invalid = 0,        // nothing is valid.
    cellsValidListsInvalid = 1,  // only the cell structure is valid, but the lists are not.
    cellsAndListsValid = 2   // the lists are valid
  };
```
But I think this illustrates why I would split it into two books. So what I probably would like most is having something like:
```C++
class VCLValidityState {
    bool cellsValid = false;
    bool listsValid = false;
  public:
    bool getEverythingValid() {
        return cellsValid and listsValid;
    }
    void setListsValid() {
        if (not cellsValid) {
            exception....
        } else {
            listsValid = true;
        }
    }
    // and so on....
};
```
This could prevent potential misuse, misunderstandings and provides sanity checks.

PR #429, opened 2020-02-25
various cmake fixes

SteffenSeckler:
# Description

Does some changes to cmake config of AutoPas.

- [x] build harmony on make time and not on config time.
- [x] fix and write version info of autopas.


cmake/modules/harmony.cmake
```diff
@@ -1,86 +1,58 @@
 message(STATUS "harmony - using bundled version")
 
-# Enable FetchContent CMake module
-include(FetchContent)
+# Enable ExternalProject CMake module
```
FG-TUM: explain why you use `ExternalProject` here instead of e.g. `FetchContent`

cmake/modules/harmony.cmake
```diff
@@ -1,86 +1,58 @@
...
 
-    # Bring the populated content into the build
-    add_library(
-        harmony
-        STATIC
-        IMPORTED
-        GLOBAL
-    )
+# target_link_libraries(harmony INTERFACE
+# ${CMAKE_CURRENT_BINARY_DIR}/harmony/include/lib/libharmony.a)
```
FG-TUM: if this is not needed remove it

PR #433, opened 2020-03-25
Remove unneeded includes from Bayesian and FullSearch

SteffenSeckler:
# Description

I was wondering, why the TuningsStrategyFactory was always rebuilt, this fixes that.


SteffenSeckler:
@FG-TUM review please, should be fast enough ;)

PR #440, opened 2020-04-01
Predictive Tuning

julianpelloth:
# Description

- [x] Added selectOptimalSearchSpace() - Creates the SearchSpace by adding the configurations with an optimal prediction or that have not been tested for a while.

- [x] Added calculatePredictions() - in the future it should be possible to select between different extrapolation methods. Right now there is only a function call to linePrediction().

- [x] Added linePrediction() - simple prediction that puts a line trough the last to time points and calculates the prediction with it.

- [x] Added reselectOptimalSearchSpace() - is only called if every configuration in optimalSearchSpace is invalid. Creates a new optimalSearchSpace in a similar way as selectPredictedConfigurations() but without any invalid configuration. This has to be tested if this is a efficient way or if there is a simpler solution.

- [x] Adapted reset(), populateSearchSpace() and tune() to the new tuning class

- [ ] Add PredictiveTuning in TuningStrategyFactory and iteration pointer in constructor call

## Related Pull Requests

- none

## Resolved Issues

- [x] fixes #440 

# How Has This Been Tested?

- [x] Same tests as FullSearch to make sure that the other functions still work
- [x] selectOptimalSearchSpace() - tested with four different tests. One for the main function to select the optimal prediction. One for the first iteration, that checks if the hole searchSpace is selected. And two for the selection of the right configurations, the first when another config is near the optimum and and the second when a config has not been tested the last five tuning phases.
- [x]  reselectOptimalSearchSpace() - tested with two tests. The first tests if the function finds a new optimalSeachSpace when the function is called the first time and the second tests the same but the function gets called two times in the same tuning phase.
- [x] tune() and linePrediction() are tested with the other tests.


src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -0,0 +1,370 @@
...
+    _optimalSearchSpace = _searchSpace;
+    return;
+  }
+
+  predictConfigurations();
+
+  auto optimum = std::min_element(_traversalPredictions.begin(), _traversalPredictions.end(),
+                                  [](std::pair<Configuration, size_t> a, std::pair<Configuration, size_t> b) -> bool {
+                                    return a.second < b.second;
+                                  });
```
FG-TUM, line 217: since you use this `min_element` call at least twice you could write a wrapper for readability like:
```C++
template<class M>
  auto minMapValue(M map) {
    return std::min_element(std::begin(map), std::end(map), [](auto a, auto b) -> bool {
      return a.second < b.second;
    });
  }
```
Maybe you even find a better name than `minMapValue` for better readability.

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -0,0 +1,370 @@
...
+
+void PredictiveTuning::selectPredictedConfigurations() {
+  if (_searchSpace.size() == 1 || _timer == 0 || _timer == 1) {
+    _optimalSearchSpace = _searchSpace;
+    return;
+  }
+
+  predictConfigurations();
+
+  auto optimum = std::min_element(_traversalPredictions.begin(), _traversalPredictions.end(),
```
FG-TUM, line 214:
```suggestion
  const auto optimum = std::min_element(_traversalPredictions.begin(), _traversalPredictions.end(),
```

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -0,0 +1,370 @@
...
+                                    return a.second < b.second;
+                                  });
+
+  _optimalSearchSpace.emplace(optimum->first);
+
+  // selects configurations that are near the optimal prediction or have not been tested in the last five iterations
+  for (auto &configuration : _searchSpace) {
+    auto vector = _traversalTimesStorage[configuration];
+    if ((_timer - vector[vector.size() - 1].first) >= 5 ||
+        ((float)_traversalPredictions[configuration] / optimum->second <= 1.2)) {
```
FG-TUM, line 225: You use you threshold of 1.2 on multiple points in your strategy. Make it a variable (preferably a `constexpr double`) for a) readability and b) if we change it we change it everywhere.

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -0,0 +1,370 @@
...
+
+  // sanity check
+  if (_optimalSearchSpace.empty()) {
+    autopas::utils::ExceptionHandler::exception("PredicitveTuning: No possible configuration prediction found!");
+  }
+
+  _currentConfig = _searchSpace.begin();
+  while (_optimalSearchSpace.count(*_currentConfig) == 0 && _currentConfig != _searchSpace.end()) {
+    ++_currentConfig;
+  }
```
FG-TUM, line 290: here you could also point `_currentConfig` to `_optimalSearchSpace`

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -0,0 +1,370 @@
...
+                                  const std::set<TraversalOption> &allowedTraversalOptions,
+                                  const std::set<DataLayoutOption> &allowedDataLayoutOptions,
+                                  const std::set<Newton3Option> &allowedNewton3Options);
+
+  inline void selectOptimalConfiguration();
+
+  /**
+   * Selects the configuration that are going to be tested.
+   */
+  inline void selectPredictedConfigurations();
```
FG-TUM, line 112:
```suggestion
  /**
   * Selects the configurations that are going to be tested.
   */
  inline void selectOptimalSearchSpace();
```
(don't apply this suggestion through github but make the refactoring manually)

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -0,0 +1,370 @@
...
+
+  /**
+   * Provides different extrapolation methods for the prediction of the traversal time.
+   */
+  inline void predictConfigurations();
```
FG-TUM, line 117:
```suggestion
  inline void calculatePredictions();
```

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -0,0 +1,370 @@
...
+  for (auto &configuration : _searchSpace) {
+    std::vector<std::pair<int, size_t>> vector;
+    _traversalTimesStorage.emplace(configuration, vector);
+  }
+
+  _currentConfig = _searchSpace.begin();
+}
+
+void PredictiveTuning::selectPredictedConfigurations() {
+  if (_searchSpace.size() == 1 || _timer == 0 || _timer == 1) {
```
FG-TUM, line 207:
```suggestion
  if (_searchSpace.size() == 1 || _timer < 2) {
```
So if I understand this correctly, this condition is responsible for this strategy only shrinking the search space starting from the third tuning phase?
If so, make the `2` a variable.
Else explain what I missed ;)

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -0,0 +1,370 @@
...
+  auto optimum = std::min_element(_traversalPredictions.begin(), _traversalPredictions.end(),
+                                  [](std::pair<Configuration, size_t> a, std::pair<Configuration, size_t> b) -> bool {
+                                    return a.second < b.second;
+                                  });
+
+  _optimalSearchSpace.emplace(optimum->first);
+
+  // selects configurations that are near the optimal prediction or have not been tested in the last five iterations
+  for (auto &configuration : _searchSpace) {
+    auto vector = _traversalTimesStorage[configuration];
```
FG-TUM, line 223:
```suggestion
    const auto &vector = _traversalTimesStorage[configuration];
```

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -0,0 +1,370 @@
...
+    if ((_timer - vector[vector.size() - 1].first) >= 5 ||
```
FG-TUM, line 224:
```suggestion
    if ((_timer - vector.back().first) >= 5 ||
```
also add a comment what this does, and what the magic number 5 stands for (maybe even replace it with a variable)

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -0,0 +1,370 @@
...
+  * default:
+
+  }*/
+}
+
+void PredictiveTuning::linePrediction() {
+  for (auto &configuration : _searchSpace) {
+    auto vector = _traversalTimesStorage[configuration];
+    auto tmp1 = vector[vector.size() - 1];
+    auto tmp2 = vector[vector.size() - 2];
```
FG-TUM, line 244: tmp is never a good name ;)

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -0,0 +1,370 @@
...
+  /*switch (method) {
+  * case line: linePrediction(); break;
+  * default:
+
+  }*/
+}
+
+void PredictiveTuning::linePrediction() {
+  for (auto &configuration : _searchSpace) {
+    auto vector = _traversalTimesStorage[configuration];
```
FG-TUM, line 242:
```suggestion
    const auto &vector = _traversalTimesStorage[configuration];
```
also `const` everything else that can be `const`

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -0,0 +1,370 @@
...
+    auto tmp1 = vector[vector.size() - 1];
+    auto tmp2 = vector[vector.size() - 2];
+
+    auto prediction = tmp1.second + (tmp1.second - tmp2.second) / (tmp1.first - tmp2.first) * (_timer - tmp1.first);
+    _traversalPredictions[configuration] = prediction;
```
FG-TUM, line 247:
```suggestion
    _traversalPredictions[configuration] = tmp1.second + (tmp1.second - tmp2.second) / (tmp1.first - tmp2.first) * (_timer - tmp1.first);
```
No need for a temporary here.
Please add a small comment on what this calculation is. E.g. like `t1 + (t1-t2) / (iter1 - iter2) * (tuningPhase - iter1)` or something more understandable....

FG-TUM: So if I understand this correctly, you make your prediction base on the number of the tuning Iteration (`_timer`). However, wouldn't it be more precise if it was based on the actual iteration number, since the tuning iterations might not be equally long?

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -0,0 +1,370 @@
...
+  // sanity check
+  if (_currentConfig == _searchSpace.end()) {
+    autopas::utils::ExceptionHandler::exception(
+        "PredicitveTuning: Optimal configuration not found in list of configurations!");
+  }
+
+  // measurements are not needed anymore
+  _traversalTimes.clear();
+  _traversalPredictions.clear();
+  _optimalSearchSpace.clear();
```
FG-TUM, line 344: true, but isn't this already handled by `reset()`? (I would probably have the same comment in `FullSearch()` so please check and fix this there too if applicable).

tests/testAutopas/tests/selectors/tuningStrategy/PredictiveTuningTest.cpp
```diff
@@ -0,0 +1,484 @@
...
+ * @file PredictiveTuningTest.cpp
+ * @author Julian Pelloth
+ * @date 01.04.2020
+ */
+
+#include "PredictiveTuningTest.h"
+
+#include <gmock/gmock-matchers.h>
+#include <gmock/gmock-more-matchers.h>
+
```
FG-TUM, line 11: please add small comments to every test to briefly describe what it tests and how.

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -0,0 +1,439 @@
...
+    return;
+  }
+  if (_searchSpace.size() == 1) {
+    _optimalConfig = _searchSpace.begin();
+    return;
+  }
+
+  // select the tested traversal times for the current tuning phase
+  std::unordered_map<Configuration, size_t, ConfigHash> traversalTimes;
+  // for (auto &configuration : _optimalSearchSpace) does not work, but would be way more efficient
```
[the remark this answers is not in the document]
julianpelloth: I do not know why it does not work, but all tests fail.
Apparently traversalTimes is empty with the commented line of code and the exception gets thrown.

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -0,0 +1,439 @@
...
+        return true;
+      }
+    } else {
+      reselectOptimalSearchSpace();
+    }
+  } else if (_currentConfig == _longNotTestedSearchSpace.end()) {
+    _currentConfig = _optimalConfig;
+    _tuningIterationsCounter++;
+    return false;
+  }
```
FG-TUM, line 372: I might misinterpret this but as far as I see, you do not consider the case that a long untested configuration might be actually the best, do you? This is because you call `selectOptimalConfiguration()` before finishing all tests. So you would need to move the call to `selectOptimalConfiguration()` inside of `if(_longNotTestedSearchSpace.empty())` AND this `else if` block.

julianpelloth: I already changed that.

tests/testAutopas/tests/selectors/tuningStrategy/PredictiveTuningTest.cpp
```diff
@@ -0,0 +1,584 @@
...
+  EXPECT_NO_THROW(predictiveTuning.removeN3Option(autopas::Newton3Option::enabled));
+  EXPECT_FALSE(predictiveTuning.searchSpaceIsEmpty());
+  EXPECT_FALSE(predictiveTuning.searchSpaceIsTrivial());
+}
+
+/*
+ * Tests the prediction and the selection of the right optimum.
+ * Three different configurations:
+ *      - C08 that is not optimal but it getting getting less expensive.
```
FG-TUM, line 68:
```suggestion
 *      - C08 that is not optimal but it getting less expensive.
```

tests/testAutopas/tests/selectors/tuningStrategy/PredictiveTuningTest.cpp
```diff
@@ -0,0 +1,584 @@
...
+  unsigned int iteration = 0;
+  autopas::PredictiveTuning predictiveTuning(
+      {autopas::ContainerOption::linkedCells}, {1.},
+      {autopas::TraversalOption::c08, autopas::TraversalOption::c01, autopas::TraversalOption::sliced},
+      {autopas::DataLayoutOption::soa}, {autopas::Newton3Option::disabled}, iteration);
+
+  predictiveTuning.reset();
+
+  EXPECT_EQ(autopas::Configuration(autopas::ContainerOption::linkedCells, 1., autopas::TraversalOption::c08,
+                                   autopas::DataLayoutOption::soa, autopas::Newton3Option::disabled),
```
FG-TUM, line 83: for readability, you should probably create variables for all the configurations that appear here that have concise names pointing out their differences. E.g.
```C++
const auto configurationC08 = autopas::Configuration(autopas::ContainerOption::linkedCells, 
                                                     1.,
                                                     autopas::TraversalOption::c08, 
                                                     autopas::DataLayoutOption::soa, 
                                                     autopas::Newton3Option::disabled);
const auto configurationSliced = ...
```

tests/testAutopas/tests/selectors/tuningStrategy/PredictiveTuningTest.cpp
```diff
@@ -0,0 +1,584 @@
...
+            predictiveTuning.getCurrentConfiguration());
+  predictiveTuning.addEvidence(20, iteration);
+  ++iteration;
+
+  predictiveTuning.tune();
+  EXPECT_EQ(autopas::Configuration(autopas::ContainerOption::linkedCells, 1., autopas::TraversalOption::sliced,
+                                   autopas::DataLayoutOption::soa, autopas::Newton3Option::disabled),
+            predictiveTuning.getCurrentConfiguration());
+
+  predictiveTuning.reset();
```
FG-TUM, line 107:
```suggestion
  // end of first tuning phase
  predictiveTuning.reset();
```

tests/testAutopas/tests/selectors/tuningStrategy/PredictiveTuningTest.cpp
```diff
@@ -0,0 +1,584 @@
...
```
FG-TUM, line 134:
```suggestion
  // end of second tuning phase
  predictiveTuning.reset();
```

tests/testAutopas/tests/selectors/tuningStrategy/PredictiveTuningTest.cpp
```diff
@@ -0,0 +1,584 @@
...
+
+  EXPECT_EQ(autopas::Configuration(autopas::ContainerOption::linkedCells, 1., autopas::TraversalOption::c08,
```
FG-TUM, line 136:
```suggestion
  // this tests the actual prediction
  EXPECT_EQ(autopas::Configuration(autopas::ContainerOption::linkedCells, 1., autopas::TraversalOption::c08,
```

tests/testAutopas/tests/selectors/tuningStrategy/PredictiveTuningTest.cpp
```diff
@@ -0,0 +1,584 @@
...
+  unsigned int iteration = 0;
+  autopas::PredictiveTuning predictiveTuning(
+      {autopas::ContainerOption::linkedCells}, {1.},
+      {autopas::TraversalOption::c08, autopas::TraversalOption::c01, autopas::TraversalOption::sliced},
+      {autopas::DataLayoutOption::soa}, {autopas::Newton3Option::disabled}, iteration);
+
+  predictiveTuning.reset();
+
+  EXPECT_EQ(autopas::Configuration(autopas::ContainerOption::linkedCells, 1., autopas::TraversalOption::c08,
+                                   autopas::DataLayoutOption::soa, autopas::Newton3Option::disabled),
```
FG-TUM, line 152: same as above concerning variables for configurations. You could also move them into `PredictiveTuningTest.h` so you only have to define them once.

tests/testAutopas/tests/selectors/tuningStrategy/PredictiveTuningTest.cpp
```diff
@@ -0,0 +1,584 @@
...
```
FG-TUM, line 196: aaaand again :)

Issue #410, opened 2020-02-05
Parallel Tuning Strategies


SteffenSeckler:
**Is your feature request related to a problem? Please describe.**
Currently, AutoPas always searches in each rank through the entire search space.
If a completely homogeneous domain is used one could parallelize this search over multiple AutoPas instances.

**Describe the solution you'd like**
- [x] MPI-parallel Full-Search
- [x] Parallel search independent of tuning-strategy
- [x] Parallel search with Active Harmony
- [x] Inhomogeneous domains
  - [x] Normalization?
  - [x] Clustering of similar AutoPas instances for comparison.
- Compare:
  - [ ] Test similar configurations on same ranks (less switching)
  - [ ] Test different configurations on same ranks (better load balancing?)

SteffenSeckler:
clustering:
can potentially be done by actual 2d clustering by measuring the speed for each process for e.g. one static and one dynamic linked cells traversal

wothie:
Considering a combination of "Parallel search independent of tuning-strategy" and "Parallel search with Active Harmony" by just providing a utils header with all necessary (and as general as possible) MPI functionalities regarding configurations, search spaces, finding optima, etc.
This would best work by changing the search spaces to std::vector and providing additional iterators  for tuning strategies not based on search spaces.

FG-TUM:
closed as completed with the PRs linked above

PR #446, opened 2020-04-16
FullSearchMPI tuning strategy and WrapMPI to handle mpi calls inside of AutoPas

wothie:
# Description

- FullSearchMPI works similar to FullSearch, but splits the search space between all ranks and uses MPI calls to determine the optimal configuration.
- AutoPasConfigurationCommunicator was thus added as a helper for serialization and sending of configurations.
- WrapMPI is used internally to provide non-mpi versions of all necessary mpi calls in case MPI cannot be used. Otherwise it simply wraps the respective MPI calls without additional computation.
- The MPI communicator for AutoPas can now be set externally through AutoPas::setMPICommunicator
- A new testing folder has been added with tests for MPI methods. Currently only a test suite for FullSearchMPI exists

# Resolved Issue
 - None, but works on issue #410 

# How Has This Been Tested?

- [x] sph-main-mpi with AUTOPAS_MPI=ON/OFF and different numbers of processes (1, 2, or 4)
- [x] runMPITests with AUTOPAS_MPI=ON/OFF and different numbers of processes (1, 2, or 4)
- [x] Tested with both OpenMPI and MPICH
- [ ] Tested on cluster


src/autopas/utils/WrapMPI.h
```diff
@@ -0,0 +1,499 @@
...
+  if (not initialized) {
+    result1 = MPI_Init(argc, argv);
+  }
+  result2 = MPI_Comm_dup(MPI_COMM_WORLD, &AUTOPAS_MPI_COMM_WORLD);
+  if (result1 == MPI_SUCCESS) {
+    return result2;
+  } else {
+    return result1;
+  }
+}
```
SteffenSeckler, line 297: move the communicator to AutoPas.h as setComm().

src/autopas/utils/WrapMPI.h
```diff
@@ -0,0 +1,499 @@
...
+
+inline int AutoPas_MPI_Finalize() {
+  int finalized;
+  int result1 = MPI_SUCCESS;
+  int result2 = MPI_SUCCESS;
+  MPI_Finalized(&finalized);
+
+  if (not finalized) {
+    result1 = MPI_Comm_free(&AUTOPAS_MPI_COMM_WORLD);
```
SteffenSeckler, line 306: move to new AutoPas.finalize()

src/autopas/AutoPas.h
```diff
@@ -75,13 +76,14 @@ class AutoPas {
         _relativeOptimumRange(1.2),
         _maxTuningPhasesWithoutTest(5),
         _acquisitionFunctionOption(AcquisitionFunctionOption::lowerConfidenceBound),
-        _tuningStrategyOption(TuningStrategyOption::fullSearch),
+        _tuningStrategyOption(TuningStrategyOption::fullSearchMPI),
```
SteffenSeckler, line 79: undo

src/autopas/AutoPas.h
```diff
@@ -101,17 +102,27 @@ class AutoPas {
...
+            _maxTuningPhasesWithoutTest, _acquisitionFunctionOption, _autopasMPICommunicator)),
         _selectorStrategy, _tuningInterval, _numSamples);
     _logicHandler =
         std::make_unique<autopas::LogicHandler<Particle, ParticleCell>>(*(_autoTuner.get()), _verletRebuildFrequency);
   }
 
+  /**
+   * Free the AutoPas MPI communicator.
+   * DO call if: No MPI_Comm was not externally provided or is not freed externally
+   * Do NOT call if: Externally provided MPI_Comm is freed externally
```
SteffenSeckler, line 122: this is NOT a good idea!
this should always be called consistently.
as you are doing an `AutoPas_MPI_Comm_dup` it should be safe to free it, right?

src/autopas/AutoPas.h
```diff
@@ -553,6 +564,23 @@ class AutoPas {
     _tuningStrategyOption = tuningStrategyOption;
   }
 
+// Only define the interface for the MPI communicator if AUTOPAS_MPI=ON
+// The internal implementation will use _autopasMPICommunicator with WrapMPI regardless of AUTOPAS_MPI
+#if defined(AUTOPAS_MPI)
+  /**
+   * Setter for the MPI communicator that AutoPas uses for potential MPI calls.
+   * If not set, MPI_COMM_WOLRD will be used.
```
SteffenSeckler, line 572:
```suggestion
   * If not set, MPI_COMM_WORLD will be used.
```

tests/MPIParallelAutoPasTests/tests/selectors/tuningStrategy/FullSearchMPITest.cpp
```diff
@@ -0,0 +1,141 @@
...
+      {autopas::DataLayoutOption::soa, autopas::DataLayoutOption::aos}, {autopas::Newton3Option::enabled});
+
+  EXPECT_THROW(fullSearchMPI.removeN3Option(autopas::Newton3Option::enabled),
+               autopas::utils::ExceptionHandler::AutoPasException);
+}
+
+TEST_F(FullSearchMPITest, testGlobalOptimumAndReset) {
+  int rank;
+  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+// manually set every rank to 0, because we cannot use WrapMPI here
```
SteffenSeckler, line 50: please add a short description that makes the idea of the test easily understandable.

src/autopas/utils/WrapMPI.h
```diff
@@ -0,0 +1,501 @@
...
+
+/**
+ * Dummy for MPI_Datatype.
+ * initialize values to the size of the respective type in bytes.
+ */
+enum AutoPas_MPI_Datatype {
+  AUTOPAS_MPI_BYTE = 1,
+  AUTOPAS_MPI_CXX_BOOL = sizeof(bool),
+  AUTOPAS_MPI_INT = sizeof(int),
+  AUTOPAS_MPI_LONG_INT = sizeof(long) + sizeof(int),
```
SteffenSeckler, line 86: this is not necessarily true, see here:
https://godbolt.org/z/n5KCur

Best is, to remove this data type, it's anyways not used...

src/autopas/utils/WrapMPI.h
```diff
@@ -0,0 +1,501 @@
...
+/** Dummy for MPI_STATUS_IGNORE */
+#define AUTOPAS_MPI_STATUS_IGNORE nullptr
+
+/**
+ * Dummy for MPI_Request.
+ */
+enum AutoPas_MPI_Request {
+  AUTOPAS_MPI_REQUEST_NULL,
+  _AUTOPAS_MPI_COMPLETED_REQUEST,
+  _AUTOPAS_MPI_INCOMPLETE_REQUEST,
```
SteffenSeckler, line 125: why did you add `_` here?

src/autopas/selectors/tuningStrategy/FullSearchMPI.h
```diff
@@ -0,0 +1,318 @@
...
+      : _containerOptions(allowedContainerOptions),
+        _optimalConfig(Configuration()),
+        _configurationCommunicator(AutoPasConfigurationCommunicator()),
+        _localOptimalTime(0),
+        _request(AUTOPAS_MPI_REQUEST_NULL),
+        _allLocalConfigsTested(false),
+        _allGlobalConfigsTested(false),
+        _autopasMPICommunicator(comm) {
+    // @todo distribute the division process, so that not every rank has to traverse all configs
+    // @todo consider using MPI_Scatterv for dividing the search space
```
SteffenSeckler, line 50: are these todos still valid?

src/autopas/AutoPas.h
wothie: I'm only calling AutoPas_MPI_Comm_dup if has not been set until AutoPas::init() is called.
In finalize I don't know if the communicator has been created by the AutoPas_MPI_Comm_dup in init() or if it has been provided by the client. Thus if I free it blindly, the client implementation would crash if it also attempted to free it.
So either I have these rules or I explicitely store whether the communicator came externally.

src/autopas/utils/WrapMPI.h
wothie: Because these do not correspond to anything defined by mpi. Ideally I would just want to not export them at all, but I chose this solution so that they at least wouldn't look like the regular dummy values.

src/autopas/selectors/tuningStrategy/FullSearchMPI.h
wothie: > are these todos still valid?

The idea is still relevant, but considering that FullSearchMPI will be deleted soon, I suppose they aren't. I removed them.

FG-TUM:
@wothie It seems the `AutoPasConfigurationCommunicator` is broken because we recently added a new element to the configurations. Can you please update it and add some instructions on what to do in `.github\CONTRIBUTING.md` maybe in a new subsection "Making an Option tuneable" (I will fill in the rest on what to do in such a case later.)?

VincentF98:
@wothie My guide is up; see https://github.com/AutoPas/AutoPas/pull/487

PR #453, opened 2020-04-28
Balanced sliced traversal

VincentF98:
# Description

This adds a new subclass of SlicedBasedTraversal which calculates the slice thicknesses to balance out the load between all threads. Heuristics for estimating the load can be added to the namespace autopas::loadEstimators. The file src/autopas/containers/loadEstimators/cellBasedHeuristics.h contains functions specifically for cell based containers. Currently one heuristic is implemented which sums up the squared number of particles of all cells within the region to be estimated. For evaluation a constant heuristic is also included as a controll.

# How Has This Been Tested?
Currently only existing generated tests have been used.
Later tests for calculating cut positions should be added

PR #465, opened 2020-05-15
Update to use new archer image

SteffenSeckler:
# Description

Update Jenkinsfile to use the new Archer image.
This archer image uses the new llvm-integrated archer version:
https://github.com/llvm/llvm-project/tree/master/openmp/tools/archer/

# Additional Info
This update becomes needed, as the old version provided false errors in #453.
Archer is now fully integrated into the thread-sanitizer, so we more or less only use that now and can remove any Archer specific functions from the cmake files.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,16 +33,6 @@ if (NOT CMAKE_CXX_COMPILER_ID IN_LIST SUPPORTED_COMPILERS)
     message(WARNING "Compiler " ${CMAKE_CXX_COMPILER_ID} " not supported!")
 endif ()
 
-# ~~~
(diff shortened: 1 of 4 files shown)


SteffenSeckler:
archer is now integrated into llvm's libomp library and is loaded automatically when tsan is started and a program using openmp is found. Thus enabling tsan is sufficient for archer to be enabled.

only problem is that the archer library is somehow not correctly packaged into the libomp package from apt.llvm.org

the archer stuff from the cmake is no longer needed. 

the export statement for `print_ompt_counters` had to be removed as archer no longer uses that variable built in.

archer without openmp: doesnt make sense, but you can use the docker image also for building wothout openmp

PR #466, opened 2020-05-19
update doxygen links in AutoPas.h

SteffenSeckler:


src/autopas/AutoPas.h
```diff
@@ -452,7 +452,9 @@ class AutoPas {
   [[nodiscard]] AcquisitionFunctionOption getAcquisitionFunction() const { return _acquisitionFunctionOption; }
 
   /**
-   * Set acquisition function for tuning
+   * Set acquisition function for tuning.
+   * For possible acquisition function choices see options::AcquisitionFunctionOption::Value.
+   * @note This function is only relevant for the bayesian search.
```
FG-TUM, line 457:
```suggestion
   * @note This function is only relevant for the bayesian based searches.
```
Da wird es voraussichtlich mehrere geben ;)

src/autopas/AutoPas.h
```diff
@@ -465,7 +467,7 @@ class AutoPas {
 
   /**
    * Set the selector configuration strategy.
```
FG-TUM, line 469: Needs to be more descriptive. Something like this:
```suggestion
   * Set the strategy of how to select a performance value for a piece of evidence from multiple time measurements (=samples).
```

src/autopas/options/DataLayoutOption.h
```diff
@@ -11,7 +11,7 @@
 #include "autopas/options/Option.h"
 
 namespace autopas {
-
+inline namespace options {
```
FG-TUM, line 14: We need [nested inline namespaces](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1094r2.html) so lets upgrade to C++20 😈  /s

SteffenSeckler: yes, to skrew up doxygen even more xD

PR #468, opened 2020-05-20
Fixes one potential memory leak and add fakedlclose library

SteffenSeckler:
# Description
Fixes a potential leak in sph-mpi.

Fixes leak and adds basic library to overwrite `dlclose()`.
This library can be preloaded (LD_PRELOAD) to executables built with tsan to disable `dlclose()` which can enable better traces.


SteffenSeckler:
> before you merge this document how to use this

am doing this right now :)

SteffenSeckler:
added doc

.github/CONTRIBUTING.md
```diff
@@ -67,6 +67,19 @@ docker run -v ${PathToAutoPasRoot}/:/autopas -it autopas/autopas-build-cuda \
...
+
+## Jenkins/CI
+A continuous integration setup (CI) is automatically run for each open pull request and for the master.
+The executed tests are defined within the Jenkinsfile in the root AutoPas directory.
+These tests include:
+* Formatting and documentation checks 
+* Builing of all targets and execution of the provided ctest tests.
+* Sanitizer runs (Address+Leak sanitizer, Thread Sanitizer)
+
+If you encounter problems within these tests check whether you can reproduce them.
```
FG-TUM, line 80:
```suggestion
If you encounter problems within these tests check whether you can reproduce them locally. Have a look at the [Jenkinsfile](https://github.com/AutoPas/AutoPas/blob/master/Jenkinsfile) for how the tools are used with AutoPas. If you do not have the respective compiler installed you can use the [AutoPas docker images](https://hub.docker.com/u/autopas).
```

.github/CONTRIBUTING.md
```diff
@@ -67,6 +67,19 @@ docker run -v ${PathToAutoPasRoot}/:/autopas -it autopas/autopas-build-cuda \
...
+For encountering "unknown module" problems with the thread sanitizer, a library to override `dlclose()` can be found in the `libs` directory. Said library can be used to get better stack traces that are caused by dynamically loaded libraries (using `dlopen()`).
```
FG-TUM, line 81:
```suggestion
To circumvent "unknown module" problems with the thread sanitizer, a library to override `dlclose()` can be found in the `libs` directory. Said library can be used to get better stack traces that are caused by dynamically loaded libraries (using `dlopen()`).
```
Or do you want to trigger these problems with the fake lib?

libs/fake-dlclose/README.md
```diff
@@ -0,0 +1,16 @@
+This shared library provides a fake `dlclose()` function which does not actually close a library.
+
+You can use this library if you want to get better traces for LSAN (leak sanitizer) using
+
+LD_PRELOAD=PATH_TO_FAKE_DL_CLOSE_SHARED_LIBRARY ./executable
```
FG-TUM, line 5: I want to suggest putting this line in triple '`' characters but this breaks the suggestion syntax :D

PR #469, opened 2020-05-22
fancy help messages, doc and completions

FG-TUM:
# Description

- [x] restructure md-flexible folder
- [x] add doxygen for md-flexible
- [x] clean up all doxygen warnings
- [x] rework generation of cli options to generate a [GNU conform `--help`](https://www.gnu.org/prep/standards/html_node/_002d_002dhelp.html) option.
- [x] generate completions file for `zsh`
- [x] huge parser switch case now is safer bc it uses variables not literals
- [x] fix exit code of md-flex when encountering help to be 0
- [x] generate `getoptChar`s from names. (Maybe not easily doable, see [comment below](https://github.com/AutoPas/AutoPas/pull/469#discussion_r431270944))
-> workaround use `__LINE__`

## ~Resolved Issues~

# How Has This Been Tested?

compiling and running md-flex


SteffenSeckler:
codacy still complains

examples/md-flexible/src/parsing/CLIParser.cpp
```diff
@@ -8,62 +8,80 @@
...
+      config.vtkFileName,
+      config.vtkWriteFrequency,
+      config.selectorStrategy,
+      config.yamlFilename,
+      config.distributionStdDev,
+      MDFlexConfig::MDFlexOption<std::string>("", "zsh-completions", false, 'Z', "Generate completions file for zsh."),
+      MDFlexConfig::MDFlexOption<std::string>("", "help", false, 'h', "Display this message."),
+  };
+
+  // TODO: verify getOptSwitchChars are unique
```
SteffenSeckler, line 60: todo

examples/md-flexible/src/parsing/CLIParser.cpp
```diff
@@ -569,3 +575,67 @@ void CLIParser::inputFilesPresent(int argc, char **argv, MDFlexConfig &config) {
 
   opterr = opterrBefore;
 }
+
+void MDFlexParser::CLIParser::printHelpMessage(
+    std::ostream &ostream, const std::string &relPathOfExecutable,
+    const std::vector<MDFlexConfig::MDFlexOptionInterface> &relevantOptions) {
+  ostream << "Usage: " << relPathOfExecutable << "\n";
```
SteffenSeckler, line 582: comment, how you can maintain this to be zsh completion conform

examples/md-flexible/src/parsing/CLIParser.cpp
```diff
@@ -569,3 +575,67 @@ void CLIParser::inputFilesPresent(int argc, char **argv, MDFlexConfig &config) {
...
+                "# format : \"--arg[description.]: : (list of options)\"\n"
+                "\n"
+                "_arguments -s -S \\\n";
+
+  // add all cli options to the auto completion file
+  for (const auto &option : cliOptions) {
+    // the keyword to look for value suggestions
+    const std::string keywordForValues{"Possible Values:"};
+    std::string choices;
+    // if the option requires arguments and the description suggests some parse them
```
SteffenSeckler, line 629:
```suggestion
    // If the option requires arguments and the description suggests some, parse them.
```

examples/md-flexible/src/parsing/CLIParser.cpp
```diff
@@ -569,3 +575,67 @@ void CLIParser::inputFilesPresent(int argc, char **argv, MDFlexConfig &config) {
...
+    if (option.requiresArgument and option.description.find("Possible Values:") != std::string::npos) {
+      // start after (+1) first occurrence of '[' after "Possible Values:"
+      const auto startPossibleValues{option.description.find('(', option.description.find("Possible Values:")) + 1};
+      // end is the first ']' after the start. The lenght of the substr is the index difference to start.
```
SteffenSeckler, line 633:
```suggestion
      // end is the first ']' after the start. The length of the substr is the index difference to start.
```

examples/md-flexible/src/parsing/CLIParser.cpp
FG-TUM: We do not derive the completions from `--help` anymore but generate a completion file directly. See `createZSHCompletionFile()`.

examples/md-flexible/src/parsing/CLIParser.cpp
FG-TUM: done in 05c7ab99afc1508946c42fe5097f8a8e2fc49345

FG-TUM:
the codacity complaint is a known false positive:
https://trac.cppcheck.net/ticket/9237
-.-

FG-TUM:
> The building of the documentation of the example should probably be done in Jenkins as well, right?

Would probably be sensible, yes, but I don't care too much since we change this rather rarely.

examples/md-flexible/src/parsing/CLIParser.cpp
```diff
@@ -8,62 +8,72 @@
...
   optind = 1;
   bool displayHelp = false;
   for (int cliOption = 0, cliOptionIndex = 0;
-       (cliOption = getopt_long(argc, argv, "", long_options, &cliOptionIndex)) != -1;) {
+       (cliOption = getopt_long(argc, argv, "", long_options.data(), &cliOptionIndex)) != -1;) {
     string strArg;
     if (optarg != nullptr) strArg = optarg;
     transform(strArg.begin(), strArg.end(), strArg.begin(), ::tolower);
     switch (cliOption) {
       case '3': {
```
FG-TUM, line 74: I still would like to have a variable here instead of the literal. However, I currently see only three options and I like none of them ;) 
1. make all option members of MDFlexConfig static, then we could write something like:
```C++
case MDFlexConfig::newton3option.getoptChar() :
```
Currently this is not possible because newton3option is not available at compile time.
2.  make constexpr variables for all `getoptChars` next to the option members. So in MDFlexConfig.h something like:
```C++
  constexpr static char newton3OptionsGetoptChar{'3'};
  /**
   * newton3Options
   */
  MDFlexOption<std::set<autopas::Newton3Option>, newton3OptionsGetoptChar> newton3Options{
      autopas::Newton3Option::getAllOptions(), "newton3", true,
      "List of newton3 options to use. Possible Values: " +
          autopas::utils::ArrayUtils::to_string(autopas::Newton3Option::getAllOptions(), " ", {"(", ")"})};
```
then we can here use `case newton3OptionsGetoptChar: `
3. replacing the `switch` with a large `if else` tree

FG-TUM: fixed via `decltype(option)::getoptChar` in d52640c2

examples/md-flexible/src/parsing/MDFlexConfig.h
```diff
@@ -0,0 +1,476 @@
...
+   */
+  enum class GeneratorOption { grid, uniform, gaussian, sphere };
+
+  //  All options in the config
+  //  Make sure that the description is parsable by `CLIParser::createZSHCompletionFile()`!
+
+  /**
+   * yamlFilename
+   */
+  MDFlexOption<std::string, 'Y'> yamlFilename{"", "yaml-filename", true, "Path to input file."};
```
FG-TUM, line 130: It now would be totally awesome if we could drop specifying the getoptChar (here 'Y') and generate it from e.g. the name (here "yaml-filename"). For this to work, however, the name needs to be available at compile-time, because the getoptChar needs to be available at compile time. [This Stackoverflow post](https://stackoverflow.com/a/1826505/7019073) seems to suggest that it is possible but I do not see how this can be done here easily.

SteffenSeckler: This will be supported in c++20 via [P0732R2](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0732r2.pdf) "Class types in non-type template parameters". This is in gcc9 and msvc, but not yet in clang.

SteffenSeckler: (at least partially)

SteffenSeckler: for now, I'd suggest to use `__LINE__`.

FG-TUM: done with 8e93ac8

PR #428, opened 2020-02-21
Test tuning

FG-TUM:
# Description

Small python tools to test and plot tuning behavior.

md-flex:
- [x] flag to suppress creation of end-config
- [x] fix duplicated brackets in output
- [x] print more info about iterations

other stuff:
- [x] test tuning script
- [x] ~several~ one example input
- [x] fancy plot script

Example plot:
newplot

The plot is actually interactive and provides all information about a configuration at each datapoint. Further improvements are conceivable but not now...

## ~Resolved Issues~

# How Has This Been Tested?

- [x] angrily looking at the code

PR #470, opened 2020-05-24
Improve plotting

SabrinaKrallmann:
# Description

Improve plotting of md-flexible examplen output files. Enables plot script to use directories as input and plot per traversal, per number of particles and per further parameters.

Plot per number of particles
image

Plot per density of particles
image

Plot per boxsize
image


FG-TUM:
can you add an example of a new plot to the PR like here #428 ?

examples/md-flexible/scripts/plotTuning.py
```diff
@@ -14,13 +14,22 @@
...
     if "--help" in arg:
-        print("Usage: ./plotTuning.py [path/To/mdFlex/std.out ...]")
+        print("Usage: ./plotTuning.py [path/To/mdFlex/std.out ...] or [path/To/mdFlex/directoryWithOutput]")
         print("If no input is given the script looks for the latest testTuning directory in the current directory.")
         exit(0)
 
 # take all input files as source for a plot
 if len(sys.argv) > 1:
-    datafiles=sys.argv[1:]
+    if ".out" in arg:
```
FG-TUM, line 23: better check if the argument is a directory or file, otherwise, this will only work properly if one always goes for the `.out` filename, which is completely arbitrary.

examples/md-flexible/scripts/plotPerTraversalAndParameter.py
```diff
@@ -0,0 +1,253 @@
...
+    else:
+        datadirs = sys.argv[2:]
+        datadirs = list(datadirs)
+        datadirs.sort(reverse=True)
+        datafiles = os.listdir(datadirs[0])
+        datafiles = filter(lambda s: s.endswith('.out'), datafiles)
+        datafiles = map(lambda s: datadirs[0] + "/" + s, datafiles)
+        datafiles = list(datafiles)
+
+#    TODO: if nothing is given search for the latest test dir
```
FG-TUM, line 45: todo still valid?

examples/md-flexible/scripts/plotPerTraversalAndParameter.py
```diff
@@ -0,0 +1,253 @@
...
+        exit(0)
+    elif parameterIdentifierString in arg:
+        parameterArg = arg.split('=', 1)[1]
+        containsParameterArg = True
+    elif not containsParameterArg and not parameterIdentifierString in arg:
+        print("Please specify the parameter for the x-axis: parameter=[number, size, density]")
+        exit(0)
+
+# take all input files as source for a plot
+if len(sys.argv) > 1:
```
FG-TUM, line 33: better check if the argument is a directory or file, otherwise, this will only work properly if one always goes for the `.out` filename, which is completely arbitrary.

examples/md-flexible/input/testTimePerTraversal/TestTimePerTraversal1.yaml
```diff
@@ -0,0 +1,22 @@
...
+functor                          :  Lennard-Jones (12-6)
+cutoff                           :  2
+deltaT                           :  0
+# 138 tuning iterations + 50 to observe choses configuration
+iterations                       :  50
+periodic-boundaries              :  false
+Objects:
+  CubeGrid:
+    0:
+      particles-per-dimension    :  [10, 10, 10]
```
FG-TUM, line 13: I think it would be simpler, to only have one yaml file per test type and then overwrite the parameter that is scaled with command line arguments.

examples/md-flexible/scripts/plotPerTraversalAndParameter.py
```diff
@@ -0,0 +1,253 @@
...
+    meanYaxis.sort()
+    meanYaxis = [y / divisor for y in meanYaxis]
+    meanXaxis.sort()
+
+    # TODO: sort depending
+
+# ---------------------------------------------- Plot ---------------------------------------------
+
+# build the full rgb color space
+colorrange = ['rgb(  0, 0, 0)', 'rgb(0, 100,   100)', 'rgb(255, 219,   0)', 'rgb(255, 0,   0)', 'rgb(  0, 147, 255)', 'rgb(71, 0, 255)', 'rgb(73, 255,   0)', 'rgb(255,   0, 221)']
```
FG-TUM, line 208: the comment does not fit anymore ;)
I think it would be nicer if the number of colors is generated so that even if we add more containers there will still be enough colors.

examples/md-flexible/scripts/plotPerTraversalAndParameter.py
```diff
@@ -0,0 +1,253 @@
...
+    density = True
+    xAxisTitle = 'Density of particles'
+
+allContainers[0] = ["directSum"]  # directSum
+allContainers[1] = ["c01", "c08", "c18", "sliced", "c01-combined-SoA", "c04", "c04SoA", "c04HCP"]  # linkedCells
+allContainers[2] = ["verlet-clusters", "verlet-clusters-coloring", "verlet-clusters-static"]  # verletClusterLists
+allContainers[3] = ["verlet-sliced", "verlet-c18", "verlet-c01"]  # verletListsCells
+allContainers[4] = ["verlet-cluster-cells"]  # verletClusterCells
+allContainers[5] = ["var-verlet-lists-as-build"]  # varVerletListsAsBuild
+allContainers[6] = ["verlet-lists"]  # verletLists
```
FG-TUM, line 83: This does not look very sustainable with regard to new containers or traversals. You could generate these arrays from the output of md-flex when you run it once without particles and the `all` arguments.

FG-TUM (approved):

PR #472, opened 2020-05-24
Add a version of sliced traversal with 2 colors

VincentF98:
# Description

This adds a new version of the sliced traversal, which uses a 2-coloring to prevent race conditions, instead of locking the starting layers. This could also be described as a c02-traversal, however unlike the other cxx-traversals, the color-blocks may have varying sizes. 

# TODO
- [x] add support for VerletListsCells container

# How Has This Been Tested?

Up until now nothing new was added, that isn't covered by autogenerated tests.


src/autopas/containers/cellPairTraversals/CSlicedBasedTraversal.h
```diff
@@ -0,0 +1,235 @@
...
+ * is however not derived from CBasedTraversal, as that would not allow varying slice thicknesses,
+ * and would prevent us from selecting the dimension in which we cut the slices.
+ *
+ * @tparam ParticleCell The type of cells.
+ * @tparam PairwiseFunctor The functor that defines the interaction of two particles.
+ * @tparam dataLayout
+ * @tparam useNewton3
+ */
+template <class ParticleCell, class PairwiseFunctor, DataLayoutOption::Value dataLayout, bool useNewton3>
+class CSlicedBasedTraversal : public CellPairTraversal<ParticleCell> {
```
FG-TUM, line 35: this class seems to have some overlap to `SlicedBasedTraversal`. Maybe a common base class or one inheriting from the other would be reasonable to reduce code duplication.

src/autopas/containers/cellPairTraversals/CSlicedBasedTraversal.h
```diff
@@ -0,0 +1,235 @@
...
+template <class ParticleCell, class PairwiseFunctor, DataLayoutOption::Value dataLayout, bool useNewton3>
+template <bool allCells, typename LoopBody>
+void CSlicedBasedTraversal<ParticleCell, PairwiseFunctor, dataLayout, useNewton3>::cSlicedTraversal(
+    LoopBody &&loopBody) {
+  using std::array;
+
+  auto numSlices = _sliceThickness.size();
+  // 0) check if applicable
+
+  std::array<size_t, 2> overLapps23{_overlap[_dimsPerLength[1]], _overlap[_dimsPerLength[2]]};
```
FG-TUM, line 197:
```suggestion

  // check if applicable
  std::array<size_t, 2> overLapps23{_overlap[_dimsPerLength[1]], _overlap[_dimsPerLength[2]]};
```

src/autopas/containers/cellPairTraversals/CSlicedBasedTraversal.h
```diff
@@ -0,0 +1,235 @@
...
+  if (allCells) {
+    overLapps23 = {0ul, 0ul};
+    _sliceThickness.back() += _overlapLongestAxis;
+  }
+
+  for (size_t offset = 0; offset < 2; offset++) {
+#ifdef AUTOPAS_OPENMP
+// although every thread gets exactly one iteration (=slice) this is faster than a normal parallel region
+#pragma omp parallel for schedule(dynamic, 1)
```
FG-TUM, line 206: wouldn't a static scheduling make more sense or is there no performance difference?

src/autopas/containers/linkedCells/traversals/CSlicedTraversal.h
```diff
@@ -0,0 +1,73 @@
...
+#include "autopas/utils/WrapOpenMP.h"
+
+namespace autopas {
+
+/**
+ * This class provides the colored sliced traversal.
+ *
+ * The traversal finds the longest dimension of the simulation domain and cuts
+ * the domain in one slice (block) per thread along this dimension. Unlike the regular
+ * sliced traversal, this version uses a 2-coloring to prevent race conditions, instead of
```
FG-TUM, line 26: same as in the other class

src/autopas/containers/verletListsCellBased/verletListsCells/traversals/CSlicedTraversalVerlet.h
```diff
@@ -0,0 +1,72 @@
...
+
+namespace autopas {
+
+/**
+ * This class provides the colored sliced traversal.
+ *
+ * The traversal finds the longest dimension of the simulation domain and cuts
+ * the domain in one slice (block) per thread along this dimension. Unlike the regular
+ * sliced traversal, this version uses a 2-coloring to prevent race conditions, instead of
+ * locking the starting layers.
```
FG-TUM, line 25: same as above, also explain how this matches to the verlet lists

src/autopas/containers/cellPairTraversals/CSlicedBasedTraversal.h
```diff
@@ -0,0 +1,235 @@
...
+  void init(const std::array<unsigned long, 3> &dims);
+
+  /**
+   * The main traversal of the colored sliced traversal.
+   *
+   * @copydetails C01BasedTraversal::c01Traversal()
+   *
+   * @tparam allCells Defines whether or not to iterate over all cells with the loop body given as argument. By default
+   * (allCells=false) it will not iterate over all cells and instead skip the last few cells, because they will be
+   * covered by the base step. If you plan to use the default base step of the traversal on this function, use
```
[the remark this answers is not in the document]
VincentF98: This depends on the specific version for LinkedCells or VerletListsCells containers and is the same as for regular sliced.
ie c08 for LinkedCells and c18 for VerletListsCells if I'm not mistaken.

FG-TUM:
@VincentF98 the merge with the master broke something. Probably due to some renaming. Please fix :)

src/autopas/containers/cellPairTraversals/CSlicedBasedTraversal.h
```diff
@@ -50,7 +50,15 @@ class CSlicedBasedTraversal : public SlicedBasedTraversal<ParticleCell, Pairwise
...
    * The main traversal of the colored sliced traversal.
    * This provides the structure of the loops and its parallelization.
-   * @copydetails SlicedBasedTraversal::slicedTraversal()
+   *
+   * @copydetails C01BasedTraversal::c01Traversal()
+   *
+   * @tparam allCells Defines whether or not to iterate over all cells with the loop body given as argument. By default
+   * (allCells=false) it will not iterate over all cells and instead skip the last few cells, because they will be
+   * covered by the base step. If you plan to use the default base step of the traversal on this function, use
+   * allCells=false, if you plan to just iterate over all cells, e.g., to iterate over verlet lists saved within the
```
FG-TUM, line 59:
```suggestion
   * allCells=false. If you plan to just iterate over all cells, e.g., to iterate over verlet lists saved within the
```

src/autopas/containers/cellPairTraversals/CSlicedBasedTraversal.h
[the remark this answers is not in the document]
FG-TUM: Actually there is no default base step for a container. Sliced uses the c08 base step because it uses a c08 cell handler. VerletListsCells always processes everything in it's neighbor lists. Depending on whether they were built with Newton3 they contain only forward neighbors, resulting in a C18 base step or all neighbors resulting in C01.
So when can one set `allCells=false`? Actually always when doing a force calculation right? Because we never want to calculate interactions within the halo?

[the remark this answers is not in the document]
VincentF98: As far as I can tell allCells=false isn't actually used anywhere. I can check if removing it breaks something :P

src/autopas/containers/cellPairTraversals/CSlicedBasedTraversal.h
```diff
@@ -0,0 +1,133 @@
...
+   * This provides the structure of the loops and its parallelization.
+   *
+   * @copydetails C01BasedTraversal::c01Traversal()
+   *
+   * @tparam allCells Defines whether or not to iterate over all cells with the loop body given as argument. By default
+   * (allCells=false) it will not iterate over all cells and instead skip the last few cells, because they will be
+   * covered by the base step. If you plan to use the default base step of the traversal on this function, use
+   * allCells=false. If you plan to just iterate over all cells, e.g., to iterate over verlet lists saved within the
+   * cells, use allCells=true. For the sliced step if allCells is false, iteration will not occur over the last layer of
+   * cells (for _overlap=1) (in x, y and z direction).
```
FG-TUM, line 61:
```suggestion
   * @tparam allCells Defines whether or not to iterate over all cells with the loop body given as argument.
   * If set to false, loopBody will not be applied to the last layer of cells (times overlap) in each dimension.
   * This is sufficient for base steps that only cover spacially forward neighbors, like c08.
   * For base steps that only cover index-forward neighbors like c18, it is necessary to also apply the base
   * step to the last cells, because it calculates interactions with cells spacially behind the base cell.
```

PR #485, opened 2020-06-10
all

FG-TUM:
# Description

Add keyword to activate all choices


FG-TUM:
@kruegener @SabrinaKrallmann this may or may not be useful for

kruegener:
@FG-TUM:

I'm just quickly using this PR as a more real world example for the performance testing ci, as it is already closed and shouldn't mess with performance.

_Sneak Preview: Adding the label 'test-performance' to a PR will trigger the pre-Alpha version of performance CI_

PR #474, opened 2020-05-26
DummyParticles: Main: Mark particles as deleted on deletion.

SteffenSeckler:
# Description

- [x] removed RMM because why not...
- [x] don't actually delete particles, but instead, mark them as dummy/deleted.
- [x] [Jenkinsfile] remove threadsanitizer because it is fully covered by archer
- [x] dummy particles need to be removed on a rebuild of the container.
- [x] Fixes incorrect handling of masks in `LJFunctorAVX::SoAKernel()`. `upot` was used instead of the masked version (only for non-cell-wise-owned-state)
- Adapt functors to ignore dummy particles:
   - [x] LJFunctor
   - [x] LJFunctorAVX
   - [x] LJFunctorCuda
   - [x] LJFunctorCudaGlobals
   - [ ] FlopCounterFunctor <-- do we need that? I don't really care about it...
   - [x] SPHDensityFunctor
   - [x] SPHForceFunctor
- [x] Deleting particles using `AutoPas::deleteParticle` no longer invalidates the neighbor lists.

## Resolved Issues

- Needed for #434 (parallel_for, ...)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes.

- [x] Test if `updateContainer()` removes dummy particles.
- [x] Test if newly deleted particles are not still used with SoA <-- This is tested through the test `TraversalComparison`. While the particles are still inside of the SoA, they are explicitly skipped inside of the SoAFunctors.
- [x] Test with dummies to LJFunctorAVXTest

PR #490, opened 2020-06-17
DummyParticles: use char for OwnershipState instead of int64_t in #474

SteffenSeckler:
# Description
Back to using chars (unsigned) for ownershipstate.
Requires almost black magic in SoAFunctorAVX.

### DIFF
--- a/src/autopas/molecularDynamics/LJFunctorAVX.h
+++ b/src/autopas/molecularDynamics/LJFunctorAVX.h
@@ -469,13 +469,13 @@ class LJFunctorAVX : public Functor<Particle, ParticleCell, typename Particle::S
           // without the switch a call to std::memcpy is generated, which we do not want!
           // using a switch the compiler knows what rest can be 1, 2 or 3, and can optimize things.
         case 1:
-          std::memcpy(ownedStateJ.data(), &ownedStatePtr2[j], 4 - rest);
+          std::memcpy(ownedStateJ.data(), &ownedStatePtr2[j], rest);
           break;
         case 2:
-          std::memcpy(ownedStateJ.data(), &ownedStatePtr2[j], 4 - rest);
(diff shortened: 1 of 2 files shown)


src/autopas/molecularDynamics/LJFunctorAVX.h
```diff
@@ -467,13 +468,35 @@ class LJFunctorAVX : public Functor<Particle, ParticleCell, typename Particle::S
...
+    } else {
+      std::memcpy(ownedStateJ.data(), &ownedStatePtr2[j], 4);
+      //_mm256_load_si256(reinterpret_cast<const __m256i *>(&ownedStatePtr2[j]));
+    }
+    // This requires that OwnershipState::dummy is zero!
+    static_assert(static_cast<unsigned char>(OwnershipState::dummy) == 0,
+                  "OwnershipState::dummy has to have the value 0.");
+    // const int32_t dummyMask = reinterpret_cast<int &>(*ownedStateJ.data());
+    const __m256i ownershipStateJ = _mm256_set_epi64x(ownedStateJ[3], ownedStateJ[2], ownedStateJ[1], ownedStateJ[0]);
+    const __m256d dummyMask = _mm256_castsi256_pd(_mm256_cmp_pd(ownershipStateJ, _zero, _CMP_NEQ_OQ));
```
FG-TUM, line 497: 1. If you do this multiple times make a function (that can be inlined!) out of it.
2. ownershipStateJ is a _m256i but then used as a _m256d argument and the cast is then too late

SteffenSeckler: done in https://github.com/AutoPas/AutoPas/pull/490/commits/93d3e63017ad9b9bc9f2f83b935ca4796779a3ab

SteffenSeckler:
closing due to clearly slower performance than int64_t variant

PR #492, opened 2020-06-18
Minor fixes

FG-TUM:
# Description

Some small stuff I stumbled upon...

### DIFF
--- a/examples/md-flexible/CMakeLists.txt
+++ b/examples/md-flexible/CMakeLists.txt
@@ -33,6 +33,9 @@ add_subdirectory(tests)
 
 # add check for current target
 # cmake-format: off
+
+# randomly generated imbalanced scenario tested with LC only
+# also tests the flop functor
 add_test(
     NAME md-flexible.test-static
     COMMAND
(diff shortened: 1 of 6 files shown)


FG-TUM:
@SteffenSeckler please review asap because this fixes a bug in the gauss generator that @kruegener will need :roll_eyes:

src/autopas/containers/linkedCells/LinkedCells.h
```diff
@@ -287,12 +287,6 @@ class LinkedCells : public ParticleContainer<ParticleCell, SoAArraysType> {
    */
   std::vector<ParticleCell> &getCells() { return this->_cells; }
 
-  /**
-   * @copydoc getCells()
-   * @note const version
-   */
-  const std::vector<ParticleCell> &getCells() const { return this->_cells; }
-
```
SteffenSeckler, line 295: is this not needed? I'd leave it, as it makes sort of sense...

examples/md-flexible/src/Generator.h
```diff
@@ -83,7 +83,7 @@ void Generator::cubeGauss(autopas::AutoPas<Particle, ParticleCell> &autopas, con
   dummyParticle.setTypeId(object.getTypeId());
   autopasTools::generators::GaussianGenerator::fillWithParticles(
       autopas, object.getBoxMin(), object.getBoxMax(), object.getParticlesTotal(), dummyParticle,
-      object.getDistributionMean(), object.getDistributionMean());
+      object.getDistributionMean(), object.getDistributionStdDev());
```
SteffenSeckler, line 86: This change makes `md-flexible.test-sim` (defined in `examples/md-flexible/CMakeLists.txt:59`) fail, as there are really really fast particles.
Quick fix would be to change the stddeviation of that test...

src/autopas/containers/linkedCells/LinkedCells.h
FG-TUM: This is provided by the parent class. Thus this function only shadows an in fact identical function. Now that I see it, we should move the non-const version to the interface too.

examples/md-flexible/src/Generator.h
FG-TUM: we probably should not use a randomly generated scenario with deltaT != 0 as a test then ;)

examples/md-flexible/CMakeLists.txt
```diff
@@ -88,24 +90,24 @@ add_test(
...
         md-flexible
-        --container
-        VerletClusterCells
-        --verlet-cluster-size
-        32
+        --container VerletClusterCells
+        --verlet-cluster-size 32
+        --no-end-config
         --no-flops
     CONFIGURATIONS checkExamples
```
FG-TUM, line 108: so we still need these tests?

SteffenSeckler: can go away, I guess...

PR #503, opened 2020-06-26
Csf only for cell based container

FG-TUM:
# Description

Only test different cell size factors for containers that actually use it
- [x] Set based tuning strategies
- [ ] Bayesian strategies
- [ ] Active Harmony (not sure this is even possible)

## ~Resolved Issues~

# ~How Has This Been Tested?~


SteffenSeckler:
I would rather solve this by modifying the `isApplicable` method or sth. similar.

FG-TUM:
But the existing `isApplicable` for traversals as this is for containers. We could think about introducing a `static isApplicable()` for containers because the decision should be made before it is actually built.

How do you want to solve this with an `isApplicable` that maps from one Configuration to true/false? If let's say only CSF == 0.5 is allowed all containers should be applicable (read available), however e.g. VerletClusterLists just ignore it.

SteffenSeckler:
possibly a container should provide supported features...
like `supportsCellSizes()` or sth.

This would be similar to the functors, where we have at least sth. like `isSupportedClusterSize`.

What I would have ideally liked would be sth. like hierarchical tuning options, i.e., a container/traversal, ... specifies which options it can take and this would be built hierarchically.
That way you wouldn't even try building wrong traversals with wrong containers, ...
You could also "easily" specify an additional tuning option for a specific container or traversal, like the verlet-build-variant, or sth. like that

But this would probably hinder any reasonable tuning... As you would sort of need to tune hierarchical...

FG-TUM:
feature postponed -> #557

PR #504, opened 2020-06-26
Fix issue #500

### LINKED ISSUE #500: activeHarmony crash after longer simulations
**Description**
When running the sph-examples with activeHarmony, the time steps get progressively longer (not in simulation time, I suspect the time is spent in tuning). ```ActiveHarmony::reset: Error connecting to Harmony session.``` thrown after around 120-140 time steps in sph and sph-mpi. On occassion it will not throw the exception, but instead finish the tuning phase. However the slow-down always happens

**To Reproduce**
1. Master branch at commit ```918ce7c356250df60b9a1e217482a0c8d4f7bf0d```. This was commited on June 16th
2. Increase number of time steps in sph-main.cpp or sph-main-mpi.cpp. 
Example values: end_time=1; let loop in sph-main-mpi.cpp::main run to 550. I also decreased dx to 1/28
3. set the tuningStrategy to activeHarmony in AutoPas.h
4. Execute with or without mpirun, doesn't matter.
5. See error

**Build environment:**
* OS: Linux. 5.6.16-1-MANJARO
* Compiler: gcc 10.1.0
* Build System: CMake 3.17.3
* IDE: Clion 2020.1.2

**Additional context**
I suspect that the problem is not dependent on the sph-example specifically, but this is as far as I've tested up until now. I'm also not sure if activeHarmony is supposed to slow down during the tuning phase, but I assume that this is a bug.

wothie:
# Description

The in #500 described bug was caused by an implicit conversion from long to double. The invalidateConfiguration method thus reported good times to Active Harmony which in turn learned to return invalid configurations. So the only way for a valid configuration to be chosen would have been to have very short time steps or be lucky and not have to invalidate any configs.

## Resolved Issues

- [X] fixes #500 

# How Has This Been Tested?
- [X] locally, with sph-main-mpi.

# Considerations before merging
* Since the bug described in #500 could not be observed universally, someone else should test this fix on their machine.

### DIFF
--- a/src/autopas/selectors/tuningStrategy/ActiveHarmony.h
+++ b/src/autopas/selectors/tuningStrategy/ActiveHarmony.h
@@ -230,7 +230,7 @@ bool ActiveHarmony::tune(bool currentInvalid) {
   if (currentInvalid) {
     if (ah_converged(htask)) {
       AutoPasLog(debug, "Active Harmony converged to invalid configuration; resetting active-harmony server.");
-      resetHarmony();
+      reset(0);
     } else {
       invalidateConfiguration();
     }
--- a/src/autopas/selectors/tuningStrategy/ActiveHarmony.h
+++ b/src/autopas/selectors/tuningStrategy/ActiveHarmony.h
@@ -214,7 +214,7 @@ void ActiveHarmony::fetchConfiguration() {
 }
 
 void ActiveHarmony::invalidateConfiguration() {
-  auto worstPerf = std::numeric_limits<double>::max();
+  auto worstPerf = std::numeric_limits<long>::max();
   addEvidence(worstPerf, 0);
   AutoPasLog(debug, "ActiveHarmony::invalidateConfiguration: {}", _currentConfig.toString());
 }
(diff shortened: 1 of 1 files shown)


SteffenSeckler:
this appears to be ending up in an indefinite tuning phase, isn't it?

wothie:
> this appears to be ending up in an indefinite tuning phase, isn't it?

Assuming that Active Harmony always converges against invalid configurations, yes. Like I said, I don't know why Active Harmony chooses the way it does, so I can't tell if it always does or if it just sometimes takes 1000 tries before getting a valid one.

wothie:
But at least in my case, an indefinite tuning phase is still a lot better than the alternative. I don't know about other systems though, maybe there the fix would actually cause problems.

SteffenSeckler:
does exchanging the numeric_limits value have any effect?

wothie:
> does exchanging the numeric_limits value have any effect?

On my machine, it's the difference between almost always converging against an invalid configuration and almost always converging against a valid one. I guess it wouldn't make a difference on systems that didn't suffer the bug from #500, but it makes a huge difference for me.
Even if it didn't make a difference, I think since addEvidence takes a long, we shouldn't provide it with a double instead.

SteffenSeckler:
ok, if this helps, it's good.
Can you check, whether setting
```
REJECT_METHOD=Random
```
helps? (I don't really have an idea what it does, though...)

wothie:
> Can you check...

Will do.

PR #508, opened 2020-07-02
Add blacklist for bad configurations

julianpelloth:
# Description

- [x] add blacklist
- [x] add functionality that bad configuration get no longer tested after the first tuning phase
- [x] adapt md-flexible
- [x] changes in python scipts
- [x] additional refactoring

# How Has This Been Tested?

- [x] Old UnitTests


examples/md-flexible/src/parsing/MDFlexConfig.h
```diff
@@ -231,9 +231,17 @@ class MDFlexConfig {
    * maxTuningPhasesWithoutTest
    */
   MDFlexOption<unsigned int, __LINE__> maxTuningPhasesWithoutTest{
-      5, "max-tuning-phases-without-test", true,
+      10, "max-tuning-phases-without-test", true,
```
SteffenSeckler, line 234: why has the 5 been changed to a 10?

examples/md-flexible/src/parsing/MDFlexConfig.h
```diff
@@ -231,9 +231,17 @@ class MDFlexConfig {
...
       "For predictive based tuning strategies: Maximal number of "
       "tuning phases a configurations can be excluded from testing."};
+  /**
+   * relativeRangeBlacklist
+   */
+  MDFlexOption<unsigned int, __LINE__> relativeRangeBlacklist{
+      0, "relative-range-blacklist", true,
+      "For predictive based tuning strategies: Relative range to the optimum in which the first evidence of a "
+      "configuration needs to be ot not get put on the blacklist. If the blacklist should not be used give 0 as an "
```
SteffenSeckler, line 243: the sentence with "ot" does not make to much sense.

SteffenSeckler:
@FG-TUM at some point it probably makes sense to not put all options for tuning strategies into the AutoPas class, but instead, expose the tuning strategy or make it settable from the outside?
Alternatively, do sth. like a "params" parameter list, i.e., sth. like a hierarchical list of options that are then only applied to specific tuning strategies. This would allow for a bit less cluttered interface. (even though it isn't type-safe anymore and auto-complete will not work...)

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -231,7 +252,19 @@ void PredictiveTuning::setTestsUntilFirstPrediction(unsigned int testsUntilFirst
 
 void PredictiveTuning::selectOptimalSearchSpace() {
   if (_searchSpace.size() == 1 or _tuningIterationsCounter < _evidenceFirstPrediction + 1) {
```
FG-TUM, line 254:
```suggestion
  // Special case: Search space is trivial or more evidence is needed.
  if (_searchSpace.size() == 1 or _tuningIterationsCounter < _evidenceFirstPrediction + 1) {
```

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -284,6 +319,9 @@ void PredictiveTuning::calculatePredictions() {
 
 void PredictiveTuning::linePrediction() {
   for (const auto &configuration : _searchSpace) {
+    if (_blacklistSearchSpace.count(configuration) > 0) {
+      continue;
+    }
```
FG-TUM, line 324: shouldn't this loop go over `_optimalSearchSpace`? (Same for the other prediction functions.)

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -610,14 +656,31 @@ void PredictiveTuning::selectOptimalConfiguration() {
...
+void PredictiveTuning::selectBlackList() {
+  auto optimumTime = _traversalTimesStorage[getCurrentConfiguration()].back().second;
+  std::set<Configuration> oldNotTestedYet = _notTestedYet;
+
+  for (const auto &configuration : oldNotTestedYet) {
+    if (_lastTest[configuration] == _tuningIterationsCounter) {
+      if (_traversalTimesStorage[configuration].back().second / optimumTime > _relativeRangeForBlacklist) {
+        _blacklistSearchSpace.emplace(configuration);
+      }
+      _notTestedYet.erase(configuration);
```
FG-TUM, line 679: if you use a real iterator (`auto iter = set.begin()`) you can call `iter = set.erase(iter)` and directly work on `_notTestedYet` without copying it.

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
julianpelloth: The _optimalSearchSpace got cleared with the prior reset and gets filled after the predictions got calculated.
_searchSpace is the only set that contains all configurations. There is no set, that only contains the not blacklisted functions.

examples/md-flexible/src/parsing/MDFlexConfig.h
```diff
@@ -246,8 +254,10 @@ class MDFlexConfig {
    * extrapolationMethodOption
    */
   MDFlexOption<autopas::ExtrapolationMethodOption, __LINE__> extrapolationMethodOption{
-      autopas::ExtrapolationMethodOption::linePrediction, "extrapolation-method", true,
-      "For predictive based tuning strategies: The extrapolation method that calculates the prediction."};
+      autopas::ExtrapolationMethodOption::linearRegression, "extrapolation-method", true,
+      "For predictive based tuning strategies: The extrapolation method that calculates the prediction. Possible "
+      "Values: " +
+          autopas::utils::ArrayUtils::to_string(autopas::ExtrapolationMethodOption::getAllOptions(), " ", {"(", ")"})};
   /**
```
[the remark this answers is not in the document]
julianpelloth: I can include it in the yaml parser. The reason I did not do it, is that it was not done for the other options.

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
FG-TUM: Then maybe there should be one ;)
I think it could make sense instead of having a set containing blacklisted configs to have a set with all not blacklisted configs. Then you could always loop over this and remove the checks whether a config is blacklisted.

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -610,14 +657,30 @@ void PredictiveTuning::selectOptimalConfiguration() {
...
 
   // sanity check
-  if (_currentConfig == _searchSpace.end() or _currentConfig == _optimalSearchSpace.end()) {
+  if (_currentConfig == _searchSpace.end() or _currentConfig == _optimalSearchSpace.end() or
+      _currentConfig == _tooLongNotTestedSearchSpace.end()) {
     autopas::utils::ExceptionHandler::exception(
         "PredicitveTuning: Optimal configuration not found in list of configurations!");
   }
 
+  if (_relativeRangeForBlacklist != 0 and not _notTestedYet.empty()) fillBlackList();
```
FG-TUM, line 666:
```suggestion
  if (_relativeRangeForBlacklist != 0 and not _notTestedYet.empty()) {
    fillBlackList();
  }
```

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
julianpelloth: That would be easier in some way, but if we want so keep `_searchSpace`, which we should.
I have a few options:
- replace `_searchSpace` with `_whitelistSearchSpace` in the loops and `_whitelistSearchSpace` would contain `_searchSpace` if blacklisting was not used.
- create a iterationSearchSpacePointer, I would think of a better name, which would point to the searchSpace, that should be looped through with this it can be selected which should be used. Maybe useful for future functions.
- or something different I did not think of
But I would not want to select the searchSpace to loop through right before each loop.

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -246,9 +262,9 @@ void PredictiveTuning::selectOptimalSearchSpace() {
   for (const auto &configuration : _searchSpace) {
     // Adds configurations that have not been tested for _maxTuningIterationsWithoutTest or are within the
     // _relativeOptimumRange
-    if ((float)_configurationPredictions[configuration] / optimum->second <= _relativeOptimumRange) {
+    if ((double)_configurationPredictions[configuration] / optimum->second <= _relativeOptimumRange) {
```
FG-TUM, line 265:
```suggestion
    if (static_cast<double>(_configurationPredictions[configuration]) / optimum->second <= _relativeOptimumRange) {
```

PR #514, opened 2020-07-15
Indicate tuning

FG-TUM:
# Description

Move tuning indication to AutoPas logger output so that it appears the same way everywhere AutoPas is used.


FG-TUM:
@VincentF98 this should expose the info you need right?

VincentF98:
> @VincentF98 this should expose the info you need right?

Yes, looks good. I will try this out later. Thanks :)

PR #515, opened 2020-07-16
Measurement of homogeneity.

SabrinaKrallmann:
# Description

Measuring the homogeneity of a scenario by using the standard deviation of the particle density.


examples/md-flexible/src/Simulation.h
```diff
@@ -512,3 +521,55 @@ void Simulation<Particle, ParticleCell>::writeVTKFile(unsigned int iteration,
...
+        densityPerCell.push_back(counter / size);
+      }
+    }
+  }
+
+  double mean = autopas.getNumberOfParticles() / (endCorner[0] * endCorner[1] * endCorner[2]);
+  double variance = 0.0;
+
+  for(int r = 0; r < densityPerCell.size(); ++r){
+    variance += (pow((densityPerCell.at(r) - mean), 2.0) / densityPerCell.size());
```
SteffenSeckler, line 569:
```suggestion
    variance += (densityPerCell.at(r) - mean)*(densityPerCell.at(r) - mean) / densityPerCell.size();
```

SteffenSeckler: using 2.0 is mostly ok, because it is almost always optimized out, but still not so good.

FG-TUM: Actually, in this case, I would prefer the `pow` version since it is more readable.
For the case that the exponent is 2 `pow` and the multiplication have the same performance (as you can see [here](https://quick-bench.com/q/8Tak-Luy597pDB2AtvUeXt5zjds)). Whenever the exponent is not 2 the difference in speed is significant. double or int does not matter in either case.

SabrinaKrallmann: So do I leave the pow-version as it is, or shall I change it?

SteffenSeckler: as you like. pow might be more readable. alternatively, you can introduce an intermediate variable `distance` or so and then use `distance*distance`.

examples/md-flexible/src/Simulation.h
```diff
@@ -508,3 +522,69 @@ void Simulation<Particle, ParticleCell>::writeVTKFile(unsigned int iteration,
...
+    std::array<double, 3> region = iter->getR();
+    std::array<long, 3> index = {};
+    for (int i = 0; i < region.size(); i++) {
+      index[i] = region[i] / cellLength;
+    }
+    const unsigned long cellIndex = autopas::utils::ThreeDimensionalMapping::threeToOneD(index, cellsPerDimension);
+    allCells[cellIndex] += 1;
+    allVolumes[cellIndex] = (index[0] == cellsPerDimension[0] - 1) ? outerCellSizePerDimension[0] : cellLength;
+    allVolumes[cellIndex] *= (index[1] == cellsPerDimension[1] - 1) ? outerCellSizePerDimension[1] : cellLength;
+    allVolumes[cellIndex] *= (index[2] == cellsPerDimension[2] - 1) ? outerCellSizePerDimension[2] : cellLength;
```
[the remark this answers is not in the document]
FG-TUM:
```suggestion
    // calculate the size of the current cell
    allVolumes[cellIndex] = 0;
    for (int i = 0; i < cellsPerDimension.size(); ++i) {
      // the last cell layer has a special size
      if (index[i] == cellsPerDimension[i] - 1) {
        allVolumes[cellIndex] *= outerCellSizePerDimension[i];
      } else {
        allVolumes[cellIndex] *=  cellLength;
      }
    }
```
Also use a for loop instead of repeating lines + ternary operator ;)

examples/md-flexible/src/Simulation.h
```diff
@@ -508,3 +522,69 @@ void Simulation<Particle, ParticleCell>::writeVTKFile(unsigned int iteration,
...
+
+  // add particles accordingly to their cell to get the amount of particles in each cell
+  for (auto iter = autopas.begin(); iter.isValid(); ++iter) {
+    std::array<double, 3> region = iter->getR();
+    std::array<long, 3> index = {};
+    for (int i = 0; i < region.size(); i++) {
+      index[i] = region[i] / cellLength;
+    }
+    const unsigned long cellIndex = autopas::utils::ThreeDimensionalMapping::threeToOneD(index, cellsPerDimension);
+    allCells[cellIndex] += 1;
```
FG-TUM, line 564: rename to particlesPerCell

examples/md-flexible/src/Simulation.h
```diff
@@ -508,3 +522,69 @@ void Simulation<Particle, ParticleCell>::writeVTKFile(unsigned int iteration,
...
+
+  // calculate variance
+  for (int r = 0; r < densityPerCell.size(); ++r) {
+    double distance = densityPerCell.at(r) - mean;
+    variance += (distance * distance / densityPerCell.size());
+  }
+
+  // finally calculate standard deviation
+  double sd = sqrt(variance);
+  return sd;
```
FG-TUM, line 589:
```suggestion
  return sqrt(variance);;
```

examples/md-flexible/src/Simulation.h
```diff
@@ -508,3 +522,69 @@ void Simulation<Particle, ParticleCell>::writeVTKFile(unsigned int iteration,
...
+        (allCells[i] == 0) ? 0 : (allCells[i] / allVolumes[i]);  // make sure there is no division of zero
+  }
+
+  // get mean and reserve variable for variance
+  double mean = numberOfParticles / volume;
+  double variance = 0.0;
+
+  // calculate variance
+  for (int r = 0; r < densityPerCell.size(); ++r) {
+    double distance = densityPerCell.at(r) - mean;
```
FG-TUM, line 583:
```suggestion
    double distance = densityPerCell[r] - mean;
```

PR #534, opened 2020-09-11
Optimizations for VCL

### LINKED ISSUE #438: Optimize RegionParticleIterators for VerletClusterLists container
Optimize the RegionParticleIterators by supplying only the needed towers to the iterators.

FG-TUM:
# Description

Make Region Iterators great (again?)!

- [x] Only search towers in Region of Interest
- [x] preallocate memory within towers upon generation

### Iterators
Scenario: `md-flexible --no-end-config --no-flops --container verletclusterlists --data-layout aos --newton3 enabled --particles-per-dimension 25 --traversal vcl_c06`

| Version | Threads  | Boundaries  | % of Simulation |
|---|---|---|---|
| old  | 1  | 19.644s  | 98.370% |
| old  | 8  | 19.847s  | 99.503% |
| new  | 1  |  0.150s | 28.162% |
| new  | 8  | 0.165s  | 50.928% |
| lc_c04_hcp | 1 | 0.040 | 11.095% |
| lc_c04_hcp | 8 | 0.048 | 34.968% |

Region Iterators are expected to be slower than for LC because instead of cells whole towers need to be searched.

### Preallocation

| Version | Threads  | rebuildNeighborLists [ns] |
|---|---|---|
| reserve 0  | 8  | 460477244 |
| reserve *2.7  | 8  |  384966463 |

## Resolved Issues

- [x] fixes #438

# How Has This Been Tested?

- [x] Jenkins


src/autopas/containers/verletClusterLists/VerletClusterLists.h
```diff
@@ -703,6 +696,55 @@ class VerletClusterLists : public ParticleContainerInterface<FullParticleCell<Pa
...
+    }
+
+    std::vector<size_t> towersOfInterest(towersOfInterstPerDim[0] * towersOfInterstPerDim[1]);
+
+    auto cellsOfInterestIterator = towersOfInterest.begin();
+    for (size_t i = 0; i < towersOfInterstPerDim[1]; ++i) {
+      std::iota(cellsOfInterestIterator, cellsOfInterestIterator + towersOfInterstPerDim[0],
+                firstTowerIndex + i * _towersPerDim[0]);
+      cellsOfInterestIterator += towersOfInterstPerDim[0];
+    }
```
SteffenSeckler, line 743:
```suggestion
    auto towersOfInterestIterator = towersOfInterest.begin();
    for (size_t i = 0; i < towersOfInterstPerDim[1]; ++i) {
      std::iota(towersOfInterestIterator, towersOfInterestIterator + towersOfInterstPerDim[0],
                firstTowerIndex + i * _towersPerDim[0]);
      towersOfInterestIterator += towersOfInterstPerDim[0];
    }
```

src/autopas/containers/verletClusterLists/VerletClusterLists.h
```diff
@@ -703,6 +696,55 @@ class VerletClusterLists : public ParticleContainerInterface<FullParticleCell<Pa
...
+    auto firstTowerCoords = _builder->getTowerCoordinates(lowerCornerInBounds);
+    auto firstTowerIndex = _builder->towerIndex2DTo1D(firstTowerCoords[0], firstTowerCoords[1]);
+    auto lastTowerCoords = _builder->getTowerCoordinates(upperCornerInBounds);
+
+    std::array<size_t, 2> towersOfInterstPerDim;
+    for (size_t dim = 0; dim < towersOfInterstPerDim.size(); ++dim) {
+      // use ternary operators instead of abs because these are unsigned values
+      towersOfInterstPerDim[dim] = firstTowerCoords[dim] > lastTowerCoords[dim]
+                                       ? firstTowerCoords[dim] - lastTowerCoords[dim]
+                                       : lastTowerCoords[dim] - firstTowerCoords[dim];
```
SteffenSeckler, line 729: I'm not certain that this is the correct way to do it.

Shouldn't you rather swap `lowerCorner` and `upperCorner` to handle this case?

Currently, it looks like you would at least need to swap `firstTowerCoords` and `lastTowerCoords` as well.

src/autopas/containers/verletClusterLists/VerletClusterListsRebuilder.h
```diff
@@ -414,24 +418,36 @@ class VerletClusterListsRebuilder {
...
+   * @param location The 3D coordinates.
+   * @return Tower reference.
+   */
+  std::array<size_t, 2> getTowerCoordinates(std::array<double, 3> location) {
+    std::array<size_t, 2> towerIndex2D{};
 
     for (int dim = 0; dim < 2; dim++) {
       const auto towerDimIndex =
-          (static_cast<long int>(floor((location[dim] - _haloBoxMin[dim]) * _towerSideLengthReciprocal))) + 1l;
+          (static_cast<long int>(floor((location[dim] - _haloBoxMin[dim]) * _towerSideLengthReciprocal)));
```
SteffenSeckler, line 437:
```suggestion
          static_cast<long int>(floor((location[dim] - _haloBoxMin[dim]) * _towerSideLengthReciprocal));
```

SteffenSeckler: there might also be problems with rounding when choosing sth. directly at a boundary of a cell.

tests/testAutopas/tests/autopasInterface/IteratorTest.cpp
```diff
@@ -36,6 +36,13 @@ void defaultInit(AutoPasT &autoPas) {
   autoPas.init();
 }
 
+/**
+ * Iterate over all particles, generate a region iterator for each that spans a tiny space around them and check if this
```
SteffenSeckler, line 40: for each `?WHAT?`
```suggestion
 * Iterate over all particles, generate a region iterator for each particle that spans a tiny space around them and check if this
```

SteffenSeckler (approved):
some minor things.
please also check rounding

src/autopas/containers/verletClusterLists/VerletClusterListsRebuilder.h
FG-TUM: seems to be fine in the unit tests that should check this

src/autopas/containers/verletClusterLists/VerletClusterLists.h
FG-TUM: Not sure I get what you mean. This is basically 
`towersOfInterstPerDim[dim] = abs(firstTowerCoords[dim] - lastTowerCoords[dim])` 
because we need to know how many towers per dimension are in the area of interest. So we need the difference on the indices in the tower grid and not domain coordinates.

SteffenSeckler: lasttowercoords - firsttowercoords should always be positive.
if it is negative, then the for loop over the `iota` should have a negative direction, shouldn't it?

FG-TUM: I don't think we enforce that the lower so firstTowerCoors is actually not guaranteed to be the first if the user messes up the input.
Since we do not make any guarantees about the order of iteration of the region iterator I would say the direction of the loop does not matter. However, I suppose instead of starting at `firstTowerIndex` we should start at `std::min(firstTowerIndex, lastTowerIndex)`.

src/autopas/containers/verletClusterLists/VerletClusterLists.h
```diff
@@ -738,7 +739,7 @@ class VerletClusterLists : public ParticleContainerInterface<FullParticleCell<Pa
     auto towersOfInterestIterator = towersOfInterest.begin();
     for (size_t i = 0; i < towersOfInterstPerDim[1]; ++i) {
       std::iota(towersOfInterestIterator, towersOfInterestIterator + towersOfInterstPerDim[0],
-                firstTowerIndex + i * _towersPerDim[0]);
+                std::min(firstTowerIndex, lastTowerIndex) + i * _towersPerDim[0]);
```
SteffenSeckler, line 742: jo, das sieht besser aus :)

SteffenSeckler: btw. such accesses (e.g. `getRegioniterator({10,10,10},{5,5,5})` are never tested and will probably lead to a lot of problems...

FG-TUM: -> #536

PR #537, opened 2020-09-14
unaligned loads and stores for LJFunctorAVX

### LINKED ISSUE #535: lc_combined* Traversals segfault with AVX Functor
**Describe the bug**

When using the lc_combined (`lc_c04_combined_SoA` and `lc_c01_combined_SoA` traversals in combination with the intrinsics functor sometimes segfaults happen in memory load operations.

**To Reproduce**
`md-flexible --no-end-config --no-flops --functor lennard-jones-AVX2 --deltaT 0 --particle-generator uniform --log-level debug --traversal lc_c04_combined_SoA  --particles-total 71`

**Expected behavior**
No segfault 🙃 

~**Screenshots**~

* OS: WSL, CooLMUC2
* Compiler: (e.g. gcc 9, clang 11)
* Build System: make

**Additional context**
This might be due to these traversals not using SoAs directly but views. These views work as offsets from normal views which might break the alignment...?

SteffenSeckler:
# Description

Replaces unaligned loads and stores in LJFunctorAVX to be able to simulate with non-aligned SoAs from SoAViews (combined SoAs)

## Todo

Should probably be implemented using a template parameter to decide whether aligned or unaligned intrinsics should be used.

## Resolved Issues

- fixes md-flexible not being build if mpi is not present.
- [x] fixes #535 

# How Has This Been Tested?

- [x] Ran example:
```bash
   ./md-flexible --no-end-config --no-flops --functor lennard-jones-AVX2 --deltaT 0 --particle-generator uniform --log-level debug --traversal lc_c04_combined_SoA --particles-total 71
```
- [x] Added the above example to the ctest tests

### DIFF
--- a/examples/md-flexible/CMakeLists.txt
+++ b/examples/md-flexible/CMakeLists.txt
@@ -5,15 +5,18 @@ file(
         "src/*.h"
 )
 
-message(STATUS "searching for mpi...")
-find_package(MPI)
+if(AUTOPAS_MPI)
+    message(STATUS "AUTOPAS_MPI set, searching for mpi to link against md-flexible ...")
+    find_package(MPI)
 
-if (NOT ${MPI_CXX_FOUND})
-    message(STATUS "cxx mpi not found, not building sph-main-mpi")
-    return()
-else ()
-    message(STATUS "cxx mpi found: ${MPI_CXX_COMPILER}")
-endif ()
+    if (NOT ${MPI_CXX_FOUND})
+        message(FATAL_ERROR "cxx mpi could not be found, even though AUTOPAS_MPI was set.")
+    else ()
+        message(STATUS "cxx mpi found: ${MPI_CXX_COMPILER}")
+    endif ()
+else()
+    message(STATUS "AUTOPAS_MPI not set, not linking MPI to md-flexible.")
+endif(AUTOPAS_MPI)
 
 include_directories(SYSTEM ${MPI_INCLUDE_PATH})
 
@@ -23,7 +26,7 @@ target_include_directories(md-flexible PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
 
 include(autopas_yaml-cpp)
 
-target_link_libraries(md-flexible PUBLIC autopas autopasTools yaml-cpp MPI::MPI_CXX)
+target_link_libraries(md-flexible PUBLIC autopas autopasTools yaml-cpp $<$<BOOL:${AUTOPAS_MPI}>:MPI::MPI_CXX>)
 
 # --- copy script files to build dir ---
 file(
--- a/examples/md-flexible/CMakeLists.txt
+++ b/examples/md-flexible/CMakeLists.txt
@@ -80,6 +80,22 @@ add_test(
         CONFIGURATIONS checkExamples
 )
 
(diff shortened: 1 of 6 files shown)


FG-TUM:
What is the performance impact of this? According to Intel Intrinsics Guide both `load` and `loadU` are the same speed but I guess since `loadU` has to potentially fetch more than one cache line it might be slower if that is the case?

SteffenSeckler:
lets use our performance testing for this :P

FG-TUM:
IIRC it uses the regular LJFunctor so you might want to check things manually or talk to @kruegener if we can have some options here ;)

kruegener:
I mean the perf measurements without selecting a specific functor ran fine ;)
I'll add an option to specify any additional CLI options for the upcoming web interface in october.

Do the CLI parameters overwrite yaml settings if both define e.g. a functor?

FG-TUM:
yes, md-flex was intentionally designed so that CLI parameters always overwrite yaml parameters.

kruegener:
> yes, md-flex was intentionally designed so that CLI parameters always overwrite yaml parameters.

Perfect. Will be easy then

SteffenSeckler:
I've added a test to the examples (ctest). This test fails for the master and runs successfully on this PR.

SteffenSeckler:
without testing too deep: performance looks equal.

SteffenSeckler:
according to https://community.intel.com/t5/Intel-ISA-Extensions/SSE-and-AVX-behavior-with-aligned-unaligned-instructions/m-p/1170006/highlight/true#M6647 it appears that since haswell, large performance impacts are not to be expected.

SteffenSeckler:
@FG-TUM ideally I would like a unit test for this :)

FG-TUM:
@SteffenSeckler feel free to write one :P

SteffenSeckler:
> @SteffenSeckler feel free to write one :P

done

examples/md-flexible/CMakeLists.txt
```diff
@@ -92,10 +95,55 @@ add_test(
         CONFIGURATIONS checkExamples
 )
 
+set(HAS_AVX true)
```
FG-TUM, line 98: Write a short justification comment for this whole code block. Would be nice to know what is coming without reading 50 lines

FG-TUM (approved):
just a bit more doc

examples/md-flexible/CMakeLists.txt
```diff
@@ -95,27 +95,28 @@ add_test(
...
-        # set to false, as we don't know...
-        set(HAS_AVX false)
-    endif()
-else()
+    try_compile(
+            HAS_AVX
+            ${CMAKE_BINARY_DIR}
+            ${AUTOPAS_SOURCE_DIR}/cmake/tests/has_avx_test.cpp
+            # We are abusing COMPILE_DEFINITIONS here. There is no other sane way of passing "-march=native" ...
+            COMPILE_DEFINITIONS "-march=native"
```
SteffenSeckler, line 116: found similar thing on `apple.com` so passing this using COMPILE_DEFINITIONS seems right...

https://opensource.apple.com/source/clang/clang-700.1.81/src/projects/compiler-rt/cmake/config-ix.cmake.auto.html

PR #541, opened 2020-10-09
dot product: 

FG-TUM:
# Description

The previous implementation of the dot product used `stl::inner_product`. Looking at the assembly output this turned out to be inefficient. The new implementation uses template parameter pack expansion for a compile-time unrolled explicit summation.

## ~Resolved Issues~

# How Has This Been Tested?

- [x] Jenkins


SteffenSeckler (approved):
good, but please wait for performance result :)

SteffenSeckler:
the thread-local values for the flop counter were missing, right?

FG-TUM:
> the thread-local values for the flop counter were missing, right?

True, but this was not the point of this PR anyway.

PR #560, opened 2020-11-19
More N3 tests

FG-TUM:
# Description

- [x] Let the Newton3OnOffTest test more Configuations that actually should be testable here.
- [x] Use `isApplicable()` instead on manually throwing out configurations.
  Non-Applicable configurations do not generate tests now.

For some reason, clang-format decided to change the indent on a large part of the file so review with hiding whitespaces

## ~Resolved Issues~

# How Has This Been Tested?

Ran the tests and they did not fail :innocent:

### DIFF
--- a/tests/testAutopas/tests/autopasInterface/Newton3OnOffTest.cpp
+++ b/tests/testAutopas/tests/autopasInterface/Newton3OnOffTest.cpp
@@ -95,7 +95,7 @@ void Newton3OnOffTest::countFunctorCalls(autopas::ContainerOption containerOptio
   }
 
   autopas::ContainerSelector<Particle> containerSelector(getBoxMin(), getBoxMax(), getCutoff());
-  autopas::ContainerSelectorInfo containerInfo(getCellSizeFactor(), getVerletSkin(), 64,
+  autopas::ContainerSelectorInfo containerInfo(getCellSizeFactor(), getVerletSkin(), getClusterSize(),
                                                autopas::LoadEstimatorOption::none);
 
   containerSelector.selectContainer(containerOption, containerInfo);
--- a/tests/testAutopas/tests/autopasInterface/Newton3OnOffTest.cpp
+++ b/tests/testAutopas/tests/autopasInterface/Newton3OnOffTest.cpp
@@ -6,94 +6,80 @@
 
 #include "Newton3OnOffTest.h"
 
+#include "autopas/selectors/ContainerSelector.h"
+#include "autopas/selectors/TraversalSelector.h"
 #include "autopas/utils/Logger.h"
 #include "autopas/utils/StaticCellSelector.h"
 
-using ::testing::_;  // anything is ok
+using ::testing::_;
 using ::testing::Combine;
 using ::testing::Return;
 using ::testing::ValuesIn;
 
 // Parse combination strings and call actual test function
 TEST_P(Newton3OnOffTest, countFunctorCallsTest) {
-  auto [containerTraversalTuple, dataLayoutOption] = GetParam();
-  auto [containerOption, traversalOption] = containerTraversalTuple;
-
-  // skip combinations that are for now not applicable
-  if ((traversalOption == autopas::TraversalOption::vlc_c18 or
-       traversalOption == autopas::TraversalOption::vlc_sliced or
-       traversalOption == autopas::TraversalOption::vlc_sliced_balanced or
-       traversalOption == autopas::TraversalOption::vlc_sliced_c02) and
-      dataLayoutOption == autopas::DataLayoutOption::soa) {
-    auto skipMessage = "Traveral " + traversalOption.to_string() + " is not applicable with SoA";
-    GTEST_SKIP_(skipMessage.c_str());
-  }
+  auto [containerOption, traversalOption, dataLayoutOption] = GetParam();
 
   countFunctorCalls(containerOption, traversalOption, dataLayoutOption);
 }
 
-// Generate Unittests for all Container / Traversal / Datalayout combinations
+// Generate Unittests for all Configurations that support both Newton3 modes
 INSTANTIATE_TEST_SUITE_P(
     Generated, Newton3OnOffTest,
-    Combine(ValuesIn([]() -> std::vector<std::tuple<autopas::ContainerOption, autopas::TraversalOption>> {
-              // needed because CellBlock3D (called when building containers) logs always
-              autopas::Logger::create();
+    ValuesIn(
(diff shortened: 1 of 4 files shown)


FG-TUM:
@tirgendetwas These changes to the Newton3OnOff Test should allow you to test your stuff too.

FG-TUM:
@SteffenSeckler For configurations that do not support N3::enabled AND N3::disabled: Do we want to skip these tests or not generate them in the first place?

SteffenSeckler:
> @SteffenSeckler For configurations that do not support N3::enabled AND N3::disabled: Do we want to skip these tests or not generate them in the first place?

What does it support then? Nothing?

tests/testAutopas/tests/autopasInterface/Newton3OnOffTest.cpp
```diff
@@ -19,59 +19,70 @@ TEST_P(Newton3OnOffTest, countFunctorCallsTest) {
...
+                auto container = containerSelector.getCurrentContainer();
+
+                for (auto traversalOption : container->getAllTraversals()) {
+                  // exclude traversals that do not support Newton3
+                  if (traversalOption == autopas::TraversalOption::lc_c01 or
+                      traversalOption ==
+                          autopas::TraversalOption::lc_c01_combined_SoA /*and autopas::autopas_get_max_threads() > 1*/
+                      or traversalOption == autopas::TraversalOption::vlc_c01 or
+                      traversalOption == autopas::TraversalOption::vcl_cluster_iteration or
+                      traversalOption == autopas::TraversalOption::vcl_c01_balanced) {
```
SteffenSeckler, line 69: Didn't you say that you wanted to use `isApplicable` here?

tests/testAutopas/tests/autopasInterface/Newton3OnOffTest.cpp
```diff
@@ -19,59 +19,70 @@ TEST_P(Newton3OnOffTest, countFunctorCallsTest) {
...
+              autopas::ContainerSelector<Particle> containerSelector({0., 0., 0.}, {10., 10., 10.}, 1.);
+              autopas::ContainerSelectorInfo containerInfo(1., 0., 64, autopas::LoadEstimatorOption::none);
+
+              // generate for all containers, even those to come
+              for (auto containerOption : autopas::ContainerOption::getAllOptions()) {
+                // skip containers that do not work with both newton modes
+                // @TODO: let verlet lists support Newton 3
+                if (containerOption == autopas::ContainerOption::verletLists or
+                    containerOption == autopas::ContainerOption::varVerletListsAsBuild or
+                    containerOption == autopas::ContainerOption::verletClusterCells) {
```
SteffenSeckler, line 54: use `isApplicable` instead?

tests/testAutopas/tests/autopasInterface/Newton3OnOffTest.cpp
```diff
@@ -19,59 +19,70 @@ TEST_P(Newton3OnOffTest, countFunctorCallsTest) {
   auto [containerTraversalTuple, dataLayoutOption] = GetParam();
   auto [containerOption, traversalOption] = containerTraversalTuple;
 
+  // skip combinations that are for now not applicable
+  if ((traversalOption == autopas::TraversalOption::vlc_c18 or
+       traversalOption == autopas::TraversalOption::vlc_sliced or
+       traversalOption == autopas::TraversalOption::vlc_sliced_balanced or
+       traversalOption == autopas::TraversalOption::vlc_sliced_c02) and
+      dataLayoutOption == autopas::DataLayoutOption::soa) {
+    auto skipMessage = "Traveral " + traversalOption.to_string() + " is not applicable with SoA";
```
SteffenSeckler, line 28: maybe use isApplicable instead?

tests/testAutopas/tests/autopasInterface/Newton3OnOffTest.cpp
FG-TUM: This is what happens when you review an unfinished draft you were not asked to review.

FG-TUM:
> > @SteffenSeckler For configurations that do not support N3::enabled AND N3::disabled: Do we want to skip these tests or not generate them in the first place?
> 
> What does it support then? Nothing?

No only one of them.

tests/testAutopas/tests/autopasInterface/Newton3OnOffTest.cpp
SteffenSeckler: I unticked the tick of the change :P

SteffenSeckler:
> > > @SteffenSeckler For configurations that do not support N3::enabled AND N3::disabled: Do we want to skip these tests or not generate them in the first place?
> > 
> > 
> > What does it support then? Nothing?
> 
> No only one of them.
> Think of it as: They support enabled XOR disabled

Ah ok, I would go for skipping. That way we don't have to manually disable them, right?

FG-TUM:
> > > > @SteffenSeckler For configurations that do not support N3::enabled AND N3::disabled: Do we want to skip these tests or not generate them in the first place?
> > > 
> > > 
> > > What does it support then? Nothing?
> > 
> > 
> > No only one of them.
> > Think of it as: They support enabled XOR disabled
> 
> Ah ok, I would go for skipping. That way we don't have to manually disable them, right?

No, nothing happens manually anymore.
Option A: Move the check for valid tests to the generator so only valid tests are generated
Option B: Move the check for valid tests to the actual test so it can call `GTEST_SKIP` when finding an invalid configuration

Invalid here refers to a configuration only supporting Newton3 enabled XOR disabled

SteffenSeckler:
> > > > > @SteffenSeckler For configurations that do not support N3::enabled AND N3::disabled: Do we want to skip these tests or not generate them in the first place?
> > > > 
> > > > 
> > > > What does it support then? Nothing?
> > > 
> > > 
> > > No only one of them.
> > > Think of it as: They support enabled XOR disabled
> > 
> > 
> > Ah ok, I would go for skipping. That way we don't have to manually disable them, right?
> 
> No, nothing happens manually anymore.
> Option A: Move the check for valid tests to the generator so only valid tests are generated
> Option B: Move the check for valid tests to the actual test so it can call `GTEST_SKIP` when finding an invalid configuration
> 
> Invalid here refers to a configuration only supporting Newton3 enabled XOR disabled

Generator seems better -> less tests, faster compile times...

SteffenSeckler (approved):

SteffenSeckler:
good by me, but archer complains because of race conditions...

PR #561, opened 2020-11-23
Evidence smoothing

FG-TUM:
# Description

- [x] implemented a smoothing algorithm (based on loess)
- [x] pass smoothed values to tuning strategies

## ~Resolved Issues~

# How Has This Been Tested?

- [x] New tests for the smoothing algorithm
- [x] Existing tests (might change expected values for some tuning strategy tests)
- [x] Plotting values:
Plot_SmoothVSReducedValue
The dashed line represents the "reduced values" (what we used so far), the normal line the smoothed values. We can see that due to external stress on the system the reduced values spike frantically and sometimes lead to a wrong ordering. The smoothed lines are much more robust and always stay in the same vertical order.
 Simulation parameters: 
`md-flexible --deltaT 0 --particle-generator uniform --cutoff 1 --box-length 10 --particles-total 20000 --traversal vlc_c01,lc_c08 --tuning-interval 10 --tuning-phases 100`

### DIFF
--- a/src/autopas/selectors/AutoTuner.h
+++ b/src/autopas/selectors/AutoTuner.h
@@ -30,11 +30,12 @@ namespace autopas {
  * 1. Measuring time of the iteration.
  * 2. Selecting an appropriate configuration for the pairwise iteration.
  *
- * The tuner can be in one of two states, depending on whether it currently should look for a new optimum,
- * which is called a tuning phase. During a tuning phase,for each Configuration multiple measurements can be taken,
+ * The tuner can be in one of two states. If it currently should look for a new optimum, it is in the 
+ * so-called tuning phase. During a tuning phase, for each Configuration, multiple measurements can be taken,
  * which are called samples. To reduce noise, the samples for one configuration are then condensed to one value for
- * the current tuning phase, called evidence. These evidence are handed on to a tuningStrategy, which selects a) what
+ * the current tuning phase, called evidence. The evidences are handed on to a tuningStrategy, which selects a) what
  * Configuration to test next and b) which configuration is the best in this tuning phase.
+ * If it should not look for a new optimum it is not in a tuning phase.
  *
  * @tparam Particle
  * @tparam ParticleCell
--- a/src/autopas/selectors/AutoTuner.h
+++ b/src/autopas/selectors/AutoTuner.h
@@ -10,6 +10,7 @@
 #include <memory>
 #include <set>
 
+#include "Smoothing.h"
 #include "autopas/options/DataLayoutOption.h"
 #include "autopas/options/Newton3Option.h"
 #include "autopas/options/TraversalOption.h"
@@ -25,10 +26,15 @@
 namespace autopas {
 
 /**
- * Automated tuner for optimal iteration performance.
+ * Calls to the iteratePairwise() method are passed through this class for two reasons:
+ * 1. Measuring time of the iteration.
+ * 2. Selecting an appropriate configuration for the pairwise iteration.
  *
- * This class offers an interface to the iteratePairwise method.
- * Internally it chooses the best container, traversal etc for the current simulation.
+ * The tuner can be in one of two states, depending on whether it currently should look for a new optimum,
+ * which is called a tuning phase. During a tuning phase,for each Configuration multiple measurements can be taken,
+ * which are called samples. To reduce noise, the samples for one configuration are then condensed to one value for
+ * the current tuning phase, called evidence. These evidence are handed on to a tuningStrategy, which selects a) what
+ * Configuration to test next and b) which configuration is the best in this tuning phase.
  *
  * @tparam Particle
  * @tparam ParticleCell
@@ -141,6 +147,8 @@ class AutoTuner {
    */
   template <class PairwiseFunctor>
   void addTimeMeasurement(PairwiseFunctor &pairwiseFunctor, long time) {
+    const auto &currentConfig = _tuningStrategy->getCurrentConfiguration();
+
     if (pairwiseFunctor.isRelevantForTuning()) {
       if (_samples.size() < _maxSamples) {
         AutoPasLog(trace, "Adding sample.");
@@ -148,16 +156,21 @@ class AutoTuner {
         // if this was the last sample:
         if (_samples.size() == _maxSamples) {
           auto reducedValue = OptimumSelector::optimumValue(_samples, _selectorStrategy);
-          _tuningStrategy->addEvidence(reducedValue, _iteration);
+
+          _evidences[currentConfig].emplace_back(_iteration, reducedValue);
+          auto smoothedValue = Smoothing::smoothLastPoint(_evidences[currentConfig], .25);
+
+          _tuningStrategy->addEvidence(smoothedValue, _iteration);
 
           // print config, times and reduced value
           if (autopas::Logger::get()->level() <= autopas::Logger::LogLevel::debug) {
             std::ostringstream ss;
             // print config
-            ss << _tuningStrategy->getCurrentConfiguration().toString() << " : ";
+            ss << currentConfig.toString() << " : ";
             // print all timings
             ss << utils::ArrayUtils::to_string(_samples, " ", {"[ ", " ]"});
-            ss << " Reduced value: " << reducedValue;
+            ss << " Reduced value: " << reducedValue;  // TODO: remove this. Should not break performance testing tool.
+            ss << " Smoothed value: " << smoothedValue;
             AutoPasLog(debug, "Collected times for  {}", ss.str());
           }
         }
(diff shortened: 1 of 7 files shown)


SteffenSeckler:
Do you happen to have any measurements that show improvements using smoothing?

src/autopas/selectors/Smoothing.cpp
```diff
@@ -0,0 +1,157 @@
...
+      return points[0].second;
+    }
+    return 0;
+  }
+
+  // pick at least two points and not more than total number of points
+  size_t pointsPerEstimation = std::max(static_cast<size_t>(span * points.size()), 2ul);
+
+  // only fit last point
+  size_t i = points.size() - 1;
```
SteffenSeckler, line 141: rename `i`

src/autopas/selectors/Smoothing.cpp
```diff
@@ -0,0 +1,157 @@
...
+      weight /= sumOfWeights;
+    }
+  }
+  return std::make_tuple(weights, fitIsOk);
+}
+
+/**
+ * Calculates the smoothed y-value of the last point in the vector.
+ * The fitted value is the sum of projections of the y-values of the neighbors in the chosen interval.
+ * Each porjection is the sum of the respective weight and the residuals proportion of the weighted sum of squared
```
SteffenSeckler, line 78:
```suggestion
 * Each projection is the sum of the respective weight and the proportion of the residuals of the weighted sum of squared
```

src/autopas/selectors/AutoTuner.h
```diff
@@ -25,10 +26,15 @@
...
+ * 1. Measuring time of the iteration.
+ * 2. Selecting an appropriate configuration for the pairwise iteration.
  *
- * This class offers an interface to the iteratePairwise method.
- * Internally it chooses the best container, traversal etc for the current simulation.
+ * The tuner can be in one of two states, depending on whether it currently should look for a new optimum,
+ * which is called a tuning phase. During a tuning phase,for each Configuration multiple measurements can be taken,
+ * which are called samples. To reduce noise, the samples for one configuration are then condensed to one value for
+ * the current tuning phase, called evidence. These evidence are handed on to a tuningStrategy, which selects a) what
+ * Configuration to test next and b) which configuration is the best in this tuning phase.
```
SteffenSeckler, line 37:
```suggestion
 * The tuner can be in one of two states. If it currently should look for a new optimum, it is in the 
 * so-called tuning phase. During a tuning phase, for each Configuration, multiple measurements can be taken,
 * which are called samples. To reduce noise, the samples for one configuration are then condensed to one value for
 * the current tuning phase, called evidence. The evidences are handed on to a tuningStrategy, which selects a) what
 * Configuration to test next and b) which configuration is the best in this tuning phase.
 * If it should not look for a new optimum it is not in a tuning phase.
```

The first sentence is awkward:
- depending -> which is called.
- The second state is somehow missing
-> added

src/autopas/selectors/AutoTuner.h
```diff
@@ -141,23 +147,30 @@ class AutoTuner {
...
           // print config, times and reduced value
           if (autopas::Logger::get()->level() <= autopas::Logger::LogLevel::debug) {
             std::ostringstream ss;
             // print config
-            ss << _tuningStrategy->getCurrentConfiguration().toString() << " : ";
+            ss << currentConfig.toString() << " : ";
             // print all timings
             ss << utils::ArrayUtils::to_string(_samples, " ", {"[ ", " ]"});
-            ss << " Reduced value: " << reducedValue;
+            ss << " Reduced value: " << reducedValue;  // TODO: remove this. Should not break performance testing tool.
```
SteffenSeckler, line 172: todo?

src/autopas/selectors/AutoTuner.h
```diff
@@ -200,22 +213,45 @@ class AutoTuner {
...
+   *
+   * @note Initialized with size of _maxSamples to start tuning at start of simulation.
    */
   std::vector<size_t> _samples;
+
+  /**
+   * For each configuration the collection of all evidence (smoothed values) collected so far and in which iteration.
+   * Configuration -> vector< iteration, time >
+   */
+  std::map<Configuration, std::vector<std::pair<size_t, size_t>>> _evidences;
```
SteffenSeckler, line 254: are these evidences cleared or is their size reduced at any point? Currently, it looks like an ever growing vector...

tests/testAutopas/tests/selectors/SmoothingTest.cpp
```diff
@@ -0,0 +1,27 @@
...
+  std::vector<std::pair<size_t, size_t>> obs;
+  obs.reserve(xvals.size());
+  for (size_t i = 0; i < xvals.size(); ++i) {
+    obs.emplace_back(xvals[i], yvals[i]);
+  }
+
+  // YS values with F = .25, NSTEPS = 0, DELTA = 0.0
+  {
+    auto out = autopas::smoothing::smoothLastPoint(obs, .25);
+    // out should be an integer (or long)
```
SteffenSeckler, line 24: What is it now?

tests/testAutopas/tests/selectors/SmoothingTest.cpp
```diff
@@ -0,0 +1,27 @@
...
+    EXPECT_EQ(out, 6);
```
SteffenSeckler, line 25: Why is it 6? Please explain.

src/autopas/selectors/Smoothing.cpp
```diff
@@ -0,0 +1,157 @@
...
+  auto maxDistFromIntervalEdgeHigh = maxDistFromIntervalEdge * .999;
+  auto maxDistFromIntervalEdgeLow = maxDistFromIntervalEdge * .001;
+
+  double sumOfWeights = 0.0;
+
+  // compute all weighs that can be assumed to be non-zero
+  for (size_t j = 0; j < weights.size(); ++j) {
+    auto xj = points[j + firstIndex].first;
+    // residual = |xi - xj|
+    size_t residual = xj > xi ? xj - xi : xi - xj;
```
SteffenSeckler, line 43:
```suggestion
    size_t residual = std::abs(xj - xi);
```

src/autopas/selectors/Smoothing.cpp
```diff
@@ -0,0 +1,157 @@
...
+                                                             size_t pointsPerEstimation,
+                                                             size_t maxDistFromIntervalEdge) {
+  const size_t i = points.size() - 1;
+  const size_t firstIndex = i - pointsPerEstimation + 1;
+  // initialize all weights with 0
+  std::vector<double> weights(pointsPerEstimation);
+
+  const auto &xi = points[i].first;
+
+  // define thresholds for shortcuts: if residuals are beyond these values the
```
SteffenSeckler, line 32:
```suggestion
  // Define thresholds for shortcuts: If residuals are beyond these values, they
```

src/autopas/selectors/Smoothing.cpp
```diff
@@ -0,0 +1,157 @@
...
+  // are assumed to be 0 respectively 1
```
SteffenSeckler, line 33:
```suggestion
  // are assumed to be 0, respectively 1.
```

SteffenSeckler (commented):
Some comments. Did not look into the math...

src/autopas/selectors/Smoothing.cpp
FG-TUM: this actually does not work since the subtraction of two `size_t` is always positive due to overflow.

src/autopas/selectors/AutoTuner.h
FG-TUM: True, it is an ever-growing vector. For the smoothing, we could restrict it to the number of points that we use for the fitting. However, we might want to move or expose this vector also to the tuning strategies to avoid redundancy, since some already collect this information too.

tests/testAutopas/tests/selectors/SmoothingTest.cpp
FG-TUM: A `size_t` which is totally fine :D

FG-TUM:
> Do you happen to have any measurements that show improvements using smoothing?

See updated description.

src/autopas/selectors/AutoTuner.h
```diff
@@ -141,23 +148,30 @@ class AutoTuner {
...
       if (_samples.size() < _maxSamples) {
         AutoPasLog(trace, "Adding sample.");
         _samples.push_back(time);
         // if this was the last sample:
         if (_samples.size() == _maxSamples) {
           auto reducedValue = OptimumSelector::optimumValue(_samples, _selectorStrategy);
-          _tuningStrategy->addEvidence(reducedValue, _iteration);
+
+          _evidences[currentConfig].emplace_back(_iteration, reducedValue);
+          auto smoothedValue = smoothing::smoothLastPoint(_evidences[currentConfig], 5);
```
SteffenSeckler, line 162: so right now, 5 pieces of evidence are used for the smoothing and it is always enabled, correct?

FG-TUM: Up to 5, depending on how many are available but yes.
I actually see no advantage in disabling this feature. We could make the 5 a changeable parameter from the outside if you like but it is difficult to compare multiple runs anyway.

SteffenSeckler (approved):

PR #565, opened 2020-12-13
Resize AutoPas

FG-TUM:
# Description

- [x] resize() for changing the size of the bounding box without deleting the AutoPas object
- [x] Slight refactoring of container tests

## ~Resolved Issues~

# How Has This Been Tested?

- [x] New tests for resize

### DIFF
--- a/src/autopas/LogicHandler.h
+++ b/src/autopas/LogicHandler.h
@@ -101,7 +101,7 @@ class LogicHandler {
       if (not pIter->isOwned()) {
         utils::ExceptionHandler::exception(
             "LogicHandler::resizeBox() encountered non owned particle. "
-            "When calling resizeBox() these should be already delete. "
+            "When calling resizeBox() these should be already deleted. "
             "This could be solved by calling updateContainer() before resizeBox().");
       }
       // owned particles that are now outside are removed from the container and returned
--- a/src/autopas/LogicHandler.h
+++ b/src/autopas/LogicHandler.h
@@ -53,6 +53,17 @@ class LogicHandler {
   }
 
   /**
+   * Pass values to the towards the actual container.
+   * @param boxMin
+   * @param boxMax
+   */
+  void resizeBox(const std::array<double, 3> &boxMin, const std::array<double, 3> &boxMax) {
+    // TODO: add sanity checks here?
+    // TODO: add warning when box size changes too drastically?
+    _autoTuner->resizeBox(boxMin, boxMax);
+  }
+
+  /**
    * @copydoc AutoPas::addParticle()
    */
   void addParticle(const Particle &p) {
--- a/src/autopas/LogicHandler.h
+++ b/src/autopas/LogicHandler.h
@@ -62,6 +62,11 @@ class LogicHandler {
     const auto &oldMin = _autoTuner.getContainer()->getBoxMin();
     const auto &oldMax = _autoTuner.getContainer()->getBoxMax();
 
+    // if nothing changed do nothing
+    if (oldMin == boxMin and oldMax == boxMax) {
+      return {};
+    }
+
     // sanity check that new size is actually positive
     for (size_t i = 0; i < boxMin.size(); ++i) {
       if (boxMin[i] >= boxMax[i]) {
--- a/src/autopas/LogicHandler.h
+++ b/src/autopas/LogicHandler.h
@@ -58,8 +58,35 @@ class LogicHandler {
    * @param boxMax
    */
   void resizeBox(const std::array<double, 3> &boxMin, const std::array<double, 3> &boxMax) {
-    // TODO: add sanity checks here?
-    // TODO: add warning when box size changes too drastically?
+    const auto &oldMin = _autoTuner.getContainer()->getBoxMin();
+    const auto &oldMax = _autoTuner.getContainer()->getBoxMax();
+
+    // sanity check that new size is actually positive
+    for (size_t i = 0; i < boxMin.size(); ++i) {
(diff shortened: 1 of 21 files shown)


src/autopas/LogicHandler.h
```diff
@@ -52,6 +52,44 @@ class LogicHandler {
...
+    // warn if domain changes too drastically
+    const auto newLength = utils::ArrayMath::sub(boxMax, boxMin);
+    const auto oldLength = utils::ArrayMath::sub(oldMax, oldMin);
+    const auto relDiffLength = utils::ArrayMath::div(newLength, oldLength);
+    for (size_t i = 0; i < newLength.size(); ++i) {
+      // warning threshold is set arbitrary and up for change if needed
+      if (relDiffLength[i] > 1.3 or relDiffLength[i] < 0.7) {
+        AutoPasLog(warn,
+                   "LogicHandler.resize(): Domain size changed drastically in dimesion {}! Gathered AutoTuning "
+                   "information might by not applicable anymore!\n"
```
SteffenSeckler, line 82:
```suggestion
                   "LogicHandler.resize(): Domain size changed drastically in dimension {}! Gathered AutoTuning "
                   "information might not be applicable anymore!\n"
```

src/autopas/LogicHandler.h
```diff
@@ -52,6 +52,44 @@ class LogicHandler {
     }
   }
 
+  /**
+   * Pass values to the towards the actual container.
```
SteffenSeckler, line 56:
```suggestion
   * Pass values to the actual container.
```

src/autopas/containers/verletClusterLists/ClusterTower.h
```diff
@@ -181,7 +181,13 @@ class ClusterTower : public ParticleCell<Particle> {
    * Returns the number of particles in the tower that are not dummies.
    * @return the number of particles in the tower that are not dummies.
    */
-  [[nodiscard]] size_t getNumActualParticles() const { return _particlesStorage.numParticles() - _numDummyParticles; }
+  [[nodiscard]] size_t getNumActualParticles() const {
+    // This can actually fail in some tests. Also can't do max() because these are size_t
```
SteffenSeckler, line 185: Why would this fail, shouldn't in these cases numDummyParticles be reset?

src/autopas/selectors/ContainerSelector.h
```diff
@@ -53,6 +53,27 @@ class ContainerSelector {
...
+    _boxMax = boxMax;
+
+    // depening on their position convert particles to the corresponding ownership state
+    for (auto &p : *_currentContainer) {
+      if (utils::inBox(p.getR(), _boxMin, _boxMax)) {
+        p.setOwnershipState(OwnershipState::owned);
+      } else {
+        p.setOwnershipState(OwnershipState::halo);
+      }
+    }
```
SteffenSeckler, line 72: NO!
Previously owned particles should be removed from the container!
And halo particles should probably not be allowed at this stage.

src/autopas/utils/ArrayMath.h
```diff
@@ -99,6 +99,23 @@ template <class T, std::size_t SIZE>
   return result;
 }
 
+/**
+ * Multiplies two array's element wise and returns the result.
```
SteffenSeckler, line 103:
```suggestion
 * Divides two array's element-wise and returns the result.
```

src/autopas/utils/ArrayMath.h
```diff
@@ -99,6 +99,23 @@ template <class T, std::size_t SIZE>
...
+ * @tparam T floating point type
+ * @tparam SIZE size of the arrays
+ * @param a
+ * @param b
+ * @return element-wise multiplication of a and b
```
SteffenSeckler, line 108:
```suggestion
 * @return element-wise quotient of a and b
```

tests/testAutopas/tests/containers/AllContainersTests.h
```diff
@@ -37,5 +38,39 @@ class AllContainersTests : public AutoPasTestBase, public ::testing::WithParamIn
...
+      // boxMin + i/2 * boxMax
+      auto pos =
+          autopas::utils::ArrayMath::add(nearBoxMin, autopas::utils::ArrayMath::mulScalar(nearBoxLength, i * 0.5));
+      Particle particle(pos, {0., 0., 0.}, 1);
+      container->addParticle(particle);
+      addedParticles.push_back(particle);
+    }
+    return addedParticles;
+  }
+
```
SteffenSeckler, line 74: Unused.
```suggestion
```

SteffenSeckler (changes requested):
The resize method should not be allowed to change the ownership state of particles.
Instead, previously owned particles that are now no longer owned should be returned from AutoPas.

src/autopas/selectors/ContainerSelector.h
FG-TUM: done 5a7fab2

src/autopas/containers/verletClusterLists/ClusterTower.h
SteffenSeckler: ?

FG-TUM: I swear I had a test failing because this function returned an underflow but I can not reproduce it anymore. Either I fixed It by changing something else or I'm going crazy :D Either way, I will revert this change so we get a failing test again if `numDummyParticles` is not handled properly.

SteffenSeckler (approved):
only minor things

PR #568, opened 2021-01-26
disable building of tests for lgtm

SteffenSeckler:
# Description

- disables building of tests for lgtm.
- removes a few configurations from jenkinsfile
- minor @file fixes

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -8,6 +8,8 @@ extraction:
         - cmake
         - libssl-dev
     after_prepare:    # Customizable step used by all languages.
+      # installs cmake!
+      - cmake --version
       - wget -O cmake.tar.gz https://github.com/Kitware/CMake/releases/download/v3.16.4/cmake-3.16.4.tar.gz
       - tar zxf cmake.tar.gz
       - cd cmake-*/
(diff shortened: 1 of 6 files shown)


.lgtm.yml
```diff
@@ -8,6 +8,8 @@ extraction:
         - cmake
         - libssl-dev
     after_prepare:    # Customizable step used by all languages.
+      # installs cmake!
+      - cmake --version
```
FG-TUM, line 12: version twice?

Jenkinsfile
```diff
@@ -184,28 +184,6 @@ pipeline {
...
-                    steps {
-                        container('autopas-gcc7-cmake-make') {
-                            dir("build-addresssanitizer-release") {
-                                sh "cmake -DCCACHE=ON -DCMAKE_BUILD_TYPE=Release -DAUTOPAS_ENABLE_ADDRESS_SANITIZER=ON .."
-                                sh "entrypoint.sh make -j 4 > buildlog.txt 2>&1 || (cat buildlog.txt && exit 1)"
-                                sh './tests/testAutopas/runTests'
-                            }
-                        }
-                    }
-                }
```
FG-TUM, line 208: you remove this because it seems redundant with the clang address sanitizer stages?

FG-TUM (commented):
two questions

.lgtm.yml
SteffenSeckler: check before and after. maybe at some point we don't need the manual install anymore ;)

Jenkinsfile
SteffenSeckler: yes

FG-TUM (approved):

examples/md-flexible/tests/CMakeLists.txt
```diff
@@ -1,3 +1,8 @@
+if (NOT AUTOPAS_BUILD_TESTS)
+    message(STATUS "AUTOPAS_BUILD_TESTS is disabled. Not building MDFlex tests!")
+    return()
+endif ()
```
SteffenSeckler, line 4: @FG-TUM: I've had to disable the md-flex tests in case of AUTOPAS_BUILD_TESTS=OFF, as it depends on things defined in the autopas tests.
Is it intended that way or are the dependencies unwanted?

FG-TUM: We never thought about having the example tests without the actual tests but I guess it would make sense that this is possible.

SteffenSeckler: => #569

PR #570, opened 2021-01-29
Msvc support

SteffenSeckler:
# Description

Please include a summary of the changes.

## Related Pull Requests

- #PR

## Resolved Issues

- [ ] fixes #Issue

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes.

- [ ] Test A
- [ ] Test B

PR #572, opened 2021-02-01
Arbitrary Fixes

SteffenSeckler:
# Description

A number of fixes discovered when adding MSVC support.


FG-TUM (approved):

FG-TUM:
so we can close #570 now?

PR #575, opened 2021-02-09
Flush logger always

FG-TUM:
# Description

- [x] flush tuning result logger always instantly so it does print its information if autopas is killed

Since this logger only writes once per tuning phase there is no significant (or measurable) performance drop.

## ~Resolved Issues~

# How Has This Been Tested?

- [x] Ran it, killed it, looked at output


SteffenSeckler (commented):
two questions:
1. it appears that you are now using a synchronous logger. What's the reason for that?
2. <strike>always flushing normally significantly decreases performance. is this fine here, as the logger is called seldom enough?</strike> -- ok this is already mentioned in the description...

FG-TUM:
> 1. it appears that you are now using a synchronous logger. What's the reason for that?

I really want to make sure that the information is written to the log file immediately. Calling flush on an asynchronous logger does not flush it immediately.

SteffenSeckler (approved):

PR #577, opened 2021-02-17
Progress Bar

FG-TUM:
# Description

- [x] Progress bar for md-flex
- [x] Progress bar adapts itself to terminal width
- [x] additional progress info
- [x] toggleable via cli
- [x] toggleable via yaml

## ~Resolved Issues~

# How Has This Been Tested?

- [x] Looked at the output with bar enabled
- [x] Looked at the output with bar disabled

## Screenshots 
When number of iterations is only an estimate:
progress_1
When number of iterations is known exactly:
progress_32
progress_100

### DIFF
--- a/examples/md-flexible/src/Simulation.h
+++ b/examples/md-flexible/src/Simulation.h
@@ -729,7 +729,10 @@ void Simulation<Particle>::printProgress(size_t iterationProgress, size_t maxIte
 template <class Particle>
 std::tuple<size_t, bool> Simulation<Particle>::estimateNumIterations() const {
   if (_config->tuningPhases.value > 0) {
-    auto estimate = _config->tuningInterval.value * _config->tuningPhases.value + 100;
+    // @TODO: this can be improved by considering the tuning strategy
+    // 30 is just a random number for the number of tested configurations per tuning phase
+    auto estimate = (_config->tuningPhases.value - 1) * _config->tuningInterval.value +
+                    (_config->tuningPhases.value * _config->tuningSamples.value * 30);
     return {estimate, false};
   } else {
     return {_config->iterations.value, true};
--- a/examples/md-flexible/src/Simulation.h
+++ b/examples/md-flexible/src/Simulation.h
@@ -5,7 +5,11 @@
  */
 #pragma once
 
+#include <sys/ioctl.h>
+#include <unistd.h>
+
 #include <fstream>
+#include <iomanip>
 #include <iostream>
 
 #include "BoundaryConditions.h"
@@ -97,6 +101,12 @@ class Simulation {
   [[nodiscard]] bool needsMoreIterations() const;
 
   /**
+   * Gives an estimate of how many iterations the simulation will do in total.
+   * @return The estimate and true iff this is the correct number and not only an estimate.
+   */
+  [[nodiscard]] std::tuple<size_t, bool> estimateNumIterations() const;
+
+  /**
    * Prints statistics like duration of calculation etc of the Simulation.
    * @param autopas
    */
@@ -116,6 +126,16 @@ class Simulation {
   double calculateHomogeneity(autopas::AutoPas<Particle> &autopas);
 
  private:
+  /**
+   * Print a progressbar and progress information to the console.
+   * @note Calling this function deletes the whole current line in the terminal.
+   *
+   * @param iteration
+   * @param maxIterations
+   * @param maxIsPrecise Indicate whether maxIterations is precise (true) or an estimate (false).
+   */
+  void printProgress(size_t iteration, size_t maxIterations, bool maxIsPrecise);
+
   using ParticlePropertiesLibraryType = ParticlePropertiesLibrary<double, size_t>;
   constexpr static bool _shifting = true;
   constexpr static bool _mixing = true;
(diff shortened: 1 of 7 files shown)


SteffenSeckler:
can we get a screenshot?

SteffenSeckler:
fyi: some tests are failing

examples/md-flexible/src/Simulation.h
```diff
@@ -650,3 +679,46 @@ double Simulation<Particle>::calculateHomogeneity(autopas::AutoPas<Particle> &au
...
+
+  // actual progress bar
+  std::stringstream progressbar;
+  progressbar << "[";
+  // get current terminal width
+  struct winsize w;
+  ioctl(STDOUT_FILENO, TIOCGWINSZ, &w);
+  auto terminalWidth = w.ws_col;
+  // the bar should fill the terminal window so subtract everything else (-2 for "] ")
+  size_t maxBarWidth = terminalWidth - info.str().size() - progressbar.str().size() - 2;
```
SteffenSeckler, line 708: error could happen here, if `terminalWidth == 0`, i.e., no terminal was detected

FG-TUM:
> can we get a screenshot?

You and your majesty now may look at the updated description :P

SteffenSeckler:
is it possible to automatically detect if the output is not going to the terminal, but to a file? in that case, automatically disabling the progress bar would be nice :)
otherwise, the lines would be printed multiple times (see jenkins).

FG-TUM:
> is it possible to automatically detect if the output is not going to the terminal, but to a file? in that case, automatically disabling the progress bar would be nice :)
> otherwise, the lines would be printed multiple times (see jenkins).

Probably yes but tbh I can't be bothered at the moment. I would disable the output of this in jenkins via `--no-progress-bar` since it is only eye-candy anyway. However, I will add a few checks to catch errors when querying the terminal fails...

SteffenSeckler:
how about disabling it by default?

FG-TUM:
> how about disabling it by default?

The idea of md-flexible is to be a quick testing/demonstration tool. So it is primarily intended to be used manually and there the progress bar is helpful. Also if you manually need to switch it on you probably will rarely do so, however, it is no big effort to manually disable it if it causes problems.

examples/md-flexible/src/Simulation.h
```diff
@@ -650,3 +679,62 @@ double Simulation<Particle>::calculateHomogeneity(autopas::AutoPas<Particle> &au
...
+  // print everything
+  std::cout << progressbar.str() << info.str() << std::flush;
+}
+template <class Particle>
+std::tuple<size_t, bool> Simulation<Particle>::estimateNumIterations() const {
+  if (_config->tuningPhases.value > 0) {
+    // @TODO: this can be improved by considering the tuning strategy
+    // 30 is just a random number for the number of tested configurations per tuning phase
+    auto estimate = (_config->tuningPhases.value - 1) * _config->tuningInterval.value +
+                    (_config->tuningPhases.value * _config->tuningSamples.value * 30);
```
SteffenSeckler, line 735:
```suggestion
    // estimatedNumConfigs is just a random number for the number of tested configurations per tuning phase.
    size_t estimatedNumConfigs = 30;
    auto estimate = (_config->tuningPhases.value - 1) * _config->tuningInterval.value +
                    (_config->tuningPhases.value * _config->tuningSamples.value * estimatedNumConfigs);
```

30 appears to be a bit small for fullsearch (the default):
image
and a bit big for bayesian-search:
image

but, well, doesn't matter anyways...

examples/md-flexible/src/Simulation.h
```diff
@@ -381,6 +406,10 @@ void Simulation<Particle>::simulate(autopas::AutoPas<Particle> &autopas) {
       }
     }
   }
+  // final update for a full progress bar
+  if (not _config->dontShowProgressBar.value) {
+    printProgress(iteration, maxIterationsEstimate, maxIterationsIsPrecice);
```
SteffenSeckler, line 411:
```suggestion
    // the last update is precise, so we know the number of iterations.
    printProgress(iteration, iteration, true);
```

examples/md-flexible/src/Simulation.h
```diff
@@ -319,11 +339,16 @@ void Simulation<Particle>::simulate(autopas::AutoPas<Particle> &autopas) {
   this->homogeneity = Simulation::calculateHomogeneity(autopas);
   _timers.simulate.start();
 
+  auto [maxIterationsEstimate, maxIterationsIsPrecice] = estimateNumIterations();
```
SteffenSeckler, line 342:
```suggestion
  auto [maxIterationsEstimate, maxIterationsIsPrecise] = estimateNumIterations();
```

examples/md-flexible/src/Simulation.h
```diff
@@ -319,11 +339,16 @@ void Simulation<Particle>::simulate(autopas::AutoPas<Particle> &autopas) {
...
+
   // main simulation loop
   for (; needsMoreIterations(); ++iteration) {
     if (autopas::Logger::get()->level() <= autopas::Logger::LogLevel::debug) {
       std::cout << "Iteration " << iteration << std::endl;
     }
+    if (not _config->dontShowProgressBar.value) {
+      printProgress(iteration, maxIterationsEstimate, maxIterationsIsPrecice);
```
SteffenSeckler, line 350:
```suggestion
      printProgress(iteration, maxIterationsEstimate, maxIterationsIsPrecise);
```

SteffenSeckler (changes requested):
You appear to have been influenced by https://github.com/precice/precice ;)
There is also a newline missing after the progress bar (visible after resizing the window):

image

SteffenSeckler:
needs formatting

examples/md-flexible/src/Simulation.h
```diff
@@ -381,6 +406,13 @@ void Simulation<Particle>::simulate(autopas::AutoPas<Particle> &autopas) {
       }
     }
   }
+  // final update for a full progress bar
+  if (not _config->dontShowProgressBar.value) {
+    // The last update is precise, so we know the number of iterations.
+    printProgress(iteration, iteration, true);
+    // The progress bar does not end the line. Since this is the last progress bar end the line here.
```
SteffenSeckler, line 413:
```suggestion
    // The progress bar does not end the line. Since this is the last progress bar, end the line here.
```

SteffenSeckler (approved):
once comma

SteffenSeckler (approved):

PR #579, opened 2021-02-20
Fix Race Condition in VCL

### LINKED ISSUE #574: Exploding Liquid with VCL crashes
**Describe the bug**
Using ls1 with AutoPas the exploding liquid scenario crashes as soon as VCL is used. With every other container, the simulation runs as expected.

**To Reproduce**
TODO

**Expected behavior**
Valid simulation.

**Additional context**

FG-TUM:
# Description

Fixes data race when inserting particles in VCL.
- [x] turn _particlesToAdd from `vector` to `vector` of thread buffers
- [x] adapt iterators to process thread buffers in parallel
- [x] builder still takes `vector` (now of accumulated buffers)

 ## Related Pull Requests
This PR includes:
- [x] #580 (better iterator tests)
- [x] #583 (forceSequential mode)
- ls1: [Resize AutoPas](https://github.com/ls1mardyn/ls1-mardyn/pull/169)

## Resolved Issues
- [x] fixes #574 via #583

# How Has This Been Tested?

- [x] Old unit tests
- [x] New unit tests that test the particle buffer
- [x] exploding liquid (previously deadlocks)


SteffenSeckler:
asan detects heap-buffer-overflow

FG-TUM:
Expl liquid now works with VCL but updating the halo is excruciatingly slow...

src/autopas/iterators/ParticleIterator.h
```diff
@@ -13,14 +13,27 @@
...
+ * ParticleIterator class to access particles inside of a cell based container.
+ * The particles can be accessed using "iterator->" or "*iterator". The iterator is incremented using the ++operator.
+ *
+ * Incrementing this iterator works by jumping through the underlying particle cells from particle to particle
+ * incrementing until itfinds a particle that matches the iteration criteria (e.g. it continues to loop until it finds
+ * a halo particle if _behavior==halo). Additionally there might be additional particle vectors which are not part
+ * of the cell structure and contain arbitrary unsorted particles. These buffers result from containers that
+ * have buffers they only sometimes merge with the actual data structure. Typically there is one buffer per OpenMP
+ * thread, however, there is no guarantee that the iterator region has the same number of therads so the number of
+ * buffers is considered arbitrary.
```
SteffenSeckler, line 31:
```suggestion
 * a halo particle if _behavior==halo). Additionally, there might be additional particle vectors that are not part
 * of the cell structure and contain arbitrary unsorted particles. These buffers result from containers that
 * have buffers they only sometimes merge with the actual data structure. Typically there is one buffer per OpenMP
 * thread, however, there is no guarantee that the iterator region has the same number of threads so the number of
 * buffers is considered arbitrary.
```

src/autopas/iterators/ParticleIterator.h
```diff
@@ -40,18 +54,23 @@ class ParticleIterator : public ParticleIteratorInterfaceImpl<Particle, modifiab
    *
    * @param cont Linear data vector of ParticleCells.
    * @param flagManager The CellBorderAndFlagManager that shall be used to query the cell types.
-   * Can be nullptr if the behavior is haloAndOwned.
+   * Can be nullptr if the behavior is ownedOrHalo.
    * @param behavior The IteratorBehavior that specifies which type of cells shall be iterated through.
-   * @param additionalParticleVectorToIterate Additional Particle Vector to iterate over.
+   * @param additionalVectorsToIterate Thread buffers of additional Particle Vector to iterate over.
```
SteffenSeckler, line 59:
```suggestion
   * @param additionalVectorsToIterate Thread buffers of additional Particle vector to iterate over.
```

SteffenSeckler (changes requested):
Some things:
* change title of PR
* see comments
* some more doc for forceSequential might be good.
* maybe prevent user from using forceSequential through LogicHandler? -- is this possible with tests?

FG-TUM:
> * change title of PR
Any suggestions? :P

> * maybe prevent user from using forceSequential through LogicHandler? -- is this possible with tests?
Why would we? If you have an iterator loop and within that region iterator you already need it and I think it is too far-fetched to say that this will never be relevant for user-code.

SteffenSeckler (approved):

SteffenSeckler:
> > * change title of PR
> 
> Any suggestions? :P
Hmm, not really. Maybe you find sth. maybe you don't, either case you can merge it.

> 
> > * maybe prevent user from using forceSequential through LogicHandler? -- is this possible with tests?
> 
> Why would we? If you have an iterator loop and within that region iterator you already need it and I think it is too far-fetched to say that this will never be relevant for user-code.
ok, convinced. is there some way to ensure that you can only use them in the inner scope?

FG-TUM:
> > > * change title of PR
> > 
> > 
> > Any suggestions? :P
> 
> Hmm, not really. Maybe you find sth. maybe you don't, either case you can merge it.

TBH the title describes pretty well what the intent and effect of the PR is. Everything else that happens here is part of other PRs which are mentioned and linked. Therefore I would leave it as it is.

> > > * maybe prevent user from using forceSequential through LogicHandler? -- is this possible with tests?
> > 
> > 
> > Why would we? If you have an iterator loop and within that region iterator you already need it and I think it is too far-fetched to say that this will never be relevant for user-code.
> 
> ok, convinced. is there some way to ensure that you can only use them in the inner scope? probably not if iterators are used, right?

Nothing simple I can think of. Anyways iterators will be replaced sooner or later so I see no point in investing much more in them.

PR #580, opened 2021-02-23
Refactor Iterator Tests

### LINKED ISSUE #536: RegionIteratorTests where lower corner is actually the upper corner
Region Iterators expect a lower-front-left corner and an upper-back-right corner to define a region to iterate over. However, it is never actually enforced that these are the expected corners. Probably only some containers have their Iterators written in a way that is agnostic to this.

Unit test for all containers that check all possible combinations of corners.

Enforce corners to be the expected ones. This however would unnecessarily restrict the interface.

~**Additional context**~

FG-TUM:
# Description

- [x] Change `IteratorBehavior` `enum` to an `Option` class.
- [x] Refactor `RegionParticleIteratorTest`
- [x] Merge common functionality of `IteratorTest`, `RegionParticleIteratorTest`, and `ParticleIteratorTest`
- [x] Add Test for Iterator copying (copy constructor, copy assignment)
- [x] Throw out `TouchableParticle`
- [x] Rename `IteratorTest` to `ParticleIteratorInterfaceTest`
- [x] Discourage particle insertion while iterating (see below)

Currently, we don't explicitly say that we do not support particle addition while iterating. However, I strongly suggest discouraging this! While most containers (except LCR) correctly handle the case that a particle is inserted in a cell that the iterator of the same thread is currently covering or will cover in the future, insertion by thread A into a cell that is currently iterated by thread B is unsafe! This could be circumvented if all containers work with insertion buffers, or even better when the logic handler buffers particle insertion.

## Related Pull Requests

- merges into #579 

## Resolved Issues

- [x] closes #536 -> not supported and checked in the logic handler

# State of the tests before this PR

## IteratorTest

* uses the AutoPas interface
* all parametrized for container, csf, const, priorForceCalc

* ParticleAdditionAndIteratorTestNormal
    * parametrized for container, csf, const, priorForceCalc
    * testAdditionAndIteration
        * Add some strategically placed owned (3) and halo (7) particles around critical spots of the domain
        * Do or don't do a force calculation
        * Iterate with normal iterator and every behavior and check the number of found particles
        * Iterate domain+halo with region iterator and every behavior
        * checkRegionIteratorForAllParticles for every behavior
            * use regular Iterator over container
            * start a region iter around every particle and check dereferenced iters for equivalence
            * check that every particle is only found once
        * Move particles by skin/2
        * checkRegionIteratorForAllParticles for every behavior

* RangeBasedIterator
    * parametrized for container, csf, const, priorForceCalc
    * testRangeBasedIterator
        * Add some strategically placed owned (3) and halo (7) particles around critical spots of the domain
        * Do or don't do a force calculation
        * iterates all particles with range based and iterator based loop and compares what is found
        * NOTE: Sketchy! Comparison what was found should be done better

* testOpenMP${Region}Iterators${Behavior}
    * parametrized for container, csf, const, priorForceCalc
    * testOpenMPIterators
        * Add 500 random particles
        * Add 50 random halo particles
        * Do or don't do a force calculation
        * Iterate with OpenMP and touch everything according to behavior
        * Again without OpenMP and check that everything was touched once
    
* RegionIteratorDeletion
    * parametrized for container, csf, const, priorForceCalc
    * testRegionIteratorDeletion
        * fill AutoPas object with 200 random owned particles (might be in halo)
        * Do or don't do a force calculation
        * Save all generated particles in a vector
        * Use region iterator to delete all particles from a region and save what was deleted
        * Use regular Iterator to check everything that should be there is there
        * NOTE: looks improveable

## ParticleIteratorTest

* All tests use OpenMP if available

* testFullIterator_EFEFFEEFEF
    * parametrized for numThreads, numAdditionalParticleVectors
    * testAllParticlesFoundPattern
        * Create vector of cells, some filled with N particles according to pattern
        * Create some additional particle vectors
        * Use ParticleIterator loop, collect all found IDs
        * Check if all expected IDs were found

* testFullIterator_FEFEEFFEFE
    * same as above but different pattern

* testFullIterator_deletion
    * Create vector of cells, some filled with N particles according to pattern
    * Use iterator to loop over cells and delete all particles
    * Use iterator to loop over cells and confirm that there is nothing left

* testFullIterator_mutable
    * Create vector of cells, some filled with N particles according to pattern
    * Use iterator over cells to modify velocity data
    * Use iterator over cells to confirm modification

* testIteratorBehavior${Container}
    * Create container
    * Add two particles, one inner and one in halo
    * For every behavior check that the respective particles are found by the iterator

## RegionParticleIteratorTest

* uses container objects

* test${Container}RegionParticleIterator${Behavior}${Threads}
    * fill a container with 100 particles
    * touch them with a region iterator 
    * check if everything that is supposed to be touched is touched

* test${Container}RegionParticleIteratorEmpty
    * create an empty LC container
    * call iterator loop
    * check that the iterator is invalid

* testLinkedCellsRegionParticleIteratorCopyConstructor
    * fill a LC container with 100 particles
    * create two iterators where iterator2 is created by a copy constructor
    * touch all particles with iterator2
    * iterate with iterator1 and check that everything was touched

* testLinkedCellsRegionParticleIteratorCopyAssignment
    * same as above but with copy assignment instead of copy constructor

* testLinkedCellsRegionParticleIteratorSparseDomain
    * strategically place some particles in the domain
    * mark which should be touched and which not
    * iterate container and check that expectations match via inBox()


SteffenSeckler (commented):
some minor things

SteffenSeckler:
This PR should probably be renamed :)

README.md
```diff
@@ -175,10 +175,13 @@ for(auto iter = autoPas.getRegionIterator(lowCorner, highCorner); iter != autoPa
 Both `begin()` and `getRegionIterator()` can also take the additional parameter `IteratorBehavior`,
 which indicates over which particles the iteration should be performed. See [autopas::IteratorBehavior
 ](https://autopas.github.io/doxygen_documentation/git-master/namespaceautopas.html#a520fefd51e4555074cd16e7c3fd19c42) for possible options and details.
-The default parameter is `haloAndOwned`, which is also used for range-based for loops.
+The default parameter is `ownedOrHalo`, which is also used for range-based for loops.
 
 Analogously to `begin()`, `cbegin()` is also defined, which guarantees to return a `const_iterator`.
 
+Iterators are not guaranteed to be valid after particle insertion. 
+However particle deletion while iterating is supported via `autoPas.deleteParticle(iterator)`.
```
SteffenSeckler, line 183:
```suggestion
However, particle deletion while iterating is supported via `autoPas.deleteParticle(iterator)`.
```

src/autopas/options/IteratorBehavior.h
```diff
@@ -13,32 +13,55 @@
...
      * Iterate only over halo particles.
      */
-    haloOnly,
+    halo = 0b0010,
     /**
-     * Iterate only over owned particles.
+     * Iterate over both halo and owned particles. Defined fore ease of access.
      */
-    ownedOnly,
+    ownedOrHalo = 0b0011,
```
SteffenSeckler, line 41: Could theoretically be:
```suggestion
    ownedOrHalo = owned | halo,
```
No idea which is less error-prone...

src/autopas/options/IteratorBehavior.h
```diff
@@ -13,32 +13,55 @@
...
+     * Iterate only over dummy particles.
      */
-    haloAndOwned,
+    dummy = 0b0100,
     /**
-     * Iterate over both halo and owned particles and also dummy particles.
+     * Iterate over both halo and owned particles and also dummy particles. Defined fore ease of access.
      */
-    haloOwnedAndDummy,
+    ownedOrHaloOrDummy = 0b0111,
```
SteffenSeckler, line 49: Could theoretically be:
```suggestion
    ownedOrHaloOrDummy = owned | halo | dummy,
```
No idea which is less error-prone...

src/autopas/options/IteratorBehavior.h
```diff
@@ -50,6 +73,14 @@ class IteratorBehavior : public Option<IteratorBehavior> {
    */
   constexpr IteratorBehavior(Value option) : _value(option) {}
 
+  /**
+   * Constructor from number value.
+   * This is useful when combining values and directly using the result as argument of type IteratorBehavior.
+   * This is necessary since e.g. owned & halo results in an object of Value_t instead of Value.
+   * @param option
+   */
+  constexpr IteratorBehavior(Value_t option) : _value(static_cast<IteratorBehavior::Value>(option)) {}
```
SteffenSeckler, line 82: Alternatively, one could define the `operator |` for this class:
https://godbolt.org/z/q4GjYndT4

src/autopas/utils/logging/Logger.h
```diff
@@ -6,6 +6,7 @@
 
 #pragma once
 
+#include <spdlog/fmt/bundled/ranges.h>
```
SteffenSeckler, line 9: after including this, lots of calls to `arrayutils::to_string` could, theoretically, be removed.
Would result in different formatting, though, and might require the adaption of some scripts.

tests/testAutopas/tests/iterators/ParticleIteratorInterfaceTest.cpp
```diff
@@ -192,22 +162,24 @@ TEST_P(ParticleIteratorInterfaceTest, findAllParticlesAroundBoundaries) {
...
       break;
     }
-    case autopas::IteratorBehavior::haloOwnedAndDummy: {
-      GTEST_FAIL() << "IteratorBehavior::haloOwnedAndDummy should not be tested through this test"
-                      " as container behavior with dummy particles is not uniform.";
+    default: {
+      GTEST_FAIL() << "IteratorBehavior::" << behavior
+                   << "  should not be tested through this test!\n"
+                      "Container behavior with dummy particles is not uniform.\n"
+                      "forceSequential alone makes no sense.";
```
SteffenSeckler, line 182:
```suggestion
                      "Using forceSequential is not supported.";
```
even if `forceSequential` is not specified alone, the default branch will be hit.

tests/testAutopas/tests/iterators/ParticleIteratorInterfaceTest.cpp
```diff
@@ -245,21 +217,23 @@ TEST_P(ParticleIteratorInterfaceTest, deleteParticles) {
...
```
SteffenSeckler, line 236:
```suggestion
                      "Using forceSequential is not supported.";
```

tests/testAutopas/tests/iterators/RegionParticleIteratorTest.cpp
```diff
@@ -66,22 +63,24 @@ TEST_P(RegionParticleIteratorTest, testRegionAroundCorner) {
...
       break;
     }
-    case autopas::IteratorBehavior::haloOwnedAndDummy: {
-      GTEST_FAIL() << "IteratorBehavior::haloOwnedAndDummy should not be tested through this test"
-                      " as container behavior with dummy particles is not uniform.";
+    default: {
+      GTEST_FAIL() << "IteratorBehavior::" << behavior
+                   << "  should not be tested through this test!\n"
+                      "Container behavior with dummy particles is not uniform.\n"
+                      "forceSequential alone makes no sense.";
```
SteffenSeckler, line 83:
```suggestion
                      "Using forceSequential is not supported.";
```

SteffenSeckler (approved):
some minor things

src/autopas/options/IteratorBehavior.h
FG-TUM: I actually considered this, however, for consistency all operators should then be defined and I just couldn't be bothered ;)

src/autopas/utils/logging/Logger.h
FG-TUM: and is out of scope for this PR

src/autopas/options/IteratorBehavior.h
FG-TUM: Didn't know you could use the enums while defining them but I would go with the bit pattern for clarity. The Ferrer safety still comes with the `static_assert`s further down.

PR #584, opened 2021-03-12
remove unnecessary cast

FG-TUM:
# Description

remove unnecessary cast


SteffenSeckler (approved):
all good if the tests don't complain

PR #587, opened 2021-03-24
MPI_rank Suffix

JockelTUM:
# Description

Added a Suffix to the AUTOPAS_LOG and .vtk files in md-flexible, if AUTOPAS_MPI is enabled.
The Suffix describes to which MPI rank the files belongs.

# How Has This Been Tested?
The Suffix should only be added when AUTOPAS_MPI is enabled and should correspond to the number of ranks.

- [x] With AUTOPAS_MPI not enabled Suffix is not added.
- [x] for "mpirun -n 4" there are 4 corresponding .csv and .vtk files

### DIFF
--- a/examples/md-flexible/src/Simulation.cpp
+++ b/examples/md-flexible/src/Simulation.cpp
@@ -375,13 +375,13 @@ bool Simulation::needsMoreIterations() const {
 
 std::string Simulation::getMPISuffix() const {
   std::string suffix;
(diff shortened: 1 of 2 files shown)


FG-TUM (approved):

SteffenSeckler:
Formatting is needed, see:
https://github.com/AutoPas/AutoPas/blob/master/.github/CONTRIBUTING.md#code-style
requires: clang-format-9

PR #588, opened 2021-03-24
Homogeneity not inf

JockelTUM:
# Description

To ensure that the function "calculateHomogeneity" does not return "inf" in certain but otherwise normal cases, it was necessary to ensure that for the calculation of "densityPerCell" no value is divided by zero.


SteffenSeckler (approved):
all good, please format:
https://github.com/AutoPas/AutoPas/blob/master/.github/CONTRIBUTING.md#code-style
or simply:
```
make formatall -j
```

PR #593, opened 2021-04-16
Fix prediction underflow

### LINKED ISSUE #578: Predictive Tuning suffers from underflow
**Describe the bug**
Predictive tuning uses unsigned longs for its predictions. When the gradient is falling very strong this can lead to negative predictions (which is a problem in itself) that result in an underflow. Relevant line:
https://github.com/AutoPas/AutoPas/blob/40dad5f0f5cffedeaeae025f7edc2781f3255641/src/autopas/selectors/tuningStrategy/PredictiveTuning.h#L370-L372
Here `yIntercept` is the prediction as a double. The actual prediction is saved in `_configurationPredictions[configuration]` as a `size_t`.
This example shows the linear regression prediction but this is true for all prediction functions.

**Expected behavior**
???

**Proposed Solution**
- Store predictions signed (we can not really find a sensible value)
- Disregard negative prediction values when choosing the configuration that we set as reference (as they are fishy)
- Always test configurations with negative predictions since we need more information about them but they seem to improve rapidly.

**Additional context**
Observed with the exploding liquid experiment.

FG-TUM:
# Description

Predictive tuning can run in underflows resulting in super high predictions for configurations that should be fast. This PR aims at finding and eliminating them all.

- [x] prevent underflows in predictions
- [x] prevent overflows in predictions
- [x] remove lagrange interpolation due to numerical instability
- [x] add arithmetic functions which prevent over / underflow
- [x] add tests
- [x] add `-Wextra` for clang

`-Wextra` for clang adds warnings for incorrect signed comparisons.

## Resolved Issues

- [x] fixes #578 

# How Has This Been Tested?

- [x] Refactored tests
- [x] New tests


src/autopas/utils/Math.h
```diff
@@ -16,6 +17,165 @@ namespace autopas::utils::Math {
...
+ * @note Underflow here refers to a value more negative than representable and not the underflow gap around zero.
+ *
+ * @tparam T
+ * @param a
+ * @param b
+ * @param valUnderflow Return value in case of underflow.
+ * @param valOverflow Returns value in case of overflow.
+ * @return Sum or valUnderflow or valOverflow.
+ */
+template <class T, std::enable_if_t<std::is_floating_point_v<T>, bool> = true>
```
SteffenSeckler, line 59: would be so much nicer with c++20 Concepts/Constraints :P
```suggestion
template <std::floating_point T>
```
it only requires gcc10 or clang10 and libstdc++10 or libc++13 xD

maybe add a todo?

src/autopas/utils/Math.h
```diff
@@ -16,6 +17,165 @@ namespace autopas::utils::Math {
...
+ * @param valOverflow Return value in case of overflow.
+ * @return Product or valUnderflow or valOverflow.
+ */
+template <class T, std::enable_if_t<std::is_integral_v<T>, bool> = true>
+T safeMul(const T &a, const T &b, const T &valUnderflow = std::numeric_limits<T>::min(),
+          const T &valOverflow = std::numeric_limits<T>::max()) {
+  T result;
+  bool overflow = __builtin_mul_overflow(a, b, &result);
+  if (overflow) {
+    // if both args are negative this is an underflow.
```
SteffenSeckler, line 142:
```suggestion
    // if exactly one arg is negative this is an underflow.
```

src/autopas/utils/Math.h
```diff
@@ -16,6 +17,165 @@ namespace autopas::utils::Math {
...
+  if (std::isinf(result)) {
+    if (result > 0) {
+      result = valOverflow;
+    } else {
+      result = valUnderflow;
+    }
+  }
+  return result;
+}
+
```
SteffenSeckler, line 178: in theory, safeDevision is missing for floating-point types ;)
e.g. 10 / 0.00000000000001 or so ;)

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -312,25 +328,33 @@ void PredictiveTuning::linePrediction() {
       const auto delta = _firstIterationOfTuningPhase - _traversalTimesStorage[configuration].back().first;
 
       // gradient * delta + last point
-      _configurationPredictions[configuration] = functionParams[0] * delta + functionParams[1];
+      _configurationPredictions[configuration] =
+          static_cast<long>(functionParams[0] * static_cast<double>(delta) + functionParams[1]);
```
SteffenSeckler, line 332: safeMath needed here?

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -344,77 +368,64 @@ void PredictiveTuning::linearRegression() {
...
-        iterationMultTime += iteration * time;
+        long iterationMultTimeI = utils::Math::safeMul(static_cast<long>(iteration), time);
+        iterationMultTime = utils::Math::safeAdd(iterationMultTime, iterationMultTimeI);
+        // if any of the safe operations overflow we can directly move to the next config
+        if (iterationMultTime == std::numeric_limits<decltype(iterationMultTime)>::max()) {
+          numericOverflow = true;
+          break;
+        }
         iterationSum += iteration;
+        // this will overflow at iteration 3 037 000 499
```
SteffenSeckler, line 390:
```suggestion

```

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -344,77 +368,64 @@ void PredictiveTuning::linearRegression() {
...
+      const double denominator =
+          static_cast<double>(iterationSquareSum) - static_cast<double>(iterationSum) * iterationMeanValue;
 
       // ((Sum iteration_i * time_i) - n * iterationMeanValue * timeMeanValue) / ((Sum iteration_i^2) - n *
       // iterationMeanValue ^ 2)
-      const auto gradient = numerator / denominator;
-      const auto yIntercept = timeMeanValue - gradient * iterationMeanValue;
+      const auto gradient = static_cast<double>(numerator) / denominator;
+
+      const auto change = static_cast<long>(gradient * (_firstIterationOfTuningPhase - iterationMeanValue));
```
SteffenSeckler, line 414: safeMath?

SteffenSeckler (approved):
only a few minor things.
mainly:
- c++20 concepts would be nice, but probably not reasonable, yet...
- sometimes it feels like things could (theoretically) overflow, but safeMath isn't used.

SteffenSeckler:
fyi:  benchmarks of the __builtin_*_overflow functions:
https://quick-bench.com/q/lwFayIAHUCruAhNlV-_MKY7acfM

They aren't that bad :)

FG-TUM:
> fyi: benchmarks of the __builtin_*_overflow functions:
> https://quick-bench.com/q/lwFayIAHUCruAhNlV-_MKY7acfM
> 
> They aren't that bad :)

Benchmark might be misleading. You need to test the influence actual overflow has on the operation. Then you should modify the benchmark to have a known percentage of overflow/underflow operations.

src/autopas/utils/Math.h
FG-TUM: true. I didn't care since I didn't need it. If someone needs it they can implement it ;)

PR #596, opened 2021-04-28
Doxygen Github Actions Rework

FG-TUM:
# Description

- [x] added Doxygen version recommendation in cmake
- [x] Reworked github actions to use our docker container
- [x] removed building doc from jenkins

## ~Resolved Issues~

# How Has This Been Tested?

Locally. I also tested Versions `1.9.0`, `1.9.1` but they give me warnings from our `Math.h`.

### DIFF
--- a/.github/workflows/doxygen-md-flexible.yml
+++ b/.github/workflows/doxygen-md-flexible.yml
@@ -1,45 +1,43 @@
-# This is a basic workflow to help you get started with Actions
-
+# Workflow for generating, checking, and deploying our doxygen documentation.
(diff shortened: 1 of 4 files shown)


FG-TUM:
Our github actions use a different docker container for building Doxygen which has an older version. We should switch this to our container.

SteffenSeckler (approved):
fyi:
I remembered the reason, why we left the doxygen inside of Jenkins.
We left it, to short-circuit a build if doxygen is broken, which reduces stress on the test nodes.

Up to you, whether to reenable it, or not ;)

PR #597, opened 2021-04-29
Md flex/cart decomposition

J0llyver:
# Description
## General description of the changes
This pull request contains the implementation of the cartesian grid domain decomposition and all the refactoring that went along with it.

The previously existing simulation is still present in the MDFlexSingleRank class. Additionally, a new class MDFlexMPI has been created, uses a regular grid domain decomposition to distribute the workload to multiple processors.

The domain decomposition and all the MPI communication is handled by the RegularGrid class.
The class MDFlexSingleRank uses a dummy domain decomposition class called SingleDomain.

The initialization of the particles has been refactored to be able to distribute the particles to the different MPI processes. For this purpose, the 'generate' functions of all the Object classes has been updated, so it does not require the an AutoPas container, which is not available at this point. The logic for the 'generate' functions has been derived from the respective generators in the autopasTools folder.

Additionally a lot of the initialization, which was floating around in the main function has been moved to appropriate constructors.

The moudlarity of the code has been improved, in general.

## Communication pattern
The communication of halo particles and particles which migrate between domains is handled for each dimension separately. Therefore, the communication is not bound to three dimensions.

The two-step scheme is used to send particles to every neighbour (including diagonal neighbours). 

To send data to neighbours, the non-blocking MPI\_Isend function is used. To receive data, the blocking MPI\_Recv function is used.

# How Has This Been Tested?
The single domain simulation has been tested by running it with the fallingDrop.yaml configuration with 90000 iterations.

The MPI simulation has been tested by running it with the fallingDrop.yaml configuration with 5000 iterations. Because the paralellized simulation currently does not create vtk-files, the validity of the simulation has not been tested.

No tests have been used to verify the implementation, because the test build is broken due to a very obscure linking issue. New tests for several classes still need to be implemented.


FG-TUM:
looks like you merged master in this branch (a6f1b73) but not the target. Please merge master in the target branch too so the diff becomes smaller and easier to review ;)

examples/md-flexible/src/ParticleSerializationTools.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+  std::vector<char> attributesVector;
+  attributesVector.resize(sizeof(ParticleAttributes));
+
+  std::memcpy(&attributesVector[0], &attributes, sizeof(ParticleAttributes));
+
+  ParticleType deserializedParticle;
+  deserializeParticleData(attributesVector.data(), deserializedParticle);
+
+  serializedParticles.insert(serializedParticles.end(), attributesVector.begin(), attributesVector.end());
```
FG-TUM, line 36: couldn't you make this simpler and at the same time eliminate the ParticleAttributes class? I'm thinking about something like this:

```cpp
template<class T>
void emplaceInBuffer(std::vector<unsigned char> buffer, T value) {
  const bufferSizeBefore = buffer.size();
  const sizeOfValue = sizeof(value);
  buffer.resize(bufferSizeBefore + sizeOfValue;
  std::memcpy(&buffer[bufferSizeBefore], &value, sizeOfValue);
}

// and then use like:
emplaceInBuffer(buff, particle.getR());
emplaceInBuffer(buff, particle.getV());
emplaceInBuffer(buff, particle.getID());

```
Mind typos etc, I'm tired ;)

FG-TUM (changes requested):
Review part 1

Some general points:
- Needs more documentation (<- I guess this is always true lol)
- Use more clang-tidy. If I open this code in Clion it lights up like a Christmas tree.
- Any chance you can rearrange the methods in a similar way as they were before? Would increase review-ability.
- Check for dead / commented out code and remove it
- Check if the new todos make sense to be resolved immediately. If not add your name to them and if you do not start working on them in the near future create an issue for them.

examples/md-flexible/src/Simulation.cpp
```diff
@@ -455,140 +217,152 @@ void Simulation::writeVTKFile(autopas::AutoPas<ParticleType> &autopas) {
...
-    // This is just a randomly guessed number but seems to fit roughly for default settings.
-    size_t configsTestedPerTuningPhase = 90;
-    if (_config->tuningStrategyOption.value == autopas::TuningStrategyOption::bayesianSearch or
-        _config->tuningStrategyOption.value == autopas::TuningStrategyOption::bayesianClusterSearch) {
-      configsTestedPerTuningPhase = _config->tuningMaxEvidence.value;
+  for (auto &particle : _configuration.getParticles()) {
+    ParticleType autoPasParticle;
+    if (_domainDecomposition.isInsideLocalDomain(particle.position)) {
+      autoPasParticle = ParticleSerializationTools::convertParticleAttributesToParticle(particle);
+      _autoPasContainer->addParticle(autoPasParticle);
```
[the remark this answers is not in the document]
J0llyver: I will mark this as a todo for future-me

examples/md-flexible/src/ParallelVtkWriter.cpp
```diff
@@ -0,0 +1,127 @@
...
+  if (_mpiRank == 0) {
+    tryCreateSessionFolder(_sessionName, outputFolder);
+  }
+
+  int sessionFolderPathLength = _sessionFolderPath.size();
+  AutoPas_MPI_Bcast(&sessionFolderPathLength, 1, AUTOPAS_MPI_INT, 0, AUTOPAS_MPI_COMM_WORLD);
+
+  if (_mpiRank != 0) {
+    _sessionFolderPath.resize(sessionFolderPathLength);
+  }
```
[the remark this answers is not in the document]
J0llyver: MPI_Bcast writes into sessionDolderPath. If we don't make sure, there is enough space reserved for the string, we get an memory access violation.

What do you mean by "for shortening", here?

examples/md-flexible/src/ParticleSerializationTools.cpp
J0llyver: Does this really make it simpler? And I would also have to implement some function for deserialization, right? And without the ParticleAttributes the deserialization would get a lot messier. 

And always resizing the buffer for every single attribute seems like a lot of additional work.

FG-TUM: > Does this really make it simpler?
Debatable. I think it is more direct since you do not have the additional indirection with the `ParticleAttributes` object.

> And I would also have to implement some function for deserialization, right? 
Sure but this is analogous.

> And without the ParticleAttributes the deserialization would get a lot messier.
Hmm this way you would have to make sure to push / pop values in the correct order.

> And always resizing the buffer for every single attribute seems like a lot of additional work.
If you do one big `buffer.reserve()` beforehand the `resize()` calls are cheap.

I see merit in both ways. I do not like that we have to co-maintain the `ParticleAttributes`, however, paying attention to the order of pushing / popping values to the buffer is also weird. Probably a more elegant solution would be the push / pop approach but using the `MoleculeLJ::AttributeNames` and respective `get`. You would have to only define one tuple with all `AttributeNames` that should be sent and traverse it forward for serialization and backward for deserialization. There already is a function for iterating over a tuple in `TupleUtils.h`.

examples/md-flexible/src/ParticleAttributes.h
```diff
@@ -0,0 +1,51 @@
...
+
+#include <array>
+
+#include "autopas/particles/OwnershipState.h"
+
+/*
+ * A struct containing all properties of autopas::MoleculeLJ<double>.
+ * This can be used to align the attributes of a particle in memory to make serialization and deserialization easier.
+ */
+struct ParticleAttributes {
```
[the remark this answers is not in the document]
J0llyver: During initialization of MDFlexConfig and when sending / receiving particles using MPI, we only require the actual particle data. 
In MDFlexConfig, we avoid any additional overhead that might come with initializing a MoleculeLJ particle.
For MPI communication the ParticleAttributes serve as an "interface" to the particle data, which make de-/serialization easier.

examples/md-flexible/src/configuration/objects/Object.h
```diff
@@ -39,24 +40,28 @@ class Object {
...
-   *
-   * @param autopas
+   * @note Particle IDs for the new particles will start at particles.size().
+   * @param particles The container to which the new particles will be appended to.
    */
-  virtual void generate(autopas::AutoPas<ParticleType> &autopas) const = 0;
+  virtual void generate(std::vector<ParticleAttributes> &particles) const = 0;
 
   /**
    * Create a particle that acts as blueprint for all particles to be created for the object.
```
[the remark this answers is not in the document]
J0llyver: Particles are only generated in derivations of Object.h. Each derivation has to overwrite the generate function.

examples/md-flexible/src/configuration/MDFlexConfig.h
```diff
@@ -553,6 +561,35 @@ class MDFlexConfig {
...
    */
   static constexpr size_t valueOffset{33};
+
+ private:
+  /**
+   * Stores the particles generated based on the provided configuration file
+   * These particles can be added to the respective autopas container,
+   * but have to be converted to the respective particle type, first.
+   */
+  std::vector<ParticleAttributes> _particles;
```
FG-TUM, line 571: two things here:
1. After the AutoPas object is initialized you can probably flush this vector
2. In the long run it does not seem reasonable to store the particles in the config but rather directly create them in the respective AutoPas objects. If you want to do it as it is implemented now please create an issue and reference it with a TODO here  so that we do not forget about it.

examples/md-flexible/src/domainDecomposition/DomainTools.cpp
```diff
@@ -0,0 +1,111 @@
...
+                    std::array<double, 3> &boxMax) {
+  bool isInsideLocalDomain = true;
+  for (int i = 0; i < coordinates.size(); ++i) {
+    if (!isInsideLocalDomain) {
+      break;
+    }
+    isInsideLocalDomain = coordinates[i] >= boxMin[i] && coordinates[i] < boxMax[i];
+  }
+  return isInsideLocalDomain;
+}
```
FG-TUM, line 33:
```suggestion
template<class T>
bool isInsideDomain(const T &coordinates, std::array<double, 3> &boxMin,
                    std::array<double, 3> &boxMax) {
  bool isInsideLocalDomain = true;
  for (int i = 0; i < coordinates.size(); ++i) {
    if (!isInsideLocalDomain) {
      break;
    }
    isInsideLocalDomain = coordinates[i] >= boxMin[i] and coordinates[i] < boxMax[i];
  }
  return isInsideLocalDomain;
}
```

Similar with `getDistanceToDomain()` to avoid code duplication. However, this then has to go to the header.

FG-TUM: On second thought: Why do you need the vector variant anyways?

examples/md-flexible/src/domainDecomposition/DomainTools.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+double getDistanceToDomain(const std::array<double, 3> &coordinates, std::array<double, 3> &boxMin,
+                           std::array<double, 3> &boxMax) {
+  if (coordinates.size() == boxMin.size() && coordinates.size() == boxMax.size()) {
+    std::vector<double> differences(coordinates.size(), 0.0);
+    for (int i = 0; i < coordinates.size(); ++i) {
+      if (coordinates[i] < boxMin[i]) {
+        differences[i] = boxMin[i] - coordinates[i];
+      } else if (coordinates[i] > boxMax[i]) {
+        differences[i] = coordinates[i] - boxMax[i];
```
FG-TUM, line 65: `std::abs()`?

examples/md-flexible/src/domainDecomposition/DomainTools.cpp
```diff
@@ -0,0 +1,111 @@
...
+      }
+    }
+
+    double distance = 0.0;
+    for (const auto &difference : differences) {
+      distance += std::pow(difference, 2.0);
```
FG-TUM, line 71:
```suggestion
      distance += difference * difference;
```

FG-TUM: most of this could be rewritten with the math functions from `autopas::ArrayMath()`

examples/md-flexible/src/domainDecomposition/DomainTools.cpp
```diff
@@ -0,0 +1,111 @@
...
+    primeFactors.sort();
+    auto firstElement = primeFactors.front();
+    primeFactors.pop_front();
+    primeFactors.front() *= firstElement;
+  }
+
+  oDecomposition.resize(dimensionCount);
+
+  for (auto &dimensionSize : oDecomposition) {
+    if (primeFactors.size() > 0) {
```
FG-TUM, line 103:
```suggestion
    if (not primeFactors.empty()) {
```

examples/md-flexible/src/domainDecomposition/DomainTools.cpp
```diff
@@ -0,0 +1,111 @@
...
+    for (const auto &difference : differences) {
+      distance += std::pow(difference, 2.0);
+    }
+
+    return std::pow(distance, 1.0 / differences.size());
+  }
+  return -1;
+}
+
+void generateDecomposition(unsigned int subdomainCount, int dimensionCount, std::vector<int> &oDecomposition) {
```
FG-TUM, line 79: is dimensionCount necessary? Is this ever not 3?

examples/md-flexible/src/domainDecomposition/DomainTools.h
```diff
@@ -0,0 +1,65 @@
...
+ */
+double getDistanceToDomain(const std::array<double, 3> &coordinates, std::array<double, 3> &boxMin,
+                           std::array<double, 3> &boxMax);
+
+/**
+ * Generates a decomposition with a secific number of subdomains.
+ * This function uses prime factorization to determine the number of subdomains in each dimension.
+ * @param subdomainCount The number of subdomains in the resulting decomposition
+ * @param dimensionCount The number of dimensions in the simulation.
+ * @param oDecomposition The resulting decomposition.
```
FG-TUM, line 62:
```suggestion
 * @param oDecomposition Vector containing the number of subdomains per dimension.
```
is this interpretation correct?
Should probably be an array?
And why "o" ?

FG-TUM (changes requested):
Review 2

What is the reason behind eliminating the parsing folder?

No only the review of the tests should be missing

examples/md-flexible/CMakeLists.txt
```diff
@@ -4,29 +4,31 @@ file(
...
     message(STATUS "AUTOPAS_INTERNODE_TUNING set, searching for mpi to link against md-flexible ...")
     find_package(MPI)
 
     if (NOT ${MPI_CXX_FOUND})
         message(FATAL_ERROR "cxx mpi could not be found, even though AUTOPAS_INTERNODE_TUNING was set.")
     else ()
         message(STATUS "cxx mpi found: ${MPI_CXX_COMPILER}")
+        set(MD_FLEXIBLE_USE_MPI ON)
+        add_definitions(-DAUTOPAS_INTERNODE_TUNING)
+        add_definitions(-DAUTOPAS_INCLUDE_MPI)
```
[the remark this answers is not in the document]
J0llyver: AUTOPAS_INTERNODE_TUNING is still being used in WrapMPI.h
The goal is to replace it at most locations and only to keep it where actual internode communication is required.
The definition AUTOPAS_INCLUDE_MPI aims  to replace AUTOPAS_INTERNODE_TUNING at every other place.
Because I did not yet get to properly replace AUTOPAS_INTERNODE_TUNING we need both definitions here.

examples/md-flexible/src/Checkpoint.h
```diff
@@ -1,91 +0,0 @@
-/**
- * @file Checkpoint.h
```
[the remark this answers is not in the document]
J0llyver: I think I moved the checkpoint code somewhere else at some point. But because I cannot find it anymore, I will reinclude the original file.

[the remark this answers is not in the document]
J0llyver: Ah nevermind, the code moved to MDFlexConfig as it is only required during the initialization of the particles. (See MDFlexConfig::loadParticlesFromCheckpoint. If a proper checkpoint file (old vtk format) exists, it still can be loaded, even for multiple ranks. Therefore it is unnecessary to throw an exception. Each rank then will only include the particles which are located in it's respective domain.

examples/md-flexible/src/domainDecomposition/DomainTools.cpp
J0llyver: This logic cannot be represented by std::abs. In the first case we use boxMin[i] and in the second case boxMax[i]. So even if we would use std::abs we still would neet the two branches.

examples/md-flexible/src/domainDecomposition/DomainTools.cpp
J0llyver: I will use autopas::utils::ArrayMath for the std::array implementation of getDistanceToDomain.

examples/md-flexible/src/domainDecomposition/DomainTools.cpp
J0llyver: I initially created this with an arbitrary number of domains in mind. I am very hesitant to limit it to 3 domains, but cannot really argue against it. Although consequently some refactoring might be necessary.

examples/md-flexible/src/domainDecomposition/DomainTools.h
J0llyver: "o" is a convention to mark "output" parameters, i. e. funciton arguments which will be changed by a function. We use this convention where I work. I will remove it.
The interpretation is rather vague, I agree. This vector only contains the number of subdomains along each dimension.
I chose the type vector with an arbitrary amount of dimensions in mind.

examples/md-flexible/src/domainDecomposition/DomainTools.cpp
HobbyProgrammer: `std::clamp()` might also be useful here. Something like: `differences[i] = std::abs(coordinates[i] - std::clamp(coordinates[i], boxMin[i], boxMax[i]));`.  `std::abs()` is theoretically not even necessary here, since the value is squared later anyways.

examples/md-flexible/src/ParallelVtkWriter.cpp
```diff
@@ -0,0 +1,202 @@
...
+  timestepFile << "      <PointData>" << std::endl;
+
+  // print velocities
+  timestepFile << "        <DataArray Name=\"velocities\" NumberOfComponents=\"3\" format=\"ascii\" type=\"Float32\">"
+               << std::endl;
+  for (auto particle = autoPasContainer.begin(autopas::IteratorBehavior::owned); particle.isValid(); ++particle) {
+    auto v = particle->getV();
+    timestepFile << "        " << v[0] << " " << v[1] << " " << v[2] << std::endl;
+  }
+  timestepFile << "        </DataArray>" << std::endl;
```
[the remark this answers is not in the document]
J0llyver: We could do this.  But I would rather not loop over each particle for every property but use multiple stringstreams (one for each attribute) which then will be included in the output stream after looping over the particles once.
But this is something I would mark as a todo as the creation of the timestep files currently serves our purpose.

examples/md-flexible/src/Objects/CubeClosestPacked.h
```diff
@@ -1,72 +0,0 @@
-/**
- * @file CubeClosestPacked.h
```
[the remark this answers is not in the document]
J0llyver: I will try to remember that. This would reduce the number of changed lines in the pull request, right?

examples/md-flexible/src/configuration/objects/Sphere.h
```diff
@@ -87,53 +91,74 @@ class Sphere : public Object {
...
+   */
+  void generate(std::vector<ParticleAttributes> &particles) const override {
+    ParticleAttributes particle = getDummyParticle(particles.size());
     iteratePositions([&](auto pos) {
-      dummyParticle.setR(pos);
-      autopas.addParticle(dummyParticle);
-      dummyParticle.setID(dummyParticle.getID() + 1);
+      particle.position = {pos[0], pos[1], pos[2]};
+      particles.push_back(particle);
+      particle.id++;
```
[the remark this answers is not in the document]
J0llyver: The particle is getting initialized with particles.size(). This is already the correct id for the first new particle. There is a test to check exactly this problem and they run successful :)
I won't remove ParticleAttributes for now. See the discussion on the particle serialization for the reasons.

examples/md-flexible/src/BoundaryConditions.h
```diff
@@ -158,13 +158,16 @@ template <class Particle>
...
-  }
+
+  // auto [leavingParticles, updated] = autoPas.updateContainer(forceUpdate);
+
+  // if (updated) {
+  // 2. apply periodic wrap by shifting positions of leaving particles to positions of periodic images.
+  // wrapPositionsAroundBoundaries(autoPas, leavingParticles);
+  // 2b. re-insert shifted particles
+  // addEnteringParticles(autoPas, leavingParticles);
+  //}
```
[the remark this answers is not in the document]
J0llyver: I don't think the BoundaryCondition library is compatible with the 2-step scheme. So I would remove it.

examples/md-flexible/src/ParticleSerializationTools.cpp
J0llyver: I don't think the solution using AttributeNames is possible.
AttributeNames contains as first attribute the pointer to the object, which we don't want to send.
because the ParticleBase::get<>() function uses a template to identify the attribute, we need to define the getNeededAttribute as constexpr.
Therefore it is not possible to skip the pointer element in the Attribute names.

Additionally, because the current solution is working and because my thesis does not focus on finding the optimal way to serialize data for MPI I would mark this as a todo.
I dont' think it is reasonable to invest more time, here, even if ParticleAttributes needs to be maintained in the future.

examples/md-flexible/src/Thermostat.h
```diff
@@ -118,8 +118,10 @@ auto calcTemperatureComponent(const AutoPasTemplate &autopas,
   // AutoPas works always on 3 dimensions
   constexpr unsigned int dimensions{3};
 
-  for (auto [kinEIter, numParIter] = std::tuple{kineticEnergyMul2Map.begin(), numParticleMap.begin()};
-       kinEIter != kineticEnergyMul2Map.end(); ++kinEIter, ++numParIter) {
+  auto kineticEnergyAndParticleMaps = std::make_tuple(kineticEnergyMul2Map.begin(), numParticleMap.begin());
+
+  for (auto [kinEIter, numParIter] = kineticEnergyAndParticleMaps; kinEIter != kineticEnergyMul2Map.end();
+       ++kinEIter, ++numParIter) {
```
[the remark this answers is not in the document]
J0llyver: This is just for readability. Initializing the tuple within the for loop definition obscures the logic of the loop.

examples/md-flexible/src/Thermostat.h
```diff
@@ -118,8 +118,10 @@ auto calcTemperatureComponent(const AutoPasTemplate &autopas,
...
     kinEIter->second /= numParIter->second * dimensions;
   }
   return kineticEnergyMul2Map;
```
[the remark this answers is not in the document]
J0llyver: There is already a todo in the documentation for the function Simulation::updateThermostat for this issue. Though it might be reasonlable to move the todo.

examples/md-flexible/src/configuration/CLIParser.cpp
```diff
@@ -41,7 +41,7 @@ MDFlexParser::exitCodes MDFlexParser::CLIParser::parseInput(int argc, char **arg
       config.verletRebuildFrequency, config.vtkFileName, config.vtkWriteFrequency, config.selectorStrategy,
       config.yamlFilename, config.distributionStdDev, config.globalForce, zshCompletionsOption, helpOption)};
 
-  constexpr auto relevantOptionsSize = std::tuple_size_v<decltype(relevantOptions)>;
+  constexpr auto relevantOptionsSize = std::tuple_size<decltype(relevantOptions)>{};
```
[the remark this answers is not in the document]
J0llyver: clang-tidy did complain that std::tuple_size_v does not exist in the respective header.

examples/md-flexible/src/configuration/CLIParser.h
```diff
@@ -18,7 +18,8 @@
 /**
  * Parser for input from the command line.
  */
-namespace MDFlexParser::CLIParser {
+namespace MDFlexParser {
+namespace CLIParser {
```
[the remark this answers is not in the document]
J0llyver: This was also a suggestion by clang-tidy. I think I even used the configuration you sent me when running clang-tidy.

examples/md-flexible/src/configuration/MDFlexConfig.cpp
```diff
@@ -3,10 +3,72 @@
...
+ * @param file
+ * @param word
+ */
+void findWord(std::ifstream &file, const std::string &word) {
+  std::string currentWord;
+  while (currentWord != word) {
+    file >> currentWord;
+  }
+}
+}  // namespace
```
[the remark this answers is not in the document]
J0llyver: MDFlexConfig currently handles particle innitialization. This is part of the particle initialization so I moved it into the same context as the particle initialization using objects to have it in the same place.

examples/md-flexible/CMakeLists.txt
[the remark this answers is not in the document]
FG-TUM: Do you plan to do this in this PR? Otherwise add a TODO and maybe an issue

examples/md-flexible/src/Objects/CubeClosestPacked.h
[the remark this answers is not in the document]
FG-TUM: Probably. However they would show up as moved and not deleted + added which is more understandable during review.

examples/md-flexible/src/configuration/MDFlexConfig.h
```diff
@@ -553,6 +561,35 @@ class MDFlexConfig {
...
+  std::shared_ptr<ParticlePropertiesLibraryType> _particlePropertiesLibrary;
+
+  /**
+   * Initializes the ParticlePropertiesLibrary
+   */
+  void initializeParticlePropertiesLibrary();
+
+  /**
+   * Iniitalizes all particles present at the start of the simulation.
+   * The particles are contained in a normalized domain where every coordinate is between 0 and 1.
```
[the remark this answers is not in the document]
J0llyver: This line is wrong. I had a version, where I did this but decided against normalizing the particle positions. I will remove this line.

examples/md-flexible/src/domainDecomposition/DomainTools.cpp
J0llyver: Again this is because of me having multiple dimensions in mind. Therefore I am using vectors where arrays might be sufficient. I guess I will limit the whole logic to 3 dimensions to avoid such helper functions and other problems.

examples/md-flexible/src/domainDecomposition/DomainTools.cpp
FG-TUM: Totally missed that min max was mixed here, my bad. `std::clamp` looks better.

examples/md-flexible/src/TimeDiscretization.h
```diff
@@ -1,69 +1,52 @@
...
+void calculatePairwiseForces(autopas::AutoPas<ParticleType> &autoPasContainer,
+                             ParticlePropertiesLibraryType &particlePropertiesLibrary, const double &deltaT,
+                             MDFlexConfig::FunctorOption functorOption, bool &wasTuningIteration);
 
+/**
+ * Adds global forces to the particles in the container.
+ * @param autoPasContainer The container for which to update the particle forces.
+ * @param globalForce The global force which will be applied to each particle in the container.
+ */
+void calculateGlobalForces(autopas::AutoPas<ParticleType> &autoPasContainer, std::array<double, 3> &globalForce);
```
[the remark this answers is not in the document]
J0llyver: Why have the update of velocities and positions here, but not the updates for forces? 
I found the name "TimeDiscretization" quite vague and interpreted it as in "Functions particle updates which occure each timestep". Following this interpretation the functions are at the right place.

examples/md-flexible/src/configuration/MDFlexConfig.h
```diff
@@ -553,6 +561,35 @@ class MDFlexConfig {
...
+   * Stores the particles generated based on the provided configuration file
+   * These particles can be added to the respective autopas container,
+   * but have to be converted to the respective particle type, first.
+   */
+  std::vector<ParticleAttributes> _particles;
+
+  /**
+   * Stores the physical properties of the particles used in the an MDFlexSimulation
+   */
+  std::shared_ptr<ParticlePropertiesLibraryType> _particlePropertiesLibrary;
```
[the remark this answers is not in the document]
J0llyver: It does not work, because std::unique_ptr deletes the copy constructor which is why we cannot create a getter for the _particlePropertiesLibrary, which we currently need.

examples/md-flexible/src/configuration/objects/Object.h
```diff
@@ -39,24 +40,28 @@ class Object {
...
    * @return
    */
-  [[nodiscard]] ParticleType getDummyParticle(const autopas::AutoPas<ParticleType> &autopas) const {
-    ParticleType dummyParticle;
-    dummyParticle.setV(_velocity);
-    dummyParticle.setID(autopas.getNumberOfParticles());
-    dummyParticle.setTypeId(_typeId);
-    return dummyParticle;
+  [[nodiscard]] ParticleAttributes getDummyParticle(const int &particleId) const {
+    ParticleAttributes particle;
```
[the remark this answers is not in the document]
J0llyver: During initialization it might as well be considered a particle. I feel like renaming it would cause more confusion as anything else.

examples/md-flexible/src/configuration/MDFlexConfig.h
J0llyver: I will flush the vector.
I am currently not sure if the way I initialize the particles will work for octrees. So the particle initialization might change soon, anyway

examples/md-flexible/CMakeLists.txt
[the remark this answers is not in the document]
J0llyver: I will do it in this pull request.

examples/md-flexible/src/TimeDiscretization.cpp
```diff
@@ -0,0 +1,98 @@
+/**
+ * @file TimeDiscretization.cpp
+ * @author N. Fottner
+ * @date 13/05/19
```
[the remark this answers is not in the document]
J0llyver: This should be true. I did not create this file. I don't even remember moving it, not sure why git marked it as new.

examples/md-flexible/src/configuration/objects/Object.h
```diff
@@ -39,24 +39,27 @@ class Object {
...
+  [[nodiscard]] ParticleType getDummyParticle(const size_t &particleId) const {
+    ParticleType particle{};
+    particle.setID(particleId);
+    particle.setTypeId(_typeId);
+    particle.setOwnershipState(autopas::OwnershipState::owned);
+    particle.setV(_velocity);
+    particle.setF({0.0, 0.0, 0.0});
+    particle.setOldF({0.0, 0.0, 0.0});
+
+    return particle;
```
FG-TUM, line 60:
```suggestion
  [[nodiscard]] ParticleType getDummyParticle(const size_t &particleId) const {
    ParticleType particle{};
    particle.setID(particleId);
    particle.setTypeId(_typeId);
    particle.setV(_velocity);
    return particle;
```

Why repeat what the constructor does anyway?

examples/md-flexible/src/configuration/MDFlexConfig.h
```diff
@@ -23,18 +23,24 @@
...
-#include "src/Objects/CubeClosestPacked.h"
-#include "src/Objects/CubeGauss.h"
-#include "src/Objects/CubeGrid.h"
-#include "src/Objects/CubeUniform.h"
-#include "src/Objects/Sphere.h"
+#include "objects/CubeClosestPacked.h"
+#include "objects/CubeGauss.h"
+#include "objects/CubeGrid.h"
+#include "objects/CubeUniform.h"
+#include "objects/Sphere.h"
```
FG-TUM, line 30:
```suggestion
#include "src/configuration/objects/CubeClosestPacked.h"
#include "src/configuration/objects/CubeGauss.h"
#include "src/configuration/objects/CubeGrid.h"
#include "src/configuration/objects/CubeUniform.h"
#include "src/configuration/objects/Sphere.h"
```

full include paths are a bit clearer

FG-TUM (changes requested):
Please also consider the suggestions from deepcode

examples/md-flexible/CMakeLists.txt
```diff
@@ -4,29 +4,30 @@ file(
...
 )
+add_executable(md-flexible ${MDFlex_SRC})
+
+target_include_directories(md-flexible PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
 
-if (AUTOPAS_INTERNODE_TUNING)
-    message(STATUS "AUTOPAS_INTERNODE_TUNING set, searching for mpi to link against md-flexible ...")
+include(autopas_yaml-cpp)
+
+option(AUTOPAS_INCLUDE_MPI "Enable this variable to use MPI." OFF)
```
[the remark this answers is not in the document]
J0llyver: In this case we would need to activate MD_FLEXIBLE_USE_MPI and AUTOPAS_INCLUDE_MPI because the latter is used by the WrapMPI header. This header is not part of MD-Flexible's context which is why I don't want to use MD_FLEXIBLE_USE_MPI in this header. 
And requiring two flags to enable MPI will definitely cause confusion.

examples/md-flexible/src/ParallelVtkWriter.cpp
```diff
@@ -0,0 +1,211 @@
...
+  if (not timestepFile.is_open()) {
+    throw std::runtime_error("Simulation::writeVTKFile(): Failed to open file \"" + filename.str() + "\"");
+  }
+
+  timestepFile << "<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"no\" ?>\n";
+  timestepFile << "<VTKFile byte_order=\"LittleEndian\" type=\"PUnstructuredGrid\" version=\"0.1\">\n";
+  timestepFile << "  <PUnstructuredGrid GhostLevel=\"0\">\n";
+  timestepFile << "    <PPointData>\n";
+  timestepFile << "      <PDataArray Name=\"velocities\" NumberOfComponents=\"3\" format=\"ascii\" type=\"Float32\"/>"
+               << std::endl;
```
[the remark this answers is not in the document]
J0llyver: I thought I did use a search & replace pattern which catches every occurrence. Sorry for having to point at this multiple times.

examples/md-flexible/src/domainDecomposition/DomainTools.cpp
```diff
@@ -0,0 +1,70 @@
...
+      differences[i] = std::clamp(coordinates[i], boxMin[i], boxMax[i]);
+    }
+
+    return autopas::utils::ArrayMath::L2Norm(differences);
+  }
+  return -1;
+}
+
+void generateDecomposition(unsigned int subdomainCount, std::array<int, 3> &decomposition) {
+  std::list<int> primeFactors;
```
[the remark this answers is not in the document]
J0llyver: I don't see an advantage with using std::deque. We don't really need an indexed sequence or insertion at both ends.

examples/md-flexible/CMakeLists.txt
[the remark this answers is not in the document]
FG-TUM: > And requiring two flags to enable MPI will definitely cause confusion.

Agreed. So we just leave it as is. However, in that case, maybe add something to the md-flex readme that explains this.

examples/md-flexible/src/domainDecomposition/DomainTools.cpp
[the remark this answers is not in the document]
FG-TUM: Ok but add the doc anyways ;)

FG-TUM:
Jenkins can't build this as gcc7 does not come with the `filesystem` header (and the clang container probably uses the same stdlib?). There are two ways to resolve this:
1. Do not use `filesystem` and write something yourself
2. Drop support for gcc7, update the docker containers to gcc8

I would actually prefer 2. and afaik all targeted platforms have at least gcc8 (just checked smucNG, CoolMUC2/3, HAWK).
@SteffenSeckler your opinion?

SteffenSeckler:
probably gcc9 is better as gcc8 still requires linking with libstdc++fs.
otherwise, you would have to add the correct linking options for that...

examples/md-flexible/src/Simulation.cpp
```diff
@@ -21,525 +22,189 @@ extern template bool autopas::AutoPas<ParticleType>::iteratePairwise(autopas::LJ
...
+Simulation::~Simulation() { _timers.total.stop(); }
 
-std::string Simulation::timerToString(const std::string &name, long timeNS, size_t numberWidth, long maxTime) {
-  // only print timers that were actually used
-  if (timeNS == 0) {
-    return "";
+void Simulation::run() {
+  const int iterationsPerSuperstep = _configuration.verletRebuildFrequency.value;
+  _timers.simulate.start();
+  for (int i = 0; i < _configuration.iterations.value; i += iterationsPerSuperstep) {
```
FG-TUM, line 146: I just noticed that you basically changed the meaning of configuration values. "iterations" is actually intended to be the number of iterations of the simulation. So if I do `md-flexible --iterations 2` I expect the simulation to do 2 iterations.

examples/md-flexible/src/TimeDiscretization.h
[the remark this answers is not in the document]
FG-TUM: "Time" or "Temporal" discretization refers to the update of properties that is done via time-stepping (aka `deltaT`). This is only the case for the position and velocity. Forces are independent of time and only relies on the positions of particles. Hence it has nothing to do with time discretization.

examples/md-flexible/src/ParticleSerializationTools.cpp
FG-TUM: -> continued in PR #630

FG-TUM:
It seems you also removed support for the `--tuning-phases` parameter. This one is pretty important.

FG-TUM (changes requested):
small doc fix

FG-TUM:
Add a warning for everything that does not work if used with MPI (e.g. Thermostat)

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
```diff
@@ -0,0 +1,401 @@
...
+
+  if (_subdomainCount == 1) {
+    _mpiIsEnabled = false;
+  }
+
+  if (_mpiIsEnabled) {
+    std::cout << "MPI will be used." << std::endl;
+  } else {
+    std::cout << "MPI will not be used." << std::endl;
+  }
```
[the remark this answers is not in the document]
J0llyver: I will move this to the main function on the branch md-flex/user-time-timers

examples/md-flexible/src/Simulation.cpp
```diff
@@ -382,226 +272,103 @@ std::string Simulation::timerToString(const std::string &name, long timeNS, size
...
-  auto barWidth =
-      std::max(std::min(static_cast<decltype(maxBarWidth)>(maxBarWidth * (fractionDone)), maxBarWidth), 1ul);
-  // don't print arrow tip if >= 100%
-  if (iterationProgress >= maxIterations) {
-    progressbar << std::string(barWidth, '=');
-  } else {
-    progressbar << std::string(barWidth - 1, '=') << '>' << std::string(maxBarWidth - barWidth, ' ');
+void Simulation::calculateGlobalForces(const std::array<double, 3> &globalForce) {
+#ifdef AUTOPAS_OPENMP
+#pragma omp parallel shared(autoPasContainer)
```
FG-TUM, line 365:
```suggestion
#pragma omp parallel shared(_autoPasContainer)
```

examples/md-flexible/src/Simulation.cpp
```diff
@@ -21,346 +22,235 @@ extern template bool autopas::AutoPas<ParticleType>::iteratePairwise(autopas::LJ
...
+  _autoPasContainer->setTuningInterval(_configuration.tuningInterval.value);
+  _autoPasContainer->setTuningStrategyOption(_configuration.tuningStrategyOption.value);
+  _autoPasContainer->setMPIStrategy(_configuration.mpiStrategyOption.value);
+  _autoPasContainer->setVerletClusterSize(_configuration.verletClusterSize.value);
+  _autoPasContainer->setVerletRebuildFrequency(_configuration.verletRebuildFrequency.value);
+  _autoPasContainer->setVerletSkin(_configuration.verletSkinRadius.value);
+  _autoPasContainer->setAcquisitionFunction(_configuration.acquisitionFunctionOption.value);
+  _autoPasContainer->init();
+
+  autopas::Logger::get()->set_level(_configuration.logLevel.value);
```
FG-TUM, line 118:
```suggestion
  autopas::Logger::get()->set_level(_configuration.logLevel.value);
  _autoPasContainer->init();
```

the `init()` function has a logger call that would be unaffected if the level is set afterward.

examples/md-flexible/src/Simulation.cpp
```diff
@@ -21,346 +22,235 @@ extern template bool autopas::AutoPas<ParticleType>::iteratePairwise(autopas::LJ
...
+  if (_createVtkFiles) {
+    _vtkWriter->recordTimestep(_iteration, *_autoPasContainer);
   }
 }
 
-void Simulation::simulate(autopas::AutoPas<ParticleType> &autopas) {
-  this->_homogeneity = Simulation::calculateHomogeneity(autopas);
-  _timers.simulate.start();
+void Simulation::executeSupersteps(const int iterationsPerSuperstep) {
+  for (int i = 0; i < iterationsPerSuperstep && needsMoreIterations(); ++i) {
```
FG-TUM, line 158:
```suggestion
  for (int i = 0; i < iterationsPerSuperstep and needsMoreIterations(); ++i) {
```
consistency with the rest of the project

examples/md-flexible/src/Simulation.cpp
FG-TUM:
```suggestion
  while(needsMoreIterations()) {
```
That should do it. I still think this "superstep" loop is weird. Any reason not to merge it with the loop from line 158?

FG-TUM (changes requested):
Also, see the unresolved conversation [here](https://github.com/AutoPas/AutoPas/pull/597/files?file-filters%5B%5D=.cpp#r679789941).

examples/md-flexible/src/Simulation.cpp
J0llyver: I will remove  the executeSuperStep function and implement your suggestion. I am not sure anymore what I had in mind when implementing it.

FG-TUM:
@J0llyver Tests are ~failing~ not compiling because the tests for `calculatePairwiseForces` and `calculateGlobalForces` are still in `TimeDiscretizationTest.cpp`. Move them please.

J0llyver:
I will just uncomment the Tests. Just "moving" them would mean setting up tests for the Simulaton class. 
I don't think I would be able to finish this today and therefor create another todo on my side to write reimplement those tests.
Are you ok with this?

FG-TUM:
> I will just comment out the Tests. Just "moving" them would mean setting up tests for the Simulaton class.
> I don't think I would be able to finish this today and therefor create another todo on my side to write reimplement those tests.
> Are you ok with this?

Not too happy about it tbh. Isn't it just copying the tests, creating a Simulation object, and feeding it a simple `MDFlexConfig` object? If it is indeed more complicated then I'm ok with postponing, however, please also create an issue so that we don't forget this.

examples/md-flexible/src/ParticleSerializationTools.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+namespace ParticleSerializationTools {
+void serializeParticle(const ParticleType &particle, std::vector<char> &serializedParticles) {
+  ParticleAttributes attributes{.position{particle.getR()},
+                                .velocity{particle.getV()},
+                                .force{particle.getF()},
+                                .id = particle.getID(),
+                                .ownershipState = particle.getOwnershipState(),
+                                .typeId = particle.getTypeId(),
+                                .oldForce{particle.getOldF()}};
```
SteffenSeckler, line 16: these use dedicated initialization which is a c++20 extension (and previously a gnu extension).
while I see the benefit, they sadly are not yet usable, because we still use c++17.

J0llyver: the struct ParticleAttributes has been removed in branch md-flex/particle-serialization which will be reviewed after md-flex/cart-decomposition has been merged, as it depends on it.

SteffenSeckler: PRs will not be merged with failing compilation...

SteffenSeckler: as it stands, #630's target is this branch anyways. so it might make sense to merge that one first into this branch.

J0llyver:
It is not tthis simple. 
I would need to make the functions calculatePairwiseForces / GlobalForces public which don't like. I would need to create a getter fur the autoPasContainer in the Simulation, which is only used for testing which is also not nice. Sure these are still minor changes, 
but I'd rather spend the time to think about a proper way to do the tests for the force updates. (moving them to their own header, probably.) I only a little more than an hour left to work today and I don't think this is enough time.

examples/md-flexible/src/ParallelVtkWriter.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+  timestepFile.close();
+}
+
+void ParallelVtkWriter::tryCreateSessionAndDataFolders(const std::string &name, std::string location) {
+  time_t rawTime;
+  time(&rawTime);
+
+  struct tm timeInformation;
+  gmtime_r(&rawTime, &timeInformation);
```
FG-TUM, line 136: Why did you change this from `localtime`? Local timezones seem a bit less confusing than GMT.

FG-TUM (approved):
Remaining stuff will be addressed in future PRs

examples/md-flexible/src/ParallelVtkWriter.cpp
J0llyver: Using localtime was considered dangerous by one of the bots. Although this introduced i minor bug (time is 2 hours off) I am planning to find a way to generate the local time without using localtime.

PR #601, opened 2021-05-04
Fix switching to 32-bit floating point precision

### LINKED ISSUE #600: Switch to 32-bit floating point precision in md-flexible gives compilation errors
**Describe the bug**
According to the comment, using float instead of double for md-flexible should be as easy as changing the type here:
https://github.com/AutoPas/AutoPas/blob/afd607d0be64f4a4a135f041c0936b059e5b8bc1/examples/md-flexible/src/TypeDefinitions.h#L12-L15
This is not the case. Changing this gives compilation errors. Listing all compilation problems here does not make sense, it is kind of a bottomless pit. However, they aren't limited to md-flexible code, but also in the AutoPas library itself. 

**To Reproduce**
1. Change to the type from `float` to `double`.
2. Compile.

**Expected behavior**
The compilation should not fail.

HobbyProgrammer:
# Description
This pull request fixes the compilation errors that happen when switching to 32-bit floating point precision in AutoPas or md-flexible. A lot of changes were necessary and some choices had to be made. I will post some discussion points shortly.

## Resolved Issues
- [ ] fixes #600 

# How Has This Been Tested?
Running the test suite.

## To Do
- [ ] Add a test that compiles AutoPas with a particle that uses 32-bit precision.
- [ ] Add 32-bit precision implementation for the AVX intrinsics stuff.

### DIFF
--- a/examples/md-flexible/src/BoundaryConditions.h
+++ b/examples/md-flexible/src/BoundaryConditions.h
@@ -107,7 +107,8 @@ std::vector<Particle> identifyNewHaloParticles(autopas::AutoPas<Particle> &autoP
         for (auto iter = autoPas.getRegionIterator(min, max, autopas::IteratorBehavior::owned); iter.isValid();
              ++iter) {
           auto particleCopy = *iter;
-          particleCopy.addR(shiftVec);
+          particleCopy.addR(autopas::utils::ArrayUtils::static_cast_array<typename Particle::ParticleSoAFloatPrecision>(
+              shiftVec));
           haloParticles.push_back(particleCopy);
         }
       }
@@ -117,7 +118,7 @@ std::vector<Particle> identifyNewHaloParticles(autopas::AutoPas<Particle> &autoP
 }
 
 /**
- * Insertes the given entering particles into the AutoPas object.
+ * Inserts the given entering particles into the AutoPas object.
  * @tparam Particle
  * @param autoPas
  * @param enteringParticles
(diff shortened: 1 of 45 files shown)


HobbyProgrammer:
Discussion points and requests for help:
- The functors need a good look, especially for the globals stuff. I don't know where double precision is still wanted.
- Currently, all boundary calculations and box calculations are still done in double precision. This makes a lot of casts necessary. Is this the way to go?
- For some tests, it seems like they used both TypeDefinitions.h and commonTypedefs.h types. One example are the GridCube tests. I was confused with what to use there.
- Critical: In ArrayMath.h, I added a second float type to some methods to allow operations on data with different precisions. The return type currently is always chosen to be the type of the first array parameter. This might be counterintuitive for some applications.
- In order to have the `MoleculeLJ` actually use the float type given as a template parameter, I changed 
``using Particle = ParticleFP64;``
 to 
``template<typename floatType>
using Particle = ParticleBase<floatType, unsigned long>;`` in [Particle.h](https://github.com/AutoPas/AutoPas/blob/6f8e444b0d33c5b87cfe8d285b43d05d9c02f305/src/autopas/particles/Particle.h). Wherever `Particle` was used before, I changed it to `ParticleFP64` directly. This happened mostly in the tests. I am not sure what the reason was to define this `Particle` type in the first place, so maybe, this should have been solved in a different way.

FG-TUM:
Will not be merged but should not be deleted as a reminder of where to start if we want this feature.

Issue #606, opened 2021-05-31
CPU and Memory usage at compilation is horrible


SteffenSeckler:
**Describe the bug**
Memory and needed time for compilation is horrible!

**To Reproduce**
One of:
- compile tests
- compile md-flex
- compile ls1

**Expected behavior**
Memory and needed time is not horrible.

**Additional context**
Investigate using:
- https://github.com/mikael-s-persson/templight + https://github.com/mikael-s-persson/templight-tools/
- clang's -ftime-trace

SteffenSeckler:
[archive.zip](https://github.com/AutoPas/AutoPas/files/6574657/archive.zip)

This file might help. It contains the time needed for different tasks when compiling ls1's AutoPasContainer class with clang. Generated using -ftime-trace + execution of process.sh

File structure:
time occurrences type what

The time is the accumulated time for all occurrences.

HobbyProgrammer:
As @SteffenSeckler mentioned in #607 removing the functor template parameter, would definitely be a working and easy solution for compile time improvement. Most of the overhead comes from compiling the traversals multiple times. Normally, each traversal has ParticleCell, dataLayout and newton3 as template parameters besides the functor type. Currently, each of these three parameters have at least 2 different values, so it is already x8 instantiations. If, as in md-flexible, we have 4 functors, one traversal is instantiated 32 times. (At the moment, even if a traversal does not support a particular config, it is still instantiated and isApplicable() is called at runtime to check if it is supported)

Here is a visualization of that:
![grafik](https://user-images.githubusercontent.com/4396025/120779177-1b754a00-c527-11eb-91f1-366504578235.png)
This shows the time the compiler uses for template instantiation. The four big blocks seen here are the four different functors used in md-flexible. In each of these blocks, we first see the division into the two different ParticleCells, FullParticleCell and ReferenceParticleCell. Then, for each of these particle cells, we see 6 instantiations below, which correspond to the three data layouts AoS, SoA and Cuda and the two newton3 options. 

Here is a picture zoomed in to one of these configurations:
![grafik](https://user-images.githubusercontent.com/4396025/120780052-fcc38300-c527-11eb-9f93-a6e1e236a2b8.png)
Each of the small blue flames there is a unique_ptr to a traversal. Each of the small green flames on the right is the make_unique implementation for a traversal. This shows that for each of the 4 functor options * 2 cell options  * 3 data layout options * 2 newton3 options = 48 options, all traversals are instantiated once.

[...]

PR #607, opened 2021-06-02
Optimize compile time for md-flexible

HobbyProgrammer:
# Description

This pull request aims to improve the compile time for md-flexible. There are two main points that have been improved:
1. Changing the source code of md-flexible in main.cpp and Simulation.cpp does no longer lead to a full recompilation of whole AutoPas. Once the AutoPas compilation units are compiled, recompiling md-flexible on release mode is a matter of seconds now, not minutes.
2. The AutoPas compilation for md-flexible has been split up in multiple compilation units to enable parallel compilation. There is one compilation unit for the main AutoPas template and one for each iteratePairwise() template instantation, that is one for each functor that is used, so four in total at the moment. This was done because each iteratePairwise() instantiation basically instantiates a whole copy of the library for the given functor, dominating the compile time. This gives a factor of roughly around x3 improvement in compile time, if enough cores are used.

Main code changes:
1. Use explicit instantiation declarations in main.cpp and Simulation.cpp for the AutoPas template and its iteratePairwise() function templates. The corresponding instantiation definitions are outsourced to different compilation units.
2. Split of AutoPas.h in AutoPasDecl.h for the declarations and simple implementations and AutoPasImpl.h for method implementations that depend on AutoTuner.h and LogicHandler.h. While this does not change the time spent for optimization and code generation, it heavily reduces the time to parse and to instantiate the template declarations when AutoPasDecl.h is included instead of AutoPas.h.

Additionally, a new cmake option for generating compile time profiling information has been added.

## Related Issues

-  #606 

# How Has This Been Tested?

Here is a comparison of the compilation of md-flexible on release mode on my laptop before and after. (Generated with clang's ftime-trace and visualized with chrome)

**_Please notice the timeline on the top._** 

### Before:
Simulation.cpp
Simulation.h

### After:
Simulation.cpp
Simulation.h

+ four times in parallel different AutoPas template instantiations
AutoPasTemplateInstantiations

When building everything in parallel, there is a decrease in compile of around a factor of 3. When only building Simulation.cpp, the improvement is around x45.


src/autopas/AutoPasImpl.h
```diff
@@ -0,0 +1,181 @@
+/**
+ * @file AutoPasImpl.h
+ * Contains all the missing implementations from AutoPasDecl.h that cannot be implemented there because AutoTuner and
+ * LogicHandler are only forward declared.
+ */
+
+#pragma once
+
+#include "autopas/AutoPas.h"
```
SteffenSeckler, line 9:
```suggestion
#include "autopas/AutoPasDecl.h"
```

HobbyProgrammer:
Maybe someone can help with the problem that doxygen produces warnings for the template instantiations. I have tried to reproduce the problem locally, but at least on the most up to date version of doxygen (ca24685fbbb1b861a1bd7cc6324407ca0424b325), no warnings are produced. Is this maybe a problem of the doxygen version used on the build server? (ae0844009961225fe7b9755cb5e3e0a2ee90f00c from May 03)

src/autopas/AutoPasImpl.h
HobbyProgrammer: Thank you!

HobbyProgrammer:
Nevertheless, I would be grateful for a hint on how to resolve the doxygen problem. The only solution that worked for me so far is to disable the doxygen warnings for the template instantiations using `//! @cond Doxygen_Suppress`.

SteffenSeckler:
hmm, I wouldn't be surprised if it were a false-positive.
You might try to create a simple reproducer and report it at the [doxygen github page](https://github.com/doxygen/doxygen/issues).
For now, you can simply suppress it.

HobbyProgrammer:
The latest commit removes all includes of AutoPas.h in files included in Simulation.h. This would make it possible to remove the explicit instantiation declarations in Simulation.cpp while still getting the good compile time, as the compilation unit just does not contain any definitions for the expensive templates anymore. Instantiating the templates in one or multiple other compilation units still would be required of course.

The advantage of that would be less effort for maintainance and cleaner code.

However, if we would like to implement it without these explicit instantiation declarations, developers of md-flexible would have to be very careful to not include anything that includes AutoPas.h itself, as that would lead to the very high compile time again. Having the explicit instantiation declarations prevents the high compile time even in the presence of the template definitions in AutoPas(Impl).h.

In both solutions, when adding new functors to md-flexible, the developer has to be aware how to add it. The important point is to not just add AutoPas.h as a new include. While this works perfectly, the compile time would increase again.
With the explicit instantiation declarations, the developer has to add another one of them and also another compilation unit containing the real instantiation. Without these explicit instantiation declarations, only the instantiation has be be added to another compilation unit.

SteffenSeckler:
I have actually thought about the possibility to define at least some `extern template` in the AutoPas library itself.
But I think that this is complicated as the functor is not actually known.
Maybe, it is possible to somehow remove the need for the functor where possible, s.t., fewer templates are needed.

HobbyProgrammer:
I also thought about that, but for me, it seems even impossible as I think there is no way to use template parameters for explicit instantiation declarations. What would be possible on the other hand is providing some kind of auto generation of the other compilation units outside of C++, maybe through cmake, but I am not sure if we want to go down that road.

  Your second solution, removing the functor template parameter, would definitely be the best and easiest solution for compile time improvement. I have also thought about that and will explain some of my thoughts in #606 in a moment.

src/autopas/AutoPasImpl.h
```diff
@@ -0,0 +1,181 @@
...
+ * Contains all the missing implementations from AutoPasDecl.h that cannot be implemented there because AutoTuner and
+ * LogicHandler are only forward declared.
+ */
+
+#pragma once
+
+#include "autopas/AutoPasDecl.h"
+#include "autopas/LogicHandler.h"
+#include "autopas/selectors/AutoTuner.h"
+#include "autopas/selectors/tuningStrategy/TuningStrategyFactory.h"
```
FG-TUM, line 12: probably some includes missing. Here are some I spotted:
```suggestion
#include <memory>
#include <ostream>
#include <type_traits>

#include "autopas/AutoPasDecl.h"
#include "autopas/LogicHandler.h"
#include "autopas/selectors/AutoTuner.h"
#include "autopas/selectors/tuningStrategy/TuningStrategyFactory.h"
```

They are probably in `AutoPasDecl.h` and might be removable there.

src/autopas/AutoPasDecl.h
```diff
@@ -0,0 +1,752 @@
...
+#include "autopas/options/IteratorBehavior.h"
+#include "autopas/options/LoadEstimatorOption.h"
+#include "autopas/options/MPIStrategyOption.h"
+#include "autopas/options/Newton3Option.h"
+#include "autopas/options/SelectorStrategyOption.h"
+#include "autopas/options/TraversalOption.h"
+#include "autopas/options/TuningStrategyOption.h"
+#include "autopas/selectors/Configuration.h"
+#include "autopas/utils/NumberSet.h"
+#include "autopas/utils/WrapMPI.h"
```
FG-TUM, line 29: Check whether all of those are still necessary

FG-TUM (commented):
I have not yet looked into the doxygen problem. However, I would be fine with suppressing the warning.

FG-TUM (approved):

PR #610, opened 2021-06-09
Octree Implementation

neftlon:
# Description

* Created a new container that stores particles in two octrees: One for the owned particles and one for the halo particles.
* Adapt the iteration patterns for containers (`begin`, `getRegionIterator` and `iteratePairwise`)
* Added test cases that verify the validity of the internal octree structure
* Added a simple C01 traversal that is capable of iterating the octree structure pairwise using the Samet-neighbor finding approach
* Started to add a C18 traversal, that is currently disabled (see #621 for more information)

The code provides two methods of finding leaves in the octree:

* A neighbor finding algorithm that uses the spacial structure of the octree (relationship between parents/children and careful crafting of the connecting pointers) according to [the paper](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.80.5114&rep=rep1&type=pdf) from Hanan Samet
* A range query approach that includes all leaves that are found in a given region

These methods are used in the traversal functions and inside `getRegionIterator` and provide fast access to the leaf nodes of the octree. 

# How Has This Been Tested?

- [x] Add predefined particles to the octree and check whether children are correctly generated and spaced using the `testChildIndexing` test
- [x] Create an octree with random particles and check for each leaf whether found neighbors really touch the leaf in the intended way in the `testNeighborLocator`.
- [x] Check if the range-query finding of neighbor nodes in `testRangeNeighborFinding`
- [x] Verification that leaves only split if their size allows them to (tested in `testUnableToSplit` and `testAbleToSplit`)
- [x] Compare the interactions that are generated from the octree's `iteratePairwise` function with a reference implementation (`LinkedCells`) in `testCustomParticleDistribution`


tests/testAutopas/tests/selectors/AutoTunerTest.cpp
```diff
@@ -82,6 +82,10 @@ TEST_F(AutoTunerTest, testAllConfigurations) {
   //                        vlp_c01                     (AoS <=> SoA, noNewton3)                             = 2
   configsPerContainer[autopas::ContainerOption::pairwiseVerletLists] = 18;
 
+  // Octree:                ot_c01                      (AoS <=> SoA, noNewton3)                             = 2
+  //                        ot_naive                    (disabled)                                           = 0
```
FG-TUM, line 86: why disabled?

tests/testAutopas/tests/containers/AllContainersTests.cpp
```diff
@@ -64,8 +64,9 @@ TEST_P(AllContainersTests, testParticleAdding) {
                        boxMax[2] + .5, boxMax[2] + 1.5}) {
         autopas::Particle p({x, y, z}, {0., 0., 0.}, id++);
         if (x == -1.5 or y == -1.5 or z == -1.5 or x == 11.5 or y == 11.5 or z == 11.5) {
-          EXPECT_ANY_THROW(container->addParticle(p));     // outside, therefore not ok!
-          EXPECT_NO_THROW(container->addHaloParticle(p));  // much outside, still ok because it is ignored!
+          EXPECT_ANY_THROW(container->addParticle(p));  // outside, therefore not ok!
+          EXPECT_NO_THROW(container->addHaloParticle(p))
+              << " x=" << x << " y=" << y << " z=" << z;  // much outside, still ok because it is ignored!
```
FG-TUM, line 69: you can also directly print the whole particle. I think like this:
```suggestion
              << p.tostring();  // much outside, still ok because it is ignored!
```

src/autopas/options/TraversalOption.h
```diff
@@ -77,6 +77,14 @@ class TraversalOption : public Option<TraversalOption> {
      */
     lc_sliced_c02,
 
+    // Octree Traversals:
+    // TODO(johannes): Documentation
```
FG-TUM, line 81: TODO ;)

FG-TUM (commented):
Still lots of TODOs in there ;) Check whether they all are still valid then we can discuss them in our next meeting.

Besides that remove or revert the `#if 0` blocks and add documentation where you still intend to do so.

When this resolved I will have a closer look at the code.

tests/testAutopas/tests/containers/TraversalComparison.cpp
```diff
@@ -293,6 +294,35 @@ auto TraversalComparison::getTestParams() {
...
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+#endif
```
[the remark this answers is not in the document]
neftlon: Reverted in 02f7712

tests/testAutopas/tests/containers/octree/OctreeTests.cpp
```diff
@@ -0,0 +1,791 @@
...
+      auto box1Max = box1->getBoxMax();
+      auto box2Min = box2->getBoxMin();
+      auto box2Max = box2->getBoxMax();
+      ASSERT_GE(box1Max[axis], box2Min[axis]);
+      ASSERT_LT(box1Min[axis], box2Min[axis]);
+    }
+  }
+}
+
+#if 0
```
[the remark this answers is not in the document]
neftlon: Yes, you are right. This is from a test case that used some deprecated tree indexing. I removed the test case in c9a9267.

tests/testAutopas/tests/containers/octree/OctreeTests.cpp
```diff
@@ -0,0 +1,791 @@
...
+
+  // Obtain a compatible traversal
+  auto traversal =
+      autopas::utils::withStaticCellType<Molecule>(container->getParticleCellTypeEnum(), [&](auto particleCellDummy) {
+        return autopas::TraversalSelector<decltype(particleCellDummy)>::generateTraversal(
+            traversalOption, mockFunctor, container->getTraversalSelectorInfo(), dataLayoutOption, newton3Option);
+      });
+
+  if (doSlightShift) {
+    // TODO(johannes): Shift the particles
```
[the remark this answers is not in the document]
neftlon: This is also an artifact from adopting the code from the TraversalComparison test. However, since this is not part of the scope that is tested here, I removed this in f2b59fc. The slight-shifting capabilities of the container are tested in the TraversalComparison test and do not need to be tested here.

tests/testAutopas/tests/containers/octree/OctreeTests.cpp
```diff
@@ -0,0 +1,791 @@
...
+  }
+
+  // std::cout << "Container switch to " << autopas::ContainerOption::getOptionNames()[containerOption] << "\n";
```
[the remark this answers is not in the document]
neftlon: Removed in f2b59fc

tests/testAutopas/tests/containers/AllContainersTests.cpp
neftlon: Since this was only used for debugging purposes, I reverted it back to its original state in da1abe0.

src/autopas/options/TraversalOption.h
neftlon: Added documentation in e2c016e

tests/testAutopas/tests/selectors/AutoTunerTest.cpp
neftlon: As discussed, I added a new issue and referenced it in a comment at that place in the code in dc8f1dc.

FG-TUM:
Can you add a link to the paper you mention in the description of the PR?

src/autopas/containers/octree/Octree.h
```diff
@@ -0,0 +1,240 @@
...
+
+    // leaving: all outside boxMin/Max
+
+    // TODO(johannes): Make this less indirect. (Find a better way to iterate all particles inside the octree to change
+    //  this function back to a function that actually copies all particles out of the octree.)
+    //  The problem is captured by https://github.com/AutoPas/AutoPas/issues/622
+    std::vector<Particle *> particleRefs;
+    this->_cells[CellTypes::OWNED].appendAllParticles(particleRefs);
+    std::vector<Particle> particles;
+    auto result = std::vector<ParticleType>();
```
FG-TUM, line 93:
```suggestion
    std::vector<Particle> particles{};
    particles.reserve(particleRefs.size());
    std::vector<Particle> result{};
```
maybe also rename `result` to `invalidParticles` or similar?
Also: Most particles will end up in the `particles` vector, hence I think it would make sense to do some reserving.

src/autopas/containers/octree/Octree.h
```diff
@@ -0,0 +1,240 @@
...
+    }
+    this->deleteAllParticles();
+
+    for (auto &particle : particles) {
+      addParticleImpl(particle);
+    }
+
+    // This should happen according to the documentation in the ParticleContainerInterface#updateContainer
+    deleteHaloParticles();
+
```
FG-TUM, line 109: true but calling this only makes sense if the `addParticleImpl()` call might have added any halo particles. Otherwise the call to `deleteAllParticles()` some lines above should already have taken care of the halos

src/autopas/containers/octree/Octree.h
```diff
@@ -0,0 +1,240 @@
...
+    return ParticleIteratorWrapper<ParticleType, false>(
+        new internal::RegionParticleIterator<ParticleType, ParticleCell, false>(
+            &this->_cells, lowerCorner, higherCorner, cellsOfInterest, this, behavior));
+  }
+
+  /**
+   * @copydoc ParticleContainerInterface::getTraversalSelectorInfo()
+   */
+  [[nodiscard]] TraversalSelectorInfo getTraversalSelectorInfo() const override {
+    std::array<unsigned long, 3> dims = {1, 1, 1};
```
FG-TUM, line 203:
```suggestion
    // this is a dummy since it is not actually used
    std::array<unsigned long, 3> dims = {1, 1, 1};
```

src/autopas/containers/octree/Octree.h
```diff
@@ -0,0 +1,240 @@
...
+
+    // This should happen according to the documentation in the ParticleContainerInterface#updateContainer
+    deleteHaloParticles();
+
+    return result;
+  }
+
+  void iteratePairwise(TraversalInterface *traversal) override {
+    auto *traversalInterface = dynamic_cast<OTTraversalInterface<ParticleCell> *>(traversal);
+    if (traversalInterface) {
```
FG-TUM, line 115:
```suggestion
        if (auto *traversalInterface = dynamic_cast<OTTraversalInterface<ParticleCell> *>(traversal)) {
```

src/autopas/containers/octree/Octree.h
```diff
@@ -0,0 +1,240 @@
...
+
+  /**
+   * @copydoc ParticleContainerInterface::getContainerType()
+   */
+  [[nodiscard]] ContainerOption getContainerType() const override { return ContainerOption::octree; }
+
+  /**
+   * @copydoc ParticleContainerInterface::getParticleCellTypeEnum()
+   */
+  [[nodiscard]] CellType getParticleCellTypeEnum() override { return CellType::FullParticleCell; }
```
FG-TUM, line 132: I think it might be worth mentioning that this container does not actually use the FullParticleCell but something that is sufficiently similar.

FG-TUM (commented):
Review Part 1 (mostly the Octree data structure)
- Added some comments for myself
- Check for `const` correctness

src/autopas/containers/octree/Octree.h
```diff
@@ -0,0 +1,240 @@
...
+   * @param cellSizeFactor The cell size factor
+   */
+  Octree(std::array<double, 3> boxMin, std::array<double, 3> boxMax, const double cutoff, const double skin,
+         const double cellSizeFactor)
+      : CellBasedParticleContainer<ParticleCell>(boxMin, boxMax, cutoff, skin) {
+    // TODO(johannes): Obtain this from a configuration, reported in https://github.com/AutoPas/AutoPas/issues/624
+    int unsigned treeSplitThreshold = 16;
+
+    double interactionLength = this->getInteractionLength();
+    this->_cells.push_back(
```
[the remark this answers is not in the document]
neftlon: I (hopefully) cleaned this up in 3b63d16 by first allocating memory inside the `_cells` vector and then putting the octrees into their respective slots (`HALO` and `OWNED`) manually. Does that work for you?

src/autopas/containers/octree/Octree.h
```diff
@@ -0,0 +1,240 @@
...
+        OctreeNodeWrapper<Particle>(boxMin, boxMax, treeSplitThreshold, interactionLength, cellSizeFactor));
+
+    // Extend the halo region with cutoff + skin in all dimensions
+    auto haloBoxMin = utils::ArrayMath::subScalar(boxMin, this->getInteractionLength());
+    auto haloBoxMax = utils::ArrayMath::addScalar(boxMax, this->getInteractionLength());
+    this->_cells.push_back(
```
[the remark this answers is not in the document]
neftlon: Should be fixed in 3b63d16 as well.

src/autopas/containers/octree/Octree.h
```diff
@@ -0,0 +1,240 @@
...
+
+  [[nodiscard]] std::vector<ParticleType> updateContainer() override {
+    // This is a very primitive and inefficient way to recreate the container:
+    // 1. Copy all particles out of the container
+    // 2. Clear the container
+    // 3. Insert the particles back into the container
+
+    // leaving: all outside boxMin/Max
+
+    // TODO(johannes): Make this less indirect. (Find a better way to iterate all particles inside the octree to change
```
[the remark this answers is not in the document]
neftlon: Fixed in ca54cff.

src/autopas/containers/octree/Octree.h
```diff
@@ -0,0 +1,240 @@
...
+    // TODO(johannes): Obtain this from a configuration, reported in https://github.com/AutoPas/AutoPas/issues/624
+    int unsigned treeSplitThreshold = 16;
+
+    double interactionLength = this->getInteractionLength();
+    this->_cells.push_back(
+        OctreeNodeWrapper<Particle>(boxMin, boxMax, treeSplitThreshold, interactionLength, cellSizeFactor));
+
+    // Extend the halo region with cutoff + skin in all dimensions
+    auto haloBoxMin = utils::ArrayMath::subScalar(boxMin, this->getInteractionLength());
+    auto haloBoxMax = utils::ArrayMath::addScalar(boxMax, this->getInteractionLength());
```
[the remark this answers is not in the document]
neftlon: Fixed in 24685e2.

src/autopas/containers/octree/Octree.h
neftlon: You are right, I remove the duplication in c51cfb6.

src/autopas/containers/octree/Octree.h
neftlon: Adopted changes in 981febf.

src/autopas/containers/octree/OctreeNodeInterface.h
```diff
@@ -0,0 +1,543 @@
...
+   * @param p The particle to insert
+   * @return A std::unique_ptr to a newly created subtree or nullptr if the subtree did not change
+   */
+  virtual std::unique_ptr<OctreeNodeInterface<Particle>> insert(Particle p) = 0;
+
+  /**
+   * Put all particles that are below this node into the vector.
+   * @param ps A reference to the vector that should contain the particles after the operation
+   */
+  virtual void appendAllParticles(std::vector<Particle *> &ps) = 0;
```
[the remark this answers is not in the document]
neftlon: Marked const in 2e28162.

src/autopas/containers/octree/Octree.h
neftlon: Do you mean that I should add a comment that says that or should I add another custom `CellType`?

src/autopas/containers/octree/OctreeNodeWrapper.h
```diff
@@ -0,0 +1,193 @@
...
+
+namespace autopas {
+/**
+ * This class wraps the functionality provided by the octree leaves and inner nodes in a structure that adheres to the
+ * ParticleCell concept. This is necessary to use the octree nodes as particle containers.
+ *
+ * @tparam Particle The particle class that should be used in the octree cell.
+ */
+template <typename Particle>
+class OctreeNodeWrapper : public ParticleCell<Particle> {
```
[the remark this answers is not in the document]
neftlon: In my opinion this would not necessarily lead to cleaner code: If I inherited from `OctreeNodeInterface<Particle>` in `OctreeNodeWrapper`, I would have to implement the virtual methods that are defined in the `OctreeNodeInterface<Particle>` as well (for instance `insert`). However, this interface should not be exposed to the outside. (`insert` returns a pointer to the child itself, which should only be used internally by the octree container. In my opinion, this is solved by using the wrapper.)

I think you are very right that a method calling the same method does not make some beautiful code. But in my opinion, compared to the downside of having to implement the virtual method from `OctreeNodeInterface<Particle>` in the `OctreeNodeWrapper` (and let those only throw an exception whenever called), this is the smaller evil.

src/autopas/containers/octree/Octree.h
FG-TUM: just expand the documentation.

src/autopas/containers/octree/OctreeNodeWrapper.h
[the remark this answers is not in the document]
FG-TUM: Ok sounds reasonable. However, I think it would also be reasonable to explain this reasoning in the documentation of the class ;)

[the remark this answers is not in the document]
neftlon: I added the reasoning to the documentation in d7a2c2c:

https://github.com/AutoPas/AutoPas/blob/d7a2c2c349a1eb571268ebf72b92d489f316cbb5/src/autopas/containers/octree/OctreeNodeWrapper.h#L17-L37

Let me know if this works for you! :)

src/autopas/containers/octree/Octree.h
neftlon: In 90c931d I added a note to the documentation: 

https://github.com/AutoPas/AutoPas/blob/90c931d1df57dbc85912c3d37cc09ba1055a4016/src/autopas/containers/octree/Octree.h#L129-L135

Let me know whether that's ok for you :)

src/autopas/containers/octree/OctreeDirection.h
```diff
@@ -0,0 +1,532 @@
+/**
+ * @file OctreeInnerNode.h
```
FG-TUM, line 2: this stuff feels so special I think it should go in a dedicated namespace. E.g. `OctreeDirection` Don't forget to also place the namespace's doc in `namespaces.h`

src/autopas/containers/octree/OctreeDirection.h
```diff
@@ -0,0 +1,532 @@
...
+  D = 3,
+  U = 4,
+  B = 5,
+  F = 6,
+};
+
+/**
+ * A datatype that is wide enough to hold faces, edges or vertices.
+ */
+using Any = int unsigned;
```
FG-TUM, line 28:
```suggestion
/**
 * A datatype that is wide enough to hold faces, edges or vertices.
 */
using Any = int unsigned;

/**
 * This enum can be used to index the faces of a cube including an "invalid" face.
 */
enum Face : Any {
  O = 0,  // omega/unknown
  L = 1,
  R = 2,
  D = 3,
  U = 4,
  B = 5,
  F = 6,
};
```

That way you make the enum have the underlying type `Any`.

src/autopas/containers/octree/OctreeDirection.h
```diff
@@ -0,0 +1,532 @@
...
+  RUF = buildVertex<R, U, F>(),
+};
+
+/**
+ * Get all available faces for a cube.
+ *
+ * @return A pointer to a static table. The last element is the "invalid" face O.
+ */
+inline Face *getFaces() {
+  static Face table[] = {L, R, D, U, B, F, O};
```
FG-TUM, line 97: If you wrap all of this in a namespace the probably better alternative would be to make these tables directly `constexpr static` variables and get rid of these getters.

src/autopas/containers/octree/OctreeDirection.h
```diff
@@ -0,0 +1,532 @@
...
+ * Check if a given list of elements contains the element to test.
+ *
+ * @tparam T The base type of the list elements
+ * @param all A pointer to the start element in the list
+ * @param stop A pointer to the element which is one past the last element in the list
+ * @param test The element to check
+ * @return true if the list contains the test element, false otherwise
+ */
+template <typename T>
+inline bool contains(T *all, T stop, Any test) {
```
FG-TUM, line 155: C++ style would be to do this with start / end iterators

src/autopas/containers/octree/OctreeDirection.h
```diff
@@ -0,0 +1,532 @@
...
+/**
+ * This function implements a LUT obtained from the Samet paper:
+ * "ADJ(I,O) is true if and only if octant O is adjacent to the Ith face, edge, or vertex of O's containing block"
+ *
+ * @param direction The direction I to search the adjacent neighbor
+ * @param octant The octant O from the paper
+ * @return true if the neighbor is adjacent, false otherwise
+ */
+inline bool ADJ(Any direction, Vertex octant) {
+  static std::array<std::array<bool, 8>, 1 << 9> table;
```
FG-TUM, line 211: this table too could be a static member in the namespace. Actually you could also create an nameless namespace within the namespace. Everything in there is "private" for the stuff in this file.

FG-TUM (changes requested):
Mini-review for OctreeDirection

src/autopas/containers/octree/OctreeNodeInterface.h
```diff
@@ -0,0 +1,544 @@
...
+  // Check precondition
+  if (!contains(VERTICES(), OOO, I)) {
+    throw std::runtime_error("[OctreeNodeInterface.h] Received invalid vertex.");
+  }
+
+  auto null = [](OctreeNodeInterface<Particle> *T) { return T == nullptr; };
+
+  // Find a common ancestor
+  OctreeNodeInterface<Particle> *Q, *P = this;
+  if (null(FATHER(P))) {
```
FG-TUM, line 525:
```suggestion
  if (not FATHER(P)) {
```
should also work

src/autopas/containers/octree/OctreeNodeInterface.h
```diff
@@ -0,0 +1,544 @@
...
+    Q = FATHER(P)->GTEQ_VERTEX_NEIGHBOR(I);
+  } else if (Edge commonEdge = COMMON_EDGE(I, SONTYPE(P)); commonEdge != OO) {
+    Q = FATHER(P)->GTEQ_EDGE_NEIGHBOR(commonEdge);
+  } else if (Face commonFace = COMMON_FACE(I, SONTYPE(P)); commonFace != O) {
+    Q = FATHER(P)->GTEQ_FACE_NEIGHBOR(commonFace);
+  } else {
+    Q = FATHER(P);
+  }
+
+  if ((not null(Q)) and GRAY(Q)) {
```
FG-TUM, line 537: nullptr == false in a condition

src/autopas/containers/octree/traversals/OTC01Traversal.h
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  /**
+   * Set the cells to iterate.
+   * @param cells A list of octree roots that should be used during iteration
+   */
+  void setCells(std::vector<OctreeNodeWrapper<Particle>> *cells) override { _cells = cells; }
+
+ private:
+  void loadBuffers(OctreeNodeWrapper<Particle> *wrapper, std::vector<OctreeLeafNode<Particle> *> &leaves) {
```
FG-TUM, line 112: add doc also for private methods

src/autopas/containers/octree/traversals/OTC01Traversal.h
```diff
@@ -0,0 +1,162 @@
...
+    wrapper->appendAllLeaves(leaves);
+
+    for (OctreeLeafNode<Particle> *leaf : leaves) {
+      leaf->clearAlreadyProcessedList();
+      _dataLayoutConverter.loadDataLayout(*leaf);
+    }
+  }
+
+  void unloadBuffers(std::vector<OctreeLeafNode<Particle> *> &leaves) {
```
FG-TUM, line 121: add doc

src/autopas/containers/octree/traversals/OTNaiveTraversal.h
```diff
@@ -0,0 +1,133 @@
...
+   */
+  using ParticleCell = OctreeLeafNode<Particle>;
+
+  /**
+   * Constructor for the Octree traversal.
+   * @param pairwiseFunctor The functor that defines the interaction of two particles.
+   * @param cutoff cutoff (this is enough for the octree traversal, please don't use the interaction length here.)
+   */
+  explicit OTNaiveTraversal(PairwiseFunctor *pairwiseFunctor, double cutoff)
+      : CellPairTraversal<ParticleCell>({2, 1, 1}),
```
FG-TUM, line 45: explain magic numbers

src/autopas/containers/octree/traversals/OTNaiveTraversal.h
```diff
@@ -0,0 +1,133 @@
...
+  void traverseParticlePairs() override {
+    // Get neighboring cells for each leaf
+    for (OctreeLeafNode<Particle> *leaf : _leaves) {
+      // Process cell itself
+      _cellFunctor.processCell(*leaf);
+
+      // Process connection to all neighbors
+      auto uniqueNeighboringLeaves = leaf->getNeighborLeaves();
+      for (OctreeLeafNode<Particle> *neighborLeaf : uniqueNeighboringLeaves) {
+        if (!leaf->alreadyProcessed(neighborLeaf) && !neighborLeaf->alreadyProcessed(leaf)) {
```
FG-TUM, line 96:
```suggestion
        if (not leaf->alreadyProcessed(neighborLeaf) and not neighborLeaf->alreadyProcessed(leaf)) {
```

src/autopas/containers/octree/traversals/OTNaiveTraversal.h
```diff
@@ -0,0 +1,133 @@
...
+  utils::DataLayoutConverter<PairwiseFunctor, dataLayout> _dataLayoutConverter;
+
+  std::vector<OctreeNodeWrapper<Particle>> *_cells;
+
+  /**
+   * A list of all leaves in the octree
+   */
+  std::vector<OctreeLeafNode<Particle> *> _leaves;
+};
+
```
FG-TUM, line 132:
```suggestion
```

FG-TUM (changes requested):
Review of all the rest :D

(The checkboxes are for you to tick when you addressed or resolved the points)

Some things I did not explicitly mark everywhere, so please check them again yourself:
- [ ] Check your `const` correctness. Many function/constructor arguments, local variables, or loop variables can be marked `const`. This helps the compiler as well as the reader.
- [ ] You often pass by copy when you could pass by reference
- [ ] Prefer `not` over `!`, `and` over `&&`, and `or` over `||`. There is neither a real difference, nor is one considered better than the other but in AutoPas we decided at some point to use the words (for readability) and now we strive for consistency ;)
- [ ] Check for commented-out code

FG-TUM (approved):
Looks reasonable. The only thing that could be done is to use actual `for_each`es in the new interface instead of the legacy iterators. But I guess this is something that should be done in a new PR. Since your project is done you are of course not obliged to do this but as we are an open-source project of course you can if you want ;)

PR #615, opened 2021-06-28
AutoPas Standalone build

FG-TUM:
# Description

1. `AUTOPAS_STANDALONE_BUILD`
AutoPas tries to determine whether it is built standalone (=for dev/demo purposes) or as a dependency. This works by doing `AUTOPAS_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR`. Depending on this, we disable and hide some CMake options via the `CMakeDependentOption` module.

2. Introduce a switch to enable / disable:
* `clangformat`
* `cmakeformat`
* `formatall`

The rationale is to avoid creating these targets when the library is only used in another project since they are not needed then. They could also create naming conflicts since each target name needs to be unique in CMake.

3. Prefix all custom CMake modules with `autopas_`. On the one hand side for consistency, on the other to avoid loading the incorrect module.

## ~Resolved Issues~

# How Has This Been Tested?

- [x] Jenkins


CMakeLists.txt
```diff
@@ -46,18 +46,24 @@ if (NOT CMAKE_BUILD_TYPE)
...
-endif (NOT CMAKE_BUILD_TYPE)
+endif ()
 
 enable_testing()
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules/)
 
+option(AUTOPAS_FORMATTING_TARGETS
+       "Adds targets for code formatting in AutoPas. Highly useful when developing AutoPas but might cause conflicts when only using it."
+       OFF)
```
FG-TUM, line 57: This changes the default behavior of AutoPas to not generate these targets when not explicitly told to. This is fine from a user's perspective but might be annoying from a developer's perspective. Hence I'm open to arguments to change it to `ON`.

SteffenSeckler: How about setting this variable depending on whether AutoPas is built stand-alone or inside a project?
e.g. by comparing CMAKE_SOURCE_DIR (directory of the top-level CMakeLists.txt) with AUTOPAS_SOURCE_DIR (directory of the autopas project)?
That way it would automatically be disabled if used inside of another project ;)

as this might come in handy elsewhere, we should probably define a variable AUTOPAS_STANDALONE or so, s.t., examples, tests, etc. could also be disabled automatically.

FG-TUM: That was a bigger idea than expected. See 46a03c5

cmake/modules/autopas_clang-sanitizers.cmake
```diff
@@ -1,6 +1,18 @@
...
+cmake_dependent_option(AUTOPAS_ENABLE_MEMORY_SANITIZER "Adds clang's memory sanitizer to the build options"
+        OFF
+        AUTOPAS_STANDALONE_BUILD
+        OFF
+)
+cmake_dependent_option(AUTOPAS_ENABLE_THREAD_SANITIZER "Adds clang's thread sanitizer to the build options"
+        OFF
+        AUTOPAS_STANDALONE_BUILD
+        OFF
+)
```
SteffenSeckler, line 15: These can actually make sense when used in other programs.
mardyn, e.g., sets those.
if they cannot be set from outside, potential internal errors of autopas' compiled files cannot be found anymore.

cmake/modules/autopas_eigen.cmake
```diff
@@ -1,4 +1,4 @@
-option(Eigen3_ForceBundled "Do not look for an installed version, always use bundled." OFF)
+option(Eigen3_ForceBundled "Do not look for an installed version, always use bundled." ON)
```
SteffenSeckler, line 1: is this intended?

cmake/modules/autopas_spdlog.cmake
```diff
@@ -1,4 +1,4 @@
-option(spdlog_ForceBundled "Do not look for an installed version, always use bundled." OFF)
+option(spdlog_ForceBundled "Do not look for an installed version, always use bundled." ON)
```
SteffenSeckler, line 1: is this intended?

cmake/modules/autopas_yaml-cpp.cmake
```diff
@@ -1,4 +1,4 @@
-option(yaml-cpp_ForceBundled "Do not look for an installed version, always use bundled." OFF)
+option(yaml-cpp_ForceBundled "Do not look for an installed version, always use bundled." ON)
```
SteffenSeckler, line 1: is this intended?

SteffenSeckler (commented):
maybe it would be better to use sth. else than CMakeDependentOption.
Especially for the sanitizers, I would prefer sth. that only marks the options as advanced instead of not setting them at all.
Potentially, we could modify the code of [CMakeDependentOption](https://github.com/Kitware/CMake/blob/master/Modules/CMakeDependentOption.cmake), s.t., we also have a version that marks it as advanced instead of removing the option entirely.

But really only needed for the sanitizers.

cmake/modules/autopas_clang-sanitizers.cmake
FG-TUM: They can be set from the outside, they are just not visible in any cmake gui until you set `AUTOPAS_STANDALONE_BUILD=ON` which you can also do explicitly. But we maybe should document this somewhere.

cmake/modules/autopas_eigen.cmake
FG-TUM: yes. I often ended up having to set those explicitly because the lib from another build folder was found.

cmake/modules/autopas_spdlog.cmake
FG-TUM: see above

cmake/modules/autopas_yaml-cpp.cmake
FG-TUM: see above

cmake/modules/autopas_OpenMP.cmake
```diff
@@ -1,4 +1,5 @@
-option(AUTOPAS_OPENMP "Activates OpenMP shared memory parallelization." OFF)
+set(AUTOPAS_OPENMP_DOC "Activates OpenMP shared memory parallelization. (requires OpenMP 4.5)")
+option(AUTOPAS_OPENMP ${AUTOPAS_OPENMP_DOC} ON)
```
FG-TUM, line 2: THIS CHANGES DEFAULT BEHAVIOR! But AutoPas without OpenMP is kind of pointless anyway...

cmake/modules/autopas_clang-sanitizers.cmake
FG-TUM: See custom implementation in 2b782bc and its application in e0b2429

CMakeLists.txt
```diff
@@ -46,21 +56,29 @@ if (NOT CMAKE_BUILD_TYPE)
             STRING "Choose the type of build, options are: Debug Release RelWithDebInfo MinSizeRel."
             FORCE
     )
-endif (NOT CMAKE_BUILD_TYPE)
+endif ()
 
 enable_testing()
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules/)
```
SteffenSeckler, line 63: do you still need this line?

SteffenSeckler (approved):

PR #619, opened 2021-07-05
Sortedcellviewfix

### LINKED ISSUE #418: Sorting is broken
**Describe the bug**
Sorting is broken, especially with cell size factors < 1.
I have temporarily disabled sorting altogether, by setting
```c++
constexpr static unsigned long _startSorting = std::numeric_limits<unsigned long>::max();
```

sorting is broken at least for the c01 and for the c08based traversals.

**To Reproduce**

1. In master (after #301 is merged)
  1. enable sorting again 
2. Minimal scenario
  1. A minimal scenario can be found in branch `c08sortingminimalscenario`
  2. run tests

**Expected behavior**
Correct

**ToDo**
Once corrected, enable sorting again by resetting `_startSorting` in CellFunctor.
```c++
constexpr static unsigned long _startSorting = 8;
```

**Additional context**
I have tried fixing the bug in branch `c08sortingtriedfix`, but it didn't work out.

js97:
# Description

This PR should fix the bugs encountered when using `SortedCellViews` for all affected traversals (c01, c04, c08, c18).

The bug was basically that the wrong offsets were used for the calculation of the direction vector. To check if two cells interact with each other, we calculate the minimum distance between the two cells and check if this distance is smaller than the `interactionLength`. For directly adjacent cells this distance is 0 in at least one dimension (edge between both cells).
This distance (`pos`) vector was also used for the calculation of the sorting-direction vector, which led to wrong results (i.e. neighbored cells clearly shouldn't have sorting direction 0, 0, 0).

Fix: The sorting direction (`sortingDir`) is now calculated separately.

For all traversals that use the `CellFunctor`, a constructor argument has been added with which the sorting can be switched on and off. The default value is true (switch sorting on). This is only used in the `TraversalComparison` test to calculate the reference forces without sorting and then compare these values when sorting is enabled.

Some tests had to be adapted to work correctly with Sorting:
 - `SlicedTraversalTest`: The number of functor calls was counted, which led to a wrong result with sorting, because the functor is not called when the `_sortingCutoff` is exceeded. Now only the number of interactions that actually lead to force calculations is counted.
 - `TraversalTest`: The cutoff had to be increased slightly due to rounding errors so that sorting leads to the same number of interactions as without sorting.
 - `OctreeTest`: For the calculation of the reference forces (with LinkedCells and c08 + sorting) the functor may not be called. Therefore, `Times(testing::AtLeast(1))` was set to `Times(testing::AtLeast(containerOption != ContainerOption::octree ? 0 : 1))` for this case.

Some measurements were performed to find out if `_startSorting = 8` in the CellFunctor is still an acceptable value. Following scenario was used:
```md-flexible --no-end-config --container LinkedCells --traversal lc_c08 --newton3 enabled --data-layout AoS --particle-generator uniform --cutoff 2.5 --box-length 40 --particles-total NUMPARTICLES --iterations 20 --deltaT 0```
with `NUMPARTICLES = [172, 343, 686, 1029, 1372, 1715, 2058]` (~1 particle per two cells, ~1 particle per cell,  ~2 particles per cell, ...)
The test scenario generates 343 owned cells.
Processor: Intel(R) Xeon(R) CPU E5-2650 v2 @ 2.60GHz
`OMP_NUM_THREADS = 8` (8 HW-threads)

| NumParticles | NoSort: iteratePairwise (ns) | Sort: iteratePairwise (ns) |
|-------------:|-----------------------------:|---------------------------:|
| 172          | 1529889                     | 1983317                   |
| 343          | 2007090                     | 3024986                   |
| 686          | 2673262                     | 4221964                   |
| 1029         | 4537538                     | 5996018                   |
| 1372         | 7427471                     | 6277184                   |
| 1715         | 9832377                     | 8402543                   |
| 2058         | 13833886                    | 10043737                  |

As can be seen, the switching point is at 1372 particles which are ~4 particles per cell. Therefore, the value `_startSorting=8` (sum of the number of particles for two cells) is still a good value and has been adopted.

## Resolved Issues

- [x] fixes #418 

Taken from js97: 
>The reason this was only discovered at low cell size factors was that the particles in the tests were generated randomly and low cell size factors decreased the expected number of particles in each cell, reducing start-end-projection-difference (decreased projection value intervals). The problem with wrong sorting directions was, that the projection was iterated in the wrong direction, i.e. comparing a particle with projection 2.31 against particles of another cell with projections in [1.1, 2.1] resulted in comparing the biggest projection difference first instead of last. Thus, the probability of this "wrong" projection was so low (because with smaller intervals, it is more likely both start and end projection are either both outside or both inside the accepted projection distance) it didn't arise in some random tests or with small cell size factors (even with the larger tests, it still was sufficiently unlikely). (Example: 2.31 and [1.89, 2.1] will be successful, as well as 2.31 and [1.1, 1.30] with cutoff radius 1, or also [1.1, 1.32] with high probability since the expected distance to the projection line drawn from the other particle is often much bigger than 0)
Though, I was able to also provoke tests with CSF 1 and 2 that had the same issue.

# Testing

- [x] The TraversalComparison test was modified so that sorting is switched off for the calculation of the reference values. These values are then compared with those for which sorting is activated.
- [x] All existing tests

### CODE REVIEW

File: src/autopas/pairwiseFunctors/CellFunctor.h [hunk 1]
```diff
@@ -111,7 +111,7 @@ class CellFunctor {
    * Min. number of particles to start sorting.
    * @todo Currently, this is disabled because of https://github.com/AutoPas/AutoPas/issues/418
    */
-  constexpr static unsigned long _startSorting = std::numeric_limits<unsigned long>::max();
+  constexpr static unsigned long _startSorting = 1;
```
FG-TUM, line 114: Previously this threshold was set to 8 since sorting was not worth it for fewer particles. Maybe run some tests and see if this is still a reasonable heuristic.

dmartin13: See PR description.

src/autopas/containers/linkedCells/traversals/LCC18Traversal.h
```diff
@@ -84,7 +87,8 @@ class LCC18Traversal : public C18BasedTraversal<ParticleCell, PairwiseFunctor, d
   /**
    * CellFunctor to be used for the traversal defining the interaction between two cells.
    */
-  internal::CellFunctor<typename ParticleCell::ParticleType, ParticleCell, PairwiseFunctor, dataLayout, useNewton3>
+  internal::CellFunctor<typename ParticleCell::ParticleType, ParticleCell, PairwiseFunctor, dataLayout, useNewton3,
+                        true>
```
FG-TUM, line 91:
```suggestion
  internal::CellFunctor<typename ParticleCell::ParticleType, ParticleCell, PairwiseFunctor, dataLayout, useNewton3,
                        /*bidirectional*/ true>
```

src/autopas/pairwiseFunctors/CellFunctor.h
```diff
@@ -177,7 +180,7 @@ template <class Particle, class ParticleCell, class ParticleFunctor, DataLayoutO
 template <bool newton3>
 void CellFunctor<Particle, ParticleCell, ParticleFunctor, DataLayout, useNewton3, bidirectional>::processCellAoS(
     ParticleCell &cell) {
-  if (cell.size() > _startSorting) {
+  if (cell.size() > _sortingThreshold and _useSorting) {
```
FG-TUM, line 183: please move the check for `_useSorting` to the front in all those `if`s. For two reasons: first, it is the easiest and most predictable to check, thus improving branching performance. Second, it makes the code much more intuitive to read because you immediately state what the if is about.

src/autopas/tuning/selectors/TraversalSelector.h
```diff
@@ -68,12 +68,14 @@ class TraversalSelector {
...
    * @param info
+   * @param useSorting Only for traversals that use the CellFunctor: If the CellFunctor should sort particles
    * @return Smartpointer to the traversal.
    */
   template <class PairwiseFunctor, DataLayoutOption::Value dataLayout, bool useNewton3>
   static std::unique_ptr<TraversalInterface> generateTraversal(TraversalOption traversalType,
                                                                PairwiseFunctor &pairwiseFunctor,
-                                                               const TraversalSelectorInfo &info);
+                                                               const TraversalSelectorInfo &info,
+                                                               const bool useSorting = true);
```
FG-TUM, line 78: I'd argue this should not be a constructor argument. Either remove it as suggested somewhere else or make it part of the `TraversalSelectorInfo` struct

FG-TUM (changes requested):
Reading these changes, most are simply passing the `useSorting` through some constructors. Since this is always `true` except in some tests how about we do not set it in the constructors but initialize it with `true`. For the unit tests a setter could then be provided. This has the advantage of not polluting every constructor with another (only mildly useful) argument and also it might be realizable to implement this setter somewhere higher up in the inheritance structure  so we only need it once.

src/autopas/tuning/selectors/TraversalSelector.h
thesamriel: Here, I would definitely put it into the `TraversalSelectorInfo` struct.

src/autopas/containers/linkedCells/traversals/LCC08CellHandler.h
```diff
@@ -149,15 +152,29 @@ inline void LCC08CellHandler<ParticleCell, PairwiseFunctor, dataLayout, useNewto
...
               std::array<double, 3>{std::max(zero, x - one), std::max(zero, y - one), std::max(zero, z - one)} *
               _cellLength;
           const auto distSquare = utils::ArrayMath::dot(distVec, distVec);
           if (distSquare <= interactionLengthSquare) {
-            _cellPairOffsets.push_back(std::make_tuple(cellOffsets[z], offset, utils::ArrayMath::normalize(distVec)));
+            std::array<double, 3> baseCellVec = utils::ArrayUtils::static_cast_copy_array<double>(
+                utils::ThreeDimensionalMapping::oneToThreeD(cellOffsets[z], cellsPerDimension));
+
+            // Calculate the sorting direction from the base cell to the other cell.
+            std::array<double, 3> sortingDir = offsetVec - baseCellVec;
```
thesamriel, line 171: Can just use `auto` for these cases.

thesamriel (commented):
I agree with Fabios suggestion. As long as we have some testing mechanism that assures that sorting works correctly, I don't see a reason why we would not use sorting.
Therefore, I would also remove sorting from the constructors and make it true by default.

src/autopas/tuning/selectors/TraversalSelector.h
dmartin13: How about this idea: In the `CellFunctor`, give the functions `processCellAoS()`, `processCellPairAoSN3()` and `processCellPairAoSNoN3()` an additional argument `useSorting` and remove the member variable `_useSorting` from the `CellFunctor`. Give the `TraversalInterface` a protected member `_useSorting` (default to true) and a corresponding setter. The traversals that use a `CellFunctor` can then use the value of the inherited `_useSorting` when calling the process functions from `CellFunctor`.
If I leave the member variable `_useSorting` in the `CellFunctor` and write a setter in the `CellFunctor`, I see the following problem: I would have to write a virtual setter for `_useSorting` (or getter for `CellFunctor`) in the `TraversalInterface`, which would have to be implemented in every traversal.
The problem is, that in the `TraversalComparison` tests I have only access to a `TraversalInterface`, and in `CellFunctor` I don’t have access to the traversal that uses this `CellFunctor`.

If I put `useSorting` in `TraversalSelectorInfo`, I have to leave `useSorting` as a constructor argument for all traversals that use a `CellFunctor`.

FG-TUM: Do all Traversals that use the `CellFunctor` implement the same Class? Maybe something like `CellBasedTraversal`? Then you could put the getter/setter there and in the Test check if the Traversal object can be cast to that type?

dmartin13: Then I would have to use `CellPairTraversal` to also cover the Octree traversals and DirectSum. However, there would then also be some traversals that implement this class but do not use `CellFunctor`.

Maybe a new interface `CellFunctorTraversal` or something like that?

FG-TUM: Can you please quickly put together the list of CellPairTraversals that do not use the CellFunctor and the list of all traversals that do use it? I don't have the full overview :sweat_smile:

dmartin13: Here is a list of all traversals that implement `CellPairTraversal`
| Use CellFunctor    | Do not use CellFunctor |
|--------------------|------------------------|
| ds_sequential      | lc_c04_combined_SoA    |
| lc_c01             | vcl_c06                |
| lc_c04_HCP         | vcl_sliced_balanced    |
| lc_c04             | vcl_sliced_c02         |
| lc_c08             | vcl_sliced             |
| lc_c18             | vlc_c01                |
| lc_sliced_balanced | vlc_c18                |
| lc_sliced          | vlc_sliced             |
| lc_sliced_c02      | vlc_sliced_balanced    |
| lc_c01_combined_SoA | vlc_sliced_c02         |
| ot_01              | vlp_c01                |
| ot_18              | vlp_c18                |
|                    | vlp_sliced             |
|                    | vlp_sliced_balanced    |
|                    | vlp_sliced_c02         |

thesamriel: I would still add the functionality to `CellPairTraversal`.  Imo, it's not really a problem that it just does nothing in case of vcl, vlc, vlp traversals.
I would definitely try to not make the traversal inheritance structure more complex just for the sorting parameter. (Because it will already become more convoluted with upcoming 3-body traversals :D)

src/autopas/containers/linkedCells/traversals/LCC08CellHandler.h
dmartin13: done in [8081014](https://github.com/AutoPas/AutoPas/pull/619/commits/80810142d34cf7872346976089a90e1753180bb7)

src/autopas/tuning/selectors/TraversalSelector.h
dmartin13: done in [8081014](https://github.com/AutoPas/AutoPas/pull/619/commits/80810142d34cf7872346976089a90e1753180bb7)

src/autopas/pairwiseFunctors/CellFunctor.h
dmartin13: done in [8081014](https://github.com/AutoPas/AutoPas/pull/619/commits/80810142d34cf7872346976089a90e1753180bb7)

src/autopas/containers/linkedCells/traversals/LCC18Traversal.h
dmartin13: done in [8081014](https://github.com/AutoPas/AutoPas/pull/619/commits/80810142d34cf7872346976089a90e1753180bb7)

FG-TUM (approved):

src/autopas/pairwiseFunctors/CellFunctor.h
```diff
@@ -97,13 +105,24 @@ class CellFunctor {
...
+   * This value is used to switch on and off the sorting functionality of the CellFunctor. Sorting is enabled by default
+   */
+  bool _useSorting{true};
+
   /**
    * Min. number of particles to start sorting.
-   * @todo Currently, this is disabled because of https://github.com/AutoPas/AutoPas/issues/418
    */
-  constexpr static unsigned long _startSorting = std::numeric_limits<unsigned long>::max();
+  constexpr static unsigned long _sortingThreshold = 8;
```
FG-TUM, line 116: Add a link to this PR to show how we came up with the number
https://github.com/AutoPas/AutoPas/pull/619

FG-TUM (commented):
Just one minor doc thingy :innocent:

src/autopas/pairwiseFunctors/CellFunctor.h
dmartin13: doen in [9b3ad7d](https://github.com/AutoPas/AutoPas/pull/619/commits/9b3ad7dcc62319f6afd35d92288ce888e38eac12)

thesamriel (approved):

PR #627, opened 2021-07-21
Faster Particle Iteration in Octree

### LINKED ISSUE #625: Octree Iterator
**Documentation**

The octree is using a `SingleCellIterator`. This iterator requires the iterated object to have an `at` method implemented, which can access every particle inside the iterated object using an unique index. Since the octree is a tree data structure that does not store the particles inside a linear array, accessing particles by an index is a hard task. At the moment, the current code copies all particles out of the octree into a `std::vector<Particle *>` and returns a reference to the particle at the requested index. This happens (_including_ the copying) _every time_ `at` is called, which is extremely inefficient. (The `std::vector` with _all_ particles gets created and destroyed anytime someone calls `at`.)

https://github.com/AutoPas/AutoPas/blob/e2c016e9c8ec6bcfbc2033d884706f0f8bf1ebfe/src/autopas/containers/octree/OctreeNodeWrapper.h#L220-L222

**Solution**

We already discussed internally that iterators will move out of AutoPas eventually. Therefore, I did not introduce a new additional iterator that is capable of iterating the octree fast. The iterator would only have to go away in the next months.

Using the new, callback-based API, this issue can be solved trivially: The octree would only need to pass the callback down into its leaves. These can then apply the callback with their particles. Thereby, iterating the particles of an octree would become very cheap.

I tried to temporarily add a quick fix: Only load the vector `ps` that contains all `Particle *` one time (if the given `index` is zero or the list is empty). The following snippet shows how this could be achieved. `ps` would have been marked `static` in order to preserve the loaded particle pointers.

https://github.com/AutoPas/AutoPas/blob/e2c016e9c8ec6bcfbc2033d884706f0f8bf1ebfe/src/autopas/containers/octree/OctreeNodeWrapper.h#L212-L214

This hack does not work if there are many threads, since the threads would come and access the local variable at the same time, which lead to race conditions. (Unfortunately, `std::vector` is not thread-safe.) As a result, there has to be a locking mechanism that prevents threads from accessing the `ps` at the same time. 

AutoPas has this functionality already builtin in `WrapOpenMP.h` with the `AutoPasLock`. Due to architectural reasons, at least one of the `at` functions has to be marked as `const`. Therefore, it cannot modify local field, like an `AutoPasLock` and a thread-safe implementation cannot be provided.

However, in the non-const version of the `at` function, a lock could be used. It would be interesting to evaluate if it already helps if the non-const version had a lock.

**Additional context**

The issue refers to pull request #610.

neftlon:
# Description

As described in #625 in more detail, iterating the particles of the octree is very slow at the moment. This branch contains a "hack" that tries to fix the speed issues that arise from copying the particles at each access. Since the method is marked `const`, one cannot add a trivial member variable that contains an instance of a lock. However, marking a member variable as `mutable` makes it available even inside `const` methods. 

https://github.com/AutoPas/AutoPas/blob/c3d5511946ae707fe8b0c7155e2943295097de73/src/autopas/containers/octree/OctreeNodeWrapper.h#L197-L198

Now, the `getFromReloadingIterator` method can use the `mutable` field and is able to provide thread-safe particle caching for later access.

## Related Pull Requests

- Relates to the code of #610 

## Resolved Issues

- [ ] fixes #625

# How Has This Been Tested?

- [ ] Run the test suite in a parallel context

### DIFF
--- a/src/autopas/containers/octree/OctreeNodeWrapper.h
+++ b/src/autopas/containers/octree/OctreeNodeWrapper.h
@@ -183,11 +183,18 @@ class OctreeNodeWrapper : public ParticleCell<Particle> {
     // TODO(johannes): This needs severe improvement. If we just copy out all particles, the implementation becomes
     //  unsafe for threading. We need a way to iterate the octree using a better traversal idea.
     //  Referenced in https://github.com/AutoPas/AutoPas/issues/625
-    std::vector<Particle *> ps;
-    _pointer->appendAllParticles(ps);
+
+    lock.lock();
+    if(ps.empty() || index == 0) {
+      ps.clear();
+      _pointer->appendAllParticles(ps);
+    }
+    lock.unlock();
     return *ps[index];
   }
 
   std::unique_ptr<OctreeNodeInterface<Particle>> _pointer;
+  mutable std::vector<Particle *> ps;
+  mutable AutoPasLock lock;
 };
 }  // namespace autopas


FG-TUM:
@neftlon needs manual conflict resolving

FG-TUM (approved):

PR #636, opened 2021-08-02
Updates googletest to v1.11.0

SteffenSeckler:
# Description

Updates bundled googletest from v1.10.0 to v1.11.0 ([changelog](https://github.com/google/googletest/releases/tag/release-1.11.0))


cmake/modules/autopas_gtest.cmake
```diff
@@ -1,27 +1,27 @@
...
+find_package(Threads REQUIRED)
+
+# Enable FetchContent CMake module
+include(FetchContent)
+
+# Build GoogleTest and make the cmake targets available
+FetchContent_Declare(
+    autopas_gtest
+    URL ${AUTOPAS_SOURCE_DIR}/libs/googletest-1.11.0.zip
+    URL_HASH MD5=52943a59cefce0ae0491d4d2412c120b
```
FG-TUM, line 11: cool. Any reason for the upgrade?

tests/testAutopas/tests/selectors/ContainerSelectorTest.cpp
```diff
@@ -1,31 +1,22 @@
...
- * @date 22.06.18
- */
-
-#include "ContainerSelectorTest.h"
-
-using ::testing::Combine;
-using ::testing::UnorderedElementsAreArray;
-using ::testing::ValuesIn;
-
-// must be TEST_F because the logger which is called in the LC constructor is part of the fixture
```
FG-TUM, line 13: why throw out the comment?

tests/testAutopas/tests/selectors/ContainerSelectorTest.h
```diff
@@ -1,34 +1,25 @@
...
+#include <gtest/gtest.h>
+
+#include "AutoPasTestBase.h"
+#include "autopas/selectors/ContainerSelector.h"
+#include "testingHelpers/commonTypedefs.h"
+
+class ContainerSelectorTest : public AutoPasTestBase {
+ public:
+  ContainerSelectorTest() = default;
+  ~ContainerSelectorTest() override = default;
```
FG-TUM, line 18: so we even need those?

tests/testAutopas/tests/selectors/ContainerSelectorTest.cpp
```diff
@@ -1,31 +1,22 @@
...
-  for (auto containerOp : autopas::ContainerOption::getAllOptions()) {
-    containerSelector.selectContainer(containerOp, containerInfo);
-
-    EXPECT_EQ(containerOp, containerSelector.getCurrentContainer()->getContainerType());
-  }
-}
-
-INSTANTIATE_TEST_SUITE_P(Generated, ContainerSelectorTest,
-                         ::testing::ValuesIn(autopas::ContainerOption::getAllOptions()),
-                         ContainerSelectorTest::oneParamToString());
```
FG-TUM, line 31: while I agree that the test in that form was useless we probably want to add some particles to check that the conversion actually does what it is intended to do.

FG-TUM (changes requested):
1. Why is every single line in the files you touch marked as changed? Are there some Windows / Linux linebreak shenanigans?
2. We should think about purging the binary blobs. However, that would rewrite the history. So maybe check what size impact they have on the `.git` folder so far.

tests/testAutopas/tests/selectors/ContainerSelectorTest.cpp
SteffenSeckler: Because it's the same everywhere and we use members from the ContainerSelectorTest class.

tests/testAutopas/tests/selectors/ContainerSelectorTest.cpp
SteffenSeckler: noted in #637

cmake/modules/autopas_gtest.cmake
SteffenSeckler: * additional internal tests that check the correctness of definition of tests.
* no more stupid warnings when using cmake >= 3.19 (1.10 allowed the usage of deprecated cmake versions, i.e. < 2.8.12)

FG-TUM (approved):

tests/testAutopas/tests/selectors/ContainerSelectorTest.h
FG-TUM: still relevant

tests/testAutopas/tests/selectors/ContainerSelectorTest.h
```diff
@@ -12,20 +12,11 @@
 #include "autopas/selectors/ContainerSelector.h"
 #include "testingHelpers/commonTypedefs.h"
 
-class ContainerSelectorTest : public AutoPasTestBase, public ::testing::WithParamInterface<autopas::ContainerOption> {
+class ContainerSelectorTest : public AutoPasTestBase {
  public:
   ContainerSelectorTest() = default;
   ~ContainerSelectorTest() override = default;
```
SteffenSeckler, line 18:
```suggestion
```

PR #639, opened 2021-08-05
Cleanup of MPI linkings

FG-TUM:
# Description

Cleanup of MPI linkings:
 - [x] remove duplicate find_package(MPI) blocks
 - [x] more expressive log messages
 - [x] only add definitions to targets not globally

## Related Pull Requests

- merges into #597. If the former is merged into master before this one was merged (which I would be ok with) we need to point this one to master.

## ~Resolved Issues~

# How Has This Been Tested?

NOT YET! @J0llyver please test this and tick the box when you have.

- [ ] tested by @J0llyver

### DIFF
--- a/cmake/modules/autopas_mpi.cmake
+++ b/cmake/modules/autopas_mpi.cmake
@@ -1,11 +1,18 @@
 option(AUTOPAS_INTERNODE_TUNING "Activates distributed tuning in MPI parallel simulations" OFF)
 
+# messages first
 if (AUTOPAS_INTERNODE_TUNING)
-    message(STATUS "MPI enabled.")
-    find_package(MPI REQUIRED)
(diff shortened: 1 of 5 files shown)


J0llyver:
I cannot build this branch. 
When running cmake I get following error multiple times:
```
CMake Error at src/autopas/CMakeLists.txt:17 (target_link_libraries):
  Error evaluating generator expression:

    $<OR$<BOOL:OFF>,$<BOOL:>>

  Expression did not evaluate to a known generator expression
```
Is this because I am using a different version of cmake?

FG-TUM:
> I cannot build this branch.

Should be fixed now? -> 945b9f7

J0llyver (approved):

PR #640, opened 2021-08-06
Generic Particle::get() const

FG-TUM:
# Description

- Make generic `Particle::get()` `const`
- ~needs `const` cast on this ptr as we use it non-const~
- Introduce non-const getter for this

Although the second point seems a bit hacky the overall result seems to be more const-correct than before.

## ~Resolved Issues~

# How Has This Been Tested?

### DIFF
--- a/src/autopas/molecularDynamics/MoleculeLJ.h
+++ b/src/autopas/molecularDynamics/MoleculeLJ.h
@@ -58,9 +58,9 @@ class MoleculeLJ final : public Particle {
    * @note The value of owned is return as floating point number (true = 1.0, false = 0.0).
    */
   template <AttributeNames attribute>
-  constexpr typename std::tuple_element<attribute, SoAArraysType>::type::value_type get() {
+  constexpr typename std::tuple_element<attribute, SoAArraysType>::type::value_type get() const {
     if constexpr (attribute == AttributeNames::ptr) {
-      return this;
+      return const_cast<typename std::tuple_element<attribute, SoAArraysType>::type::value_type>(this);
     } else if constexpr (attribute == AttributeNames::id) {
       return getID();
     } else if constexpr (attribute == AttributeNames::posX) {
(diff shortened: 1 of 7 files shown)


SteffenSeckler (changes requested):
As long as a Particle* is returned, this is not correct.
The only "correct" way of doing this would be having a specialized get function for the ptr argument or returning a `const Particle*` which will break the verlet list implementation to some intend.
so I am strongly against this change.

FG-TUM:
> As long as a Particle* is returned, this is not correct.
> The only "correct" way of doing this would be having a specialized get function for the ptr argument or returning a `const Particle*` which will break the verlet list implementation to some intend.
> so I am strongly against this change.

I didn't say this is "correct" just MORE correct ;) The problem with the current master is, that you can not use the getter in `const` functions at all, which is not reasonable.

Here is a more sophisticated version that should actually be `const` correct and working.

src/autopas/molecularDynamics/MoleculeLJ.h
```diff
@@ -51,17 +51,24 @@ class MoleculeLJ final : public Particle {
...
 
+  /**
+   * Non-const getter for the pointer of this object.
+   * @tparam attribute Attribute name.
+   * @return this.
+   */
+  template <AttributeNames attribute, std::enable_if_t<attribute == AttributeNames::ptr, bool> = true>
+  constexpr typename std::tuple_element<attribute, SoAArraysType>::type::value_type get() {
+    return this;
+  }
```
SteffenSeckler, line 62: Potentially, one could also add a method that returns a const ptr, don't know whether that could be needed anywhere...

SteffenSeckler (approved):

src/autopas/molecularDynamics/MoleculeLJ.h
FG-TUM: sure but i would only add this if needed

PR #642, opened 2021-08-13
Linked Cells Interface

### LINKED ISSUE #460: Implement Linked-Cells like interface
The Verlet-like interface produces a couple of problems:
- Particles are sometimes inside of a container, even though you wouldn't expect it. (You have to iterate a farther than a domain of a particle is long)
- The position of particles can only be made conform to periodic boundary conditions when the container is rebuilt.
- Particles can only be added and deleted in steps where the container is rebuilt
- Tuning and Rebuilding across MPI ranks are not fixed but have to be synchronized. So flexible tuning is not possible.

A linked-cells like interface:
- Particles are stored inside of the AutoPas object they belong to.

This has the following advantages:
- Particles can be added and deleted at every step.
- Particles are conforming to boundaries.
- Verlet rebuild frequency could be made variable.
- Flexible tuning.
- No synchronization across MPI-Ranks needed.

**Implementation Details**
The idea is to still maintain the structure of the container, i.e., the particles are neither added nor deleted from the internal container.
As particles are still added and deleted to the AutoPas object (not the container) the following is needed:
- [x] Add an **additional vector**(or other structure) of particles:
    - This vector holds particles that should not be added to the container, as the container would be made invalid if they are added.
    - Theoretically, one could check whether it is safe to add a particle to the internal container and add it if it is. If it is not safe, then one adds the particle to the additional vector.
    - Possibly two vectors instead of one are better (one for owned, one for halo particles)
    - [x] The additional vector needs to participate in the force calculation!
    - [x] Iterators (or for_each #434) have to consider the additional vector.
- [x] Particles need to be "removable" from a container:
    - This can be done by marking particles as a dummy
         - either by moving the position far enough away from the domain or marking them via a flag (e.g. make the bool `owned` an enum)
- [x] Special care has to be taken if a particle that has previously been halo is added as an owned particle, as interactions with another particle could appear twice (once with the OLD halo and once with the newly added particle)! For this purpose, all halo particles should be marked as dummy after an updateContainer() call. This would make it consistent with Linked Cells.

**Possible problems**
Performance might be slower:
- <strike>some optimizations in Linked-Cells cannot be made (potentially no cell-wise owned state), this happens because particles can suddenly be dummies.</strike> This is no longer valid.
- Leaving particles have to be communicated at every step. This can no longer be skipped.

Leave as is.

**Additional context**

SteffenSeckler:
# Description

- Particle Buffers:
  - [x] Adds particle and halo particle buffers in LogicHandler.h
  - [x] Interactions between particle buffers with particles stored in the container.
  - [x] Particle buffers are considered when updateContainer is called, i.e.
    - [x] halo particle buffer is cleared.
    - [x] particle buffer is added to container. This happens only if keepListsValid == false.
    - [x] leaving particles are removed from buffer.
  - [x] particle buffers are considered in the iterators.
- updateContainer changes:
  - [x] updateContainer now always returns leaving particles and removes halo particles.
  - [x] To keep neighbor lists valid, the new parameter keepListsValid was added to the updateContainer call of the actual containers. If it is specified no particles are allowed to be deleted through the call. Instead, particles are only allowed to be marked as dummies.
- [x] Update README.md
- Interface changes (left undone to prevent merge conflicts):
  - [x] Remove forced from updateContainer -- it's no longer needed!
  - [x] Change return type of updateContainer from `std::pair<std::vector<Particle>, bool>` to `std::vector<Particle>` -- it now always returns true, so we can also just ommit it.
  - [x] Renames addOrUpdateHaloParticle to addHaloParticle.

## Todo
- [x] Check that particles are actually added to the container in the first iteration!
- [x] LogicHandler.h, l.~245 (iteratePairwise call): <strike> Currently, we assume that particles are not allowed to be moved between updateContainer and iteratePairwise. Is this fine? + check whether we ever need to add particles in the iteratePairwise call, or whether that only happens if particles are added between updateContainer and iteratePairwise calls. </strike> this only happens if we add particles between updateContainer and iteratePairwise, so we can also just ignore those few particles and everything is well defined.

## Possible further optimizations
- Replace particle buffers with actual cells. -> #652 
  - Allows reusing the CellFunctors.
- Do not use unoptimized LeavingParticleCollector. Instead, use container-specific algorithms. -> #653 

## Related Pull Requests

- original Verlet list interface: #278 

## Resolved Issues

- [x] fixes #460 

# How Has This Been Tested?

- Test adding additional particle vector to iterator:
  - [x] without particles in container
  - [x] with particles in container


src/autopas/LogicHandler.h
```diff
@@ -218,14 +184,28 @@ class LogicHandler {
...
+      for (auto &&p : _additionalParticleVector) {
+        if (not p.isDummy()) {
+          _autoTuner.getContainer()->template addParticle(p);
+        }
+      }
+      for (auto &&p : _additionalHaloParticleVector) {
+        if (not p.isDummy()) {
+          _autoTuner.getContainer()->template addHaloParticle(p);
+        }
+      }
```
SteffenSeckler, line 199: empty the vectors!

SteffenSeckler:
@FG-TUM You can already take a look at this PR.
Mainly tests are missing. It, however, appears to be mostly working. So a general opinion would be good :)

Besides that, I'm hunting down a clang bug that prevents compilation if OpenMP is disabled, so I assume that's some uninitialized value or so. (was failing with clang 12.0.1 and also the newest apt version).

SteffenSeckler:
@J0llyver once this PR is merged, you can ignore the skin in all of you halo exchange routines ;)

FG-TUM (approved):

PR #626, opened 2021-07-19
Ma lg kokkos prep replace iterators

LG-TUM:
# Description
The goal of this Pull Request is to introduce a new api endpoint for AutoPas.
These endpoints are called forEach and forEachInRegion and should in the medium term replace iterators.
This is a preparation step to integrating Kokkos into Autopas.

## Related PRs
none

## Related Issues
none

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes.

- [x] SingleCellForEach Testsuite (test forEach and forEachInRegion for FullParticleCell and ReferenceParticleCell)
- [x] ContainerForEach Testsuite (test forEach and forEachInRegion for all possible containers)
- [x] use code coverage to check that all possible variations are covered by above tests
- [x] md-flexible testsuite

PR #648, opened 2021-08-27
Remove LogicHandler.h and AutoTuner.h imports from AutoPasDecl.h in replace-iterators

HobbyProgrammer:
# Description
Here is one quick and easy possibility of removing the newly added dependencies in AutoPasDecl.h. 

It is achieved trough adding a private `AutoPas::getContainer()` method, implemented in AutoPasImpl.h, that allows all forEach*() methods to be directly implemented in `AutoPas` instead of calling the `LogicHandler`.

## Related Pull Requests

- #626 

## Underlying Issue
The underlying issue here is that the container is only accessible in `AutoPas` through the `LogicHandler` which accesses it through the `AutoTuner` which has dependencies to all traversals. 

## Alternative
Reorganize `AutoPas`, `LogicHandler`, and `AutoTuner` such that the `ContainerSelector` is not only accessible through the `AutoTuner`. Since the `LogicHandler` "takes care of the containers" (Documentation of `LogicHandler`), it might make sense to store the `ContainerSelector` there. Almost all calls to the `AutoTuner` inside of `LogicHandler` are to `AutoTuner::getContainer()`, so if we can achieve that `LogicHandler` does not have a dependency to `AutoTuner` anymore, which seems possible, the whole logic of handling the containers is split from the dependencies to the traversals. I think this would be good. The `AutoPas` class would have to handle the `AutoTuner` itself, inside the AutoPasImpl.h file.

# How Has This Been Tested?

Includes have been removed, autopas and md-flexible still compile.


SteffenSeckler:
is currently missing the include of `utils/StaticContainerSelector.h`

PR #651, opened 2021-09-10
Md flex/all load balancer

J0llyver:
# Description
Implemented the ALL load balancer as an alternative load balancing method

# How Has This Been Tested?
This has been tested by running md-flexibles falling drop scenario.
Tests on Cluster will follow.

### DIFF
--- a/examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
+++ b/examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
@@ -16,11 +16,8 @@
 #include "autopas/utils/ArrayUtils.h"
 #include "src/ParticleSerializationTools.h"
 
-RegularGridDecomposition::RegularGridDecomposition(const std::array<double, 3> &globalBoxMin,
-                                                   const std::array<double, 3> &globalBoxMax,
-                                                   const std::array<bool, 3> &subdivideDimension,
-                                                   const double &cutoffWidth, const double &skinWidth)
-    : _cutoffWidth(cutoffWidth), _skinWidth(skinWidth) {
+RegularGridDecomposition::RegularGridDecomposition(const MDFlexConfig &configuration)
+    : _cutoffWidth(configuration.cutoff.value), _skinWidth(configuration.verletSkinRadius.value) {
 #if defined(AUTOPAS_INCLUDE_MPI)
   _mpiIsEnabled = true;
 #else
@@ -39,13 +36,13 @@ RegularGridDecomposition::RegularGridDecomposition(const std::array<double, 3> &
     std::cout << "MPI will not be used." << std::endl;
   }
 
-  DomainTools::generateDecomposition(_subdomainCount, subdivideDimension, _decomposition);
+  DomainTools::generateDecomposition(_subdomainCount, configuration.subdivideDimension.value, _decomposition);
 
   initializeMPICommunicator();
 
   initializeLocalDomain();
 
-  initializeGlobalBox(globalBoxMin, globalBoxMax);
+  initializeGlobalBox(configuration.boxMin.value, configuration.boxMax.value);
 
   initializeLocalBox();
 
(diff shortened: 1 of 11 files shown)


examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
```diff
@@ -454,3 +406,95 @@ void RegularGridDecomposition::categorizeParticlesIntoLeftAndRightNeighbour(
...
+
+  std::vector<ALL::Point<double>> updatedVertices = _allLoadBalancer->getVertices();
+
+  for (int i = 0; i < 3; ++i) {
+    _localBoxMin[i] = updatedVertices[0][i];
+    _localBoxMax[i] = updatedVertices[1][i];
+  }
+}
+
+void RegularGridDecomposition::deleteAllLoadBalancer() { delete _allLoadBalancer; }
```
SteffenSeckler, line 499: please use RegularGridDecomposition's destructor for this kind of cleanup ;)

SteffenSeckler: or better: use std::unique_ptr

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.h
```diff
@@ -217,6 +223,20 @@ class RegularGridDecomposition final : public DomainDecomposition {
...
+   */
+  LoadBalancerOption _loadBalancer;
+
+#if defined(MD_FLEXIBLE_INCLUDE_ALL)
+  /**
+   * The ALL load balancer used for diffuse load balancing
+   * We cannot use a shared pointer here, because whenn the load balancer is deleted, it calls MPI_Comm_free after
+   * we call MPI_Finalize().
+   */
+  ALL::ALL<double, double> *_allLoadBalancer;
```
SteffenSeckler, line 237:
```suggestion
  std::unique_ptr<ALL::ALL<double, double>> _allLoadBalancer;
```

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
```diff
@@ -36,86 +30,44 @@ RegularGridDecomposition::RegularGridDecomposition(const std::array<double, 3> &
...
+  initializeGlobalBox(configuration.boxMin.value, configuration.boxMax.value);
 
   initializeLocalBox();
 
   initializeNeighbourIds();
+
+  _loadBalancer = configuration.loadBalancer.value;
+
+#if defined(MD_FLEXIBLE_INCLUDE_ALL)
+  _allLoadBalancer = new ALL::ALL<double, double>(ALL::STAGGERED, _dimensionCount, 0);
```
SteffenSeckler, line 48:
```suggestion
  _allLoadBalancer = std::make_unique<ALL::ALL<double, double>>(ALL::STAGGERED, _dimensionCount, 0);
```

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
J0llyver: I tried to use std::unique_ptr and to use the RegularGridDecompositions Destructor, but it did not work. When destroyed, the ALL load balancer calls "MPI_Comm_free". If I use the RegularGridDecomposition constructor, all will be deleted when the progam exits the scope of the main function, which is after the "MPI_Finalize" has been called. This results in a build error. As far as I know, a unique pointer cannot be deleted manually, but will be deleted when RegularGridDecomposition is deleted. Again this happens after calling MPI_Finalize which leaves us with the same problem as before.

SteffenSeckler: many ways to circumvent this.
1. change the scoping of the class containing the unique_ptr, s.t., it is destroyed before the MPI_Finalize call. In this case, the RegularDecomposition class used in the main.cpp file.
2. you can always "delete" the pointer stored by a unique_ptr by calling `.reset(nullptr)`.

In this case, option 1 is better! You could do that by moving everything that isn't MPI_Init/Finalize to a different function or by surrounding it with curly brackets.

J0llyver: I ended up using shared_ptr because RegularGridDecomposition is shared between the main function and the Simulation class.

PR #612, opened 2021-06-18
Mpi tuning for inhomogeneous scenarios

JohannesKroll:
# Description

Improved MPIParallelizedStrategy to work with inhomogeneous scenarios.
Ranks with similar scenarios are now distributed in buckets. Each bucket has its own search space.
Each bucket has a optimal configuration.

# TODOs

- [x] need to improve bucket distribution
- [x] also use max_density for distribution
- [x] thew relative difference with which ranks are assign into different buckets should be configurable with a good default value

# Tests

basic functionality:

- [x] successfully compiles
- [x] md-flexilbe can use strategy correctly

-------------------------------

Further additions by @FG-TUM :
- [x] MPI parallel tests in the CI

PR #654, opened 2021-09-17
Mpi tuning for inhomogeneous scenario merge master

FG-TUM:
# Description

Merge of master into Branch `mpi_tuning_for_inhomogeneous_scenario` from #612 

- [ ] Think about the calculation of homogeneity for the global simulation. Currently, every rank calculates its own and then the global homogeneity is calculated as the square root of the squared sums of the homogeneities.
This means that the grids that were used to calculate the local homogeneities can differ and the domain sizes will differ as well. Hence we get the same global homogeneity result in the following two cases:
  * a small domain with homogeneity X and a large one with homogeneity Y
  * a small domain with homogeneity Y and a large one with homogeneity X

## Related Pull Requests

- #612

## ~Resolved Issues~

# How Has This Been Tested?

- [ ] Looking at it sharply


examples/md-flexible/src/Simulation.h
```diff
@@ -1,189 +1,303 @@
...
-  constexpr static bool _shifting = true;
+  std::string timerToString(const std::string &name, long timeNS, size_t numberWidth = 0ul, long maxTime = 0ul);
+
   /**
-   * Use the variant of the LJFunctor that supports mixing of particle types.
+   * Calculate the homogeneity of the scenario by using the standard deviation.
+   * @return homogeneity
    */
-  constexpr static bool _mixing = true;
+  double calculateHomogeneity() const;
```
J0llyver, line 223: This function as been moved to autopas::utils::... so it can be removed here, right?

J0llyver (commented):
As far as I understand, because the calculateHomogeneity function has been moved to the autopas::utils, we can just directly implement the changes in the new function and remove the old. The calculation of the global Homogeineity over multiple domains is faulty anyway. I would just create a backup of the parallel function and use it as reference to implement the parallel calculation in the new function.

FG-TUM:
I will re-do this in #612

PR #655, opened 2021-09-19
Removed content name in vtk writer and limited output to single rank

J0llyver:
# Description
Removed unecessary content name in simulation record files.
Limited output of some simulation data to a single rank.

# How Has This Been Tested?
This has been tested by generating a new checkpoint and afterwards loading it.
Also the mdFlexTests have been executed.

### DIFF
--- a/examples/md-flexible/src/ParallelVtkWriter.cpp
+++ b/examples/md-flexible/src/ParallelVtkWriter.cpp
@@ -253,14 +253,14 @@ void ParallelVtkWriter::createPvtuFile(const int &currentIteration) {
   timestepFile << "    </PPointData>\n";
   timestepFile << "    <PCellData/>\n";
   timestepFile << "    <PPoints>\n";
-  timestepFile << "      <PDataArray Name=\"points\" NumberOfComponents=\"3\" format=\"ascii\" type=\"Float32\"/>\n";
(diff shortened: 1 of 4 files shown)


examples/md-flexible/src/ParallelVtkWriter.cpp
```diff
@@ -231,7 +231,7 @@ void ParallelVtkWriter::tryCreateSessionAndDataFolders(const std::string &name,
 
 void ParallelVtkWriter::createPvtuFile(const int &currentIteration) {
   std::ostringstream filename;
-  filename << _sessionFolderPath << _sessionName << "_particles_" << std::setfill('0')
+  filename << _sessionFolderPath << _sessionName << "_" << std::setfill('0')
```
FG-TUM, line 234: why do you want to remove this? Seems like a useful part of the name.
(Also applies to the other locations where you removed similar things)

FG-TUM (approved):

examples/md-flexible/src/ParallelVtkWriter.cpp
J0llyver: I think it is redundant and introduces additional maintainance effort. 
The file type already provides a good hint about the content of the file, which is why I consider this information redundant.
If for some reason someone wants to update the information to say "particles_and_other_stuff" this not only the ParallelVtkWriter requires an update but also the MDFlexConfig for loading checkpoints properly. I fell into that trap :)

PR #659, opened 2021-11-01
C08 traversal for VLCCellPairNeighborList

tirgendetwas:
# Description
C08 traversal for VLCCellPairNeighborList. The pairwise neighbor list structure stores a particle's interaction partners according to the cell they are in. This allows the traversal to pick two interacting cells and have access to each particle's interaction partners from that particular neighbor cell. More specifically, this lets us implement the C08 traversal which includes an interaction between two non-base cells.
The implementation includes the main component VLCCellPairC08Traversal, the VLCCellPairC08Handler which uses the LCC08CellHandler to compute the indices of the interacting cells, and the VLCCellPairTraversalInterface. The new traversal interface was necessary in order to distinguish traversals the can be used with both the all-cells and the cell-pair neighbor list (VLCTraversalInterface) from traversals than can only be used with the cell-pair neighbor list (VLCCellPairTraversalInterface).

# How Has This Been Tested?
auto-generated unit tests


FG-TUM (approved):
nice! Can be merged if the CI agrees

PR #661, opened 2021-11-03
Fix neighbor-list-length performance problem

HobbyProgrammer:
# Description

Ever wondered about the extraordinary bad performance of the `neighbor-list-length` load estimator in large scenarios? I did!  It seems like the whole neighbor list is, probably accidentally, copied each load estimation…

Here is the one character fix for this.

# How Has This Been Tested?
- Tried out once locally. It did improve the performance considerably.
- Trust in our unit tests


FG-TUM (approved):
Nice catch

PR #666, opened 2021-11-19
Md flex/post thesis improvements

J0llyver:
# Description
This pullrequest only contains the implementation for the configurable load balancing interval.

# How Has This Been Tested?
This has been tested manually and by running MD-Flexible's test with and without MPI.


examples/md-flexible/src/configuration/CLIParser.cpp
```diff
@@ -544,15 +544,13 @@ MDFlexParser::exitCodes MDFlexParser::CLIParser::parseInput(int argc, char **arg
       }
       case decltype(config.loadBalancer)::getoptChar: {
         auto parsedOptions = LoadBalancerOption::parseOptions(strArg);
-        if (parsedOptions.size() != 1) {
-          cerr << "Pass exactly one load balancer option." << endl
-               << "Passed: " << strArg << endl
-               << "Parsed: " << autopas::utils::ArrayUtils::to_string(parsedOptions) << endl;
-          displayHelp = true;
-        }
```
FG-TUM, line 552: Was this removed on purpose? Seems like a useful check.

examples/md-flexible/src/configuration/MDFlexConfig.h
```diff
@@ -340,6 +340,12 @@ class MDFlexConfig {
   MDFlexOption<std::array<double, 3>, 0> boxMax{
       {1, 1, 1}, "box-max", true, "Upper back right corner of the simulation box."};
 
+  /**
+   * loadBalancingInterval
+   */
+  MDFlexOption<unsigned int, 0> loadBalancingInterval{
```
FG-TUM, line 346:
```suggestion
  MDFlexOption<unsigned int, __LINE__> loadBalancingInterval{
```
This is super important, otherwise, you break the CLI parser. The number here is the identifier token for `getOpt`.

FG-TUM (changes requested):

examples/md-flexible/src/configuration/CLIParser.cpp
J0llyver: I will reinclude it.

PR #669, opened 2021-12-01
AutoTuner runtime estimation: Differentiate between iterations that rebuild neighbor lists and those that don't

HobbyProgrammer:
# Description

Please include a summary of the changes.

- [x] Track samples from rebuild iterations and regular independently
- [x] For the tuning estimate a weighted average from the samples so that we normalize to the full rebuild frequency. 

## ~Resolved Issues~

# How Has This Been Tested?
- One new test so far (not perfect, though, takes multiple seconds)

# TODO
- [x] Write this pull request
- [x] Discuss how to properly test this
- [x] Look at tuning data logger and check if my changes make sense


FG-TUM (approved):

FG-TUM:
Note to future self and @SamNewcome:

If `AutoTunerTest::testBuildNotBuildTimeEstimation` turns out to be unreliable / non deterministic revisit the sleep times. Either increase them or build some mechanism so that the tuner doesn't actually have to run something but can be fed information directly.

See d401e67 for the latest changes to the times.

PR #671, opened 2021-12-16
feature: Energy tuning

astrobeastie:
# Description

Adds a new option to AutoPas to select the configuration with the least energy usage instead of the fastest one.

# How Has This Been Tested?

Performed multiple benchmarks on the LRZ Linux Cluster using the Spinodal Decomposition example.
Mostly energy consumption is proportional to time, however, the proportionality constant may differ for 
various containers.


src/autopas/utils/RaplMeter.cpp
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  int fd = syscall(__NR_perf_event_open, &attr, -1 /*PID*/, 0 /*CPU*/, -1 /*GROUP FD*/, 0 /*FLAGS*/);
+  if (fd < 0) {
+    if (errno == EACCES) {
+      throw ExceptionHandler::AutoPasException("Failed to open perf event: Permission denied");
+    }
+    throw ExceptionHandler::AutoPasException("Failed to open perf event");
+  }
+  // TODO: Error handling
```
FG-TUM, line 38: open TODO :)

src/autopas/utils/RaplMeter.cpp
```diff
@@ -0,0 +1,167 @@
...
+  return fd;
+}
+
+void RaplMeter::init() {
+  FILE *fff = fopen("/sys/bus/event_source/devices/power/type", "r");
+  if (fff == NULL) {
```
FG-TUM, line 44:
```suggestion
  if (FILE *fff = fopen("/sys/bus/event_source/devices/power/type", "r"); not fff) {
```
1. The declaration can be scoped to the if (not 100% sure about syntax). This would make it a bit more easier to follow.
2. In most following blocks you compare `fff != NULL` but here  you compare `fff == NULL`. Probably better to always have the same logic for consistency.
3. No need to write the actual comparison. AFAIK `fff` should be a `nullptr` if `fopen` fails. `nullptr == 0 == false`

src/autopas/utils/RaplMeter.h
```diff
@@ -0,0 +1,72 @@
...
+ public:
+  ~RaplMeter();
+  /** initialisation may fail, so moved out of constructor */
+  void init();
+  /** reset perf file descriptors to start new measurement */
+  void reset();
+  /** measure power consumption since last call of reset
+   * the results can be retrieved with the get_<domain>_energy() functions.
+   */
+  void sample();
```
FG-TUM, line 41:
```suggestion
  /** 
   * Reset perf file descriptors to start a new measurement.
   */
  void reset();
  /** 
   * Measure power consumption since the last call to reset.
   * The results can be retrieved with the get_<domain>_energy() functions.
   */
  void sample();
```

Just doc style. Please adapt all others as well.

src/autopas/utils/RaplMeter.h
```diff
@@ -0,0 +1,72 @@
...
+
+  /** returns the energy consumed by the cpu package between the last call of sample() and the preceding
+   * call of reset()
+   * @return Energy in Joules
+   */
+  double get_pkg_energy();
```
FG-TUM, line 47: What are the units?

src/autopas/utils/RaplMeter.h
```diff
@@ -0,0 +1,72 @@
...
+   * reset()
+   * @return Energy in Joules
+   */
+  double get_ram_energy();
+  /** returns the energy consumed by the entire system between the last call of sample() and the preceding
+   * call of reset()
+   * @return Energy in Joules
+   */
+  double get_psys_energy();
+  // unscaled for tuning
```
FG-TUM, line 63: not sure what this comment is supposed to mean

FG-TUM (changes requested):
Just some small clarifications please :)

src/autopas/utils/RaplMeter.cpp
VincentF98: Improved this a bit. Current behavior if an error occurs during energy measurement:
If energy is selected as tuning metric -> crash immediately
if time is selected -> log a warning and stop measuring energy in following iterations

src/autopas/utils/RaplMeter.h
VincentF98: clarified this a bit more in the doc string. the raw perf event does not give values in any specific unit.
the conversion factor to Joules is read from the .scales files in /sys/bus/event_source/devices/power/events. But since we want a long value and not a double for tuning, and consistency across devices is not required, we use the unscaled value. The other functions are provided for the purpose of actual measurements and their values are logged on debug level.

src/autopas/utils/RaplMeter.cpp
FG-TUM: sounds very reasonable

FG-TUM (approved):

PR #672, opened 2022-01-12
Reset F to global force

FG-TUM:
# Description

- [x] instead of resetting the force to zero reset it to the global force
- [x] cleanup of warnings and unused functions

## ~Resolved Issues~

# ~How Has This Been Tested?~


SamNewcome (approved):
looks good

PR #678, opened 2022-02-26
Rule-Based Tuning

HobbyProgrammer:
# Description
This PR adds a new rule-based tuning strategy and lots of infrastructure around it. It enables developers and users of `AutoPas` to use domain knowledge to heavily prune the search space during tuning in a dynamic way. Additionally, new tools are added to collect data about and analyze the runtime behavior of AutoPas, as well as verify the domain knowledge during program execution on a specific machine. 

The new tuning strategy 'rule-based-tuning' uses a manually written rule file, provided at runtime, which specifies under which conditions it is known that certain strategies are better than others. Here is one example rule: 
```
if avgParticlesPerCell < 6:
    [container="VerletListsCells", dataLayout="AoS"] >= [container="VerletListsCells", dataLayout="SoA"] with same newton3, traversal, loadEstimator;
endif
```
This rule says that if the average particles per cell in the domain is less than 6, *all* configurations with the `VerletListsCells` container and the data layout `AoS` are equally fast or faster than *all* configurations with the `VerletListsCells` container and the data layout `SoA`, provided they have the same `newton3`, `traversal`, and `loadEstimator` value. Essentially, this means that under the condition, the data layout `AoS` is faster than the data layout `SoA` for the `VerletListsCells` container.

The rules in a rule file basically define a partial order of configurations, where `l >= r` means `l` is at least as good (fast) or better (faster) than `r`.

Before each tuning phase, the necessary data to evaluate the conditions is collected, and the `RuleBasedTuningStrategy` evaluates the rules in the provided rule file to find all possibly best configurations (provided the rules are correct), let's call them set `R`. If `A` is the set of all allowed configurations, and `W` the set of all configurations for which there exists a better configuration in the partial order defined above, then `R = A \ W`. On this result set, the tuning strategy performs full search to find the best possible configuration. This allows us to use domain knowledge to improve the tuning strategies, instead of relying on black box optimization algorithms.

Of course, two questions immediately pop up: How do we come up with useful rules, and how do we verify that they are actually correct?

Since their answer heavily depends on the performance of actual program execution, so user input and hardware, the key to solving them is to collect data. For this, a logging tuning strategy proxy (`TuningStrategyLoggerProxy`) has been introduced. This is a tuning strategy interposed between the `AutoTuner` and the actual tuning strategy, that is able to log all calls with their parameters to a tuning strategy during execution. Instead of rerunning the same simulations over and over again, these logs then can later be replayed using the newly introduced tool `ruleChecker` to verify new rules against this collected data. Additionally, another tool `tuningLogToSQL` is introduced, which is able to read these log files for an arbitrary number of program executions, and write the tuning data into a SQLite database for further analysis. This is very useful for finding previously unknown behaviors that allow for search space pruning with new rules. To get meaningful data, a script has been written that generates a hopefully representative set of scenarios to run in md-flexible with the tuning logger proxy enabled.

Here is a list of the most important user-facing additions:
- New tuning strategy: `rule-based-tuning`. Currently expects a file `tuningRules.rule` in the same directory (Example provided).
- New `md-flexible` option: `use-tuning-logger`(`=true`) to enable the tuning logger proxy. Writes log to a .txt file.
- `ruleChecker` executable: Reads tuning log files and verifies the data does not contradict any rules in the given rule file. Can be used to iteratively try out new rules and tune parameters.
- `tuningLogToSQL` executable: Reads tuning log files and writes them into a SQLite database. Also already provides some interesting SQL views on the data.
- `visualize_data.py`: Reads a SQLite database containing tuning data and generates an HTML report with interesting information gathered from the raw data.
- `generate-md-flex-configs.py`: Generates a hopefully representative set of scenarios to collect data from about the performance of configurations in them.

Interesting implementation details:
- Rule file parser: Implemented with `antlrcpp`. Grammar is in `RuleLanguage.g4` (generated code files are also commited currently).
- Rule file evaluator: Done in a simple VM in `RuleVM.h`.
- Information available for condition evaluation is gathered in `LiveInfo.h`.

Feel free to ask about anything in more detail! I am happy to explain more.

## TODO: 
- [ ] Write a README with detailed steps how to use all of them meaningfully.
- [x] Cleanup code.
- [x] Document code.
- [ ] Provide vim syntax highlighting file for .rule files in this repo.
- [x] Discuss how to handle `antlrcpp` generated code files. (all in `parser_generated/` subdirectory)
- [x] Make code ready for review.

## Future Work:
- Combine rule based tuning with other tuning approaches -> #759 
- Refactor all tuning code
- Improve example rule file

# How Has This Been Tested?
TODO


HobbyProgrammer:
> Just as an Idea: could / should only collect the runtime info that is actually necessary. So kind of a lazy approach

I thought about that, but actually, I think it makes more sense to always collect the same metrics. It is cleaner and easier to understand, and when collecting the runtime info in a log, it is a lot easier to analyze if there is always the same information available. If there was a noticeable performance impact of collecting the data, it might be worth more, but since a tuning phase normally takes at least few dozens of iterations with greatly suboptimal (x2 worse almost for sure) configurations, collecting this data once per tuning phase shouldn't have any noticeable impact on runtime.

FG-TUM:
@HobbyProgrammer I added `uuid-dev` to the docker images and this seems to have done the trick. A few things I spotted after quickly looking over the new logs:
- [ ] style check fails at the (generated?) antlr code -> Either generate according to `.clang-format` if possible or exclude from the style check
- [ ] Errors when accessing `std::variant` values.

HobbyProgrammer:
Code should be ready for review, only more unit tests are missing. I will add them soon, but this should not block code review.

FG-TUM:
ToDos:
- [ ] Add a README for the rule checker
- [ ] A paragraph about using the rule file in the main AutoPas README
- [ ] (nice to have but not mandatory) make antler dependency optional in CMake. Would require a switch that fully deactivates rule-based tuning.

src/autopas/selectors/tuningStrategy/LiveInfo.h
```diff
@@ -0,0 +1,325 @@
...
+      } else if constexpr (std::is_base_of_v<Option<decltype(type)>, decltype(type)>) {
+        return type.to_string();
+      }
+      return std::string{"fail"};
+    };
+    auto toString = [&](const auto &pair) { return pair.first + "=" + std::visit(typeToString, pair.second); };
+    std::string res{"Live Info: "};
+    if (not infos.empty()) {
+      res += std::accumulate(std::next(infos.begin()), infos.end(), toString(*infos.begin()),
+                             [&](std::string s, const auto &elem) { return std::move(s) + " " + toString(elem); });
```
FG-TUM, line 228: @HobbyProgrammer why `std::next()`? Doesn't this mean that the fist entry is skipped in the output?

FG-TUM:
@HobbyProgrammer FYI: [9459528](https://github.com/AutoPas/AutoPas/pull/678/commits/9459528a230429f36e889644dcea39d7e526a811) renamed `TuningStrategyLoggerProxy` to `TuningStrategyLoggerWrapper`.

src/autopas/selectors/tuningStrategy/LiveInfo.h
HobbyProgrammer: The first entry is passed initial value for the accumulation (`toString(*infos.begin())`). We don't want it twice in the output, so we skip the first element in the passed iterator. Passing an empty string as initializer would be nicer, but then we would have a space in the beginning, so we can't do it that way.

FG-TUM: Ah I missed the fact that you chose the first element as the initializer. Makes sense now. I'll add a comment.

src/autopas/selectors/tuningStrategy/TuningStrategyLogReplayer.cpp
```diff
@@ -0,0 +1,77 @@
...
+      auto cont = not traversalTimes.empty();
+      while (cont) {
+        cont = _tuningStrategy->tune();
+        try {
+          const auto &[time, iteration] = traversalTimes.at(_tuningStrategy->getCurrentConfiguration());
+          _tuningStrategy->addEvidence(time, iteration);
+        } catch (std::out_of_range &e) {
+          // std::cerr << "Could not find data for configuration " << _tuningStrategy->getCurrentConfiguration() <<
+          // std::endl;
+        }
```
FG-TUM, line 60: Silent catch

src/autopas/selectors/tuningStrategy/TuningStrategyLogReplayer.cpp
```diff
@@ -0,0 +1,77 @@
...
+    std::getline(in, line, '\n');
+
+    std::stringstream stream{line};
+    std::string type;
+    std::getline(stream, type, ' ');
+
+    if (type == "evidence") {
+      evidenceAdded = true;
+      const auto &[time, iteration, config] = tuningLogEntry::readEvidence(stream);
+      // std::cout << "Add data for configuration " << config << std::endl;
```
FG-TUM, line 42: unused code

src/autopas/selectors/tuningStrategy/TuningStrategyLoggerWrapper.cpp
```diff
@@ -0,0 +1,76 @@
...
+    : _actualTuningStrategy(std::move(actualTuningStrategy)) {
+  std::stringstream filename;
+  filename << "tuningLog-";
+  filename << utils::Timer::getDateStamp();
+  filename << '-' << outputSuffix;
+  filename << ".txt";
+  _logOut.open(filename.str());
+}
+
+TuningStrategyLoggerWrapper::~TuningStrategyLoggerWrapper() { _logOut.flush(); }
```
FG-TUM, line 25: this should probably also close the file?

src/autopas/selectors/tuningStrategy/ruleBasedTuning/RuleLanguage.g4
```diff
@@ -0,0 +1,143 @@
+// This file describes the grammar of the rule language used by rule based tuning. The parsing code is generated using
+// the ANTLR plugin for CLion. It is generated into parser_generated/ and committed to the repository. Changing this
+// file requires manual regeneration of the code into this directory, and also committing (and formatting) the generated
+// files.
+
+grammar RuleLanguage;
```
FG-TUM, line 6: this is another file we need to maintain whenever we add any new config element. Could we reasonably generate this via some script or mini c++ program like the `zsh` completions`? Otherwise this has to be mentioned in the contribution guidelines.

README.md
```diff
@@ -17,6 +17,8 @@ Alternatively you can build the documentation on your own:
 * CMake 3.14 or newer
 * make (build-essentials) or ninja
 * a C++17 compiler (gcc11, clang13, and ~~icpc 2019~~ are tested.)
+* For rule based tuning: `pkg-config`, `uuid`
```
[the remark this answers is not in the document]
HobbyProgrammer: Correct!

src/autopas/selectors/tuningStrategy/TuningStrategyLoggerWrapper.cpp
HobbyProgrammer: Should happen automatically in the destructor of std::ofstream ("destructs the basic_ofstream and the associated buffer, closes the file" from https://en.cppreference.com/w/cpp/io/basic_ofstream).

src/autopas/selectors/tuningStrategy/ruleBasedTuning/RuleLanguage.g4
HobbyProgrammer: I will think a bit if we can do it easily. Note that the generated code files are not generated automatically from this grammar at the moment. It is a manual process, so regenerating them manually after any change is required.

src/autopas/selectors/tuningStrategy/ruleBasedTuning/RuleLanguage.g4
```diff
@@ -0,0 +1,143 @@
...
+                : DIGIT+ '.' DIGIT*
+                ;
+
+unsigned_val
+                : Unsigned_val
+                | DIGIT
+                ;
+
+literal
+                : '"' Traversal_opt '"'
```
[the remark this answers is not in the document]
HobbyProgrammer: I think the grammar is ambiguous without them, because a literal could then also be a variable, and both are valid expressions. Also, this probably makes future changes to the grammar easier. For example, in the future, we might want to not include all container, data layout, etc. options in this file, use arbitrary strings in quotes here instead, and resolve to the actual option after parsing. Without quotes, there is no way to distinguish these from variables, then.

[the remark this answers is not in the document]
FG-TUM: Sounds reasonable. Feels like we should note this somewhere. At least your project report ;)

tuningLogToSQL/LogToSQLiteWriter.h
```diff
@@ -0,0 +1,542 @@
...
+    result.push_back(c);
+    if(c == '\'') {
+      result.push_back('\'');
+    }
+  }
+
+  return result;
+}
+
+class LogToSQLiteWriter {
```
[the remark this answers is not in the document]
FG-TUM: done in #759 d757c7e56db1f2248155b1baab3b06c325cfc3df

src/autopas/selectors/tuningStrategy/TuningStrategyLogReplayer.cpp
FG-TUM: resolved in #759 449fa892

FG-TUM (approved):

PR #693, opened 2022-09-05
Issue#667

### LINKED ISSUE #667: md-flexible: --tuning-phases=1 with only one configuration never terminates
**Describe the bug**
When md-flexible is started with the `--tuning-phases` parameter set to something >= 1 and only one configuration allowed, it never terminates. 

**To Reproduce**
1. Run `md-flexible --tuning-phases=1 --container=LinkedCells --traversal=lc_c08 --data-layout=SoA --newton3=enabled`
2. See that is never terminates.

**Expected behavior**
md-flexible should always terminate. In this case as soon as md-flexible detects that there is nothing to tune but the number of iterations is determined by the number of tuning phases it should abort with a meaningful error message.

christinanb:
# Description

Give MDflexible access to autopas searchspaceistrivial() function through AutoTuner::searchSpaceIsTrivial() and AutoPas::getsearchSpaceIsTrivial(). Use this to throw an exception  in simulation.cpp when MD-flexible is ran with the invalid inputs of a single configuration but multiple tuning phases (nothing to tune for).

## Resolved Issues

- fixes #667

# How Has This Been Tested?

MD-flexible is ran with different inputs. Configurations resulting in a search space size of 1 and number of tuning phases >1. This throws an error. When the number of tuning phases is zero this does not throw an error. When the search space size is greater than 1 the simulation always runs.

### DIFF
--- a/examples/md-flexible/src/Simulation.cpp
+++ b/examples/md-flexible/src/Simulation.cpp
@@ -134,8 +134,8 @@ Simulation::Simulation(const MDFlexConfig &configuration,
 
   // Throw an error if there is not more than one configuration to test in the search space but more than one tuning
   // phase is requested
-  if (_autoPasContainer->getsearchSpaceIsTrivial() && _configuration.tuningPhases.value > 0) {
-    throw std::runtime_error("Search space must not be empty if multiple tuning phases are requested");
+  if (_autoPasContainer->searchSpaceIsTrivial() and _configuration.tuningPhases.value > 0) {
+    throw std::runtime_error("Search space must not be trivia if the simulation time is limited by the number tuning phases");
   }
 
   // @todo: the object generators should only generate particles relevant for the current rank's domain
--- a/examples/md-flexible/src/Simulation.cpp
+++ b/examples/md-flexible/src/Simulation.cpp
@@ -134,7 +134,7 @@ Simulation::Simulation(const MDFlexConfig &configuration,
 
   // Throw an error if there is not more than one configuration to test in the search space but more than one tuning
   // phase is requested
-  if (_autoPasContainer->getsearchspaceistrivial() && _configuration.tuningPhases.value > 0) {
+  if (_autoPasContainer->getsearchSpaceIsTrivial() && _configuration.tuningPhases.value > 0) {
     throw std::runtime_error("Search space must not be empty if multiple tuning phases are requested");
   }
 
--- a/examples/md-flexible/src/Simulation.cpp
+++ b/examples/md-flexible/src/Simulation.cpp
@@ -132,6 +132,11 @@ Simulation::Simulation(const MDFlexConfig &configuration,
   autopas::Logger::get()->set_level(_configuration.logLevel.value);
   _autoPasContainer->init();
 
+  // Throw an error if there is not more than one configuration to test in the search space but more than one tuning phase is requested 
+  if (_autoPasContainer->getsearchspaceistrivial() && _configuration.tuningPhases.value > 0){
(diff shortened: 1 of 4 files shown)


examples/md-flexible/src/Simulation.cpp
```diff
@@ -132,6 +132,12 @@ Simulation::Simulation(const MDFlexConfig &configuration,
   autopas::Logger::get()->set_level(_configuration.logLevel.value);
   _autoPasContainer->init();
 
+  // Throw an error if there is not more than one configuration to test in the search space but more than one tuning
+  // phase is requested
+  if (_autoPasContainer->getsearchSpaceIsTrivial() && _configuration.tuningPhases.value > 0) {
```
FG-TUM, line 137:
```suggestion
  if (_autoPasContainer->getsearchSpaceIsTrivial() and _configuration.tuningPhases.value > 0) {
```
Both are equivalent. Suggestion is just for consistency with the rest of the project.

examples/md-flexible/src/Simulation.cpp
```diff
@@ -132,6 +132,12 @@ Simulation::Simulation(const MDFlexConfig &configuration,
...
+    throw std::runtime_error("Search space must not be empty if multiple tuning phases are requested");
```
FG-TUM, line 138:
```suggestion
    throw std::runtime_error("Search space must not be trivia if the simulation time is limited by the number tuning phases");
```

src/autopas/AutoPasDecl.h
```diff
@@ -458,6 +458,12 @@ class AutoPas {
    */
   [[nodiscard]] std::array<double, 3> getBoxMax() const;
 
+  /**
+   * get the bool value indicating if the search space is trivial (not more than one configuration to test).
+   * @return bool indicating if search space is trivial.
+   */
+  [[nodiscard]] bool getsearchSpaceIsTrivial();
```
FG-TUM, line 465: I think I would actually just call it `searchSpaceIsTrivial()`

src/autopas/AutoPasImpl.h
```diff
@@ -203,4 +203,9 @@ std::shared_ptr<const autopas::ParticleContainerInterface<Particle>> AutoPas<Par
   return _autoTuner->getContainer();
 }
 
+template <class Particle>
+bool AutoPas<Particle>::getsearchSpaceIsTrivial() {
+  return _autoTuner->searchSpaceIsTrivial();
+}
+
```
FG-TUM, line 210: Have you checked whether it is possible to have this function in `AutoPasDecl.h`? I don't think it is but just for my peace of mind :)

FG-TUM (changes requested):
Some minor naming suggestions

examples/md-flexible/src/Simulation.cpp
christinanb: Fixed!

examples/md-flexible/src/Simulation.cpp
christinanb: Done!

src/autopas/AutoPasDecl.h
christinanb: Completed!

src/autopas/AutoPasImpl.h
christinanb: I already checked this and sadly couldn't get it to work. I think it has something to do with referencing AutoTuner.

FG-TUM: 👍

FG-TUM (approved):

PR #702, opened 2022-11-24
Make addParticle() ThreadSafe

FG-TUM:
# Description

`AutoPas::addParticle()` is not thread-safe since we use the logic handler buffers because `std::vector::push_back()` is not thread-safe. This PR solves this by adding one buffer per thread.

## ~Resolved Issues~

# How Has This Been Tested?

- [x] Existing CI test
- [x] new doRemainderTraversal tests

### DIFF
--- a/src/autopas/utils/NumberInterval.h
+++ b/src/autopas/utils/NumberInterval.h
@@ -102,7 +102,7 @@ class NumberInterval : public NumberSet<Number> {
     for (size_t i = 0; i < (n - 1); ++i) {
       result.push_back(_min + distance * i);
     }
-    // add max separatly, avoiding possible rounding errors
+    // add max separately, avoiding possible rounding errors
     result.push_back(_max);
 
     // randomize the sample
@@ -128,7 +128,7 @@ class NumberInterval : public NumberSet<Number> {
     for (size_t i = 0; i < (n - 1); ++i) {
       result.insert(_min + distance * i);
     }
-    // add max separatly, avoiding possible rounding errors
+    // add max separately, avoiding possible rounding errors
     result.insert(_max);
 
     return result;


src/autopas/selectors/AutoTuner.h
```diff
@@ -457,74 +459,87 @@ bool AutoTuner<Particle>::iteratePairwise(PairwiseFunctor *f, bool doListRebuild
...
-    auto min = autopas::utils::ArrayMath::subScalar(pos, containerPtr->getCutoff());
-    auto max = autopas::utils::ArrayMath::addScalar(pos, containerPtr->getCutoff());
-    for (auto iter2 = containerPtr->getRegionIterator(min, max, IteratorBehavior::ownedOrHalo); iter2.isValid();
-         ++iter2) {
-      if (newton3) {
-        f->AoSFunctor(p, *iter2, true);
-      } else {
-        f->AoSFunctor(p, *iter2, false);
-        f->AoSFunctor(*iter2, p, false);
+void doRemainderTraversal(PairwiseFunctor *f, T containerPtr, std::vector<std::vector<Particle>> &particleBuffers,
```
SamNewcome, line 468: doRemainderTraversal should have an expanded description, this is too vague.

src/autopas/selectors/AutoTuner.h
```diff
@@ -457,74 +459,87 @@ bool AutoTuner<Particle>::iteratePairwise(PairwiseFunctor *f, bool doListRebuild
...
-        f->AoSFunctor(p1, p2, true);
-      } else {
-        f->AoSFunctor(p1, p2, false);
-        f->AoSFunctor(p2, p1, false);
+  // 3. particleBuffer with itself and all other buffers
+  for (size_t bufferIdOuter = 0; bufferIdOuter < particleBuffers.size(); ++bufferIdOuter) {
+    auto &particleBufferOuter = particleBuffers[bufferIdOuter];
+    for (size_t i = 0; i < particleBufferOuter.size(); ++i) {
+      auto &&p1 = particleBufferOuter[i];
+      for (size_t bufferIdInner = bufferIdOuter; bufferIdInner < particleBuffers.size(); ++bufferIdInner) {
```
SamNewcome, line 512: Any reason why this loop is inside the "i" loop? Initial looks make it seem like these are interchangeable and the switch might help readability.

src/autopas/selectors/AutoTuner.h
```diff
@@ -457,74 +459,87 @@ bool AutoTuner<Particle>::iteratePairwise(PairwiseFunctor *f, bool doListRebuild
...
-        f->AoSFunctor(p1, p2, true);
-      } else {
-        // Here, we do not need to interact p2 with p1, because p2 is a halo particle and an AoSFunctor call with an
-        // halo particle as first argument has no effect if newton3 == false.
-        f->AoSFunctor(p1, p2, false);
+  for (size_t bufferIdOuter = 0; bufferIdOuter < particleBuffers.size(); ++bufferIdOuter) {
+    auto &particleBufferOuter = particleBuffers[bufferIdOuter];
+    for (size_t i = 0; i < particleBufferOuter.size(); ++i) {
+      auto &&p1 = particleBufferOuter[i];
+      for (size_t bufferIdInner = 0; bufferIdInner < haloParticleBuffers.size(); ++bufferIdInner) {
```
SamNewcome, line 531: Similar to above

tests/testAutopas/tests/selectors/AutoTunerTest.cpp
```diff
@@ -21,7 +22,7 @@ using ::testing::_;
 TEST_F(AutoTunerTest, testAllConfigurations) {
   std::array<double, 3> bBoxMin = {0, 0, 0}, bBoxMax = {2, 4, 2};
   // adaptive domain size so sliced is always applicable.
-  // bBoxMax[1] = autopas::autopas_get_max_threads() * 2;
+  // bBoxMax[1] = autopas::autopas::autopas_get_max_threads() * 2;
```
SamNewcome, line 25:
```suggestion
```

src/autopas/selectors/AutoTuner.h
FG-TUM: This ordering reduces the number of memory accesses. We access each outer particle (`i`) once and then the whole inner buffer. If you change the loops you access each `i` once for every buffer. If the buffers are small this should not be significant but I feel they are sometimes not small...

SamNewcome (approved):

src/autopas/selectors/AutoTuner.h
FG-TUM: see above :upside_down_face:

src/autopas/selectors/AutoTuner.h
FG-TUM: done [3709cf3](https://github.com/AutoPas/AutoPas/pull/702/commits/3709cf3bf23b442b7cd06514a1cc10f036b620a6)

Issue #714, opened 2023-01-25
Implement ghost particle reflection


SamNewcome:
**Is your feature request related to a problem? Please describe.**
See #713 
Goal is to have a more natural reflecting boundary.

**Describe the solution you'd like**
Implement ghost particle reflection - where particles near the boundary are mirrored onto the other side of the boundary and the force between the two is applied to the original.

**Additional context**
When an error is thrown due to a particle leaving the domain, we should check if the boundary is reflective and, if so, an error message should be printed where a smaller deltaT is suggested.

PR #715, opened 2023-01-29
Mirror particles for reflecting

### LINKED ISSUE #713: Particle out of box during fallingDrop simulation
The simulation of fallingDrop interrupts with an runtime_error (particle out of the box)

**To Reproduce**
  run ./md-flexible --yaml-file fallingDrop.yaml in the build directory

**Expected behavior**
  particle should be reflected of the boundaries and the simulation should run without an error.

**Screenshots**
Screenshot from 2023-01-25 11-43-21

**Build environment:**
OS: Linux

SamNewcome:
# Description

Replaces the existing reflecting boundary mechanism with one based on mirror particles. Here 'mirror' particles are placed on the opposite side of a reflecting boundary from an original particle, and the two interact via the standard Lennard-Jones functor. Interactions are limited to when particles are within a range such that any LJ interaction leads to repulsion only.

## Resolved Issues

- [x] fixes #713 
- [x] implements #714 

# How Has This Been Tested?

- [x] Standard Reflecting Boundary tests have been modified to test for correct force for the current scheme
- [x] Additional test that checks correct zoning of reflecting boundaries (i.e. that particles react only when it leads to repulsion, for several different sigma values)
- [x] Falling Drop can be run without error.

### DIFF
--- a/examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
+++ b/examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
@@ -43,7 +43,7 @@ RegularGridDecomposition::RegularGridDecomposition(const MDFlexConfig &configura
   for (auto sigmaMapElement : configuration.sigmaMap.value) {
     maxSigma = std::max(maxSigma, sigmaMapElement.second);
   }
-  _maxReflectiveSkin = std::pow(maxSigma, 1/6.);
+  _maxReflectiveSkin = sixthRootOfTwo * maxSigma;
 
   // initialize _communicator and _domainIndex
   initializeMPICommunicator();
--- a/examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
+++ b/examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
@@ -258,9 +258,11 @@ void RegularGridDecomposition::exchangeMigratingParticles(AutoPasType &autoPasCo
   }
 }
 
-void RegularGridDecomposition::reflectParticlesAtBoundaries(AutoPasType &autoPasContainer, ParticlePropertiesLibraryType &PPL) {
+void RegularGridDecomposition::reflectParticlesAtBoundaries(AutoPasType &autoPasContainer,
+                                                            ParticlePropertiesLibraryType &PPL) {
   std::array<double, _dimensionCount> reflSkinMin{}, reflSkinMax{};
-  auto functorLJ = autopas::LJFunctor<ParticleType, false, true, autopas::FunctorN3Modes::Newton3Off, false, false>(_maxReflectiveSkin * 2., PPL);
+  auto functorLJ = autopas::LJFunctor<ParticleType, false, true, autopas::FunctorN3Modes::Newton3Off, false, false>(
+      _maxReflectiveSkin * 2., PPL);
 
   for (int dimensionIndex = 0; dimensionIndex < _dimensionCount; ++dimensionIndex) {
     // skip if boundary is not reflective
@@ -271,7 +273,8 @@ void RegularGridDecomposition::reflectParticlesAtBoundaries(AutoPasType &autoPas
            p.isValid(); ++p) {
         // Check that particle is within 6th root of 2 * sigma
         const auto position = p->getR();
-        const auto distanceToBoundary = isUpper ? reflSkinMax[dimensionIndex] - position[dimensionIndex] : position[dimensionIndex] - reflSkinMin[dimensionIndex];
+        const auto distanceToBoundary = isUpper ? reflSkinMax[dimensionIndex] - position[dimensionIndex]
+                                                : position[dimensionIndex] - reflSkinMin[dimensionIndex];
         if (distanceToBoundary < sixthRootOfTwo * PPL.getSigma(p->getTypeId()) / 2.) {
           // Create mirror particle and shift it to other side of reflective boundary
           ParticleType mirrorParticle;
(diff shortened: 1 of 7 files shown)


examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
```diff
@@ -39,6 +39,12 @@ RegularGridDecomposition::RegularGridDecomposition(const MDFlexConfig &configura
 
   _decomposition = DomainTools::generateDecomposition(_subdomainCount, configuration.subdivideDimension.value);
 
+  double maxSigma{0};
+  for (auto sigmaMapElement : configuration.sigmaMap.value) {
+    maxSigma = std::max(maxSigma, sigmaMapElement.second);
+  }
```
FG-TUM, line 45:
```suggestion
  double maxSigma{0};
  for (const auto &[_, sigma] : leMap) {
    maxSigma = std::max(maxSigma, sigma);
  }
```

Imho it would be even better to wrap that loop into a lambda function so that `maxSigma` can be turned into a `const` but since the function ends soon after this I'm fine.

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
```diff
@@ -39,6 +39,12 @@ RegularGridDecomposition::RegularGridDecomposition(const MDFlexConfig &configura
...
+  _maxReflectiveSkin = sixthRootOfTwo * maxSigma / 2.;
```
FG-TUM, line 46: add a comment explaining this calculation

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
```diff
@@ -252,8 +258,11 @@ void RegularGridDecomposition::exchangeMigratingParticles(AutoPasType &autoPasCo
   }
 }
 
-void RegularGridDecomposition::reflectParticlesAtBoundaries(AutoPasType &autoPasContainer) {
+void RegularGridDecomposition::reflectParticlesAtBoundaries(AutoPasType &autoPasContainer,
+                                                            ParticlePropertiesLibraryType &PPL) {
```
FG-TUM, line 262:
```suggestion
void RegularGridDecomposition::reflectParticlesAtBoundaries(AutoPasType &autoPasContainer,
                                                            ParticlePropertiesLibraryType &particlePropertiesLib) {
```
(will require a few more renames)

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
```diff
@@ -252,8 +258,11 @@ void RegularGridDecomposition::exchangeMigratingParticles(AutoPasType &autoPasCo
...
   std::array<double, _dimensionCount> reflSkinMin{}, reflSkinMax{};
+  auto functorLJ = autopas::LJFunctor<ParticleType, false, true, autopas::FunctorN3Modes::Newton3Off, false, false>(
```
FG-TUM, line 264:
```suggestion
  auto functorLJ = autopas::LJFunctor<ParticleType, false, true, autopas::FunctorN3Modes::Both, false, false>(
```
Is there a reason to choose a mode different than the default?

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
```diff
@@ -262,28 +271,38 @@ void RegularGridDecomposition::reflectParticlesAtBoundaries(AutoPasType &autoPas
...
       for (auto p = autoPasContainer.getRegionIterator(reflSkinMin, reflSkinMax, autopas::IteratorBehavior::owned);
            p.isValid(); ++p) {
-        auto vel = p->getV();
-        // reverse velocity in dimension if towards boundary
-        if ((vel[dimensionIndex] < 0) xor isUpper) {
-          vel[dimensionIndex] *= -1;
+        // Check that particle is within 6th root of 2 * sigma
+        const auto position = p->getR();
+        const auto distanceToBoundary = isUpper ? reflSkinMax[dimensionIndex] - position[dimensionIndex]
+                                                : position[dimensionIndex] - reflSkinMin[dimensionIndex];
```
FG-TUM, line 277:
```suggestion
        const auto distanceToBoundary = std::abs(reflSkinMax[dimensionIndex] - position[dimensionIndex]);
```

FG-TUM (changes requested):
mostly readability suggestions

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
SamNewcome: No - other than Newton3On is not needed. But I see no reason why Both is not fine - if N3L is accidentally applied there are no major problems.

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
SamNewcome: See latest commit

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
SamNewcome: Note, this is not just taking the absolute but takes distance to the two different boundaries, so std::abs will only work in conjunction with the conditional operator, at which point there is no gain in readability (and maybe even a loss).

I can add a comment explaining isUpper, or explaining this step if you feel it is needed.

FG-TUM: I overlooked `Min` vs `Max`...

FG-TUM (approved):

PR #724, opened 2023-03-06
Improve remainder traversal

FG-TUM:
# Description

- [x] parallelize AutoTuner::doRemainderTraversal()
  - Done via locks 
  - No-newton3 between buffers is not trivially doable because of global values
- [x] AutoPas::reserve(numParticles)
- [x] ~split LogicHandler buffers for better particle locality~
- [x] Check out task parallelization of remainder traversal.

## Comparison Before vs After

Machine: atsccs93 (i7-10700 @ 2x8x ~4.5GHz)

Config:
```yaml
verlet-rebuild-frequency         :  15
verlet-skin-radius-per-timestep  :  0.2
data-layout                      :  [SoA]
functor                          :  Lennard-Jones (12-6) AVX intrinsics
cutoff                           :  2.5
deltaT                           :  0.0001
iterations                       :  50
boundary-type                    :  [periodic, periodic, periodic]
Objects:                         
  CubeGrid:
    0:  
      particles-per-dimension    :  [50, 50, 50]
      particle-spacing           :  1.4
      bottomLeftCorner           :  [0.7, 0.7, 0.7]
      velocity                   :  [1000, 0, 0]
      particle-type              :  0
      particle-epsilon           :  1
      particle-sigma             :  1
      particle-mass              :  1
```

A closer look at iterations where there actually are halo particles. Average over all iterations where the particle buffer size was 2500 and halo buffer 5864. Using 8 OMP Threads.

| Timer [ms] | Before (79effeca) | After (d7e07f5) | 
|:------|-----------:|-----------------:
|Total (50 iters) | 5142 | 1585 |
| One Iteration (avg total) | 102 | 31 |
| Container::iteratePairwise (avg with halos) | ~14 | ~14 | 
| Remainder Traversal (avg with halos)| 116.8 | 4.05 |

## Related Pull Requests

- Merge after #712 

## Resolved Issues

Before this PR `doRemainderTraversal()` took a significant amount of time. If a few thousand particles were in the buffers this took significantly longer than `iteratePairwise()` due to lack of parallelization.

# How Has This Been Tested?

- [x] existing unit tests
- [x] New test for `utils::ArrayUtils::balanceVectors()`


src/autopas/utils/ArrayUtils.h
```diff
@@ -146,4 +147,79 @@ std::enable_if_t<autopas::utils::ArrayUtils::is_container<Container>::value, std
...
+  // calculate vecvec statistics
+  const auto vecvecSize = vecvec.size();
+  if (vecvecSize == 0) {
+    // nothing to do
+    return;
+  }
+  const size_t numElem = std::transform_reduce(vecvec.begin(), vecvec.end(), 0, std::plus<>(),
+                                               [&](auto &vec) { return innerContainerToVec(vec).size(); });
+  const auto targetSize = static_cast<long>(numElem / vecvecSize);
+  auto rest = numElem % vecvecSize;
```
SamNewcome, line 183: `rest` is not a clear enough name

src/autopas/utils/ArrayUtils.h
```diff
@@ -146,4 +147,79 @@ std::enable_if_t<autopas::utils::ArrayUtils::is_container<Container>::value, std
 
   return os;
 }
+
+/**
+ * Given a collection of vectors, redistributes the elements of the vectors so they all have the same (or +1) size.
```
SamNewcome, line 152: Add comment/note briefly describing the distribution process and make it clear that overall ordering is not preserved. E.g. Elements are taken from the ends of too-long vectors and added to the ends of too-short vectors. The overall ordering of elements is not preserved.

tests/testAutopas/tests/selectors/AutoTunerRemainderTraversalTest.cpp
```diff
@@ -12,9 +12,298 @@
...
+ * @param particlesHaloBuffer
+ * @param n3 Newton3 on or off
+ */
+void testIteratePairwiseSteps(std::vector<Molecule> &particlesContainerOwned,
+                              std::vector<Molecule> &particlesContainerHalo,
+                              std::vector<autopas::FullParticleCell<Molecule>> &particlesBuffers,
+                              std::vector<autopas::FullParticleCell<Molecule>> &particlesHaloBuffers,
+                              autopas::Newton3Option n3) {
+  // sanity check that there are exactly two particles in the test
+  const auto numParticlesBuffers =
```
SamNewcome, line 33: Rename `numParticlesBuffers` to `numParticlesInBuffers` for clarity. Similarly `numParticlesInHaloBuffers`. You _could_ also rename `particlesBuffers` to `particlesInBuffers` for consistency - but I don't think this is that important.

tests/testAutopas/tests/utils/ArrayUtilsTest.cpp
```diff
@@ -40,3 +42,38 @@ TEST(ArrayUtilsTest, testto_string) {
...
+  const auto numElements = [&]() {
+    size_t sum = 0;
+    for (const auto &vec : vecvec) {
+      sum += vec.size();
+    }
+    return sum;
+  }();
+  // calculate expectations
+  const auto expectedPerVec = numElements / vecvec.size();
+  const auto rest = numElements % vecvec.size();
```
SamNewcome, line 70: Similar to `ArrayUtils.h`, use a different name for `rest` that's clearer

tests/testAutopas/tests/selectors/AutoTunerRemainderTraversalTest.cpp
FG-TUM: `particlesBuffers` are the actual buffers so I think this name makes more sense?

SamNewcome: Yes I see. But I still think the `numParticlesBuffers` should be `numParticlesInBuffers` to avoid confusion that this variable does not hold the number of particle buffers.

src/autopas/selectors/AutoTuner.h
```diff
@@ -452,120 +504,185 @@ bool AutoTuner<Particle>::iteratePairwise(PairwiseFunctor *f, bool doListRebuild
...
-                f->AoSFunctor(p1, p2, true);
-              } else {
-                // Here, we do not need to interact p1 with p2, because p is a halo particle and an AoSFunctor call with
-                // a halo particle as first argument has no effect if newton3 == false.
-                f->AoSFunctor(p2, p1, false);
-              }
+              const auto lockCoords =
+                  static_cast_array<size_t>(mulScalar(sub(p2.getR(), boxMin), interactionLengthInv));
+              // No need to apply anything to p1halo
+              //   -> AoSFunctor(p1, p2, false) not needed as it neither adds force nor Upot
```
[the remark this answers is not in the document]
FG-TUM: done 6685dbbb05334e01eb1001ad4b603bcceed1b05b

src/autopas/utils/ArrayUtils.h
FG-TUM: done 6685dbbb05334e01eb1001ad4b603bcceed1b05b

src/autopas/utils/ArrayUtils.h
```diff
@@ -146,4 +147,79 @@ std::enable_if_t<autopas::utils::ArrayUtils::is_container<Container>::value, std
...
+ * @param vecvec Reference to the collection of vectors to be balanced in place.
+ */
+template <class OuterContainerT>
+void balanceVectors(OuterContainerT &vecvec) {
+  balanceVectors(
+      vecvec, [](auto &innerContainer) -> auto & { return innerContainer; });
+}
+
+/**
+ * Given a collection of containers, redistributes the elements of the vectors so they all have the same (or +1) size.
```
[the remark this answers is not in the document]
FG-TUM: done 6685dbbb05334e01eb1001ad4b603bcceed1b05b

tests/testAutopas/tests/selectors/AutoTunerRemainderTraversalTest.cpp
FG-TUM: done 6685dbbb05334e01eb1001ad4b603bcceed1b05b

tests/testAutopas/tests/utils/ArrayUtilsTest.cpp
FG-TUM: done 6685dbbb05334e01eb1001ad4b603bcceed1b05b

SamNewcome (approved):

examples/md-flexible/src/TimeDiscretization.cpp
```diff
@@ -44,9 +44,8 @@ void calculatePositions(autopas::AutoPas<ParticleType> &autoPasContainer,
     const auto distanceMovedSquared = dot(displacement, displacement);
     if (distanceMovedSquared > maxAllowedDistanceMoved) {
 #pragma omp critical
-      std::cerr << "A particle moved farther than skin/2/rebuildFrequency: " << std::sqrt(distanceMovedSquared) << " > "
-                << autoPasContainer.getVerletSkin() << "/2/" << autoPasContainer.getVerletRebuildFrequency() << " = "
-                << maxAllowedDistanceMoved << "\n"
+      std::cerr << "A particle moved farther than verletSkinPerTimestep/2: " << std::sqrt(distanceMovedSquared) << " > "
+                << autoPasContainer.getVerletSkinPerTimestep() << "/2 = " << maxAllowedDistanceMoved << "\n"
```
thesamriel, line 48: If we change the calculation here, then maybe also above in the actual calculation of `maxAllowedDistanceMoved`?
`const auto maxAllowedDistanceMoved = autoPasContainer.getVerletSkinPerTimestep() / 2`

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
```diff
@@ -39,6 +40,16 @@ RegularGridDecomposition::RegularGridDecomposition(const MDFlexConfig &configura
 
   _decomposition = DomainTools::generateDecomposition(_subdomainCount, configuration.subdivideDimension.value);
 
+  // For reflection, we interact particles with their 'mirror' only if it would cause a repulsive force. This occurs
+  // only for particles within sixthRootOfTwo * sigma / 2.0 of the boundary. For minimal redundant iterating over
+  // particles, we iterate once over particles within the largest possible range where a particle might experience a
+  // repulsion, i.e. sixthRootOfTwo * maxSigma / 2.0.
```
thesamriel, line 46: Where do we/you know this from?

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
```diff
@@ -388,34 +424,37 @@ RegularGridDecomposition::categorizeParticlesIntoLeftAndRightNeighbor(const std:
...
+        if (isNear(_localBoxMax[direction], _globalBoxMax[direction])) {
+          // TODO: check if this failsafe is really reasonable.
+          //  It should only trigger if a particle's position was already inside the box?
+          const auto periodicPosition = position[direction] - globalBoxLength[direction];
+          position[direction] = std::max(_globalBoxMin[direction], periodicPosition);
+          rightNeighborParticles.back().setR(position);
+        }
+      } else {
+        uncategorizedParticles.push_back(particle);
+      }
```
thesamriel, line 450: Why the change of the if-else structure?
From 
```cpp
if () ...
else if () ...
else ...
```
to 
```cpp
if () ...
else {
  if () ...
  else ...
}
```

src/autopas/AutoPasDecl.h
```diff
@@ -130,25 +130,89 @@ class AutoPas {
...
    * @note An exception is thrown if the particle is added and it is not inside of the owned domain (defined by
    * boxmin and boxmax) of the container.
-   * @note This function is NOT thread-safe.
+   * @note This function is NOT thread-safe if the container is Octree.
    */
   void addParticle(const Particle &p);
 
+  /**
+   * Adds all particles from the collection to the container.
+   * @note This function uses addParticlesIf().
```
thesamriel, line 167: I don't see where this function uses `addParticlesIf()`

thesamriel, line 167: I don't see where this function uses `addParticlesIf()` 
```suggestion
   * @note This function uses addParticle().
```

src/autopas/AutoPasDecl.h
```diff
@@ -130,25 +130,89 @@ class AutoPas {
...
    * @note An exception is thrown if the halo particle is added and it is inside of the owned domain (defined by boxmin
    * and boxmax) of the container.
-   * @note This function is NOT thread-safe.
+   * @note This function is NOT thread-safe if the container is Octree.
    */
   void addHaloParticle(const Particle &haloParticle);
 
+  /**
+   * Adds all halo particles from the collection to the container.
+   * @note This function uses addHaloParticlesIf().
```
thesamriel, line 197:
```suggestion
   * @note This function uses addHaloParticle().
```

src/autopas/AutoPasImpl.h
```diff
@@ -111,11 +111,65 @@ size_t AutoPas<Particle>::getNumberOfParticles(IteratorBehavior behavior) const
...
 }
 
+template <class Particle>
+void AutoPas<Particle>::reserve(size_t numParticles) {
+  _logicHandler->reserve(numParticles);
+}
+
+template <class Particle>
+void AutoPas<Particle>::reserve(size_t numParticles, size_t numHaloParticles) {
+  _logicHandler->reserve(numParticles);
```
thesamriel, line 121:
```suggestion
  _logicHandler->reserve(numParticles, numHaloParticles);
```

src/autopas/CMakeLists.txt
```diff
@@ -23,9 +23,10 @@ target_link_libraries(
         spdlog::spdlog
         $<$<OR:$<BOOL:${AUTOPAS_INTERNODE_TUNING}>,$<BOOL:${MD_FLEXIBLE_USE_MPI}>>:MPI::MPI_CXX>
         $<$<BOOL:${AUTOPAS_INCLUDE_MPI}>:MPI::MPI_CXX>
+        Eigen3
     PRIVATE
         # harmony and Eigen3 are only needed privately when building AutoPas.
```
thesamriel, line 28:
```suggestion
        # harmony is only needed privately when building AutoPas.
```

src/autopas/LogicHandler.h
```diff
@@ -455,12 +490,12 @@ class LogicHandler {
...
    */
-  std::vector<std::vector<Particle>> _particleBuffer;
+  std::vector<FullParticleCell<Particle>> _particleBuffer;
 
   /**
    * Buffer to store halo particles that should not yet be added to the container. There is one buffer per thread.
    * @note This buffer could potentially be replaced by a ParticleCell.
    */
-  std::vector<std::vector<Particle>> _haloParticleBuffer;
+  std::vector<FullParticleCell<Particle>> _haloParticleBuffer;
```
thesamriel, line 499: Remove note on potentially removing the buffer with ParticleCell. Or do you plan to also replace the outer vector with a ParticleCell?

Also what is the advantage of using FullParticleCell here instead of vector<>?

src/autopas/LogicHandler.h
```diff
@@ -310,15 +346,15 @@ class LogicHandler {
...
+          additionalVectors.push_back(&(buffer._particles));
         }
       }
     }
     if (behavior & IteratorBehavior::halo) {
       for (auto &buffer : _haloParticleBuffer) {
-        if (not buffer.empty()) {
-          additionalVectors.push_back(&buffer);
+        if (not buffer.isEmpty()) {
+          additionalVectors.push_back(&(buffer._particles));
```
thesamriel, line 357: Maybe use 
```cpp
for (auto &cell : _particleBuffer) {
auto &buffer = cell._particles;
```
 to clarify that the buffer is a cell as you did in `deleteParticlesFromBuffer()`

src/autopas/cells/FullParticleCell.h
```diff
@@ -250,15 +248,14 @@ class FullParticleCell : public ParticleCell<Particle> {
...
    * Storage of the molecules of the cell.
    */
-  StorageType _particles;
+  StorageType _particles{};
 
   /**
    * SoA buffer of this cell.
    */
-  SoA<SoAArraysType> _particleSoABuffer;
+  SoA<SoAArraysType> _particleSoABuffer{};
```
thesamriel, line 256: _Side note:_

Why are these members public, but use the underscore syntax, indicating a private member.  
Maybe getters would be nicer. Or is it on purpose to avoid any overhead?

src/autopas/utils/ParticleCellHelpers.h
```diff
@@ -38,6 +38,7 @@ static bool checkParticleInCellAndUpdateByID(CellType &cell, const typename Cell
 template <class CellType>
 static bool checkParticleInCellAndUpdateByIDAndPosition(CellType &cell, const typename CellType::ParticleType &particle,
                                                         double absError) {
+  std::lock_guard<AutoPasLock> cellLock(cell._cellLock);
```
thesamriel, line 41: Do we need the lock here?
If so, I guess also in the function `checkParticleInCellAndUpdateByID()`

src/autopas/selectors/AutoTuner.h
```diff
@@ -452,120 +510,197 @@ bool AutoTuner<Particle>::iteratePairwise(PairwiseFunctor *f, bool doListRebuild
...
 
       // 2. haloParticleBuffer with owned, close particles in container
-      for (auto &&p1 : haloParticleBuffer) {
-        auto pos = p1.getR();
-        auto min = autopas::utils::ArrayMath::subScalar(pos, cutoff);
-        auto max = autopas::utils::ArrayMath::addScalar(pos, cutoff);
+      for (auto &&p1halo : haloParticleBuffer) {
+        const auto pos = p1halo.getR();
+        const auto min = autopas::utils::ArrayMath::subScalar(pos, cutoff);
+        const auto max = autopas::utils::ArrayMath::addScalar(pos, cutoff);
```
thesamriel, line 585:
```suggestion
        const auto min = subScalar(pos, cutoff);
        const auto max = addScalar(pos, cutoff);
```

src/autopas/selectors/AutoTuner.h
```diff
@@ -452,120 +510,197 @@ bool AutoTuner<Particle>::iteratePairwise(PairwiseFunctor *f, bool doListRebuild
...
-            f->AoSFunctor(p1, p2, true);
+    } else {
+#ifdef AUTOPAS_OPENMP
+      // collapse loops for better scheduling
+#pragma omp for collapse(2)
+#endif
+      for (size_t i = 0; i < particleBuffers.size(); ++i) {
+        for (size_t jj = 0; jj < particleBuffers.size(); ++jj) {
+          auto *particleBufferSoAA = &particleBuffers[i]._particleSoABuffer;
+          const auto j = (i + jj) % particleBuffers.size();
```
thesamriel, line 650: What is the reason for this? Creating diffusion for better scheduling?

examples/md-flexible/tests/domainDecomposition/ReflectiveBoundaryConditionTest.cpp
```diff
@@ -105,31 +139,167 @@ TEST_P(ReflectiveBoundaryConditionTest, simpleReflectionTest) {
...
                                     std::array<bool, 3>{false, false, false}))
     //    ,ReflectiveBoundaryConditionTest::PrintToStringParamName());
-);
\ No newline at end of file
+);
+
+/**
+ * Implements the reflective boundary zoning test.
+ * @param particlePosition
+ * @param particleType Must be either 0 or 1. 0 corresponds to a sigma of 0.5, 1 corresponds to a sigma of 1.0.
```
thesamriel, line 174:
```suggestion
 * @param particleType Must be either 0 or 1. 0 corresponds to a sigma of 0.1, 1 corresponds to a sigma of 0.2.
```

examples/md-flexible/tests/domainDecomposition/MixedBoundaryConditionTest.cpp
```diff
@@ -47,9 +65,10 @@ auto MixedBoundaryConditionTest::setUpExpectations(
           break;
         case ::options::BoundaryTypeOption::reflective:
           // if near a reflective boundary and flying towards it the velocity sign is flipped
```
thesamriel, line 67: Comment is no longer true right?
```suggestion
          // if near a reflective boundary and flying towards it we expect a repulsive force from reflection
```

examples/md-flexible/src/TimeDiscretization.cpp
FG-TUM: [d991b24](https://github.com/AutoPas/AutoPas/pull/724/commits/d991b242d7d6c4bcb195074e7cc7bc31da94f5c6)

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
FG-TUM: Ask Sam, he wrote that :P
If you look at the LJ Potential, the distance where attraction flips to repulsion this happens at $\sqrt[6]{2} \sigma$. So for a particle to have its mirror particle in that distance, it must be $\sqrt[6]{2} \sigma / 2$ away from the mirror.

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
```diff
@@ -252,8 +271,12 @@ void RegularGridDecomposition::exchangeMigratingParticles(AutoPasType &autoPasCo
   }
 }
 
-void RegularGridDecomposition::reflectParticlesAtBoundaries(AutoPasType &autoPasContainer) {
+void RegularGridDecomposition::reflectParticlesAtBoundaries(AutoPasType &autoPasContainer,
+                                                            ParticlePropertiesLibraryType &particlePropertiesLib) {
+  using autopas::utils::Math::isNear;
   std::array<double, _dimensionCount> reflSkinMin{}, reflSkinMax{};
+  auto functorLJ = autopas::LJFunctor<ParticleType, false, true, autopas::FunctorN3Modes::Both, false, false>(
+      _maxReflectiveSkin * 2., particlePropertiesLib);
```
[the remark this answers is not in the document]
FG-TUM: Since md-flexible (for now?) always does MD Simulations with the LJ potential this seems reasonable. @SamNewcome btw did we think about what the implications for the potential energy come from the boundary interactions?

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
FG-TUM: No real reason, but also the two structures should be equivalent.
[2d81b4e](https://github.com/AutoPas/AutoPas/pull/724/commits/2d81b4eba1f70d2c841e27e450b169122a5a65bc)

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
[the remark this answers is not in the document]
SamNewcome: I think hard coding is fine for now. Using an LJ functor for reflection becomes nasty with multi-site molecules anyway (and I suppose similarly for other rigid-body-type particles e.g. in DEM), so I was going to switch this to its own kernel anyway.

We never thought of the implications for potential energy from reflective boundaries. I would argue, though, that working around this missing potential energy is beyond the scope of md-flexible. I don't think the calculations are necessarily correct even with e.g. periodic boundaries (e.g. I don't think we add potential energy from global forces)

examples/md-flexible/tests/domainDecomposition/MixedBoundaryConditionTest.cpp
SamNewcome: Yes. Plus the "flying towards it" part is no longer needed.

examples/md-flexible/src/TimeDiscretization.cpp
```diff
@@ -20,9 +20,7 @@ void calculatePositions(autopas::AutoPas<ParticleType> &autoPasContainer,
   using autopas::utils::ArrayMath::dot;
   using autopas::utils::ArrayMath::mulScalar;
 
-  const auto maxAllowedDistanceMoved =
-      autoPasContainer.getVerletSkin() / (2 * autoPasContainer.getVerletRebuildFrequency());
-  const auto maxAllowedDistanceMovedSquared = maxAllowedDistanceMoved * maxAllowedDistanceMoved;
```
SamNewcome, line 25: Keep `maxAllowedDistanceMovedSquared`. There appears to be some error in the below code. On line 45 we compare `distanceMovedSquared` with `maxAllowedDistanceMoved` when it should be with `maxAllowedDistanceMovedSquared`

SamNewcome (changes requested):

examples/md-flexible/tests/domainDecomposition/MixedBoundaryConditionTest.cpp
FG-TUM: [2d81b4e](https://github.com/AutoPas/AutoPas/pull/724/commits/2d81b4eba1f70d2c841e27e450b169122a5a65bc)

examples/md-flexible/tests/domainDecomposition/ReflectiveBoundaryConditionTest.cpp
FG-TUM: [2d81b4e](https://github.com/AutoPas/AutoPas/pull/724/commits/2d81b4eba1f70d2c841e27e450b169122a5a65bc)

examples/md-flexible/src/TimeDiscretization.cpp
FG-TUM: Good catch. [9024538](https://github.com/AutoPas/AutoPas/pull/724/commits/9024538819583e7b1b486e58547f288071dcc64f)

src/autopas/selectors/AutoTuner.h
FG-TUM: Yep. If we had a regular `i : 0 -> N` and `j : 0 -> N` loop all tasks that need the same `j` are created at the same time, increasing the probability of running into a lock.

src/autopas/cells/FullParticleCell.h
FG-TUM: puh... if I recall correctly this was initially a private member, hence the underscore. At some point, we made it public for ease of use. But I think we have a few naming inconsistencies like this one in the code and I think if we want to clean this up it should be a dedicated branch.

src/autopas/LogicHandler.h
FG-TUM: Those lines were only because before I changed the buffers to cells the code was
```cpp
for (auto &buffer : _particleBuffer ) {
```
Then, after the change, I needed to access the `_particles` member but since this (Cell) has the same interface as a `std::vector` adding this alias variable was a quick way to adapt the code with minimal changes ;)

Tbh I'm not sure I would change this here, because it would muddy the meaning of "buffer" even further (we'd call a member of an object in the buffer collection a buffer...)

src/autopas/LogicHandler.h
FG-TUM: lol yes. This is exactly what was done in this PR -> 623ba0cc4

FG-TUM: The advantage is that it makes it easier to use an SoA Functor with the particles, because the cell has the logic for transforming the particle data.

src/autopas/CMakeLists.txt
FG-TUM: 2fe5fe53d

thesamriel (approved):

PR #725, opened 2023-03-08
Extend cubeClosestPacked for md_flexible

thesamriel:
# Description

- Add `getParticlesTotal()` for cubeClosestPacked 
- Add cubeClosestPacked object to md_flexible tests

# How Has This Been Tested?
Calculation has been tested with all combinations of odd or even rows, columns, layers

### Example image for better understanding:
image
`xNumRow` = 4
`xOdd` = true, da insgesamt 7 x-Positionen
`yNumEven` = 4 (dunkelblaue, vorderste layer)
`yNumOdd` = 3 (hellblauere, hintere layer)

### DIFF
--- a/examples/md-flexible/src/configuration/objects/CubeClosestPacked.h
+++ b/examples/md-flexible/src/configuration/objects/CubeClosestPacked.h
@@ -47,30 +47,30 @@ class CubeClosestPacked : public Object {
   [[nodiscard]] size_t getParticlesTotal() const override {
     // Shorter part of the bisectrix when split at the intersection of all bisectrices.
     const double xOffset = _particleSpacing * 1. / 2.;
-    // Number of particles in the first row
+    // Number of particles in the first row.
     const size_t xNumRow = std::ceil(_boxLength[0] / _particleSpacing);
-    // True if total number of x-positions is odd
+    // True if the total number of x-positions is odd.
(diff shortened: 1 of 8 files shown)


examples/md-flexible/src/configuration/objects/CubeClosestPacked.h
```diff
@@ -69,17 +103,20 @@ class CubeClosestPacked : public Object {
...
   void generate(std::vector<ParticleType> &particles) const override {
     ParticleType particle = getDummyParticle(particles.size());
-
+    // Spacing in y direction when only moving 60° on the unit circle. Or the height in an equilateral triangle.
     const double spacingRow = _particleSpacing * sqrt(3. / 4.);
+    // Spacing in z direction. Height in an equilateral tetraeder.
     const double spacingLayer = _particleSpacing * sqrt(2. / 3.);
+    // Shorter part of the bisectrix when split at the intersection of all bisectrices.
     const double xOffset = _particleSpacing * 1. / 2.;
+    // Shorter part of the bisectrix when split at the intersection of all bisectrices.
```
FG-TUM, line 112: If I see this correctly, you compute the same values in `getParticlesTotal()`. Probably makes sense to promote these values to class members and calculate them in the constructor so we avoid the code duplication of the calculation.

thesamriel: I agree, makes the methods look cleaner as well.

FG-TUM (approved):

PR #728, opened 2023-03-24
Variable verlet lists skin

pandeddy:
# Description

Verlet rebuild frequency is tunable


src/autopas/selectors/FeatureVectorEncoder.h
SamNewcome, line 1: So I've looked at this and I believe the original code to be confusing. I don't entirely know the best way to proceed with this currently. 

@pandeddy I would not worry about this file but I will write some comments for myself or whoever finishes off this branch. I also suspect these confusions might be why that test was failing.

-------------------------
The original author of the Bayesian search methods split the search space into "discrete" dimensions and "continuous" dimensions. In the "discrete" dimensions were the likes of containers, traversals, etc. In the "continuous" dimension, the only example given was cell size factor. At no point during the thesis, does he mention actual discrete number sets e.g. integers. I suspect that these were meant to be considered from an optimisation perspective as "continuous" - as this is commonplace in the field of Bayesian optimisation. In this case, this needs to be changed in this file.

src/autopas/selectors/tuningStrategy/ActiveHarmony.h
```diff
@@ -232,9 +238,10 @@ void ActiveHarmony::fetchConfiguration() {
   } else {
     cellSizeFactor = ah_get_real(htask, cellSizeFactorsName);
   }
+  int verletRebuildFrequency = _allowedVerletRebuildFrequencies->getMin();
```
SamNewcome, line 241: This should be handled like above for `cellSizeFactor`

src/autopas/selectors/tuningStrategy/BayesianClusterSearch.h
```diff
@@ -80,13 +81,15 @@ class BayesianClusterSearch : public TuningStrategyInterface {
       const std::set<TraversalOption> &allowedTraversalOptions = TraversalOption::getAllOptions(),
       const std::set<LoadEstimatorOption> &allowedLoadEstimatorOptions = LoadEstimatorOption::getAllOptions(),
       const std::set<DataLayoutOption> &allowedDataLayoutOptions = DataLayoutOption::getAllOptions(),
-      const std::set<Newton3Option> &allowedNewton3Options = Newton3Option::getAllOptions(), size_t maxEvidence = 10,
+      const std::set<Newton3Option> &allowedNewton3Options = Newton3Option::getAllOptions(),
+      const std::set<int> &allowedVerletRebuildFrequencies = std::set<int>({12, 24, 48}), size_t maxEvidence = 10,
```
SamNewcome, line 85: Is there a reason that this is not a `NumberSet`?

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -50,24 +50,28 @@ class PredictiveTuning : public SetSearchSpaceBasedTuningStrategy {
    * @param allowedDataLayoutOptions
    * @param allowedNewton3Options
    * @param allowedCellSizeFactors
+   * @param allowedVerletRebuilFrequencies
```
SamNewcome, line 53:
```suggestion
   * @param allowedVerletRebuildFrequencies
```

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -50,24 +50,28 @@ class PredictiveTuning : public SetSearchSpaceBasedTuningStrategy {
...
    */
+
   PredictiveTuning(const std::set<ContainerOption> &allowedContainerOptions,
                    const std::set<double> &allowedCellSizeFactors,
                    const std::set<TraversalOption> &allowedTraversalOptions,
                    const std::set<LoadEstimatorOption> &allowedLoadEstimatorOptions,
                    const std::set<DataLayoutOption> &allowedDataLayoutOptions,
-                   const std::set<Newton3Option> &allowedNewton3Options, double relativeOptimum,
+                   const std::set<Newton3Option> &allowedNewton3Options,
+                   const std::set<int> &allowedVerletRebuilFrequencies, double relativeOptimum,
```
SamNewcome, line 68:
```suggestion
                   const std::set<int> &allowedVerletRebuildFrequencies, double relativeOptimum,
```

src/autopas/selectors/tuningStrategy/PredictiveTuning.h
```diff
@@ -50,24 +50,28 @@ class PredictiveTuning : public SetSearchSpaceBasedTuningStrategy {
...
                    unsigned int maxTuningIterationsWithoutTest, double relativeRangeForBlacklist,
                    unsigned int testsUntilFirstPrediction, ExtrapolationMethodOption extrapolationMethodOption,
                    const std::string &outputSuffix = "")
       : SetSearchSpaceBasedTuningStrategy(allowedContainerOptions, allowedCellSizeFactors, allowedTraversalOptions,
-                                          allowedLoadEstimatorOptions, allowedDataLayoutOptions, allowedNewton3Options),
+                                          allowedLoadEstimatorOptions, allowedDataLayoutOptions, allowedNewton3Options,
+                                          allowedVerletRebuilFrequencies),
```
SamNewcome, line 74:
```suggestion
                                          allowedVerletRebuildFrequencies),
```

src/autopas/selectors/tuningStrategy/TuningStrategyFactory.cpp
```diff
@@ -60,14 +62,16 @@ std::unique_ptr<autopas::TuningStrategyInterface> autopas::TuningStrategyFactory
         fallbackCellSizeFactors = std::make_unique<autopas::NumberInterval<double>>(allowedCellSizeFactors.getMin(),
                                                                                     allowedCellSizeFactors.getMax());
       }
+      fallbackVerletRebuildFrequencies =
+          std::make_unique<autopas::NumberSetFinite<int>>(allowedVerletRebuildFrequencies.getAll());
```
SamNewcome, line 66: Please handle this like `cellSizeFactors`, where `NumberSetFinite` is only used if the set is finite, and if it is not `NumberInterval` is used.

src/autopas/selectors/tuningStrategy/TuningStrategyFactory.cpp
```diff
@@ -95,23 +99,24 @@ std::unique_ptr<autopas::TuningStrategyInterface> autopas::TuningStrategyFactory
...
+          allowedContainers, allowedCellSizeFactors, allowedTraversals, allowedLoadEstimators, allowedDataLayouts,
+          allowedVerletRebuildFrequencies, allowedNewton3Options, maxEvidence, acquisitionFunctionOption);
       break;
     }
 
     case TuningStrategyOption::bayesianClusterSearch: {
       tuningStrategy = std::make_unique<BayesianClusterSearch>(
           allowedContainers, allowedCellSizeFactors, allowedTraversals, allowedLoadEstimators, allowedDataLayouts,
-          allowedNewton3Options, maxEvidence, acquisitionFunctionOption, outputSuffix);
+          allowedNewton3Options, allowedVerletRebuildFrequencies.getAll(), maxEvidence, acquisitionFunctionOption,
```
SamNewcome, line 118: Why is `getAll()` used unlike `allowedCellSizeFactors`?

src/autopas/utils/AutoPasConfigurationCommunicator.cpp
```diff
@@ -89,7 +93,7 @@ void generateDistribution(const int numConfigs, const int commSize, const int ra
 
   ConfigurationAndRankIteratorHandler iteratorHandler(containerOptions, finiteCellSizeFactors, traversalOptions,
                                                       loadEstimatorOptions, dataLayoutOptions, newton3Options,
-                                                      numConfigs, commSize);
+                                                      verletRebuildFrequencies.getAll(), numConfigs, commSize);
```
SamNewcome, line 96: Is there a reason why this is not handled like `cellSizeFactors`? If not, please handle it like that

SamNewcome (commented):
Only about 50% through the code review. Will continue tomorrow.

tests/testAutopas/tests/selectors/FeatureVectorTest.cpp
```diff
@@ -128,14 +131,15 @@ TEST_F(FeatureVectorTest, clusterEncode) {
...
+                               NumberSetFinite<double>({cellSizeFactor}),
+                               autopas::NumberSetFinite<int>(std::set<int>({5, 15, 30})));
 
   // generate all possible combinations
   std::vector<FeatureVector> vecList;
   for (const auto &[container, traversal, estimator] : allCompatibleContainerTraversalEstimators) {
     for (const auto &dataLayout : dataLayouts) {
       for (const auto &newton3 : newtons) {
-        vecList.emplace_back(container, cellSizeFactor, traversal, estimator, dataLayout, newton3);
+        vecList.emplace_back(container, cellSizeFactor, traversal, estimator, dataLayout, newton3, 5);
```
SamNewcome, line 142: This should not be a set just consisting of `5`. Please emplace into `vecList` the other rebuild frequencies that you allow, i.e. 15 & 30. Alternatively, don't test multiple rebuild frequencies - but in this case, make only a rebuild frequency of 5 allowed on line 135.

src/autopas/utils/ConfigurationAndRankIteratorHandler.h
```diff
@@ -121,6 +124,14 @@ class ConfigurationAndRankIteratorHandler {
    */
   [[nodiscard]] inline std::set<double>::iterator getCellSizeFactorIterator() const { return _cellSizeFactorIt; }
 
+  /**
+   * Getter for the VerletRebuildFrequenzyIterator.
```
SamNewcome, line 128:
```suggestion
   * Getter for the VerletRebuildFrequencyIterator.
```

src/autopas/selectors/tuningStrategy/ActiveHarmony.h
SamNewcome: Alternatively, if it shouldn't be handled like `cellSizeFactor`, add a comment to explain why.

examples/md-flexible/input/crashingCubes.yaml
```diff
@@ -0,0 +1,47 @@
+container                        :  [VerletLists]
+traversal                        :  [lc_sliced, lc_sliced_balanced, lc_sliced_c02, lc_c01, lc_c01_combined_SoA, lc_c04, lc_c04_HCP, lc_c04_combined_SoA, lc_c08, lc_c18, vcl_c06, vcl_c01_balanced, vcl_sliced, vcl_sliced_balanced, vcl_sliced_c02, vl_list_iteration, vlc_c01, vlc_c18, vlc_sliced, vlc_sliced_balanced, vlc_sliced_c02, vvl_as_built, vlp_c01, vlp_c18, vlp_sliced, vlp_sliced_balanced, vlp_sliced_c02]
```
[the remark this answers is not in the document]
pandeddy: yes, but then you have to change traversal when switching between VerletLists and VerletListsCells as I did.

src/autopas/AutoPasDecl.h
```diff
@@ -438,7 +438,7 @@ class AutoPas {
    * @return _verletSkin
    */
   double getVerletSkin() {
-    double _verletSkin = AutoPas::_verletSkinPerTimestep * AutoPas::_verletRebuildFrequency;
+    double _verletSkin = AutoPas::_verletSkinPerTimestep * _allowedVerletRebuildFrequencies->getMedian();
     return _verletSkin;
   };
```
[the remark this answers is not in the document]
pandeddy: I can't access the current rebuildfrequency because getcurrentconfig doesn't work

src/autopas/AutoPasDecl.h
```diff
@@ -976,6 +1001,11 @@ class AutoPas {
   std::unique_ptr<NumberSet<double>> _allowedCellSizeFactors{
       std::make_unique<NumberSetFinite<double>>(std::set<double>({1.}))};
 
+  std::unique_ptr<NumberSet<int>> _allowedVerletRebuildFrequencies{
+      std::make_unique<NumberSetFinite<int>>(std::set<int>({5}))};
+
+  int _verletRebuildFrequency = 5;
```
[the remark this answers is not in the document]
pandeddy: invalid use of incomplete type ‘class autopas::AutoTuner<autopas::MoleculeLJ<> >’
  810 |   [[nodiscard]] const Configuration &getCurrentConfig() const { return _autoTuner->getCurrentConfig(); }

would have been my approach

src/autopas/selectors/tuningStrategy/BayesianClusterSearch.h
pandeddy: At one point it didn't work with Numberset and I couldn't find the reason.

src/autopas/selectors/tuningStrategy/TuningStrategyFactory.cpp
pandeddy: It is always finite. Therefore no need of the NumberInterval

src/autopas/selectors/tuningStrategy/TuningStrategyFactory.cpp
pandeddy: because i use set in the Interface. As mentioned before it didn't work with Numberset

src/autopas/utils/StringUtils.h
```diff
@@ -258,4 +289,43 @@ inline std::unique_ptr<autopas::NumberSet<double>> parseNumberSet(const std::str
...
+      }
+      return std::make_unique<autopas::NumberSetFinite<int>>(std::set<int>({numbers}));
+    } catch (const std::exception &) {
+      // try parseDoubles instead
+    }
+  }
+
+  std::set<int> values = autopas::utils::StringUtils::parseInts(setString);
+  return std::make_unique<autopas::NumberSetFinite<int>>(values);
+}
```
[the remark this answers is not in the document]
pandeddy: yes but doubles use numberinterval. I don't think this makes sense for ints (verletrebudfreuqencies).

src/autopas/selectors/tuningStrategy/BayesianClusterSearch.h
SamNewcome: What line of code did you try and what was the error?

src/autopas/selectors/tuningStrategy/TuningStrategyFactory.cpp
SamNewcome: Currently, AutoPas does not appear to be able to handle unbounded sets, so yes allowed rebuild frequencies is always finite. But I would like to be able to handle unbounded sets in the future, in which case allowed rebuild frequencies could be infinite. 

This doesn't explicitly have to be done in your pull request, so if you would like to leave your code as it is, simply add a comment above this line saying that AutoPas assumes Verlet Rebuild Frequency is bounded so it is always finite.

src/autopas/utils/StringUtils.h
```diff
@@ -251,6 +251,55 @@ inline std::set<int> parseInts(const std::string &intString) {
...
+                           "-"      // a dash
+                           "\\s*"   // maybe more whitespaces
+                           "("      // start of 2. capture
+                           + str +  // a double
+                           ")"      // end of 2. capture
+  );
+  std::smatch matches;
+
+  std::set<N> values;
+  if (typeid(N) == typeid(int)) {
```
SamNewcome, line 274:
```suggestion
  if constexpr (typeid(N) == typeid(int)) {
```

src/autopas/utils/StringUtils.h
```diff
@@ -251,6 +251,55 @@ inline std::set<int> parseInts(const std::string &intString) {
   return ints;
 }
 
+template <class N>
+inline std::unique_ptr<autopas::NumberSet<N>> parseNumberSet(const std::string &setString) {
+  std::string str = regexDoubleStr;
+  if (typeid(N) == typeid(int)) {
+    str = regexIntStr;
+  }
```
SamNewcome, line 259: It would be cleaner to handle this inside the `if`-statement below.

SamNewcome, line 259: It would be cleaner to handle this inside the `if`-statement below.

src/autopas/utils/StringUtils.h
```diff
@@ -251,6 +251,55 @@ inline std::set<int> parseInts(const std::string &intString) {
   return ints;
 }
 
+template <class N>
```
SamNewcome, line 254:
```suggestion
template <class T>
```
Nothing wrong with `N`, but `T` is conventional.

src/autopas/utils/StringUtils.h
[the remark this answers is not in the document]
SamNewcome: Okay, I will look into this error. But the combined function you've written is much cleaner than the two functions.

SamNewcome (changes requested):
Nice! There are still some issues from the first review that you haven't dealt with yet, don't forget about them.

src/autopas/AutoPasDecl.h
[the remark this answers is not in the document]
pandeddy: I mean it's the same issue as in get VerletSkin. My first Approach was using currentConfig but it is not always initialized.

src/autopas/selectors/tuningStrategy/BayesianClusterSearch.h
pandeddy: I meant, it didn't compile at all.

src/autopas/selectors/tuningStrategy/TuningStrategyFactory.cpp
pandeddy: I didn't think about this case. You are right and I tried to change this, but at the moment there is no class available for a NumberSet with no Maximum. I think the person who likes to have this case should be able to create the class as wanted.
If you think oherwise I can create a new class for this.

src/autopas/utils/StringUtils.h
pandeddy: Yeah had the same thought. But than i need one more if clause afterwards because either way I want to parse the String. But the method has to be chosen. I could changes these though if you like it more

SamNewcome:
Closing PR as quite out-of-date with `master` and tuning rebuild frequency not so interesting a concept with dynamic Verlet Lists (where, if anything, the verlet skin itself is tuned).

PR #737, opened 2023-04-24
Fix GaussianProcess Tests

FG-TUM:
# Description

After updating the compiler versions in our docker containers, some `GaussianProcessTest`s fail. This is only because they use very few samples in order to keep runtime as low as possible. This PR adjusts the numbers so that the tests pass again.

## ~Resolved Issues~

# How Has This Been Tested?
CI

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Alternatively you can build the documentation on your own:
 ## Requirements
 * CMake 3.14 or newer
 * make (build-essentials) or ninja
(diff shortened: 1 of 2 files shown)


thesamriel (approved):
The number fixes seem random, but look fine to me.

FG-TUM:
> The number fixes seem random, but look fine to me.

Very true. The increased number of samples increases the accuracy of the prediction but the change in the rng seed is just for "better luck"...

Issue #740, opened 2023-04-28
YamlParser::parseYamlFile() is missing if statement for vtk-output-file


TumBeat:
**Describe the bug**
It seems like the `parseYamlFile()` has no if-else statement for "vtk-output-folder". `MDFlexConfig.h` has a definition for the `vtkOutputFolder` property (see screenshot) but at the moment this variable is only read and never changed.
When trying to set a custom output folder via `vtk-output-folder` in a yaml file, the attached error message occurs saying that this option was not recognized (see screenshot).

**To Reproduce**
Add a line
`vtk-output-folder   : myFancyFolder`
to a chosen input yaml file and run the md-flexible program with this input file.

**Expected behavior**
The specified output folder from the yaml-file should be read and used.

**Screenshots**
Config in `MDFlexConfig.h`
![MDFlexConfig](https://user-images.githubusercontent.com/111445698/235192169-48bc9734-23e2-4322-9e4d-2f1e09a073f6.png)
Error message
![Yaml-issue](https://user-images.githubusercontent.com/111445698/235191885-8bb9982b-474f-4624-8231-12cb6a698be2.png)

**Build environment (please complete the following information):**
* OS: Linux Cluster, LRZ
* Compiler: gcc 11.2.0
* Build System: CMake

TumBeat:
I'm already on it.
I need the feature of setting a custom output folder for running the md-flexible code on the linux cluster to specify that the output folder is $SCRATCH. Just stumbled accros this when I was just trying to test something there.
But I agree, in general the importance is kind of low and for now, I can use the command line for specifying this :D

FG-TUM:
Tip for programs where you can't change this yourself: If you'd create a symlink of the destination folder `output`, which `md-flex` uses, which points to `$SCRATCH` then this would be used.
So what I'm trying to say:
```bash
ln -s ${SCATCH}/whatever ~/your/AutoPas/buildDir/output
```

FG-TUM:
fixed via #741

PR #741, opened 2023-04-28
Bugfix/yaml parser missing vtk output folder

TumBeat:
# Description
`YamlParser::parseYamlFile()`  was missing an if statement for the vtk-output-folder option. Because of this, one was not able to set a custom output folder via the yaml input file.

(Edit, 30.04.):
Same issue with the `CLIParser::parseInput()`.

## Resolved Issues

- [x] #740 

# How Has This Been Tested?

This fix was tested with new unit tests and by running the md-flexible program with the changed fallingDrop.yaml input file.

- [x] Specified a custom output folder and verified that the vtk-files were actually stored there
- [x] Wrote a new unit test to verify that the options in the input file end up in the configuration object
- [x] Tested the CLI-Option with calling md-flexible with "./md-flexible --yaml-filename fallingDrop.yaml  --vtk-output-folder fancyFolderName" and observed that the output files go into the correct folder


FG-TUM:
From the top of my head, this also has to go into the `CLIParser::relevantOptions`. Maybe also have a look at the `CONTRIBUTING.md` file to check if I forgot something else.

dmartin13:
It might also be a good idea to add this missing option to AllOptions.yaml.

TumBeat:
> From the top of my head, this also has to go into the `CLIParser::relevantOptions`. Maybe also have a look at the `CONTRIBUTING.md` file to check if I forgot something else.

Seems like the CLIParser as well does not have the relevant case statement so it was not possible to specify an custom output folder at all or did I miss something?
I will add this as well in this PR and hope this is OK the main goal of this PR is about with the YamlParser.

FG-TUM:
> so it was not possible to specify an custom output folder at all or did I miss something?

Nope, that was never a feature. 

> I will add this as well in this PR and hope this is OK the main goal of this PR is about with the YamlParser.

Awesome!

PR #764, opened 2023-07-11
Created new Container for HierarchicalGrids

hoppef:
# Description

Created (draft version of) new Container for HierarchicalGrids consisting of "_numberOfLevels" levels of LinkedCells each with a different size.

This is realized using a vector of length "_numberOfLevels" containing LinkedCells grids.

# ToDos:

## Soonish
- [ ] Finish member functions marked with todo and still containing the "virtual" keyword.
- [ ] Incorporate function to determine the level the particles belong to, i.e. implement getHierarchyLevelOfParticle(p).

## Intermediate
- [ ] Modify iteratePairwise such that it first iterates within the levels of the HierarchicalGrid and then between the different levels

## Later
- [ ] Enable tuning over HierarchicalGrid properties...

# New files: 

src/autopas/containers/hierarchicalGrids/HierarchicalGrids.h

# How Has This Been Tested?

Not at all yet


FG-TUM:
Potential idea regarding en/disabling the container: If the provided particle class has no `radius` member we disable the generation of the function bodies via SFINAE.

SamNewcome:
> Potential idea regarding en/disabling the container: If the provided particle class has no `radius` member, we disable the generation of the function bodies via SFINAE.

Less of a fan of this. E.g. consider applying the container to LJ-MD. LJ Molecules have de-facto diameters of sigma. We could consider a situation where we might want to simulate molecules with very different sizes (i.e. different sigma), we might want to use hierarchical grids. But there is no radius member but a sigma member.

I would suggest giving the responsibility of not choosing this container to the user/developer. We could even have a cmake switch to disable compilation of this.

FG-TUM:
> I would suggest giving the responsibility of not choosing this container to the user/developer.

The problem is that we have to access the particle size (whatever that is) in the container via a defined interface. A flexible way that could be always available would be a method `double ParticleBase::getSize() = 0`, which has to be overwritten by every particle type with something reasonable.

> We could even have a cmake switch to disable compilation of this.

We could have this for every container. I'd be hesitant to only introduce this for a single container since this will probably lead to treating it as something weird and mostly unsupported.

src/autopas/containers/hierarchicalGrids/HierarchicalGrids.h
```diff
@@ -0,0 +1,644 @@
...
+   * @brief Carrier level containing all particles for functions like addParticle() etc.
+   * 
+   */
+  autopas::LinkedCells<Particle> _carrierLevel;
+
+  /**
+   * @brief Number of levels (hierarchies) in the H-Grid
+   * 
+   */
+  unsigned int _numberOfLevels;
```
FG-TUM, line 620: Why do you need this? Shouldn't this just be the length of `_hierarchyLevels`?

src/autopas/containers/hierarchicalGrids/HierarchicalGrids.h
```diff
@@ -0,0 +1,644 @@
...
+  * @brief Vector containing the HGrid's different hierarchy levels of Linked Cells.
+  * 
+  */
+  std::deque<LinkedCells<Particle>> _hierarchyLevels;
+
+  /**
+   * @brief Carrier level containing all particles for functions like addParticle() etc.
+   * 
+   */
+  autopas::LinkedCells<Particle> _carrierLevel;
```
FG-TUM, line 614: as discussed, this can potentially be replaced by `_hierarchyLevels[0]`

src/autopas/containers/hierarchicalGrids/HierarchicalGrids.h
```diff
@@ -0,0 +1,644 @@
...
+
+  std::array<double, 2> _particleRadiusRange;
+
+  std::array<double, 3> _boxMin; 
+  std::array<double, 3> _boxMax; 
+  double _cutoff;
+  double _skinPerTimestep; 
+  unsigned int _rebuildFrequency;
+  double _cellSizeFactor;
+  autopas::LoadEstimatorOption _loadEstimator;
```
FG-TUM, line 630: theoretically you could get rid of all of them too because they should also be stored in `_hierarchyLevels[0]`.

src/autopas/containers/hierarchicalGrids/HierarchicalGrids.h
```diff
@@ -0,0 +1,644 @@
...
+
+  /**
+   * @brief Vector containing the upper boundaries of the grid levels. 
+   * 
+   * For example, the first element describes the upper boundary of the largest H-Grid level,
+   * i.e. the largest particles fitting in the topmost level.
+   * By design '_hierarchicalGridBorders' should have the same '_numberOfLevels' as the H-Grid.
+   * 
+   */
+  std::vector<double> _hierarchicalGridBorders;
```
FG-TUM, line 640: Why is this needed? Is `_particleRadiusRange` not enough?

src/autopas/containers/hierarchicalGrids/HierarchicalGrids.h
```diff
@@ -0,0 +1,644 @@
...
+   * @param traversal The traversal to use for the iteration.
+   */
+  void iteratePairwise(TraversalInterface *traversal) override {
+
+    // Distribute particles over the hierarchical grid levels, if more than one HG level is present
+    if (_numberOfLevels > 1) {
+      
+      distributeParticlesToHGLevels();
+
+    }
```
FG-TUM, line 425: this should go into `updateContainer()`

src/autopas/containers/hierarchicalGrids/HierarchicalGrids.h
```diff
@@ -0,0 +1,644 @@
...
+
+  }
+
+  void iterateAllHGLevels(TraversalInterface *traversal) {
+
+    //AUTOPAS_OPENMP(parallel firstprivate (traversal))
+    for (auto & hgLevel : _hierarchyLevels) {
+
+      hgLevel.iteratePairwise(traversal);
```
FG-TUM, line 340: This could be an interesting optimization: Reduce the number of threads per level and parallelize as much as possible over the levels.

But I fear this might mean some major redesign all over autopas? :shrug:

src/autopas/containers/hierarchicalGrids/HierarchicalGrids.h
```diff
@@ -0,0 +1,644 @@
...
+   * @copydoc autopas::ParticleContainerInterface::getInteractionLength()
+   */
+  [[nodiscard]] double getInteractionLength() const override { return _cutoff + _skinPerTimestep * _rebuildFrequency; }
+
+  /**
+   * @TODO !!! Funktioniert so bestimmt nicht ... 
+   * 
+   * @copydoc autopas::ParticleContainerInterface::updateContainer()
+   */
+  [[nodiscard]] std::vector<ParticleType> updateContainer(bool keepNeighborListsValid) {
```
FG-TUM, line 482:
```suggestion
  [[nodiscard]] std::vector<ParticleType> updateContainer(bool keepNeighborListsValid) override {
```

check if this is missing somewhere else too

src/autopas/containers/hierarchicalGrids/HierarchicalGrids.h
```diff
@@ -0,0 +1,644 @@
...
+   */
+  [[nodiscard]] double getVerletSkin() const override { return _skinPerTimestep * _rebuildFrequency; }
+
+  /**
+   * @copydoc autopas::ParticleContainerInterface::getInteractionLength()
+   */
+  [[nodiscard]] double getInteractionLength() const override { return _cutoff + _skinPerTimestep * _rebuildFrequency; }
+
+  /**
+   * @TODO !!! Funktioniert so bestimmt nicht ...
```
FG-TUM, line 478: :rofl:

src/autopas/containers/hierarchicalGrids/HierarchicalGrids.h
```diff
@@ -0,0 +1,644 @@
...
+    }
+    
+    return numberOfCrossLevels;
+  }
+
+  /**
+  * @brief Vector containing the HGrid's different hierarchy levels of Linked Cells.
+  * 
+  */
+  std::deque<LinkedCells<Particle>> _hierarchyLevels;
```
FG-TUM, line 608:
```suggestion
  /**
   * Vector containing the HGrid's different hierarchy levels of Linked Cells.
   * 
   * @note This can not be a std::vector<LinkedCells<Particle>> because there is no copy constructor for LinkedCells.
   */
  std::deque<LinkedCells<Particle>> _hierarchyLevels;
```

src/autopas/containers/hierarchicalGrids/HierarchicalGrids.h
```diff
@@ -0,0 +1,644 @@
...
+
+    // Loop over all hierarchy levels and create a Linked Cells container for each level.
+    //
+    // @todo !!! cellSizeFactor has to be adapted !!!
+    //
+
+
+
+      const double radiusMax = cutoff / 2.;
+      const double radiusSeg = radiusMax / numberOfLevels;
```
FG-TUM, line 54: what is this abbreviation? Use variable names without abbreviations

FG-TUM, line 54: are these assumptions reasonable? Wouldn't it make more sense to look at the number of particles in the system and allocate the levels according to their size and distribution? Aka make this more dynamic?

src/autopas/containers/hierarchicalGrids/HierarchicalGrids.h
```diff
@@ -0,0 +1,644 @@
...
+      std::cout << "Created Hierarchical Grid with " << numberOfLevels << " levels" << std::endl;
+
+    for (unsigned int level = 0; level < _numberOfLevels; level++) {
+
+      const double gridBorder = radiusMax - radiusSeg * level;
+      const double cellSize = gridBorder / radiusMax;
+      _hierarchicalGridBorders.emplace_back(gridBorder);
+
+      _hierarchyLevels.emplace_back(boxMin, boxMax, cutoff, skinPerTimestep, rebuildFrequency, 
+                                   cellSizeFactor, loadEstimator);
```
FG-TUM, line 65: call `reserve()` before you use `push_back()` or `emplace_back()` in a loop.

src/autopas/containers/hierarchicalGrids/HierarchicalGrids.h
FG-TUM: Now after reading most of the class it seems `_particleRadiusRange` is not used at all? If that is correct, get rid of it or get rid of `_hierarchicalGridBorders` but sooner or later, I think the mechanism that decides which particle needs to go where might need to be reconsidered (see the comment in the constructor about the radii variables).

FG-TUM (commented):
Some things that I stumbled upon...
Please apply clang format (see [contribution guidelines](https://github.com/AutoPas/AutoPas/blob/userDocRewrite/.github/CONTRIBUTING.md#code-style))

DEMFunctor.h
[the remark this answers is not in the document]
hoppef: Yes. Was a copy-paste mistake

applicationLibrary/discreteElementMethod/discreteElementMethodLibrary/ParticleDEM.h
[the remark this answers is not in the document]
hoppef: Done

src/autopas/AutoPasDecl.h
```diff
@@ -990,6 +990,26 @@ class AutoPas {
    * @return sorting-threshold
    */
   size_t getSortingThreshold() const { return _sortingThreshold; }
+    /**
+   * Get number of H-Grid levels.
+   * @return
```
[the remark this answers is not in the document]
hoppef: Done

src/autopas/AutoPasDecl.h
```diff
@@ -990,6 +990,26 @@ class AutoPas {
...
+   */
+  [[nodiscard]] unsigned int getNumberOfHGLevels() const { return _logicHandlerInfo.numberOfHGLevels; }
+  /**
+   * Set number of H-Grid levels.
+   * @param numberOfHGLevels
+   */
+  void setNumberOfHGLevels(unsigned int numberOfHGLevels) { _logicHandlerInfo.numberOfHGLevels = numberOfHGLevels; }
+  /**
+   * Get possible particle radii(DEM only).
+   * @return
```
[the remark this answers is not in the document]
hoppef: Done

src/autopas/containers/hierarchicalGrids/HierarchicalGrids.h
hoppef: Deque doesn't support reserve(). Added reserve() for the _hierarchicalGridBorders

src/autopas/containers/hierarchicalGrids/HierarchicalGrids.h
hoppef: Done

hoppef: Your suggestion would be the better solution and is a goal for the future. This was a first simplistic approach

src/autopas/containers/hierarchicalGrids/HierarchicalGrids.h
hoppef: Added comment that this is a possibility

src/autopas/containers/hierarchicalGrids/HierarchicalGrids.h
hoppef: Done

src/autopas/containers/hierarchicalGrids/HierarchicalGrids.h
hoppef: Done

src/autopas/containers/hierarchicalGrids/HierarchicalGrids.h
hoppef: True. Got ride of it

src/autopas/containers/hierarchicalGrids/HierarchicalGrids.h
hoppef: Did a first implementation of this, but has to be improved further

src/autopas/containers/hierarchicalGrids/HierarchicalGrids.h
hoppef: Got rid of _particleRadiusRange. Idea behind it was to account for particle radii (but I did not do it in the end).
I fully agree that there needs to be a more sophisticated logic behind the sorting and level size mechanisms.

SamNewcome:
Closed, as being supported through #1014

SamNewcome:
Closed, as being supported through https://github.com/AutoPas/AutoPas/pull/1014

PR #775, opened 2023-08-21
Improve ease of use without UUID

### LINKED ISSUE #769: Improve ease of use without UUID
# The Problem

The rule-based tuning used `antlr4` for parsing the rule files. `antlr4` has `uuid` (the dev version!) as a build-dependency which is unfortunately not present on all systems (e.g. not HSUper). Installing this package and then telling our build process about this package is a bit nasty on environments where one doesn't have a package manager at hand (see workaround below).

## What we want
Simple, out of the box `cmake` -> `make` workflow no matter the system or installed packages.

## Possible Solution
Bundle `libuuid` and build it if no local one can be detected.

### Package Manager
If `uuid` is not installed and you have access to a package manager, simply install it directly or as part of the `util-linux` package:
Ubuntu:
```bash
apt install uuid-dev
```
Arch
```
pacman -S util-linux
```

### From Source
1. Download and build libuuid.
```bash
wget https://sourceforge.net/projects/libuuid/files/latest
tar -xzf libuuid.tgz
cd libuuid-*
mkdir myBuild
./configure --prefix=$(realpath myBuild)
make
```

2. Go to your AutoPas or ls1 build directory and use `CMake` to configure it as you like. Since Antlr is only compiled in the build step, this will not cause any problems yet.

3. Point the `ExternalProject_Add` workflow and the linker to the `libuuid` installation (assumed to be in `${YOUR_LIBUUID_DIR}`):
```bash
PKG_CONFIG_PATH=${YOUR_LIBUUID_DIR}/myInstall/lib/pkgconfig:$PKG_CONFIG_PATH \
    LIBRARY_PATH=${YOUR_LIBUUID_DIR}/myInstall/lib:$LIBRARY_PATH \
    make
```

dmartin13:
# Description

The library uuid-dev is now available as bundled version and is installed locally if no system-wide installation is present.

## Todo
- [x] fix issue with ninja build system

## Resolved Issues

- [x] fixes #769 

# How Has This Been Tested?

- [x] Compiling tested on an Ubuntu 22.04 system with and without installed uuid-dev.
- [x] Compiling tested on HSUper
- [x] Compiling tested with ls1


cmake/modules/antlr4cpp.cmake
```diff
@@ -2,13 +2,52 @@ message(STATUS "antlr4cpp - using bundled version")
...
+        # pass PKG_CONFIG_PATH as a environment variable to cmake so find_package() in antlr4cpp's CMakeLists.txt can find uuid-dev if using the bundled version
+        CMAKE_COMMAND    ${CMAKE_COMMAND} -E env PKG_CONFIG_PATH=${LIBUUID_PKGCONFIG_DIR} ${CMAKE_COMMAND}
         BUILD_BYPRODUCTS ${CMAKE_BINARY_DIR}/antlr4cppPrefix/install/lib/libantlr4-runtime.a
-        CMAKE_ARGS  -DCMAKE_INSTALL_PREFIX=${CMAKE_BINARY_DIR}/antlr4cppPrefix/install -DCMAKE_CXX_FLAGS=-w
+        CMAKE_ARGS       -DCMAKE_INSTALL_PREFIX=${CMAKE_BINARY_DIR}/antlr4cppPrefix/install -DCMAKE_CXX_FLAGS=-w
+        # pass LIBRARY_PATH as a environment variable to make so the linker finds uuid-dev if using the bundled version
+        BUILD_COMMAND    ${CMAKE_COMMAND} -E env LIBRARY_PATH=${LIBUUID_LIBRARY_DIR} ${CMAKE_MAKE_PROGRAM}
+        LOG_BUILD        ON
+        LOG_INSTALL      ON
+        DEPENDS          uuid_bundled
```
FG-TUM, line 50: Check if the target can be substituted by a string that either holds the target name or is empty

cmake/modules/antlr4cpp.cmake
```diff
@@ -2,13 +2,52 @@ message(STATUS "antlr4cpp - using bundled version")
...
+pkg_check_modules(UUID QUIET uuid)
+
+# if uuid-dev was not found on system we install it locally
+if (NOT UUID_FOUND)
+    message(STATUS "UUID not found - using bundled version")
+
+    set(LIBUUID_INSTALL_DIR "${CMAKE_BINARY_DIR}/uuid/install")
+    set(LIBUUID_PKGCONFIG_DIR ${LIBUUID_INSTALL_DIR}/lib/pkgconfig:$ENV{PKG_CONFIG_PATH})
+    set(LIBUUID_LIBRARY_DIR ${LIBUUID_INSTALL_DIR}/lib:$ENV{LIBRARY_PATH})
+    set(UUID_CONFIG_PARAMS "--prefix=${LIBUUID_INSTALL_DIR}")
```
FG-TUM, line 16: Do we have to set these Vars manually or can we use something like `find_package()`?

cmake/modules/antlr4cpp.cmake
```diff
@@ -2,13 +2,52 @@ message(STATUS "antlr4cpp - using bundled version")
...
 ExternalProject_ADD(
         antlr4cpp_bundled
-        PREFIX      antlr4cppPrefix
-        URL         ${PROJECT_SOURCE_DIR}/libs/antlr4-cpp-runtime-4.9.3-source.zip
-        URL_HASH    MD5=eafa4fef583e12e963062882773461be
+        PREFIX           antlr4cppPrefix
+        URL              ${PROJECT_SOURCE_DIR}/libs/antlr4-cpp-runtime-4.9.3-source.zip
+        URL_HASH         MD5=eafa4fef583e12e963062882773461be
+        # pass PKG_CONFIG_PATH as a environment variable to cmake so find_package() in antlr4cpp's CMakeLists.txt can find uuid-dev if using the bundled version
+        CMAKE_COMMAND    ${CMAKE_COMMAND} -E env PKG_CONFIG_PATH=${LIBUUID_PKGCONFIG_DIR} ${CMAKE_COMMAND}
```
FG-TUM, line 43: check if you can move `-E env PKG_CONFIG_PATH=${LIBUUID_PKGCONFIG_DIR}` to the `CMAKE_ARGS` arguemnt. Then we could get rid of this line and only have to call `cmake` once

FG-TUM (changes requested):
Some things to try out

cmake/modules/antlr4cpp.cmake
dmartin13: I need the `DEPENDS` here to ensure that `uuid` is present before the `ExternalProject` for `antlr4cpp` is configured. Without the `DEPENDS` I had problems when I use for example `make -j8`. However, `DEPENDS` expects a target, so I need to create the dummy target above. With a empty string it did not work.

cmake/modules/antlr4cpp.cmake
dmartin13: The problem is that at this point `uuid` is not yet installed, i.e. `find_package()` doesn't find anything yet, but we need to set the variables so we can pass them to `antlr4cpp_bundled` in the next `ExternalProject_ADD`. 

Unfortunately, antlr4-cpp-runtime's CMakeLists.txt uses `pkg_check_modules()`, so we need to set the `PKG_CONFIG_PATH` and `LIBRARY_PATH` environment variables somehow.

An alternative, which would replace both setting the variables (like `LIBUUID_INSTALL_DIR`, `LIBUUID_PKGCONFIG_DIR`, ...) and the custom commands like 
- `CMAKE_COMMAND    ${CMAKE_COMMAND} -E env PKG_CONFIG_PATH=${LIBUUID_PKGCONFIG_DIR} ${CMAKE_COMMAND}` 
- `BUILD_COMMAND    ${CMAKE_COMMAND} -E env LIBRARY_PATH=${LIBUUID_LIBRARY_DIR} ${CMAKE_MAKE_PROGRAM}`)

would be to modify antlr4-cpp-runtime's CMakeLists.txt and instead of using `pkg_check_modules()` simply use a `find_library(UUID NAMES uuid uuid-dev libuuid HINTS "${CMAKE_BINARY_DIR}/../../../uuid/install/lib" REQUIRED)`. But of course this has the disadvantage that you have to adjust the CMakeLists.txt every time when upgrading antlr4cpp.

FG-TUM: Ok in that case leave it as it is

FG-TUM:
Please also check that this works with ls1 when AutoPas is a dependency

dmartin13:
Works now also with ls1.

FG-TUM (approved):

PR #778, opened 2023-08-24
Fix failing tests

dmartin13:
# Description

- indices in `qMul` in Quaternion.cpp were off-by-one
- multipleSimilarObjects.yaml had wrong syntax
- reflectiveMultiSiteZoningTest did not take MPI into account
- fixed some CMake settings

## Resolved Issues

- [x] fixes failing tests after CI-Pipeline fix

# How Has This Been Tested?

- [x] Executing failing tests


FG-TUM:
IMHO this PR should merge into #777, not master

dmartin13:
One thing I noticed: The warnings in `ParticlePropertiesLibrary.h `depend on md-flexible variables (`MD_FLEXIBLE_MODE`), although the applicationLibrary does not necessarily have to be used in combination with md-flexible. In order to pass the `MD_FLEXIBLE_MODE` to the applicationLibrary and get rid of the warnings even if md-flexible is compiled with multisite-support, I had to add the appropriate `target_compile_definitions` in the CMakeLists of applicationLibrary. This however leads to some warnings during compiling: `warning: 'MD_FLEXIBLE_MODE' macro redefined`

So my questions/suggestions:
- Maybe it would be better to move the warnings into md-flexible instead of having it in the applicationLibrary?
- Is there a better way of telling applicationLibrary, that md-flex is compiled with multisite?

applicationLibrary/CMakeLists.txt
```diff
@@ -11,6 +11,12 @@ file(
 
 add_library(molecularDynamicsLibrary STATIC ${LIB_MD_SRC})
 
+if (MD_FLEXIBLE_MODE MATCHES  "SINGLESITE|singlesite")
+    target_compile_definitions(molecularDynamicsLibrary PUBLIC MD_FLEXIBLE_MODE=0 SINGLESITE=0 MULTISITE=1)
+elseif(MD_FLEXIBLE_MODE MATCHES "MULTISITE|multisite")
```
FG-TUM, line 16: I'd say this is still not fool-prove enough ;)
Typically what you want is to lower-case the input and then compare against the expected string. So something like:
```suggestion
string( TOLOWER "${MD_FLEXIBLE_MODE}" MD_FLEXIBLE_MODE_lower )
if (MD_FLEXIBLE_MODE_lower MATCHES "singlesite")
    target_compile_definitions(molecularDynamicsLibrary PUBLIC MD_FLEXIBLE_MODE=0 SINGLESITE=0 MULTISITE=1)
elseif(MD_FLEXIBLE_MODE_lower MATCHES "multisite")
```

examples/md-flexible/CMakeLists.txt
```diff
@@ -18,20 +18,20 @@ target_link_libraries(md-flexible PUBLIC autopas autopasTools molecularDynamicsL
...
 # Define numerical values for different md-flexible modes
 target_compile_definitions(md-flexible PUBLIC SINGLESITE=0 )
 target_compile_definitions(md-flexible PUBLIC MULTISITE=1)
 
 # Define numerical value for actually compiled md-flexible mode, as per above.
-if (MD_FLEXIBLE_MODE MATCHES "SINGLE-SITE")
+if (MD_FLEXIBLE_MODE MATCHES "SINGLESITE|singlesite")
     target_compile_definitions(md-flexible PUBLIC MD_FLEXIBLE_MODE=0)
-elseif(MD_FLEXIBLE_MODE MATCHES "MULTI-SITE")
+elseif(MD_FLEXIBLE_MODE MATCHES "MULTISITE|multisite")
```
FG-TUM, line 34: Same as above

FG-TUM: But is this really necessary? Shouldn't md-flexible link against the application library and thus receive all public definitions? (Can't currently check if this is the case)

examples/md-flexible/CMakeLists.txt
```diff
@@ -50,7 +50,7 @@ target_compile_definitions(
 )
 
 # --- display warnings if single-/multi-site mode is used with a functor type that has not been implemented ---
-if (MD_FLEXIBLE_MODE MATCHES "MULTI-SITE")
+if (MD_FLEXIBLE_MODE MATCHES "MULTISITE|multisite")
```
FG-TUM, line 53: tolower

applicationLibrary/CMakeLists.txt
dmartin13: done in [53d6fad](https://github.com/AutoPas/AutoPas/pull/778/commits/53d6fad3ea476758572a183c8d9e849437fcdf9e)

examples/md-flexible/CMakeLists.txt
dmartin13: done in [53d6fad](https://github.com/AutoPas/AutoPas/pull/778/commits/53d6fad3ea476758572a183c8d9e849437fcdf9e)

examples/md-flexible/CMakeLists.txt
dmartin13: Yes that works, I just tested it. I could move all the `target_compile_definitions` for `MD_FLEXIBLE_MODE` into the CMakeLists.txt of applicationLibrary. 
However, I would find it nicer if the md-flexible related variables would stay in the CMakeLists.txt of md-flexible. But so far I haven't found a way to set the `MD_FLEXIBLE_MODE` variable in the applicationLibrary without setting the `target_compile_definitions` in the CMakeLists.txt of applicationLibrary again. 
This variable was not set so far for the applicationLibrary and therefore all the warnings appeared even if md-flexible was compiled with multisite support.

(But if moving the definitions to applicationLibrary is the only possibility, then of course I can do that.)

FG-TUM: The application library needs a definition of the variables. The application library can exist without md-flex, md-flex can't exist without the application library. Hence, the variables need to be defined in the application library.
We could think about renaming them to eliminate the semantic association but honestly I don't care too much.

Move it to the application lib to get rid of the code duplication and let's get this fix merged asap.

dmartin13: done in [fc2e569](https://github.com/AutoPas/AutoPas/pull/778/commits/fc2e569b5b29eae75b37dc5341b0fcb9b19cc58f)

PR #785, opened 2023-09-15
Minimize VCL Unnecessary interactions

### LINKED ISSUE #716: Untangle VCL and VCLRebuilder
## The Problem

VCL is designed rather differently from all other containers. The actual class holds the list of towers (cells), provides access to the particles in them etc. The rebuilding of the data structure, meaning filling the vector of towers, sorting particles into it, and creating the clusters, however, is done by VCLRebuilder.

Through this separation of concerns, there is a lot of copied state (e.g. boxSize) as well as no clear rule which is responsible for which utility functions, like calculating which tower is responsible for a given position.

## Proposed Solution
- The lifetime of VCLRebuilder should be made clear more explicitly. Probably it can be limited to the `updateContainer()` function. Then we could get rid of the smart pointer class member.
- instead of copying lots of state information from VCL to VCLRebuilder add a reference to the first in the second.
- Introduce a utility class similar to cellBlock3D to encapsulate utility functions.

FG-TUM:
# Description

VCL do a significant amount of Halo<->Halo interactions or at least distance calculations. This is due to the fact the layout of the tower structure does not allow to easily distinguish between halo and owned regions.

Also, there are a lot of Owned-Owned interactions because the bounding box estimations between clusters are not good enough.

The main points in this PR that address this are:
- [x] Align towers with owned box (fixes halo identification in X/Y dimension)
- [x] Each tower knows its first and last cluster with owned particles (fixes halo identification in Z dimension)
- [x] No neighbor lists are generated for pure-halo clusters, except for forward neighbors if newton3 is enabled.
- [x] Dedicated class `ClusterTowerBlock` to handle tower structure (similar to `CellBlock3D`)
- [x] Rewrite Cluster bounding box distance logic (TODO: check if bbox should be stored or calculated on the fly)

## Related Pull Requests

- builds on top of #781 

## Resolved Issues

- partially fixes #716 

# How Has This Been Tested?

- [x] New tests
- [x] Some md-flex runs


SamNewcome (commented):
Start of review

src/autopas/containers/verletClusterLists/ClusterTower.h
```diff
@@ -99,9 +103,39 @@ class ClusterTower : public FullParticleCell<Particle> {
...
+
     for (size_t index = 0; index < numClusters; index++) {
       _clusters[index].reset(&(this->_particles[_clusterSize * index]));
+
+      // identify the range of owned clusters
+      if (not foundFirstOwnedCluster and isAnyOfClusterOwned(_clusters[index])) {
+        _firstOwnedCluster = _clusters.begin() + index;
+        foundFirstOwnedCluster = true;
+      }
+      if (not foundFirstTailHaloCluster and foundFirstOwnedCluster and not isAnyOfClusterOwned(_clusters[index])) {
```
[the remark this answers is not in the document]
FG-TUM: Which loop are you referring to? The for loop in line 118? `isAnyOfClusterOwned()` checks a new cluster in every iteration of this loop.

src/autopas/containers/verletClusterLists/ClusterTowerBlock2D.h
```diff
@@ -0,0 +1,421 @@
...
+   * @param clusterSize
+   */
+  void addTower(size_t clusterSize) { _towers.emplace_back(clusterSize); }
+
+  /**
+   * Estimates the optimal 2D tower grid side length.
+   *
+   * sideLength = (clusterSize / (numParticles / volume))^(1/3)
+   * The side length is then fitted to exactly fill the (non-halo) box with the same number of towers
+   * as the formula above.
```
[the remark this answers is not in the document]
FG-TUM: done 2b0087973

tests/testAutopas/tests/containers/verletClusterLists/VerletClusterListsTest.cpp
```diff
@@ -236,18 +266,73 @@ TEST_F(VerletClusterListsTest, testNewton3NeighborList) {
...
+  verletLists.addHaloParticle(p1);
+  // just inside the upper corner, thus in the last inner tower
+  const Particle p2{boxMax - 0.1, {0., 0., 0.}, numParticles++};
+  verletLists.addParticle(p2);
+  // just outside the upper corner, thus in the halo
+  const Particle p3{boxMax, {0., 0., 0.}, numParticles++};
+  verletLists.addHaloParticle(p3);
+
+  // we need 256 to 500 Particles for a 5x5 grid (= tower side length of 2)
+  // so we add as man as we are missing
```
SamNewcome, line 296:
```suggestion
  // so we add as many as we are missing
```

tests/testAutopas/tests/containers/verletClusterLists/VerletClusterListsTest.cpp
```diff
@@ -236,18 +266,73 @@ TEST_F(VerletClusterListsTest, testNewton3NeighborList) {
...
+  const Particle p1{boxMin - 0.1, {0., 0., 0.}, numParticles++};
+  verletLists.addHaloParticle(p1);
+  // just inside the upper corner, thus in the last inner tower
+  const Particle p2{boxMax - 0.1, {0., 0., 0.}, numParticles++};
+  verletLists.addParticle(p2);
+  // just outside the upper corner, thus in the halo
+  const Particle p3{boxMax, {0., 0., 0.}, numParticles++};
+  verletLists.addHaloParticle(p3);
+
+  // we need 256 to 500 Particles for a 5x5 grid (= tower side length of 2)
```
SamNewcome, line 295:
```suggestion
  // we need 256 to 500 Particles for a 5x5 grid of towers with side length 2
```

SamNewcome (changes requested):

src/autopas/containers/verletClusterLists/ClusterTower.h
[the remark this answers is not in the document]
FG-TUM: This look can be in any of the following states:
- First owned cluster not found yet: The first `if` will check all encountered clusters via `isAnyOfClusterOwned()`. The second `if` will short-circuit at the second condition and not call `isAnyOfClusterOwned()`.
- First owned cluster found but not yet first tail cluster: The first `if` will short-circuit in its first condition, not calling `isAnyOfClusterOwned()`.
- The first tail cluster was found: The first `if` still short-circuits, the second one too in its first condition. We still need to finish the loop to reset all clusters.
- Edge case: The iteration where the first owned cluster is found: In this one iteration neither if short-circuits and `isAnyOfClusterOwned()` is evaluated twice. I don't think it has a measurable impact but I created a patch in 58053a6.

src/autopas/containers/verletClusterLists/ClusterTowerBlock2D.h
```diff
@@ -0,0 +1,422 @@
...
+  const std::array<double, 2> &getTowerSideLengthReciprocal() const { return _towerSideLengthReciprocal; }
+
+  bool cellCanContainHaloParticles(index_t index1d) const override {
+    // TODO
+    return true;
+  }
+
+  bool cellCanContainOwnedParticles(index_t index1d) const override {
+    // TODO
+    return true;
```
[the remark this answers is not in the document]
FG-TUM: I forgot to implement this. Done here: ef7de01

src/autopas/containers/verletClusterLists/ClusterTowerBlock2D.h
```diff
@@ -0,0 +1,422 @@
...
+
+ private:
+  /**
+   * Index of the first tower in _towers that can contain owned particles.
+   * TODO: INITIALIZE AND USE THIS
+   */
+  size_t _firstOwnedTowerIndex{};
+  /**
+   * Index of the last tower in _towers that can contain owned particles.
+   * TODO: INITIALIZE AND USE THIS
```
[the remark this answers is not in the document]
FG-TUM: same as above

src/autopas/containers/verletClusterLists/ClusterTowerBlock2D.h
```diff
@@ -349,24 +354,34 @@ class ClusterTowerBlock2D : public CellBorderAndFlagManager {
...
   bool cellCanContainOwnedParticles(index_t index1d) const override {
-    // TODO
-    return true;
+    // check if the tower is strictly in the halo region
+    if (index1d < _firstOwnedTowerIndex or index1d > _lastOwnedTowerIndex) {
+      return false;
+    }
+    const auto index2D = towerIndex1DTo2D(index1d);
+    bool isHaloTower = false;
+    for (size_t i = 0; i < index2D.size(); ++i) {
```
SamNewcome, line 368:
```suggestion
    for (size_t i = 0; i < 2; ++i) {
```
A bit more readable, and when 2D is literally in the name I don't feel like it adds any maintainability to use `.size()`

SamNewcome (changes requested):

src/autopas/containers/verletClusterLists/ClusterTowerBlock2D.h
FG-TUM: The idea is that it is clear from the head of the loop what we iterate over. Also calling `size()` on an `std::array` should be "free" as it is a constant known at compile time. However, if you insist we can hard-code the `2`.

src/autopas/utils/ArrayMath.h
```diff
@@ -186,6 +186,40 @@ template <class T, std::size_t SIZE>
   return result;
 }
 
+/**
+ * Divides two array's element-wise and returns the result.
```
thesamriel, line 190:
```suggestion
 * Divides an array element-wise by a given scalar and returns the result.
```
Check this sentence also for the other implementations.

tests/testAutopas/tests/containers/verletClusterLists/VerletClusterListsTest.cpp
```diff
@@ -114,71 +119,87 @@ auto calculateValidPairs(const std::vector<autopas::Particle *> &particles, doub
...
+                          const std::vector<std::pair<autopas::Particle *, autopas::Particle *>> &actual,
+                          const std::string &errMsg) {
+  using autopas::utils::ArrayUtils::operator<<;
+  EXPECT_EQ(expected.size(), actual.size()) << errMsg << "\n"
+                                            << "Number of pairwise interactions differs.";
+  EXPECT_THAT(actual, ::testing::UnorderedElementsAreArray(expected)) << errMsg;
 }
 
+/**
+ * Fill a VLC container with randomly placed particles and compare pairwise interactions before and after
```
thesamriel, line 132:
```suggestion
 * Fill a VCL container with randomly placed particles and compare pairwise interactions before and after
```

thesamriel (changes requested):
Some minor things

src/autopas/utils/ArrayMath.h
```diff
@@ -186,6 +186,40 @@ template <class T, std::size_t SIZE>
...
+}
+
+/**
+ * Divides two array's element-wise and returns the result.
+ * @tparam T floating point type
+ * @tparam SIZE size of the arrays
+ * @param a dividend.
+ * @param s divisor.
+ * @return element-wise quotient of a and b, i.e., `result[i] = a[i]/s`
+ */
```
FG-TUM, line 213:
```suggestion
/**
 * Divides a scalar with by every element of an array to create an array of fractions.
 * @tparam T floating point type
 * @tparam SIZE size of the arrays
 * @param a dividend.
 * @param s divisor.
 * @return element-wise quotient of s and a, i.e., `result[i] = s/a[i]`
 */
```

src/autopas/containers/verletClusterLists/VerletClusterLists.h
```diff
@@ -1156,11 +1005,12 @@ class VerletClusterLists : public ParticleContainerInterface<Particle>, public i
    */
   template <class LoopBody>
   void traverseClustersSequential(LoopBody &&loopBody) {
-    for (size_t x = 0; x < _towersPerDim[0]; x++) {
-      for (size_t y = 0; y < _towersPerDim[1]; y++) {
-        auto &tower = getTowerByIndex(x, y);
-        for (auto &cluster : tower.getClusters()) {
-          loopBody(cluster);
+    for (size_t x = 0; x < _towerBlock.getTowersPerDim()[0]; x++) {
```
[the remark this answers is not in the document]
FG-TUM: done 20af98147

src/autopas/containers/verletClusterLists/ClusterTowerBlock2D.h
```diff
@@ -0,0 +1,437 @@
...
+    using namespace autopas::utils::ArrayMath::literals;
+    using autopas::utils::ArrayMath::ceil;
+    using autopas::utils::ArrayUtils::static_cast_copy_array;
+    _towerSideLength = towerSideLength;
+    _towerSideLengthReciprocal = 1. / towerSideLength;
+    const auto numTowersPerInteractionLength2D =
+        static_cast_copy_array<int>(ceil(_interactionLength / _towerSideLength));
+    // sanity check
+    if (numTowersPerInteractionLength2D[0] != numTowersPerInteractionLength2D[1]) {
+      AutoPasLog(WARN, "Number of towers per interaction length differs in X vs Y direction! {}",
```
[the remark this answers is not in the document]
FG-TUM: Honestly, I have no idea. The thing is, that the container was built with the assumption that they don't differ, and I did not look into the details where this might become a problem. I just thought this is something we might want to warn about to avoid silent bugs.-> 10e694129

src/autopas/containers/verletClusterLists/ClusterTowerBlock2D.h
SamNewcome: I won't say that I insist, as my reasoning is only anecdotal: I had to second-guess what `index2D` was e.g. thinking that it might have been an array of indices - because `index2D.size()` is very clearly always going to be 2.

I would leave it up to you to decide if this anecdotal evidence is worthy of switching to hard-coding.

I do understand your point about making it clear what we are iterating over, and so if you would like to hard-code 2 I'd suggest adding a brief comment.

SamNewcome (approved):

src/autopas/utils/ArrayMath.h
```diff
@@ -481,6 +515,32 @@ constexpr std::array<T, SIZE> operator/(const std::array<T, SIZE> &a, const std:
   return div(a, b);
 }
 
+/**
+ * Divides two array's element-wise and returns the result.
```
thesamriel, line 519:
```suggestion
 * Divides an array element-wise by a given scalar and returns the result.
 ```

File: src/autopas/utils/ArrayMath.h [hunk 14]
```diff
@@ -481,6 +515,32 @@ constexpr std::array<T, SIZE> operator/(const std::array<T, SIZE> &a, const std:
...
+ * @param b divisor.
+ * @return element-wise quotient of a and b, i.e., `result[i] = a[i]/b`
+ */
+template <class T, std::size_t SIZE>
+constexpr std::array<T, SIZE> operator/(const std::array<T, SIZE> &a, T b) {
+  return divScalar(a, b);
+}
+
+/**
+ * Divides two array's element-wise and returns the result.
```
thesamriel, line 532:
```suggestion
 * Divides a scalar by every element of an array to create an array of fractions.
 ```

thesamriel (approved):

PR #792, opened 2023-10-18
Parallelize LeavingParticleCollector

### LINKED ISSUE #653: Optimize LeavingParticleCollector
LeavingParticleCollector can theoretically not be very fast.

- [ ] OpenMP parallelization
- [ ] Optimization for different containers

### LINKED ISSUE #790: VerletClusterList container or the use of it in md-flexible is broken
**Describe the bug**
When building md-flexible according to the README and simulating fallingDrop.yaml the progress segfaults around iteration 5260-ish.
When removing the option to use the VerletClusterList container (remove VerletClusterList in line 3 of fallingDrop.yaml) the simulation works as expected. Only allowing VerletClusterList as a container leads to a segfault much sooner (around iteration 600-ish).
Therefore it is heavily implied that in a normal execution the VerletClusterList gets used later leading to a segfault due to 
-a faulty VerletClusterList
-a faulty interaction between the VerletClusterList and another component

Commit that was broken: 261878dc92c8ecc7ad2e11bf99ff925aaa7e753d 
Commit that worked: 7d757c3af42eae60e22cd0f3deddf4cf8caaa78f
I did not check which commit in between exactely introduced the error

**To Reproduce**
1. Pull the repo
2. In the root-folder mkdir build && cd $_, cmake .., make md-flexible,  cd examples/md-flexible,  ./md-flexible --yaml-filename fallingDrop.yaml
3.: Adjust the container List in fallingDrop.yaml (line 3) to see the other previously described behaviour

FG-TUM:
# Description

- [x] merge two loops over all particles into one
- [x] parallelize this loop
- [x] change loop over all particles to region iterators over borders

## Resolved Issues

- [x] fixes #653 
- [x] fixes #790 

# How Has This Been Tested?

Especially the following unit tests check that regions work correctly:
- [x] `testUpdateContainerCloseToBoundary()`
- [x] `md-flexible.test-sim`

- [x] Test sim with config:
```yaml
verlet-rebuild-frequency         :  15
verlet-skin-radius-per-timestep  :  0.2
data-layout                      :  [SoA]
functor                          :  Lennard-Jones (12-6) AVX intrinsics
cutoff                           :  2.5
deltaT                           :  0.0001
iterations                       :  50
boundary-type                    :  [periodic, periodic, periodic]
box-min                          :  [-70,-70,-70]
box-max                          :  [70, 70, 70]
Objects:
  CubeGrid:
    0:
      particles-per-dimension    :  [50, 50, 50]
      particle-spacing           :  1.4
      bottomLeftCorner           :  [0.7, 0.7, 0.7]
      velocity                   :  [1000, 0, 0]
      particle-epsilon           :  1
      particle-sigma             :  1
      particle-mass              :  1
```
Using 16 Threads on my Workstation (i7-10700) I get the following improvements:

| Version | Commit | UpdateContainer[s] |
|---------|:------:|-------------------:|
|Baseline | 7ae9a6e21cf9456cd8db99fcc8d84418d7c70ff2 | 0.174 |
| Merged loops | 82fae0d35006838bd6c8c58154fe36673529cf1e | 0.059 |
| Region Iters | e6475b0 | 0.063 |

~Seems like merged loops beat the "complicated" region iterator approach.~

- [x] Proper scenario

Falling drop from the examples folder but only linked cells on HSUper with one node and 144 threads:

| Version | Commit | UpdateContainer[s] |
|---------|:------:|-------------------:|
|Baseline | 7ae9a6e21cf9456cd8db99fcc8d84418d7c70ff2 | 6.193 |
| Merged loops | 82fae0d35006838bd6c8c58154fe36673529cf1e | 6.037 |
| Region Iters | e6475b0 | 4.441 |

=> Region iterators are better if the domain is large enough.

### DIFF
--- a/src/autopas/LogicHandler.h
+++ b/src/autopas/LogicHandler.h
@@ -111,7 +111,7 @@ class LogicHandler {
         }
         buffer.clear();
       } else {
-        for (auto iter = buffer.begin(); iter != buffer.end();) {
+        for (auto iter = buffer.begin(); iter < buffer.end();) {
           auto &p = *iter;
 
           auto fastRemoveP = [&]() {
@@ -124,7 +124,9 @@ class LogicHandler {
             // We remove dummies!
             fastRemoveP();
           }
-          if (utils::notInBox(p.getR(), boxMin, boxMax)) {
+          // if p was a dummy a new particle might now be at the memory location of p so we need to check that.
+          // We also just might have deleted the last particle in the buffer in that case the inBox check is meaningless
+          if (not buffer.empty() and utils::notInBox(p.getR(), boxMin, boxMax)) {
             leavingBufferParticles.push_back(p);
             fastRemoveP();
           } else {
--- a/src/autopas/LogicHandler.h
+++ b/src/autopas/LogicHandler.h
@@ -124,6 +124,8 @@ class LogicHandler {
           if (p.isDummy()) {
             // We remove dummies!
             fastRemoveP();
+            // In case we swapped a dummy here, don't increment the iterator and do another iteration to check again.
+            continue;
           }
           // if p was a dummy a new particle might now be at the memory location of p so we need to check that.
           // We also just might have deleted the last particle in the buffer in that case the inBox check is meaningless
--- a/src/autopas/containers/LeavingParticleCollector.h
+++ b/src/autopas/containers/LeavingParticleCollector.h
@@ -64,6 +64,7 @@ std::vector<typename ContainerType::ParticleType> collectParticlesAndMarkNonOwne
 
   std::vector<typename ContainerType::ParticleType> leavingParticles{};
 #ifdef AUTOPAS_OPENMP
+  // custom openmp reduction to concatenate all local vectors to one at the end of a parallel region
 #pragma omp declare reduction(vecMergeParticle : std::vector<typename ContainerType::ParticleType> : omp_out.insert(omp_out.end(), omp_in.begin(), omp_in.end()))
   // this has to be a `parallel` and not `parallel for` because we are actually interested in parallelizing the inner
   // loops without having a barrier between outer loop iterations.
--- a/src/autopas/containers/LeavingParticleCollector.h
+++ b/src/autopas/containers/LeavingParticleCollector.h
@@ -28,12 +28,14 @@ std::vector<typename ContainerType::ParticleType> collectParticlesAndMarkNonOwne
   using namespace autopas::utils::ArrayMath::literals;
 
   const auto interactionLength = container.getInteractionLength();
-  const auto skin = container.getVerletSkin();
   const auto containerDimensions = container.getBoxMax() - container.getBoxMin();
   const auto lowerHaloCorner = container.getBoxMin() - interactionLength;
   const auto upperHaloCorner = container.getBoxMax() + interactionLength;
-  const auto lowerInnerCorner = container.getBoxMin() + skin;
-  const auto upperInnerCorner = container.getBoxMax() - skin;
+  // We need to extend these boundaries into the container because we need to find particles that are stored in the
+  // container but have since moved out. Extending by skin is not enough since we are interested in particles that have
+  // travelled beyond that and skin can be 0.
+  const auto lowerInnerCorner = container.getBoxMin() + interactionLength;
+  const auto upperInnerCorner = container.getBoxMax() - interactionLength;
 
   // volumes describing the halo regions on the six faces of the cuboid shaped container.
   const std::array<std::tuple<decltype(lowerHaloCorner), decltype(upperHaloCorner)>, 6> haloVolumes{{
--- a/src/autopas/containers/LeavingParticleCollector.h
+++ b/src/autopas/containers/LeavingParticleCollector.h
@@ -31,8 +31,8 @@ std::vector<typename ContainerType::ParticleType> collectParticlesAndMarkNonOwne
   const auto containerDimensions = container.getBoxMax() - container.getBoxMin();
   const auto lowerHaloCorner = container.getBoxMin() - upperBoundForMisplacement;
   const auto upperHaloCorner = container.getBoxMax() + upperBoundForMisplacement;
-  // We need to extend these boundaries into the container because we need to find particles that are stored in the
-  // container but have since moved out. Extending by skin is not enough since we are interested in particles that have
+  // We need to extend these boundaries into the container because we need to include cells that still hold particles,
+  // which now have left the cells region towards the outside of the container.
   // travelled beyond that and skin can be 0.
   const auto lowerInnerCorner = container.getBoxMin() + upperBoundForMisplacement;
   const auto upperInnerCorner = container.getBoxMax() - upperBoundForMisplacement;
--- a/src/autopas/containers/LeavingParticleCollector.h
+++ b/src/autopas/containers/LeavingParticleCollector.h
@@ -58,12 +58,14 @@ std::vector<typename ContainerType::ParticleType> collectParticlesAndMarkNonOwne
   }};
 
   std::vector<typename ContainerType::ParticleType> leavingParticles{};
+#ifdef AUTOPAS_OPENMP
 #pragma omp declare reduction(vecMergeParticle : std::vector<typename ContainerType::ParticleType> : omp_out.insert(omp_out.end(), omp_in.begin(), omp_in.end()))
   // this has to be a `parallel` and not `parallel for` because we are actually interested in parallelizing the inner
   // loops without having a barrier between outer loop iterations.
   // The way this works is that all threads iterate the outer loop but due to the way the iterators are parallelized,
   // each thread only picks its iterations of the inner loops, hence nothing is done multiple times.
 #pragma omp parallel reduction(vecMergeParticle : leavingParticles)
+#endif
   for (const auto &[regionStart, regionEnd] : haloVolumes) {
     for (auto iter = container.getRegionIterator(regionStart, regionEnd, autopas::IteratorBehavior::ownedOrHalo);
          iter.isValid(); ++iter) {
--- a/src/autopas/containers/LeavingParticleCollector.h
+++ b/src/autopas/containers/LeavingParticleCollector.h
@@ -38,22 +38,27 @@ std::vector<typename ContainerType::ParticleType> collectParticlesAndMarkNonOwne
   const auto upperInnerCorner = container.getBoxMax() - upperBoundForMisplacement;
 
   // volumes describing the halo regions on the six faces of the cuboid shaped container.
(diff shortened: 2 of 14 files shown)


src/autopas/containers/LeavingParticleCollector.h
```diff
@@ -9,38 +9,76 @@
...
-  }
+  using namespace autopas::utils::ArrayMath::literals;
+
+  const auto interactionLength = container.getInteractionLength();
+  const auto containerDimensions = container.getBoxMax() - container.getBoxMin();
+  const auto lowerHaloCorner = container.getBoxMin() - interactionLength;
+  const auto upperHaloCorner = container.getBoxMax() + interactionLength;
+  // We need to extend these boundaries into the container because we need to find particles that are stored in the
+  // container but have since moved out. Extending by skin is not enough since we are interested in particles that have
+  // travelled beyond that and skin can be 0.
```
SamNewcome, line 36: I think this could be clearer. For example, why are these particles not caught by the region iterator? Is this extension by `interactionLength` determined by something or just a heuristic?

src/autopas/iterators/ContainerIterator.h
```diff
@@ -56,24 +56,25 @@ namespace containerIteratorUtils {
 template <bool regionIter, class Particle, class Arr>
 [[nodiscard]] bool particleFulfillsIteratorRequirements(const Particle &p, IteratorBehavior behavior,
                                                         const Arr &regionMin, const Arr &regionMax) {
-  bool particleOk = false;
+  // If this is a region iterator check box condition first, otherwise race conditions can occur
+  // if multiple region iterator overlap and ownership state is touched.
+  // (e.g. during collection of leavein particles)
```
SamNewcome, line 61:
```suggestion
  // (e.g. during collection of leaving particles)
```

tests/testAutopas/tests/containers/linkedCells/LinkedCellsTest.cpp
```diff
@@ -57,7 +59,7 @@ TYPED_TEST_P(LinkedCellsTest, testUpdateContainerCloseToBoundary) {
     }
   }
   std::set<unsigned long> movedIDs;
-  // we move particles that are close to the boundary to outside of the container and remember the id's we moved
+  // we move particles that are close to the boundary to the outside of the container and remember the id's we moved
```
SamNewcome, line 62:
```suggestion
  // we move particles that are close to the boundary to outside the container and remember their IDs
```

src/autopas/containers/LeavingParticleCollector.h
FG-TUM: This is how all of our containers work. Since we rebuild every container not in every iteration but when verlet-rebuild-frequency is reached all subcontainers (neighborlists, cells, ...) are extended by the skin.

SamNewcome: No, my point was not about the extension by skin. 

I'm going to explain my understanding of the function. Correct me if I am wrong.

All particles marked halo need to be deleted. It is not simply enough to iterate over regions outside of the box since a halo particle may have moved inside the box. So that we are not iterating over the whole container, we need some heuristic for the most a particle might have moved into the container. We can't use verlet skin /2 because we might have skin=0 (i.e. we rebuild every iteration) or, in the case of Luis' dynamic verlet lists, it is not necessary to use a verlet skin such that no particle should ever move more than verlet skin / 2. So you propose `interactionLength` as a heuristic. 

It makes total sense to have this heuristic as at least cutoff because if a particle could travel further than that in one iteration then we could have particles jumping over other particles when there should be a repulsive force between them at short distances (which implies that a smaller step-size should be used anyway). This heuristic makes a lot of sense in the types of particle simulations we are currently dealing with. But at the end of the day, this is still some heuristic - and is potentially not valid for every possible simulation. 

If I am correct in my understanding then, to prevent future headaches, this should be highlighted in the comments and the reasoning for the heuristic should be provided. You could go further and provide a "debug mode" whereby the region iterator is replaced with a normal iterator and a warning is printed if a halo particle is found outside of the regions of the region iterators. But this wouldn't be necessary for this PR and an issue could be opened. Finally, is there some reason for `interactionLength` instead of cutoff?

If I am not correct in my understanding then provide clearer reasoning to the extension by `iteractionLength`.

FG-TUM: In general your understanding is correct. Just some clarifications:
- What we add/subtract from the box boundaries is not a heuristic but an upper bound for how far particles have moved inside of the container. If this is not the lowest upper bound possible, it is still correct just not as efficient as it could be.
- Theoretically, the upper bound for movement per timestep is `skinPerTimestep/2`. If `skin = 0` this doesn't help us. However, having a `skin > cutoff` also makes no sense from a performance perspective, hence `cutoff` seems a reasonable upper bound.
- Yes cutoff should be more reasonable than the full interaction length. Don't ask why I had `interactionLengt` there first -> c78a8c683b390537c95c5a617222e231ea8b502a

src/autopas/containers/LeavingParticleCollector.h
```diff
@@ -9,38 +9,76 @@
...
-    iter->setOwnershipState(OwnershipState::dummy);
-  }
+  using namespace autopas::utils::ArrayMath::literals;
+
+  const auto upperBoundForMisplacement = container.getCutoff();
+  const auto containerDimensions = container.getBoxMax() - container.getBoxMin();
+  const auto lowerHaloCorner = container.getBoxMin() - upperBoundForMisplacement;
+  const auto upperHaloCorner = container.getBoxMax() + upperBoundForMisplacement;
+  // We need to extend these boundaries into the container because we need to find particles that are stored in the
+  // container but have since moved out. Extending by skin is not enough since we are interested in particles that have
```
thesamriel, line 35: Regarding your conversation, the first comment is still a bit confusing. What I understand is this:

We extend the boundaries *outside* the container to find particles that have moved out and we extend the boundaries *into* the container to catch halo particles that moved in.

src/autopas/containers/verletClusterLists/VerletClusterLists.h
```diff
@@ -224,12 +225,13 @@ class VerletClusterLists : public ParticleContainerInterface<Particle>, public i
 #ifdef AUTOPAS_OPENMP
 #pragma omp parallel for reduction(|| : deletedSomething)
 #endif
+    // Thanks to clang 13 this has to be a index based loop instead of a range based
     for (size_t i = 0; i < _towerBlock.size(); ++i) {
       auto &tower = _towerBlock[i];
       const auto towerSize = tower.getNumActualParticles();
```
thesamriel, line 231:
```suggestion
      const auto towerSize = tower.size();
```
`getNumActualParticles` would already get number of particles without tailing dummies.

thesamriel (changes requested):
Mostly found things next to your changes :D

src/autopas/LogicHandler.h
```diff
@@ -125,7 +125,9 @@ class LogicHandler {
             // We remove dummies!
             fastRemoveP();
           }
```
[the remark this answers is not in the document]
FG-TUM: good catch -> 897aada7126857bcd112857d57eaed7807642077

src/autopas/containers/LeavingParticleCollector.h
thesamriel: Ah, now I get what I mean. Yeah, this is clearer

src/autopas/containers/LeavingParticleCollector.h
```diff
@@ -9,38 +9,76 @@
...
 
-  std::vector<typename ContainerType::ParticleType> leavingParticles;
-  // Collect leaving particles
-  for (auto iter = container.begin(autopas::IteratorBehavior::owned); iter.isValid(); ++iter) {
-    if (not utils::inBox(iter->getR(), container.getBoxMin(), container.getBoxMax())) {
-      leavingParticles.push_back(*iter);
-      iter->setOwnershipState(OwnershipState::dummy);
+  std::vector<typename ContainerType::ParticleType> leavingParticles{};
+#ifdef AUTOPAS_OPENMP
+#pragma omp declare reduction(vecMergeParticle : std::vector<typename ContainerType::ParticleType> : omp_out.insert(omp_out.end(), omp_in.begin(), omp_in.end()))
```
SamNewcome, line 62: Add a brief one-line explaination of `vecMergeParticles`

SamNewcome (changes requested):

src/autopas/containers/LeavingParticleCollector.h
FG-TUM: done d04ac70d52e340f77c38cf5429b5a76f3e3243e8

src/autopas/containers/LeavingParticleCollector.h
```diff
@@ -9,38 +9,76 @@
...
+      // right + upper edge
+      {{upperInnerCorner[0], lowerInnerCorner[1], lowerInnerCorner[2]},
+       {upperHaloCorner[0], upperInnerCorner[1], upperHaloCorner[2]}},
+      // top + left edge
+      {{lowerHaloCorner[0], lowerInnerCorner[1], upperInnerCorner[2]},
+       {upperInnerCorner[0], upperInnerCorner[1], upperHaloCorner[2]}},
+      // bottom + right edge
+      {{lowerInnerCorner[0], lowerInnerCorner[1], lowerHaloCorner[2]},
+       {upperHaloCorner[0], upperInnerCorner[1], lowerInnerCorner[2]}},
+      // front with all edges and corners
```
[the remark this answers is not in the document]
FG-TUM: 5a0cf8fd6d2fe9739748328adca3390901a20fb7

src/autopas/containers/LeavingParticleCollector.h
```diff
@@ -9,38 +9,76 @@
...
+  const auto upperHaloCorner = container.getBoxMax() + upperBoundForMisplacement;
+  // We need to extend these boundaries into the container because we need to include cells that still hold particles,
+  // which now have left the cells region towards the outside of the container.
+  // travelled beyond that and skin can be 0.
+  const auto lowerInnerCorner = container.getBoxMin() + upperBoundForMisplacement;
+  const auto upperInnerCorner = container.getBoxMax() - upperBoundForMisplacement;
+
+  // volumes describing the halo regions on the six faces of the cuboid shaped container.
+  const std::array<std::tuple<decltype(lowerHaloCorner), decltype(upperHaloCorner)>, 6> haloVolumes{{
+      // left + lower edge
```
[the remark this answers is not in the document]
FG-TUM: done 5a0cf8fd6d2fe9739748328adca3390901a20fb7

thesamriel (approved):

SamNewcome (approved):

PR #794, opened 2023-10-19
LinkedCellsReferencesFix

### LINKED ISSUE #791: LinkedCellsReferences do not reinsert leaving particles properly
**Describe the bug**
When using the `LinkedCellsReferences` container some leaving particles are not reinserted properly in the container. The sanity check in `RegularGridDecomposition::exchangeMigratingParticles()` throws an exception: "All boundaries are periodic or reflective but ... migrants could not be re-inserted".
This is behavior is independent of the number of threads and appears reproducible in the same iteration (with the yaml file below).

After some investigation it has turned out that it is a problem with leaving particles, which are then suddenly present twice as owned particles in the container. The flow is roughly as follows: In iteration n, particle x is detected as leaving particle and set as dummy by `collectParticlesAndMarkNonOwnedAsDummy()`. In iteration n+1, the same particle is then added again in `LogicHandler::updateContainer()` by `collectLeavingParticlesFromBuffer()`. Afterwards, the cell structure is rebuilt in `LinkeCellsReference::updateDirtyParticleReferences() `and then the particle is suddenly present twice with the same position and velocity as owned particles, which in iteration n+2 leads to the fact that these particles then have position and velocity of `[-nan, -nan, -nan]` after position update and trigger the exception in `RegularGridDecomposition`.

Update: After digging into the code it turned out that this line of code in `LinkedCellsReferences::updateContainer()`  is responsible for doubling the particles (but this happens only with leaving particles that are reinserted by the `LogicHandler`):

```cpp
// multi layer swap-delete
// we copy the particle behind the last pointer in the cell over the particle we want to delete
**pIter = *particleVec.back();
```

**To Reproduce**
1. Clone 261878d
2. compile with `mkdir build && cd build && cmake .. && make md-flexible`
3. execute md-flexible with the yaml file below: `./md-flexible --yaml-filename input.yaml`
4. exception should appear (for me reproducible in iteration 286)

**Expected behavior**
No exception and no doubled owned particles in the container.

**Build environment:**
* OS: Ubuntu Server 20.04
* Compiler: gcc 11
* Build System: Make

**Additional context**
- The same yaml file runs without any problems when using LinkedCells as container.
- The parameters `verlet-rebuild-frequency` and `tuning-samples` affect the iteration number in which the exception is thrown.

**Input Yaml**
```yaml
container                        :  [LinkedCellsReferences]
data-layout                      :  [A0S]
verlet-skin-radius-per-timestep  :  0.025
cutoff                           :  2
box-max                          :  [15, 60, 15]
deltaT                           :  0.00182367
iterations                       :  400
boundary-type                    : [periodic, periodic, periodic]
Sites:
  0:
Objects:                         
  CubeClosestPacked:
    0:
      box-length                 :  [15, 6, 15]
      bottomLeftCorner           :  [0, 27, 0]
      particle-spacing           :  1.
no-flops                         :  false
no-progress-bar                  :  false
vtk-filename                     :  LCvsLCRef
vtk-write-frequency              :  100
```

dmartin13:
# Description

This PR should fix the bug described in #791.

In `LinkedCellsReferences::updateContainer()`, the halo particles were deleted with `deleteHaloParticles()` before all leaving particles were detected. This function deletes all halo particles in the data structure that stores the actual particles (`_particleList`). However, as a result, the references in the cells that point to the particles in `_particleList` are no longer correct. 
One solution would be to make a call to `updateDirtyParticleReferences()` directly after deleting the halo particles. However, since `_particleList` contains owned particles that have already been moved to the halo area, an exception is thrown when assigning the particles to the halo cells. So the simplest solution is to first run the loop that filters out all leaving particles and then call `deleteHaloParticles()`.

## Resolved Issues

- [x] fixes #791

# How Has This Been Tested?

- [x] All existing tests
- [x] The md-flexible scenario from #791
- [x] Testcase that checks correctness of references after updating the container: The `LinkedCellsTest` test was adjusted, so that wrong references after removing or inserting particles are detected. Therefore owned and halo particles are added alternating  to the container. After `updateContainer` it is checked that references still point to the expected particles. Also the `OwnershipState` of particles is checked.

### DIFF
--- a/src/autopas/containers/linkedCells/LinkedCellsReferences.h
+++ b/src/autopas/containers/linkedCells/LinkedCellsReferences.h
@@ -290,7 +290,6 @@ class LinkedCellsReferences : public CellBasedParticleContainer<ReferenceParticl
     if (keepNeighborListsValid) {
       return autopas::LeavingParticleCollector::collectParticlesAndMarkNonOwnedAsDummy(*this);
     }
-    this->deleteHaloParticles();
 
     std::vector<ParticleType> invalidParticles;
 #ifdef AUTOPAS_OPENMP
@@ -313,7 +312,8 @@ class LinkedCellsReferences : public CellBasedParticleContainer<ReferenceParticl
 
         auto &particleVec = this->getCells()[cellId]._particles;
         for (auto pIter = particleVec.begin(); pIter != particleVec.end();) {
-          if (utils::notInBox((*pIter)->getR(), cellLowerCorner, cellUpperCorner)) {
+          if ((*pIter)->getOwnershipState() == OwnershipState::owned and
+              utils::notInBox((*pIter)->getR(), cellLowerCorner, cellUpperCorner)) {
             myInvalidParticles.push_back(**pIter);
 
             // multi layer swap-delete
@@ -350,6 +350,15 @@ class LinkedCellsReferences : public CellBasedParticleContainer<ReferenceParticl
                                 myInvalidNotOwnedParticles.end());
       }
     }
+
+    // we have to remove halo particles after the above for-loop since removing halo particles changes the undelying
+    // datastructure (_particleList) which invalidates the references in the cells.
+    // Note: removing halo particles before the loop and do a call to updateDirtyParticleReferences() right after won't
+    // work since there are owned particles in _particleList which fall into the halo area (they are not yet filtered
+    // out by the above loop) and can therfore not added to halo cells during particle insert in
+    // updateDirtyParticleReferences()
+    this->deleteHaloParticles();
+
     _particleList.deleteDummyParticles();
     updateDirtyParticleReferences();
     return invalidParticles;
--- a/src/autopas/containers/linkedCells/LinkedCellsReferences.h
+++ b/src/autopas/containers/linkedCells/LinkedCellsReferences.h
@@ -292,6 +292,11 @@ class LinkedCellsReferences : public CellBasedParticleContainer<ReferenceParticl
     }
 
     std::vector<ParticleType> invalidParticles;
+
+    // for exception handling in parallel region
+    bool exceptionCatched{false};
+    std::string exceptionMsg{""};
+
 #ifdef AUTOPAS_OPENMP
 #pragma omp parallel
 #endif  // AUTOPAS_OPENMP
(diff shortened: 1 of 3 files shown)


src/autopas/containers/linkedCells/LinkedCellsReferences.h
```diff
@@ -313,7 +312,8 @@ class LinkedCellsReferences : public CellBasedParticleContainer<ReferenceParticl
 
         auto &particleVec = this->getCells()[cellId]._particles;
         for (auto pIter = particleVec.begin(); pIter != particleVec.end();) {
-          if (utils::notInBox((*pIter)->getR(), cellLowerCorner, cellUpperCorner)) {
+          if ((*pIter)->getOwnershipState() == OwnershipState::owned and
+              utils::notInBox((*pIter)->getR(), cellLowerCorner, cellUpperCorner)) {
```
FG-TUM, line 316:
```suggestion
          if ((*pIter)->isOwned() and utils::notInBox((*pIter)->getR(), cellLowerCorner, cellUpperCorner)) {
```
Keep it simple ;)

FG-TUM (changes requested):
Would be nice to have a dedicated unit test that detects errors like these to avoid reintroducing them in the future.

src/autopas/containers/linkedCells/LinkedCellsReferences.h
dmartin13: done in [06acfa6](https://github.com/AutoPas/AutoPas/pull/794/commits/06acfa6ec157f4d573d6d3b989f242caf776aef8)

src/autopas/containers/linkedCells/LinkedCellsReferences.h
```diff
@@ -292,6 +292,11 @@ class LinkedCellsReferences : public CellBasedParticleContainer<ReferenceParticl
     }
 
     std::vector<ParticleType> invalidParticles;
+
+    // for exception handling in parallel region
+    bool exceptionCatched{false};
```
FG-TUM, line 297: From the top of my head, I have no better idea. Except for renaming the variable to `exceptionCaught` ;)
```suggestion
    // OpenMP regions need to handle exceptions directly, so we need to remember if
    // an exception occurred.
    bool exceptionCaught{false};
```

dmartin13: done in [3bac09c](https://github.com/AutoPas/AutoPas/pull/794/commits/3bac09ce3ea6dc898f2b09df090624a9c3a9ce10)

SamNewcome (approved):

FG-TUM (approved):

PR #804, opened 2023-12-07
OpenMP Wrapper Macro

FG-TUM:
# Description

Whenever we do something with OpenMP macros, we need to write an ugly block of `#ifdef` to avoid potential compiler warnings about unknown pragmas when compiling without OpenMP.
Example:
```c++
#ifdef AUTOPAS_OPENMP
#pragma omp parallel reduction(+:foo)
#endif
{
  ...
}
```

I was sufficiently annoyed to introduce a new macro that wraps this away, so we can now write this significantly shorter:
```c++
AUTOPAS_OPENMP(parallel reduction(+:foo)) { 
  ...
}
```

The new pragma simply replaces `pragma omp`, so it will work for every OpenMP pragma. If we compile without OpenMP, it is replaced with nothing.

The indicator macro was renamed to provoke errors if someone uses it in the old way.

- [x] Introduces a wrapper macro for OpenMP
- [x] Rename indicator macro to `AUTOPAS_USE_OPENMP`
- [x] Refactor code base
- [x] CI check for misuse of old macro. 

## Additional Notes
- The new macro only accepts one argument. However, OpenMP also accepts its clauses to be separated by `,` which would, in turn, split the argument to `AUTOPAS_OPENMP`. We could make `AUTOPAS_OPENMP` variadic, and then during stringification recreate the `,` but I couldn't be bothered... I'd keep it like that until a use case arises that requires this.
- Blocks can be started in the same line as the new macro :eyes:
- `clang-format` employs a different indentation for the new macro vs the old pragma. The latter was always pushed to the front of the line, while the former follows the indentation of the rest of the code.
- Why is this a valid use case for a macro?
  -> We are wrapping a pragma that wasn't debuggable in the first place.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -9,7 +9,7 @@ Please keep in mind the following notes while working.
 ### General Notes
 * Cpp standard: C++17. If there is a piece of code, which could be done better using a newer standard, please add a comment like `@todo C++20` including the alternative version of the code.
 * Pointers: Always use smart pointers when you are managing memory. Don't use `new` or `delete`.
-* OpenMP: Use AutoPas wrapper functions for OpenMP (`src/autopas/utils/WrapOpenMP.h`) instead of OpenMP functions to allow building without enabled OpenMP.
+* OpenMP: Use AutoPas wrapper functions and macros for OpenMP (`src/autopas/utils/WrapOpenMP.h`) instead of OpenMP functions to allow building without enabled OpenMP.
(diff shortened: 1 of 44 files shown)


thesamriel (approved):
Didn't realize this was a problem, lol, but I'm fine with the changes.
Would be nice if we could document the usage somewhere.
Maybe in [CONTRIBUTING.md](https://github.com/AutoPas/AutoPas/blob/master/.github/CONTRIBUTING.md)?

thesamriel (approved):

PR #809, opened 2023-12-26
Userdoc

FG-TUM:
# Description

An attempt to better organize our user documentation.

- [x] Move existing documentation from main `README.md` to `/doc/`
- [x] Add a tag-based system that notifies when code has changed but the corresponding doc has not.

## ~Resolved Issues~

# How Has This Been Tested?

- [x] Dummy PR: #812


docs/userdoc/SimulationLoop.md
thesamriel, line 1: Same as the comment at `ContainerInterfaceModel`

docs/userdoc/SimulationLoop.md
```diff
@@ -0,0 +1,34 @@
...
+       ```
+
+4. Perform an iteratePairwise step.
+   ```cpp
+   autoPas.iteratePairwise(functor);
+   ```
+
+## Related Files and Folders
+- AutoPasDecl.h
+- Simulation.cpp
```
thesamriel, line 34: Does the tool also check `Simulation.h` or do you assume that changes there would anyway lead to changes in `AutoPasDecl.h` or `Simulation.cpp`

docs/userdoc/ParticleOwnershipModel.md
```diff
@@ -0,0 +1,18 @@
...
+  These normally are ghost particles arising from either periodic boundary conditions or particles of a neighboring AutoPas object
+  (if you split the entire domain over multiple AutoPas objects, i.e., you use a domain decomposition algorithm).
+  The halo particles are needed for the correct calculation of the pairwise forces.
+  On update of the AutoPas container, halo particles are deleted (see [Simulation Loop](TODO).
+* **dummy**: Particles that are deleted or that act as filler for certain algorithms. These particles do not affect the force calculation.
+
+## Related Files and Folders
+- ParticleBase.h
+- OwnershipState.h
+- markParticleAsDeleted.h
```
thesamriel, line 18:
```suggestion
- markParticleAsDeleted.h
- AutoPasDecl.h
- AutoPasImpl.h
```
For changes in e.g. `updateContainer()`. Or would this trigger the doc checker too often for irrelevant changes?

thesamriel (changes requested):
Looks pretty much fine, as it is just the same information split up :D.

Right now, it is quite obvious that the docs pages are just a continuous text split up. I would aim for a more wiki-like page style. But this can be revised in a later PR.

One thing I don't like so much is that there is no comment about building in the main README. That is often the most important information for newcomers. Ideally, I would still like build instructions in the main README, but I think that doesn't work with the doc-checker. So I would at least explicitly link the build instructions.

docs/userdoc/SimulationLoop.md
FG-TUM: > Does the tool also check Simulation.h

No it only tracks names that are stated explicitly.

My reasoning was that changes to the header will probably not affect the actual simulation loop. 🤷

docs/userdoc/ParticleOwnershipModel.md
FG-TUM: `updateContainer()` in the main class does not directly affect anything discussed in this doc. However LogicHandler does. Thanks

docs/userdoc/ContainerInterfaceModel.md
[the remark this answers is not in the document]
FG-TUM: > could use a lvl 1 heading

Agreed [1b68e53](https://github.com/AutoPas/AutoPas/pull/809/commits/1b68e538a92923cbdfda0e1066c4c496fae39307)

> some introductory words

I didn't want to change the content of the doc in this PR because it would be hard to compare it to the original due to the restructuring in more files. We should make a dedicated PR where we write more / new / better doc.

docs/userdoc/SimulationLoop.md
FG-TUM: [1b68e53](https://github.com/AutoPas/AutoPas/pull/809/commits/1b68e538a92923cbdfda0e1066c4c496fae39307)

thesamriel (approved):

PR #820, opened 2024-01-29
Safer Particle Movement

### LINKED ISSUE #819: Safer Particle Movement

One of the biggest assumptions that we make is that particles never move further than `skin / 2` during two rebuild steps. However, we have no good way to check this since particle movement is usually part of the user code.

We control the interface to the particles via `Particlebase`. The only way to move a particle is to change its position (`r`) via `setR()`. Here we should add another method:
```cpp
// skinPerStepHalvedSquared = (skin / rebuildFreq / 2)^2
bool setRChecked(const std::array<double, 3> &r, double skinPerStepHalvedSquared)
```
This method should check whether the difference between the old and new positions is bigger than skin per step / 2 and return the result of that check.
Nevertheless, the method should still set the new `r` since the user is then free to choose what to do with this information (ignore / throw / ... ?).

1. Remove `setR()` to force the check always, but this seems unnecessarily harsh.
2. Instead of `ParticleBase::setRChecked()` add some utility function somewhere à la `isMovementTooFar(rOld, rNew, skinPerStepHalvedSquared)`. The disadvantage is that it involves more argument pushing.

**Additional context**
- Technically, this restriction is too strict since a particle can move back and forth or faster and slower, and we only get a problem if the total movement over the whole rebuild frequency is farther than `skin / 2`. However, this is the best approximation we can do and the user:
  - is not forced to use it.
  - can implement further checks if false is returned.
  - can give a wider skin to the check (not a good idea but possible)
- MD-flexible already does this, so we can simply move parts of the implementation from there into AutoPas.
- We should use this in ls1 and throw exceptions or at least warnings there.

FG-TUM:
# Description

Add new functions that check if moving a particle is within safe bounds.

## Resolved Issues

- [x] fixes #819 

# How Has This Been Tested?

- [x] new tests

### DIFF
--- a/src/autopas/particles/ParticleBase.h
+++ b/src/autopas/particles/ParticleBase.h
@@ -153,12 +153,47 @@ class ParticleBase {
   void setR(const std::array<double, 3> &r) { _r = r; }
 
   /**
+   * Add a distance vector to the position of the particle and check if the distance between the old and new position
+   * is less than a given max distance.
+   * This max distance usually should be the skin per timestep divided by two.
+   *
+   * @param r vector to be added
+   * @param maxDistSquared The maximum allowed movement distance squared.
+   * @return true if dot(r - _r) < skinPerTimestepHalvedSquared
+   */
+  bool setRDistanceOk(const std::array<double, 3> &r, double maxDistSquared) {
+    using namespace autopas::utils::ArrayMath::literals;
+    const auto distanceVec = r - _r;
+    const double distanceSquared = utils::ArrayMath::dot(distanceVec, distanceVec);
+    setR(r);
+    return distanceSquared < maxDistSquared;
+  }
+
+  /**
    * Add a distance vector to the position of the particle
    * @param r vector to be added
    */
   void addR(const std::array<double, 3> &r) {
     using namespace autopas::utils::ArrayMath::literals;
     _r += r;
+  }
+
+  /**
+   * Add a distance vector to the position of the particle and check if the distance between the old and new position
+   * is less than a given max distance.
+   * This max distance usually should be the skin per timestep divided by two.
+   *
+   * @param r vector to be added
+   * @param maxDistSquared The maximum allowed movement distance squared.
+   * @return true if dot(r - _r) < skinPerTimestepHalvedSquared
+   */
+  bool addRDistanceOk(const std::array<double, 3> &r, double maxDistSquared) {
+    using namespace autopas::utils::ArrayMath::literals;
+    const auto oldR = _r;
+    addR(r);
+    const auto distanceVec = _r - oldR;
+    const double distanceSquared = utils::ArrayMath::dot(distanceVec, distanceVec);
+    return distanceSquared < maxDistSquared;
   }
 
   /**
--- a/src/autopas/particles/ParticleBase.h
+++ b/src/autopas/particles/ParticleBase.h
@@ -188,17 +188,16 @@ class ParticleBase {
    * is less than a given max distance.
    * This max distance usually should be the skin per timestep divided by two.
    *
+   * @note uses setRDistanceCheck()
+   *
    * @param r vector to be added
-   * @param maxDistSquared The maximum allowed movement distance squared.
+   * @param maxDistSquared The maximum expected movement distance squared.
    * @return true if dot(r - _r) < skinPerTimestepHalvedSquared
    */
   bool addRDistanceCheck(const std::array<double, 3> &r, double maxDistSquared) {
     using namespace autopas::utils::ArrayMath::literals;
-    const auto oldR = _r;
-    addR(r);
-    const auto distanceVec = _r - oldR;
-    const double distanceSquared = utils::ArrayMath::dot(distanceVec, distanceVec);
-    return distanceSquared < maxDistSquared;
+    const auto newR = _r + r;
+    return setRDistanceCheck(newR, maxDistSquared);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/autopas/particles/ParticleBase.h
```diff
@@ -152,6 +152,23 @@ class ParticleBase {
...
+  /**
+   * Add a distance vector to the position of the particle and check if the distance between the old and new position
+   * is less than a given max distance.
+   * This max distance usually should be the skin per timestep divided by two.
+   *
+   * @param r vector to be added
+   * @param maxDistSquared The maximum allowed movement distance squared.
+   * @return true if dot(r - _r) < skinPerTimestepHalvedSquared
+   */
+  bool setRDistanceOk(const std::array<double, 3> &r, double maxDistSquared) {
```
SamNewcome, line 164: Not keen on the naming. Some alternative suggestions:

- `setRWithCheck`
- `setRWithDistanceCheck`
- `setRSafe` (except it is only "safe" if the user does something with this information)

src/autopas/particles/ParticleBase.h
```diff
@@ -161,6 +178,24 @@ class ParticleBase {
...
+  /**
+   * Add a distance vector to the position of the particle and check if the distance between the old and new position
+   * is less than a given max distance.
+   * This max distance usually should be the skin per timestep divided by two.
+   *
+   * @param r vector to be added
+   * @param maxDistSquared The maximum allowed movement distance squared.
+   * @return true if dot(r - _r) < skinPerTimestepHalvedSquared
+   */
+  bool addRDistanceOk(const std::array<double, 3> &r, double maxDistSquared) {
```
SamNewcome, line 190: Similar as with `setRDistanceOk`

src/autopas/particles/ParticleBase.h
SamNewcome, line 1: What were your thoughts on throwing an error inside the function versus leaving this to the simulator? I don't think it should be entirely on the simulator from providing some way of the user knowing that a distance set was not "safe".

Some suggestions for how to improve the situation:
* These functions should write debug-level logs if beyond the cutoff but otherwise are unchanged. (So most of the responsibility is still on the simulator but there is at least one way of AutoPas directly informing the user)
* You could add an extra, optional, argument for throwError (or similar). By default, this is enabled. If so, AutoPas throws an error inside the function. If disabled, we assume the simulator has a plan for error handling.
* Move the error handling entirely inside this function. Maybe provide an option for the simulator to switch between warnings & errors.

SamNewcome (changes requested):
Two general changes

src/autopas/particles/ParticleBase.h
FG-TUM: > These functions should write debug-level logs

Good idea. Maybe even warn-level? If the user doesn't want the log they can still use the unchecked `setR()`.

> You could add an extra, optional argument for throwError

I thought about that too. My reasoning against that was roughly like this:
1. This is not necessarily a critical failure hence throwing must be optional
2. With the throw parameter, the return parameter is still necessary, otherwise the user has no idea if the movement was too far. (Actually, they can get the information by activating throw and catching the exception but this is exception handling abuse as flow control which is horrible.)

Therefore my conclusion was if we need the return parameter anyway, only do this to avoid redundant information :shrug: 
Another option would be to return the distance the particle moved and leave everything up to the user, but I feel this would be too unspecific and not convey what this function is intended to protect against.

> Move the error handling entirely inside this function

Sure, we could also do that and say if the user wants anything else they can build it themselves. In the end the function does nothing that can't be done without `setR()` and the information one has to provide.

src/autopas/particles/ParticleBase.h
FG-TUM: I'm also not really happy with the name but I noticed that something like `setRWithCheck` doesn't nicely convey the whats happening when used in an `if`. Maybe `setRSafe` is a good middleground?


```c++
if (setRWithCheck(r)) { /* do stuff */ }
// vs
if (setRDistanceOk(r)) { /* do stuff */ }
// vs
if (setRSafe(r)) { /* do stuff */ }
```

src/autopas/particles/ParticleBase.h
SamNewcome: As far as I am concerned - I am happy with a warn-level log only. It allows us to identify that this is a problem even if the user has not written some serious error handling which is my key point.

src/autopas/particles/ParticleBase.h
SamNewcome: My concern with `setRSafe` is that it implies that the set *is* safe when in reality it just warns if the set if unsafe. I would prefer `setRDistanceCheck` as a compromise.

src/autopas/particles/ParticleBase.h
FG-TUM: done -> [0a1fecb](https://github.com/AutoPas/AutoPas/pull/820/commits/0a1fecb1a2d4ace0abbd4f800063e66d77aa0a42)

src/autopas/particles/ParticleBase.h
FG-TUM: renamed to `setRDistanceCheck` fd40e01

src/autopas/particles/ParticleBase.h
FG-TUM: addRDistanceCheck fd40e017265ee0d5e58e672f027d237fca767f03

SamNewcome (approved):

Issue #522, opened 2020-08-17
OwnershipState should be fully managed by Logic Handler


FG-TUM:
**Is your feature request related to a problem? Please describe.**
Currently, the `OwnershipState` of an added particle is managed inconsistently. Some `addParticle()` methods set it explicitly, some don't.

**Describe the solution you'd like**
Let the `LogicHandler` fully take care of this.

**Describe alternatives you've considered**
The `LogicHandler` can also not set the states but only check if the passed particles have the correct state.

PR #789, opened 2023-10-10
Dynamic Verlet Lists

TumBeat:
# Description

This branch introduces the idea of Dynamic Rebuilding to the _VerletLists_, _VerletListsCells_ and _PairwiseVerletLists_ containers.
The implemented changes follow the procedure described in [my Bachelor's thesis](https://mediatum.ub.tum.de/node?id=1720413), sorry for the advertisement xD.

# How Has This Been Tested?

- [x] Running a lot of simulations with the example input files included in this branch

PR #821, opened 2024-01-30
Dynamic VL merge

manishmishra6016:
# Description

In this pull request, the existing containers are modified to make them dynamic. So, instead of being updated at rebuild frequency, they will be updated if any of the particle moves more than half the skin length. 

Note: this is applicable to all containers, not just VL containers

Main changes:
* `ParticleBase` is modified to include `rAtRebuild`, setter function, and method to calculate displacement since rebuild, 
* `neighborListsAreValid()` in `LogicHandler.h` is updated to check condition for dynamic rebuilding
* A method to update the `rAtRebuild` to current location once neighbor lists are rebuilt is added in `LogicHandler.h` 
* added CMake option `AUTOPAS_ENABLE_DYNAMIC_CONTAINERS` to enable/disable dynamic rebuilding. Default is set to `ON`. This is done as performance of LC containers are slightly better for the static case.
* ~~Check whether a given particle is `owned` or `halo` is now done in `LogicHandler.h` before adding to the container or buffer, earlier this was separately handled by the containers~~ Merged to master in #946 
* A new `IteratorBehavior::containerOnly` is added to enable only checking particles in container for dynamic rebuild

## Related Pull Requests

- #789 (closed)

## Things to be done before merging

- [x] tests added for dynamic rebuild with single particle, particle moving across periodic boundary and being added to buffer or container
- [x] ~~setting `fastParticlesWarn` in the code and using~~ not relevant anymore, to be removed completely
- [x] verifying results
- [x] comparison with `master` for runtime


FG-TUM:
The changes we discussed on the whiteboard should close #522

manishmishra6016:
The pull request is open to review. Note that I haven't added any tests as I think most functionality added is already getting tested (maybe not all the edge cases), so if you encounter some part where you think adding a test is necessary, I will do that. I am also thinking what tests would make sense, perhaps checking if dynamic rebuild is triggered?

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
```diff
@@ -570,6 +571,8 @@ RegularGridDecomposition::categorizeParticlesIntoLeftAndRightNeighbor(const std:
         const auto justInsideOfBox = std::nextafter(_globalBoxMax[direction], _globalBoxMin[direction]);
         position[direction] = std::min(justInsideOfBox, periodicPosition);
         leftNeighborParticles.back().setR(position);
+        leftNeighborParticles.back().setRAtRebuild();  // TODO: this isn't correct, best thing would be to shift so that
+                                                       // displacement before and after remains same
```
FG-TUM, line 575:
```suggestion
        // TODO: this isn't correct, best thing would be to shift so that displacement before and after remains same
        leftNeighborParticles.back().setRAtRebuild();  
```

- [ ] Reformat so that comments are above the line they refer to
- [ ] Reformulate what exactly is not correct
- [ ] Is this something we should fix? Then create an issue. If not then this shouldn't probably even be marked as todo but just as an simplification

src/autopas/LogicHandler.h
```diff
@@ -304,24 +306,29 @@ class LogicHandler {
...
       // Check if we can update an existing halo(dummy) particle.
-      bool updated = container.updateHaloParticle(haloParticle);
+      bool updated = container.updateHaloParticle(particleCopy);
       if (not updated) {
         // If we couldn't find an existing particle, add it to the halo particle buffer.
-        _haloParticleBuffer[autopas_get_thread_num()].addParticle(haloParticle);
-        _haloParticleBuffer[autopas_get_thread_num()]._particles.back().setOwnershipState(OwnershipState::halo);
+        _haloParticleBuffer[autopas_get_thread_num()].addParticle(particleCopy);
+        //_haloParticleBuffer[autopas_get_thread_num()]._particles.back().setOwnershipState(OwnershipState::halo);
+        // should be already halo now
```
FG-TUM, line 331: ok then these lines can be removed?

src/autopas/containers/linkedCells/LinkedCells.h
```diff
@@ -80,23 +80,19 @@ class LinkedCells : public CellBasedParticleContainer<FullParticleCell<Particle>
...
+    auto cells = _cellBlock.getNearbyHaloCells(haloParticle.getR(), this->getVerletSkin());
     for (auto cellptr : cells) {
-      bool updated = internal::checkParticleInCellAndUpdateByID(*cellptr, pCopy);
+      bool updated = internal::checkParticleInCellAndUpdateByID(*cellptr, haloParticle);
       if (updated) {
         return true;
       }
     }
-    AutoPasLog(TRACE, "UpdateHaloParticle was not able to update particle: {}", pCopy.toString());
+    AutoPasLog(TRACE, "UpdateHaloParticle was not able to update particle: {}", haloParticle.toString());
```
FG-TUM, line 95: this trace message should probably be moved to the logic handler for consistency

src/autopas/containers/linkedCells/LinkedCellsReferences.h
```diff
@@ -106,16 +103,14 @@ class LinkedCellsReferences : public CellBasedParticleContainer<ReferenceParticl
...
+    auto cells = _cellBlock.getNearbyHaloCells(haloParticle.getR(), this->getVerletSkin());
     for (auto cellptr : cells) {
-      bool updated = internal::checkParticleInCellAndUpdateByID(*cellptr, pCopy);
+      bool updated = internal::checkParticleInCellAndUpdateByID(*cellptr, haloParticle);
       if (updated) {
         return true;
       }
     }
-    AutoPasLog(TRACE, "UpdateHaloParticle was not able to update particle: {}", pCopy.toString());
+    AutoPasLog(TRACE, "UpdateHaloParticle was not able to update particle: {}", haloParticle.toString());
```
FG-TUM, line 113: can be removed if the message from LC is moved to the LogicHandler

src/autopas/containers/verletClusterLists/VerletClusterLists.h
```diff
@@ -200,6 +198,7 @@ class VerletClusterLists : public ParticleContainerInterface<Particle>, public i
         // don't simply copy haloParticle over iter. This would trigger a dataRace with other regionIterators that
         // overlap with this region.
         it->setR(haloPos);
+        it->setRAtRebuild();
```
FG-TUM, line 201: is this correct? If I remember correctly, this function is like moving the halo particle so why assume we rebuild it's neighbor list?

Btw: Do we even build neighbor lists for halo particles? I think in the 3 Body case we do but I'm not sure for the 2 Body case.

src/autopas/containers/verletListsCellBased/VerletListsLinkedBase.h
```diff
@@ -185,19 +185,17 @@ class VerletListsLinkedBase : public ParticleContainerInterface<Particle> {
...
+      bool updated = internal::checkParticleInCellAndUpdateByID(*cellptr, particle);
       if (updated) {
         return true;
       }
     }
     AutoPasLog(TRACE,
                "updateHaloParticle was not able to update particle at "
                "[{}, {}, {}]",
-               pCopy.getR()[0], pCopy.getR()[1], pCopy.getR()[2]);
+               particle.getR()[0], particle.getR()[1], particle.getR()[2]);
```
FG-TUM, line 198: remove

src/autopas/particles/ParticleBase.h
```diff
@@ -152,6 +157,22 @@ class ParticleBase {
    */
   void setR(const std::array<double, 3> &r) { _r = r; }
 
+  /**
+   * update rebuild position of the particle
+   */
+  void setRAtRebuild() { _rAtRebuild = _r; }
```
FG-TUM, line 163:
```suggestion
  /**
   * Update rebuild position of the particle.
   */
  void setRAtRebuild() { _rAtRebuild = _r; }
```

Slightly controversial. This is not a setter in the typical sense because it doesn't take an argument. I'm fine with it, though.

src/autopas/particles/ParticleBase.h
```diff
@@ -152,6 +157,22 @@ class ParticleBase {
...
+
+  /**
+   * calculate distance since last rebuild
+   * this is used to check if neighbour lists are still valid inside the logic handler
+   * @return displacement vector of particle since rebuild
+   */
+  const std::array<double, 3> calculateDisplacementSinceRebuild() const {
+    using namespace autopas::utils::ArrayMath::literals;
+    auto dist = utils::ArrayMath::sub(_rAtRebuild, _r);
+    return dist;
```
FG-TUM, line 173:
```suggestion
    return utils::ArrayMath::sub(_rAtRebuild, _r);
```

src/autopas/particles/ParticleBase.h
```diff
@@ -152,6 +157,22 @@ class ParticleBase {
...
 
+  /**
+   * update rebuild position of the particle
+   */
+  void setRAtRebuild() { _rAtRebuild = _r; }
+
+  /**
+   * calculate distance since last rebuild
+   * this is used to check if neighbour lists are still valid inside the logic handler
+   * @return displacement vector of particle since rebuild
```
FG-TUM, line 168:
```suggestion
   * Calculate the distance since the last rebuild.
   * This is used to check if neighbor lists are still valid inside the logic handler
   * @return displacement vector of particle since rebuild
```
American English :P

FG-TUM (changes requested):
Do we even still need the rebuild frequency with this?

src/autopas/particles/ParticleBase.h
manishmishra6016: I made this change as I also needed an actual setter to resolve some other comment :) But still open to changing the name if necessary.

src/autopas/particles/ParticleBase.h
manishmishra6016: Done

src/autopas/particles/ParticleBase.h
manishmishra6016: Done

src/autopas/containers/linkedCells/LinkedCells.h
manishmishra6016: Done, moved the TRACE message to `LogicHandler`

src/autopas/LogicHandler.h
```diff
@@ -790,12 +842,33 @@ void LogicHandler<Particle>::checkMinimalSize() const {
...
   }
   return _neighborListsAreValid.load(std::memory_order_relaxed);
 }
 
+template <typename Particle>
+void LogicHandler<Particle>::checkNeighborListsInvalidDynamicRebuild() {
+  auto halfSkinSquare = (getContainer().getVerletSkin() * getContainer().getVerletSkin()) / 4;
+  for (auto iter = this->begin(IteratorBehavior::owned); iter.isValid(); ++iter) {
+    auto distance = iter->calculateDisplacementSinceRebuild();
+    double distanceSquare = utils::ArrayMath::dot(distance, distance);
```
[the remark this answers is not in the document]
manishmishra6016: makes sense, changed.

src/autopas/LogicHandler.h
```diff
@@ -771,9 +803,29 @@ class LogicHandler {
...
 
+  /**
+   * iteration logger
+   */
   IterationLogger _iterationLogger;
+
+  /**
+   * neighborListInvalid - true if a particle has moved more than skin/2
+   */
+  bool _neighborListInvalidDynamicRebuild{false};
```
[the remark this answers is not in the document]
manishmishra6016: Yes, I changed this. Thanks :)

src/autopas/LogicHandler.h
```diff
@@ -711,6 +741,8 @@ class LogicHandler {
    */
   unsigned int _verletClusterSize;
 
+  std::vector<int> _rebuildDistances;
```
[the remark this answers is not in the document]
manishmishra6016: done

src/autopas/containers/verletClusterLists/VerletClusterLists.h
manishmishra6016: Corrected this - setting the rebuild position as the value stored in halo particle being added

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
manishmishra6016: resolved with in-person discussion

src/autopas/LogicHandler.h
```diff
@@ -771,9 +803,29 @@ class LogicHandler {
...
+   */
+  void updateRebuildPositions();
 };
 
+template <typename Particle>
+void LogicHandler<Particle>::updateRebuildPositions() {
+  for (auto iter = this->begin(IteratorBehavior::ownedOrHaloOrDummy); iter.isValid(); ++iter) {
+    iter->setRAtRebuild();
+  }
+}
```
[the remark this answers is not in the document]
manishmishra6016: removed Halo and Dummy particles, dynamic building only checks for owned particles right now. But Halos can be added later on.

manishmishra6016:
Changed `SkinPerTimeStep` to `Skin` everywhere and adjusted the values accordingly. And the default rebuild frequency is set to 100 as discussed. Also, merged recent changes in master into the current branch. @FG-TUM @SamNewcome

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
```diff
@@ -582,6 +587,10 @@ RegularGridDecomposition::categorizeParticlesIntoLeftAndRightNeighbor(const std:
         const auto periodicPosition = position[direction] - globalBoxDimensions[direction];
         position[direction] = std::max(_globalBoxMin[direction], periodicPosition);
         rightNeighborParticles.back().setR(position);
+        // Shifting the rebuild position of the particle, so that displacement since rebuild remains consistent
```
FG-TUM, line 590: This is actually quite important and should definitely be mentioned somewhere in the user documentation because it will impact performance if ignored.
Also please check the SPH examples for this.

src/autopas/LogicHandler.h
```diff
@@ -101,11 +100,13 @@ class LogicHandler {
     for (auto &cell : _particleBuffer) {
       auto &buffer = cell._particles;
       if (insertOwnedParticlesToContainer) {
-        for (const auto &p : buffer) {
+        // Can't be const because we potentially modify particels before readding them
```
FG-TUM, line 103:
```suggestion
        // Can't be const because we potentially modify particles before re-adding them
```

src/autopas/LogicHandler.h
```diff
@@ -304,24 +304,29 @@ class LogicHandler {
...
-      container.template addHaloParticle</* checkInBox */ false>(haloParticle);
+      container.template addHaloParticle</* checkInBox */ false>(particleCopy);
     } else {
       // Check if we can update an existing halo(dummy) particle.
-      bool updated = container.updateHaloParticle(haloParticle);
+      bool updated = container.updateHaloParticle(particleCopy);
       if (not updated) {
+        AutoPasLog(TRACE,
+                   "updateHaloParticle was not able to update particle at "
+                   "[{}, {}, {}]",
```
FG-TUM, line 326:
```suggestion
                   "updateHaloParticle was not able to update particle at "
                   "[{}, {}, {}]. Adding a copy to the halo buffer.",
```

src/autopas/LogicHandler.h
```diff
@@ -711,6 +742,12 @@ class LogicHandler {
    */
   unsigned int _verletClusterSize;
 
+  /**
+   * Array of int which contains number of steps after which rebuild occurs
+   * This is used to calculate the mean rebuild frequency
```
FG-TUM, line 747:
```suggestion
   * Array of int which contains number of iterations after which rebuild occurs.
   * This is used to calculate the mean rebuild frequency.
```

src/autopas/LogicHandler.h
```diff
@@ -711,6 +742,12 @@ class LogicHandler {
...
+   */
+  std::vector<int> _rebuildDistances;
```
FG-TUM, line 749: How about renaming this to `_rebuildIntervals`? At least for me "distances" is a bit misleading here.

FG-TUM (changes requested):
I love how this PR simplifies all of the add(Halo)Particle functions.

src/autopas/LogicHandler.h
manishmishra6016: makes sense, done :)

SamNewcome (changes requested):

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
manishmishra6016: done: added the comment by Sam, user documentation for dynamic rebuild as discussed, and changes in SPH code

examples/md-flexible/src/Simulation.cpp
```diff
@@ -601,6 +601,8 @@ void Simulation::logMeasurements() {
       std::cout << "  GFLOPs/sec                         : "
                 << static_cast<double>(flops) * 1e-9 / (static_cast<double>(simulate) * 1e-9) << std::endl;
       std::cout << "  Hit rate                           : " << flopCounterFunctor.getHitRate() << std::endl;
+      std::cout << "Mean Rebuild Frequency               : " << _autoPasContainer->getMeanRebuildFrequency()
+                << std::endl;
```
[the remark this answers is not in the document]
manishmishra6016: done - moved it out. also makes sense because it is not necessarily related to force computation.

examples/md-flexible/src/configuration/MDFlexConfig.h
```diff
@@ -381,19 +381,10 @@ class MDFlexConfig {
...
-      "verletRebuildFrequency."};
-
-  /**
-   * fastParticlesThrow
-   */
-  MDFlexOption<bool, __LINE__> fastParticlesThrow{false, "fastParticlesThrow", false,
-                                                  "Decide if particles that move farther than skin/2/rebuildFrequency "
-                                                  "will throw an exception during the position update or not."};
+  MDFlexOption<double, __LINE__> verletSkinRadius{.2, "verlet-skin-radius", true,
+                                                  "Skin added to the cutoff to form the interaction length."};
```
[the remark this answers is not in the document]
manishmishra6016: done

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
```diff
@@ -500,6 +499,10 @@ void RegularGridDecomposition::collectHaloParticlesAux(AutoPasType &autoPasConta
       auto position = particleIter->getR();
       position[direction] = position[direction] + wrapAroundDistance;
       haloParticlesBuffer.back().setR(position);
+      // Shifting the rebuild position of the particle, so that displacement since rebuild remains consistent
```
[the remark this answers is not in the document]
manishmishra6016: added

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
```diff
@@ -570,6 +571,10 @@ RegularGridDecomposition::categorizeParticlesIntoLeftAndRightNeighbor(const std:
         const auto justInsideOfBox = std::nextafter(_globalBoxMax[direction], _globalBoxMin[direction]);
         position[direction] = std::min(justInsideOfBox, periodicPosition);
         leftNeighborParticles.back().setR(position);
+        // Shifting the rebuild position of the particle, so that displacement since rebuild remains consistent
```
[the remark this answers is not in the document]
manishmishra6016: added

src/autopas/AutoPasDecl.h
```diff
@@ -969,6 +961,13 @@ class AutoPas {
    */
   void setOutputSuffix(const std::string &suffix) { _outputSuffix = suffix; }
 
+  /**
+   * Getter for the mean rebuild frequency.
+   * Helpful for analyzing the behaviour of the dynamic containers.
```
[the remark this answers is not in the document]
manishmishra6016: done

src/autopas/LogicHandler.h
```diff
@@ -694,13 +713,25 @@ class LogicHandler {
...
 
+  /**
+   * Checks if any particle has moved more than skin/2.
+   * updates bool: _neighborListInvalidDoDynamicRebuild to true
+   */
+  void checkNeighborListsInvalidDynamicRebuild();
+
+  /**
+   * Checks if any particle has moved more than skin/2.
+   * updates bool: _neighborListInvalidDoDynamicRebuild to false
```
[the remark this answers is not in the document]
manishmishra6016: done

src/autopas/LogicHandler.h
```diff
@@ -711,6 +742,12 @@ class LogicHandler {
    */
   unsigned int _verletClusterSize;
 
+  /**
+   * Array of int which contains number of iterations after which rebuild occurs.
```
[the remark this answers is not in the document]
manishmishra6016: done

FG-TUM (changes requested):
I'd still like to see dedicated unit tests for this mechanic

SamNewcome (approved):
Everything looks mostly good to me.

The usability of the fix for periodic boundaries concerns me, but I think it can be handled in a separate PR. Please check issue #860 to check that my reasoning is correct.

FG-TUM:
> The usability of the fix for periodic boundaries concerns me, but I think it can be handled in a separate PR. Please check issue #860 to check that my reasoning is correct.
> 
> Alternatively, you can handle this issue within this PR - I am easy either way.

If we merge this now we shouldn't update the AutoPas version in ls1 before we resolve #860. I'm also ok either way but if this is merged now #860 should be tackled immediately, probably in combination with #861.

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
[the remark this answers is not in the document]
manishmishra6016: This is no longer necessary as with the new containerOnly iterator, we just ignore the particles in buffer.

examples/md-flexible/src/domainDecomposition/RegularGridDecomposition.cpp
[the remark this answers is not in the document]
manishmishra6016: update: with the new `containerOnly` iterator, this is no longer needed, so I reverted this change.

FG-TUM (changes requested):
Needs a few more comments for clarity, then this can be merged.

manishmishra6016:
I ran the `SpinodalDecomposition_equilibration.yaml` case on HSUper and `dynamic-VL-merge` branch seems to be around ~20% slower. This is because of large number of particles in this case and we do a check for dynamic rebuild that iterates over all the particles (in container) in every step. And as expected, cost of this check depends on the number of particles. I am not sure what is the best way to get around it. I have added `timeCheckForDynamicRebuild` in the iteration logger if someone wants to use.

----
| | Master (rf 10) | Dynamic (rf 100) | 
| --- | --- | --- |
|runtime (s) | 56680 | 67885 | 
| mean rf | 10 | 14.37 |
---


It can be done seamlessly while doing position updates but that is done in `md-flexible` which is not useful for us as it means shifting responsibility to user and autopas no longer being `a blackbox`. 

One possible solution would be to make it optional but I am not sure if we want it? @FG-TUM @SamNewcome @thesamriel

FG-TUM:
Can you profile a bit more what exactly is so expensive? 20% sounds like a lot ans significantly more than I'd expected for such a simple check! Is it maybe more efficient to use the `forEach` functions instead of iterators (the are not yet parallelized however...)?

SamNewcome:
> @SamNewcome @FG-TUM @thesamriel I have results from the individual runs (for single configurations of each container)
> ### Runtime for 10000 steps of `SpinodalDecomposition_equilibration` case
> 	LC 	VLC 	VL
> Dynamic (rf=100) 	5100 s 	13900 s 	7140 s
> Master (rf=10) 	4450 s 	55100 s 	14000 s
> 
> For LC case, running with `rf=1` gives 4750s, which is still better than dynamic LC. I was not able to run LC with `skin=0` on `Dynamic-VL-merge` branch as it had some error (probably because the whole dynamic check depends on skin). On master also, it was quite slow as there were a lot of warnings (but this could be fixed by disabling the warning).
> ### Time for rebuilding and checking for rebuild
> ## Note: time give is for one rebuild (or one check)
> Container 	CheckForDynamicRebuild (dynamic) 	RebuildNeighborList (dynamic) 	RebuildNeighborList (master)
> LC 	0.02 s 	0.02 s 	~ 0 s (100 ns)
> VLC 	0.02 s 	23 s 	22 s
> VL 	0.02 s 	2 s 	1.5 s
> 
> Conclusions from this:
> 
>     * LC on master branch takes almost no time to rebuild, this increases on dynamic branch as it involves call to `updateRebuildPositions()` which loops over all the particle in container
> 
>     * Additonally, call to `checkNeighborListDoDynamicRebuild()` is as expensive as rebuilding in case of LC, and this function is called every step.
> 
>     * For other containers where rebuilding is expensive, dynamic containers perform better.
> 
> 
> @FG-TUM also to note, I did these with the for loop parallelized with OpenMP (with 36 threads on HSUper). I think parallelization should work but I haven't confirmed it in any way.

Did you check the cost of LC without any skin?

FG-TUM:
Conclusion from the talk:
- The safety margin of region iterators should be optimized for two cases:
  - inside of `iteratePairwise()` use `maxDistMoved` because this is guaranteed to be up to date there and the best safety margin possible.
  - outside of `iteratePairwise()` use `skin` for now because these loops are not as performance critical.
- In the future (aka not in this PR) we can introduce any of the proposed optimizations to also speed up these loops. Suggestions were:
  - `p.setRNotify(newR, &autoPas)`
  - additional parameter to region iterators to limit the safety margin, passed through `autoPas.begin()`, and defaulting to the full skin. Could be based on `maxDistMoved`

manishmishra6016:
Hi again,
with some optimizations, changing safetyMargin back to `skin` and separately merging changes unrelated to dynamic containers to master, I compare again both the branches for default `SpinodalDecomposition_equilibration` case for 10k steps. 

### Case 1: skin = 0.5

---
|  | Master (rf = 10) | Dynamic (rf = 100/ mean rf = 19) |Dynamic without iter.* (rf=100, mean rf=2 :exclamation: ) |
| --- | --- | --- | --- |
| total time (s) | 5044 | 5176 | 4759 |
| Pos. update (s) | 153 | 175 |148 |
| Update Container (s) | 38 | 191 | 193|
| Force update (s) | 3522 | 3543 |2918|
| Velocity update (s) | 133 | 160 | 135 |
---
**Dynamic without iter.*** refers to removing `containerOnly` iterator that was implemented as part of this feature
:exclamation: mean_rf=2 happens because without iterator we have particles crossing boundaries causing frequent rebuild even when they are in buffer.

### Case 2: skin = 0.

---
|  | Master (rf = 1) | Dynamic (rf = 100/ mean rf = 1) |Dynamic without iter.* (rf=100, mean rf=1) |
| --- | --- | --- | --- |
| total time (s) | 4972 | 5364 | 5060 |
| Pos. update (s) | 172 | 214 |187 |
| Update Container (s) | 130 | 333 | 280|
| Force update (s) | 3162 | 3213 |3038|
| Velocity update (s) | 147 | 194 | 165 |
---
**Dynamic without iter.*** refers to removing `containerOnly` iterator that was implemented as part of this feature

Conclusion:
- performance between [master](https://github.com/AutoPas/AutoPas/commit/c3b5d2105b3fedef8157116ce9ca04a71004fe6a) and [Dynamic-VL-merge](https://github.com/AutoPas/AutoPas/commit/c3b5d2105b3fedef8157116ce9ca04a71004fe6a) is comparable. Dynamic branch takes slightly longer (~200s) in `UpdateContainer()` function due to check for dynamic rebuilding. 
- The `containerOnly` iterator seems to be expensive but removing it defeats the purpose of "dynamic" rebuilding unless we take care of particle in buffer as well, which makes the code complex and might require user to update `rAtRebuild()`.
- Region Iterator wasn't the expensive part here (as we discussed during the talk). 

I think with this, this PR is ready for a decision. There is still a some issue with respect to autotuning:
- [ ] find a good estimate of mean_rf for comparison of different algorithms during the tuning phase. --> I think for now this can be kept as the rf we provide in input file (which maynot be a good choice) and I can work on it separately.  

@SamNewcome @FG-TUM @thesamriel @schuhmaj

src/autopas/LogicHandler.h
```diff
@@ -558,11 +560,39 @@ class LogicHandler {
     if (_rebuildIntervals.empty()) {
       return 0.;
     } else {
-      return std::accumulate(_rebuildIntervals.begin(), _rebuildIntervals.end(), 0.) /
-             static_cast<double>(_rebuildIntervals.size());
+      return std::accumulate(_rebuildIntervals.begin() + 1, _rebuildIntervals.end(), 0.) /
+             static_cast<double>(_rebuildIntervals.size() - 1);
```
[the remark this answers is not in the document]
manishmishra6016: done

src/autopas/LogicHandler.h
```diff
@@ -791,14 +856,44 @@ void LogicHandler<Particle>::checkMinimalSize() const {
...
+  // e.g. neighbour list, and these are iterated over using the region iterator. Movement of particles in buffer doesn't
+  // require a rebuild of neighbor lists.
+  AUTOPAS_OPENMP(parallel reduction(or : _neighborListInvalidDoDynamicRebuild))
+  for (auto iter = this->begin(IteratorBehavior::owned | IteratorBehavior::containerOnly); iter.isValid(); ++iter) {
+    const auto distance = iter->calculateDisplacementSinceRebuild();
+    const double distanceSquare = utils::ArrayMath::dot(distance, distance);
+
+    if (distanceSquare >= halfSkinSquare) {
+      _neighborListInvalidDoDynamicRebuild = true;
+    }
```
FG-TUM, line 888:
```suggestion
    _neighborListInvalidDoDynamicRebuild |= distanceSquare >= halfSkinSquare;
```
Since you are reducing anyway we could try to remove the `if`. Avoiding branches is always a good idea to increase throughput.

FG-TUM:
I don't fully buy these measurements. Velocity update should be the exact same across the three versions, right? Yet, I see a 30% slowdown from `master` to `Dynamic`. Where have you conducted these measurements? If it wasn't a cluster, have you [disabled automatic performance scaling / set the processor to max performance](https://wiki.archlinux.org/title/CPU_frequency_scaling)? Have you done the measurements multiple times to get some idea of their variance?

Maybe the difference between the measurements falls within the uncertainty of the measurements?

src/autopas/LogicHandler.h
```diff
@@ -153,11 +152,9 @@ class LogicHandler {
    * @copydoc AutoPas::updateContainer()
    */
   [[nodiscard]] std::vector<Particle> updateContainer() {
+    this->checkNeighborListsInvalidDoDynamicRebuild();
```
FG-TUM, line 155:
```suggestion
    // update member variable _neighborListInvalidDoDynamicRebuild
    this->checkNeighborListsInvalidDoDynamicRebuild();
```

FG-TUM: also place a timer around this function and add it to your measurements.

FG-TUM (changes requested):
some minor suggestions

manishmishra6016:
> I don't fully buy these measurements. Velocity update should be the exact same across the three versions, right? Yet, I see a 30% slowdown from `master` to `Dynamic`. Where have you conducted these measurements? If it wasn't a cluster, have you [disabled automatic performance scaling / set the processor to max performance](https://wiki.archlinux.org/title/CPU_frequency_scaling)? Have you done the measurements multiple times to get some idea of their variance?
> 
> Maybe the difference between the measurements falls within the uncertainty of the measurements?

Hi Fabio, yes. the measurements vary quite a bit depending on which node the problem is allocated to (I think). I did them on a cluster (HSUper). Is this not something you normally observe? I did do some extra measurements but not enough times to take average (but I repeated all of them atleast twice but I have just listed value of 1st run). And as you said, the velocity update should be same. The main difference (which was consistent across different runs) was the `updateContainer()` call as you can see.

manishmishra6016:
Repeating runs on HSUper (max NumThreads is 72 on 1 node) to get an idea on general variation in values reported
Simulation: `SpinodalDecomposition_equilibration` run for 10000 steps

---
| No.  |  Case \ Total runtime| Run 1 (s) | Run 2 (s) |Run3 (s) |
| --- | --- | --- | --- | --- |
| 1 | skin 0.5, rf 10, NumThreads 36, master | 4872.4 | 5174.5 | 4659.1 |
| 2 | skin 0.5, rf 100 (~19), NumThreads 36, dynamic | 4852.9 | 4776.1 | 4554.4 |
| 3 | skin 0.5, rf 10, NumThreads 72, master | 3468.5 | 3382.8 | 3333.4 |
| 4 | skin 0.5, rf 100(~19), NumThreads 72, dynamic | 3727.8 | 3672.4 | 3740.9 |
| 5 | skin 0, rf 1, NumThreads 36, master | 5293.9| 4997.1 | 4989.3 |
| 6 | skin 0, rf 100(1), NumThreads 36, dynamic | 5082.7| (6220.8) | 5171.1 |
| 7 | skin 0, rf 1, NumThreads 72, master | 3747.0 | 3784.0 | 3746.0 |
| 8 | skin 0, rf 100(1), NumThreads 72, dynamic | 4253.0 | 4102.0 | 4193.0 |
---

As I observe less variation when runs are done with 72 threads (all threads available on the node), we can look at these results for comparing the branches. I also take the runs with minimum total runtime for comparison. 


### Case 1: skin = 0.5

---
|  | Master (rf = 10) | Dynamic (rf = 100/ mean rf = 19) |
| --- | --- | --- | 
| total time (s) | 3333.4 | 3727.8 |
| Pos. update (s) | 102.7 | 114.5 |
| Update Container (s) | 27.8 | 110.2 |
| Force update (s) | 1905.4 | 2095.8 |
| Velocity update (s) | 93.2 | 98.9 |
| rebuildNeighborList (s/iteration) | ~0 | ~0.011 |
---

### Case 2: skin = 0.

---
|  | Master (rf = 1) | Dynamic (rf = 100/ mean rf = 1) |
| --- | --- | --- | 
| total time (s) | 3747.0 | 4102.2 | 
| Pos. update (s) | 121.2 | 137.5 |
| Update Container (s) | 122.6 | 219.9 |
| Force update (s) | 1977.3 | 2222.0 |
| Velocity update (s) | 106.2 | 121.0 | 
| rebuildNeighborList (s/iteration) | ~0 | ~0.017 | 
---

* time for `rebuildNeighborList` includes resetting `rAtRebuild` values for each particle. This is also reflected in increased Force update timing.
* time for checking rebuild condition is reflected in increased time for Update Container call. I have not added a separate timer for it in AutoPas as all our timers are in `iteratePairwise()` function and this condition is checked in `updateContainer()`

@FG-TUM @SamNewcome I now did multiple runs for each case, to get an idea of the variation and I am now comparing the runs with minimum time.

src/autopas/LogicHandler.h
manishmishra6016: time for checking rebuild condition is reflected in increased time for Update Container call. I have not added a separate timer for it in AutoPas as all our timers are in `iteratePairwise()` function and this condition is checked in `updateContainer()`. I would need to find a way to return it and add it to logger. But as this can already be inferred, I have skipped it for now. I could look at implementing this if deemed absolutely necessary.

src/autopas/LogicHandler.h
```diff
@@ -842,22 +937,24 @@ LogicHandler<Particle>::getParticleBuffers() const {
...
+  const auto &configuration = _autoTuner.getCurrentConfig();
+  auto &container = _containerSelector.getCurrentContainer();
+
   const bool energyMeasurementsPossible = _autoTuner.resetEnergy();
   timerTotal.start();
 
   functor.initTraversal();
   if (doListRebuild) {
     timerRebuild.start();
+    this->updateRebuildPositions();
```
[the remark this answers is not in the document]
manishmishra6016: done

manishmishra6016:
@FG-TUM @SamNewcome I have added the `AUTOPAS_ENABLE_DYNAMIC_CONTAINERS` option with default as `ON`. I logged the runtime for `SpinodalDecomposition_equilibration` run on HSUper with 72 OMP threads (for LinkedCells container) as comparison.

--- 
| | run 1 (s) | run2 (s) | run3 (s) |
|---| ---| --- | --- |
| master | 3170 | 3136 | 3134 |
| Dynamic (OFF) | 3250 | 3064 | 3087 |
| Dynamic (ON) | 3372 | 3414 | 3384 |
---

Dynamic branch with dynamic containers turned off seems to run faster than master, which I think might be because of Issue #947 - the safety margin now would directly be `skin` always (instead of ```skinPerStep * iterationSinceRebuild```)

This would be a good time for a review again :)

SamNewcome (changes requested):
Nice!

My key concern is with the unsuitability of the weighted average used in the AutoTuner of iteration times spent rebuilding vs non-rebuilding:

My naive but better than nothing suggestion is to, between tuning phases, collect an average rebuild frequency which is used for the weighted average for the following tuning phase.

This could be done in a separate PR. If so, let me know and I will add an issue.

src/autopas/LogicHandler.h
```diff
@@ -587,6 +595,52 @@ class LogicHandler {
...
   getParticleBuffers() const;
 
+  /**
+   * Getter for the mean rebuild frequency.
+   * Helpful for determining the frequency for the dynamic containers
+   * @return value of the mean frequency as double
+   */
+  [[nodiscard]] double getMeanRebuildFrequency() const {
+    if (_rebuildIntervals.empty()) {
+      return 0.;
```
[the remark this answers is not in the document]
manishmishra6016: done, see [74e6d8](https://github.com/AutoPas/AutoPas/pull/821/commits/74e6d8b9a93aaa19bccd24e94131a82e50db4ae2)

docs/userdoc/ContainerInterfaceModel.md
```diff
@@ -10,6 +9,11 @@ Particles leaving the domain are always returned, and particles can be added and
 Historically this interface behavior was also called 'Linked-Cells-like'.
 For periodic boundary conditions or in an MPI-parallel simulation, the user is responsible for inserting the appropriate halo particles.
 
+### Dynamic Rebuilding
+The particle neighbor lists are updated when a particle moves more than `verletSkin / 2` or at the rebuild frequency. 
+For this, every particle stores the position at which its neighbor list was rebuilt last in a member called `rAtRebuild`.
+This was implemented in the [PR 821](https://github.com/AutoPas/AutoPas/pull/821), so refer to it for more details.
+
```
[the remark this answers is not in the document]
manishmishra6016: done, see [74e6d8](https://github.com/AutoPas/AutoPas/pull/821/commits/74e6d8b9a93aaa19bccd24e94131a82e50db4ae2)

This made me go on a tangent. I realised that I had removed the `fastParticlesThrow` mechanism when implementing the dynamic containers. I added it back for the static container case now.

manishmishra6016:
> Nice!
> 
> My key concern is with the unsuitability of the weighted average used in the AutoTuner of iteration times spent rebuilding vs non-rebuilding:
> 
> My naive but better than nothing suggestion is to, between tuning phases, collect an average rebuild frequency which is used for the weighted average for the following tuning phase.
> 
> This could be done in a separate PR. If so, let me know and I will add an issue.

This is on my To-do, but I would indeed like to do it in a separate PR. And there is already an issue about this: #954

docs/userdoc/ContainerInterfaceModel.md
```diff
@@ -1,6 +1,7 @@
 # Container Interface Model
 
-The behavior described in this section is completely hidden from AutoPas users. 
+The behavior described in this section is completely hidden from AutoPas users.
+To ensure functionality, particles must not move more than `verletSkinPerTimestep / 2` per timestep.
```
SamNewcome, line 4: This should be moved under the static rebuilding section as it does not apply to dynamic

examples/md-flexible/src/TimeDiscretization.cpp
```diff
@@ -30,7 +41,29 @@ void calculatePositionsAndResetForces(autopas::AutoPas<ParticleType> &autoPasCon
...
+      if (fastParticlesThrow) {
+        throwException = true;
+      }
+    }
+  }
+
+  if (throwException) {
+    throw std::runtime_error("At least one particle was too fast!");
+#endif
   }
```
SamNewcome, line 67: I don't really like that this `}` is for different things depending on the CMake variable. If `AUTOPAS_ENABLE_DYNAMIC_CONTAINERS`, then this is for the above `if (throwException)`; otherwise, it is for the `for` loop. I would personally keep the same `}` to end the `for` loop and then have a separate CMake branch around this `if` statement, even if it does mean extra CMake variable branching.

examples/md-flexible/src/TimeDiscretization.h
```diff
@@ -28,10 +28,13 @@ namespace TimeDiscretization {
  * @param particlePropertiesLibrary The particle properties library for the particles in the container.
  * @param deltaT The time step width.
  * @param globalForce Base force value to which every particle is reset.
+ * @param fastParticlesThrow When true throws an exception if particles moved too far for verlet technique
+ * (>skin/2/rebuildFrequency).
```
SamNewcome, line 32:
```suggestion
 * @param fastParticlesThrow When true throws an exception if particles moved too far to ensure the static containers' skins remain valid (>skin/2/rebuildFrequency). 
 * Does nothing if dynamic containers are enabled.
```
Probably is not correct for clang format

src/autopas/LogicHandler.h
```diff
@@ -597,12 +597,14 @@ class LogicHandler {
 
   /**
    * Getter for the mean rebuild frequency.
-   * Helpful for determining the frequency for the dynamic containers
+   * Helpful for determining the frequency for the dynamic containers as well as for determining fast particles by
+   * computing skinPerStep for static container
    * @return value of the mean frequency as double
    */
   [[nodiscard]] double getMeanRebuildFrequency() const {
```
SamNewcome, line 604: This is never used for static containers so is somewhat confusing.

manishmishra6016: done, see [4da4355](https://github.com/AutoPas/AutoPas/pull/821/commits/4da43550d82b376771ada1352a61fc648ff9e0d9)

I was planning to use it for static case but ended up not using it. I have removed it now and added a comment that the function is only used for dynamic containers right now.

docs/userdoc/ContainerInterfaceModel.md
manishmishra6016: done, see [4da4355](https://github.com/AutoPas/AutoPas/pull/821/commits/4da43550d82b376771ada1352a61fc648ff9e0d9)

examples/md-flexible/src/TimeDiscretization.cpp
manishmishra6016: done, see [4da4355](https://github.com/AutoPas/AutoPas/pull/821/commits/4da43550d82b376771ada1352a61fc648ff9e0d9)

SamNewcome (approved):

FG-TUM (changes requested):
Just some small suggestions

src/autopas/LogicHandler.h
```diff
@@ -169,6 +172,9 @@ class LogicHandler {
 
       // We will do a rebuild in this timestep
       if (not _neighborListsAreValid.load(std::memory_order_relaxed)) {
+#ifdef AUTOPAS_ENABLE_DYNAMIC_CONTAINERS
+        _rebuildIntervals.push_back(_stepsSinceLastListRebuild);
```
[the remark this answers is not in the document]
manishmishra6016: done, see [76b9d0b](https://github.com/AutoPas/AutoPas/pull/821/commits/76b9d0b753e0a589161b3b600102e1dfde1a61cf)

src/autopas/AutoPasDecl.h
```diff
@@ -1107,7 +1107,7 @@ class AutoPas {
   /**
    * Specifies after how many pair-wise traversals the neighbor lists are to be rebuild.
    */
-  unsigned int _verletRebuildFrequency{20};
+  unsigned int _verletRebuildFrequency{100};
```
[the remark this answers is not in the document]
manishmishra6016: done, see [76b9d0b](https://github.com/AutoPas/AutoPas/pull/821/commits/76b9d0b753e0a589161b3b600102e1dfde1a61cf)

examples/md-flexible/src/TimeDiscretization.cpp
```diff
@@ -37,6 +41,9 @@ void calculatePositionsAndResetForces(autopas::AutoPas<ParticleType> &autoPasCon
     v *= deltaT;
     f *= (deltaT * deltaT / (2 * m));
     const auto displacement = v + f;
+#ifdef AUTOPAS_ENABLE_DYNAMIC_CONTAINERS
+    iter->setR(iter->getR() + displacement);
```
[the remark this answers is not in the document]
manishmishra6016: done, see [76b9d0b](https://github.com/AutoPas/AutoPas/pull/821/commits/76b9d0b753e0a589161b3b600102e1dfde1a61cf)

src/autopas/LogicHandler.h
```diff
@@ -587,6 +595,58 @@ class LogicHandler {
...
+#ifdef AUTOPAS_ENABLE_DYNAMIC_CONTAINERS
+    if (_rebuildIntervals.empty()) {
+      return static_cast<double>(_neighborListRebuildFrequency);
+    } else {
+      // Neglecting the first entry in the vector as _stepsSinceLastListRebuild is initialized to a very large value,
+      // to trigger the first rebuild when starting the simulation,
+      // which gets stored in the vector on the very first rebuild.
+      // This large value is ignored when calculating the mean value below.
+      return std::accumulate(_rebuildIntervals.begin() + 1, _rebuildIntervals.end(), 0.) /
+             static_cast<double>(_rebuildIntervals.size() - 1);
```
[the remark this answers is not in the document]
manishmishra6016: This is a very valid point. I will do this. But I have to rework this anyway as we need to estimate a better rebuild frequency that can be used while comparison of configuration during tuning (Issue #954 ). So, I would like to handle this with that, for now I created an issue for it too #991 - I hope this is fine?

FG-TUM (approved):

Issue #807, opened 2023-12-19
Coverage Report


FG-TUM:
**Is your feature request related to a problem? Please describe.**
It would be cool if our CI could produce some test coverage report.

**Describe the solution you'd like**
Needs further research. Some options:
- [ ] [LCOV](https://github.com/linux-test-project/lcov)
- [ ] Integration into the CI
- [ ] [Coverage badge](https://github.com/GoogleCloudPlatform/lcov-coverage-badge)

**Describe alternatives you've considered**

- [GCOV](https://gcc.gnu.org/onlinedocs/gcc/Gcov.html)
- [CDash](https://cmake.org/cmake/help/book/mastering-cmake/chapter/CDash.html)

I have no idea which coverage tools are good so some research should be done beforehand.

**Additional context**
As soon as we have this, we can, and should, add more tests to increase coverage. Maybe even a CI check that rejects PRs that significantly decrease test coverage.

The MaMiCo guys currently include this, maybe we can use something similar. 
- This is their Dashboard: https://hsu-hpc.github.io/MaMiCo/coverage/index.html
- Their relevant PRs: https://github.com/HSU-HPC/MaMiCo/pull/12 and https://github.com/HSU-HPC/MaMiCo/pull/36

SamNewcome:
Could you add some more details on what you mean by this? E.g. a link to a web page, or to an example test coverage report, or to the PR from MaMiCo.

In particular, what is meant by LCOV and Coverage Badge?

FG-TUM:
See the updated description. I don't really care which tool we use in the end.

thesamriel:
Implemented in #888. Further coverage drama to be read in #927

PR #824, opened 2024-01-31
Fix CellBlock3D::getCellBoundingBox()

FG-TUM:
# Description

Fixes the value clamping in `CellBlock3D::getCellBoundingBox()` when we calculate the bounding box of a cell at the lower edge of the domain.

- [x] Fixed bug
- [x] Adapted some tangentially related documentation.

## Additional Context

- This caused the particle duplication bug in ls1 because particles were not found by the `LeavingParticleCollector`. Thus during the update, some halos were not removed and their corresponding owned partners were placed right on top of them.
- *Why was this not an issue before?* Not sure. I think when I rewrote the Iterators in #712 I started using `CellBlock3D::getCellBoundingBox()` significantly more than it was used previously. Later when I rewrote `LeavingParticleCollector` in #792 this was used even more, especially around domain borders. So maybe we just didn't use the function for these edge cases before? Seems like another important argument for introducing test coverage -> #807 

# How Has This Been Tested?

- [x] fixed bug in ls1
- [x] New unit tests

### DIFF
--- a/src/autopas/containers/CellBlock3D.h
+++ b/src/autopas/containers/CellBlock3D.h
@@ -458,8 +458,8 @@ inline std::pair<std::array<double, 3>, std::array<double, 3>> CellBlock3D<Parti
 
     boxmax[d] = boxmin[d] + this->_cellLength[d];
 
-    // This must not be an else to the if block above.
-    // If the above is true, this can also be true if there is e.g. only one cell
+    // This must not be an else to the if block above as both cases can be true.
+    // e.g. if there is only one cell
     if (index3d[d] == this->_cellsPerDimensionWithHalo[d] - _cellsPerInteractionLength - 1) {
       boxmax[d] = _boxMax[d];
     } else if (index3d[d] == this->_cellsPerDimensionWithHalo[d] - 1) {
--- a/src/autopas/containers/CellBlock3D.h
+++ b/src/autopas/containers/CellBlock3D.h
@@ -447,16 +447,18 @@ inline std::pair<std::array<double, 3>, std::array<double, 3>> CellBlock3D<Parti
   for (int d = 0; d < 3; d++) {
     // defaults
     boxmin[d] = index3d[d] * this->_cellLength[d] + _haloBoxMin[d];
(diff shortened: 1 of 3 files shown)


src/autopas/containers/CellBlock3D.h
```diff
@@ -447,16 +447,18 @@ inline std::pair<std::array<double, 3>, std::array<double, 3>> CellBlock3D<Parti
...
-      boxmax[d] = this->_cellLength[d];
     } else if (index3d[d] == _cellsPerInteractionLength) {
+      // Case: we are at the lower boundary of the non-halo box
       boxmin[d] = _boxMin[d];
     }
-    // no else, as this might ALSO be 1
+
+    boxmax[d] = boxmin[d] + this->_cellLength[d];
+
+    // This is not an else to the if block above, if there is only one cell
```
SamNewcome, line 461: Can you rephrase this? I'm not super sure what is meant.

```suggestion
    // This is not necessarily an else case to the if block above e.g. in the case that there is only one cell
```

SamNewcome (changes requested):

thesamriel (approved):

src/autopas/containers/CellBlock3D.h
FG-TUM: How about that: 4288da1e2

src/autopas/containers/CellBlock3D.h
```diff
@@ -447,16 +447,19 @@ inline std::pair<std::array<double, 3>, std::array<double, 3>> CellBlock3D<Parti
...
     } else if (index3d[d] == _cellsPerInteractionLength) {
+      // Case: we are at the lower boundary of the non-halo box
       boxmin[d] = _boxMin[d];
     }
-    // no else, as this might ALSO be 1
+
+    boxmax[d] = boxmin[d] + this->_cellLength[d];
+
+    // This must not be an else to the if block above.
+    // If the above is true, this can also be true if there is e.g. only one cell
```
SamNewcome, line 462:
```suggestion
    // This must not be an else to the if block above as both cases can be true.
    // e.g. if there is only one cell
```

SamNewcome (approved):

PR #826, opened 2024-02-06
Fix leaving particle collector halo boxes

FG-TUM:
# Description

Fixes a bug in `LeavingParticleCollector` that caused particle duplication in ls1.

The halo box did not consider the interaction length

## Related Pull Requests

- builds on top of #824

## ~Resolved Issues~

# How Has This Been Tested?

- [x] New test for `LeavingParticleCollector`
- [x] ls1 simulation (Argon example from CI) that crashed before doesn't crash anymore


src/autopas/containers/LeavingParticleCollector.h
```diff
@@ -18,25 +18,32 @@
...
-std::vector<typename ContainerType::ParticleType> collectParticlesAndMarkNonOwnedAsDummy(ContainerType &container) {
+std::array<std::tuple<const std::array<double, 3>, const std::array<double, 3>>, 6> calculateHaloVolumes(
+    const ContainerType &container) {
   using namespace autopas::utils::ArrayMath::literals;
 
-  const auto upperBoundForMisplacement = container.getCutoff();
-  const auto containerDimensions = container.getBoxMax() - container.getBoxMin();
-  const auto lowerHaloCorner = container.getBoxMin() - upperBoundForMisplacement;
-  const auto upperHaloCorner = container.getBoxMax() + upperBoundForMisplacement;
+  const auto upperBoundForMisplacement = container.getVerletSkin() / 2.0;
```
SamNewcome, line 35: I'm not happy with the longevity of this assumption: With the dynamic rebuilding containers this stops becoming a reasonable upper bound.

I would (unless there a serious performance/other problems) keep this as cutoff -> I think we can still assume that a particle will not travel further than cutoff.

tests/testAutopas/tests/containers/LeavingParticleCollectorTest.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+#include <gtest/gtest.h>
+
+#include "autopas/containers/LeavingParticleCollector.h"
+#include "autopas/containers/linkedCells/LinkedCells.h"
+#include "testingHelpers/commonTypedefs.h"
+
+/**
+ *
+ */
```
SamNewcome, line 17: No comment :(

SamNewcome (changes requested):

src/autopas/containers/LeavingParticleCollector.h
FG-TUM: This should decrease performance because we search a significantly larger region. But also, this leads to particle duplication and thus explosions in the Argon example. Honestly, I'm not completely sure why this is the case, but raising or lowering this bound leads to wrong total energies (=missed interactions) or crashes due to particle duplication.

tests/testAutopas/tests/containers/LeavingParticleCollectorTest.cpp
FG-TUM: done ae83b424f

src/autopas/containers/LeavingParticleCollector.h
FG-TUM: > With the dynamic rebuilding containers, this stops becoming a reasonable upper bound.

Shouldn't the bound still stay the same and just the number of iterations when we expect it to be hit changes?

SamNewcome: ### 1. I don't think we can assume that a particle will not move further than verlet skin /2 in the future

I would generally expect this to be true. But this requires the skin to be chosen sensibly which, with the dynamic containers, becomes a performance optimisation problem and not a requirement. 

Furthermore, this assumption also assumes that the optimal verlet skin size is larger than what any particle should be able to move: consider a scenario in which on rare occasions a particle might move quickly. It might be performant to generally still use a small verlet skin and accept that occasionally there will be two consecutive rebuilds. This becomes more likely when you consider partially rebuilding verlet lists in which a single particle exceeding the skin size does not result in a full rebuild so it is beneficial to pick an even small skin size.

### 2. I think now that what you have done here is fine

In the case that a particle moves further than verlet skin / 2 into the halo (but still less than cutoff), it should still get caught by this iterator. So I wouldn't change this back to cutoff.


### 3. I think there's a more fundamental problem 

I'm not sure why this misplacement stuff is needed in the first place? Surely this is a flaw in the region iterators if they cannot handle the misplacement of particles? We shouldn't have to account for this misplacement manually.

FG-TUM: ad 1.
Ok, I see your point, but I think we should tackle the problem when it arises because I don't really know how to do that here and now.

ad 3.
Hmm probably misplacement is the wrong term here.
The halo region only extends out of the container by the interaction length. This distance is added to the halo region, towards the outside and the inside because we want to also catch all particles that left this region (meaning their coordinates are outside. Doesn't matter where they are physically stored).
However, while writing this, I realized that there actually IS a region iterator bug that needs to be addressed for all containers :/ My suggestion is to merge this PR and address the bug in a separate one that I'll start tomorrow.

I'll update this comment with a reference

src/autopas/containers/LeavingParticleCollector.h
```diff
@@ -18,25 +18,32 @@
...
 namespace autopas::LeavingParticleCollector {
 /**
- * Collects leaving particles and marks halo particles as dummy.
- * @note This function does not move or actually delete any particles!
- * @tparam ContainerType The type of the container.
- * @param container The container from which particles should be collected.
- * @return Returns a vector of leaving particles.
+ * For a given container that supports `getCutoff()`, `getBoxMin()` and `getBoxMax()`,
+ * calculates a set of 6 non-overlapping boxes that fully cover the halo.
+ *
```
SamNewcome, line 23:
```suggestion
 * These boxes extend into the container to catch halo particles which have moved inside the container.
 *
```

src/autopas/containers/LeavingParticleCollector.h
SamNewcome: Okay, I think I just didn't understand why the boxes extended into the container as I wasn't considering halo particles moving inside the container. In which case you can ignore point 3 although I'm glad you also found a bug :D

I've added a suggestion above in the documentation to make this clearer.

SamNewcome (changes requested):

SamNewcome (approved):

PR #830, opened 2024-02-22
Fix Region Iterators

FG-TUM:
# Description

Region iterators do not consider particles that are in the region but are stored in cells that are completely outside the region. This affects all containers.

# Todo
- [x] pass timesteps since last rebuild somehow to containers
- [x] performance measurements

## Related Pull Requests

- noticed in #826 

## Performance
The following scenario with 179800 particles was used for the performance comparison with 8 HW threads:

```
./md-flexible --container LinkedCells --functor Lennard-Jones-AVX2 --traversal lc_c01 --newton3 disabled --data-layout AoS --verlet-rebuild-frequency 10 --verlet-skin-radius-per-timestep 0.025 --cutoff 2.5 --cell-size 1 --deltaT 0.001 --iterations 2000 --boundary-type reflective --fastParticlesThrow true --particle-generator closest --box-length 50 --particle-spacing 1.0 --no-progress-bar true
```

Since `getRegionIterator` is called in `updateContainer` and `reflectParticlesAtBoundaries`, the times were measured here. In this scenario, around 12500 particles are iterated by the `regionIterator` at boundaries in one timestep in `reflectParticlesAtBoundaries`.

The following table contains the times of 5 measurements in each case.

|                              | New regionIterator ([61a9789](https://github.com/AutoPas/AutoPas/pull/830/commits/61a978946ed9ff45a8f1a1455cb377a0dc326a7c))                         | avg | master ([2545a53](https://github.com/AutoPas/AutoPas/commit/2545a538de175046cf2bcd589eed98e75f9ac8be))                                 | avg |
|------------------------------|----------------------------------------|-----|----------------------------------------|-----|
| `updateContainer`              | 1.945s, 1.938s, 1.941s, 1.956s, 1.939s | 1.9438s | 1.953s, 1.945s, 1.946s, 1.953s, 1.951s | 1.9496s |
| `reflectParticlesAtBoundaries` | 9.716s, 9.723s, 9.677s, 9.726s, 9.735s | 9.7154s | 9.241s, 9.231s, 9.218s, 9.265s, 9.248s | 9.2406s

# How Has This Been Tested?

A new `regionIterator` test (`RegionParticleIteratorTest::testParticleMisplacement`) has been implemented for all containers. This test checks whether particles that have moved beyond the boundaries of a data structure element but are still stored in the old data structure element are found. This test works on the assumption that the halo box is in a new data structure element (cell, tower, ...).
Particles are placed close to the boundaries within the simulation box. The particles are then moved slightly outside the simulation box without updating the data structure. It is checked whether the particles are still found with a region that no longer includes the old data structure element in which the particles are still stored.

- [x] New region iterator test (for all containers)

### DIFF
--- a/src/autopas/containers/linkedCells/LinkedCells.h
+++ b/src/autopas/containers/linkedCells/LinkedCells.h
@@ -64,7 +64,9 @@ class LinkedCells : public CellBasedParticleContainer<FullParticleCell<Particle>
               LoadEstimatorOption loadEstimator = LoadEstimatorOption::squaredParticlesPerCell)
       : CellBasedParticleContainer<ParticleCell>(boxMin, boxMax, cutoff, skinPerTimestep * rebuildFrequency),
         _cellBlock(this->_cells, boxMin, boxMax, cutoff + skinPerTimestep * rebuildFrequency, cellSizeFactor),
-        _loadEstimator(loadEstimator) {}
+        _loadEstimator(loadEstimator),
+        _skinPerTimestep(skinPerTimestep),
+        _rebuildFrequency(rebuildFrequency) {}
 
   [[nodiscard]] ContainerOption getContainerType() const override { return ContainerOption::linkedCells; }
 
@@ -244,12 +246,17 @@ class LinkedCells : public CellBasedParticleContainer<FullParticleCell<Particle>
                                                                IteratorBehavior iteratorBehavior,
                                                                const std::array<double, 3> &boxMin,
                                                                const std::array<double, 3> &boxMax) const {
+    using namespace autopas::utils::ArrayMath::literals;
+
     // first and last relevant cell index
     const auto [startCellIndex, endCellIndex] = [&]() -> std::tuple<size_t, size_t> {
       if constexpr (regionIter) {
         // if particles might have moved extend search box for cells here
-        // TODO
-        return {_cellBlock.get1DIndexOfPosition(boxMin), _cellBlock.get1DIndexOfPosition(boxMax)};
+        const auto boxMinWithSafetyMargin = boxMin - (_skinPerTimestep * _rebuildFrequency);
+        const auto boxMaxWithSafetyMargin = boxMax + (_skinPerTimestep * _rebuildFrequency);
+
+        return {_cellBlock.get1DIndexOfPosition(boxMinWithSafetyMargin),
+                _cellBlock.get1DIndexOfPosition(boxMaxWithSafetyMargin)};
       } else {
         if (not(iteratorBehavior & IteratorBehavior::halo)) {
           // only potentially owned region
@@ -501,6 +508,8 @@ class LinkedCells : public CellBasedParticleContainer<FullParticleCell<Particle>
           (iteratorBehavior & IteratorBehavior::halo and _cellBlock.cellCanContainHaloParticles(cellIndex));
       if constexpr (regionIter) {
         // short circuit if already false
+        // TODO: should we remove the if isRelevant here, since a cell with a different ownership-state could contain a
+        // moved particle with another ownership state?
         if (isRelevant) {
           // is the cell in the region?
           const auto [cellLowCorner, cellHighCorner] = _cellBlock.getCellBoundingBox(cellIndex);
@@ -547,6 +556,10 @@ class LinkedCells : public CellBasedParticleContainer<FullParticleCell<Particle>
    * load estimation algorithm for balanced traversals.
    */
   autopas::LoadEstimatorOption _loadEstimator;
+
+  double _skinPerTimestep;
+
+  unsigned int _rebuildFrequency;
 };
 
 }  // namespace autopas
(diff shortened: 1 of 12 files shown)


dmartin13:
@FG-TUM review please.

src/autopas/containers/linkedCells/LinkedCells.h
```diff
@@ -499,6 +509,8 @@ class LinkedCells : public CellBasedParticleContainer<FullParticleCell<Particle>
           (iteratorBehavior & IteratorBehavior::halo and _cellBlock.cellCanContainHaloParticles(cellIndex));
       if constexpr (regionIter) {
         // short circuit if already false
+        // TODO: should we remove the if isRelevant here, since a cell with a different ownership-state could contain a
+        // moved particle with another ownership state?
```
FG-TUM, line 513:
```suggestion
```

No, the cells will always only contain halo or owned particles for linked cells. It might only be the case that an owned particle is in a position that is inside a halo cell.

src/autopas/containers/linkedCells/LinkedCells.h
```diff
@@ -545,6 +557,11 @@ class LinkedCells : public CellBasedParticleContainer<FullParticleCell<Particle>
    * load estimation algorithm for balanced traversals.
    */
   autopas::LoadEstimatorOption _loadEstimator;
+
+  /**
+   * Skin distance a particle is allowed to move in one time-step.
+   */
+  double _skinPerTimestep;
```
FG-TUM, line 564: isn't this relevant for all containers now and thus could be moved into the `ParticleContainerInterface` (this would also affect the container constructors)?

tests/testAutopas/tests/iterators/RegionParticleIteratorTest.cpp
```diff
@@ -14,9 +14,22 @@
...
+using ::testing::Combine;
+using ::testing::UnorderedElementsAreArray;
+using ::testing::Values;
+using ::testing::ValuesIn;
+
+// Helper struct to represent region iterator boxes.
+struct Box {
+  std::array<double, 3> min;
+  std::array<double, 3> max;
+};
```
FG-TUM, line 26: I feel like we should add helper types like Point and Box to some utils class... but not in this PR

tests/testAutopas/tests/iterators/RegionParticleIteratorTest.cpp
```diff
@@ -212,3 +218,138 @@ TEST_F(RegionParticleIteratorTest, testForceSequential) {
...
+        for (auto [z, zMirrored] : generateInteresting1DPositions(boxMin[2], boxMax[2])) {
+          if (not(isNear(x, (boxMax[0] - boxMin[0]) / 2.0) and isNear(y, (boxMax[1] - boxMin[1]) / 2.0) and
+                  isNear(z, (boxMax[2] - boxMin[2]) / 2.0))) {
+            positionsInsideBox.push_back({x, y, z});
+            positionsMirroredAtBoundary.push_back({xMirrored, yMirrored, zMirrored});
+
+            // this ensures that a value that is closer than offset to the floored integer, is floored to the next
+            // smaller integer
+            auto getFlooredValue = [offset](double value) {
+              return ((value - floor(value) < (offset - margin)) && (value - floor(value) >= 0)) ? floor(value - 1)
```
FG-TUM, line 271:
```suggestion
              return ((value - floor(value) < (offset - margin)) and (value - floor(value) >= 0)) ? floor(value - 1)
```

tests/testAutopas/tests/iterators/RegionParticleIteratorTest.cpp
```diff
@@ -212,3 +218,138 @@ TEST_F(RegionParticleIteratorTest, testForceSequential) {
...
+                                                                                                 : floor(value);
+            };
+
+            // this ensures that a value that is closer than offset to the ceiled integer, is ceiled to the next greater
+            // integer
+            auto getCeiledValue = [offset](double value) {
+              return ((ceil(value) - value < (offset - margin)) && (ceil(value) - value >= 0)) ? ceil(value + 1)
```
FG-TUM, line 278:
```suggestion
              return ((ceil(value) - value < (offset - margin)) and (ceil(value) - value >= 0)) ? ceil(value + 1)
```

src/autopas/containers/linkedCells/LinkedCells.h
dmartin13: done in [ca2c857](https://github.com/AutoPas/AutoPas/pull/830/commits/ca2c857cd3043cf4d91f215e152900d71de0cea7)

dmartin13:
@FG-TUM review again please.

tests/testAutopas/tests/iterators/RegionParticleIteratorTest.cpp
FG-TUM: -> #844

SamNewcome (approved):

Issue #721, opened 2023-02-24
VarVerletListsAsBuild don't correctly construct their neighbour lists during the first neighbour list build after AutoPas has switched to this container


SamNewcome:
**Describe the bug**
During the first neighbour list build, after AutoPas has switched to this container, VarVerletLists don't perform as expected. The neighbour list build takes significantly less time than expected, from looking at subsequent rebuilds of lists and from comparing to alternative runs of a simulation where VarVerletLists have already been used prior to the iteration which is causing problems. The side effect of this (beyond any potential incorrect force calculations during one rebuild cycle), is that AutoPas selects algorithmic configurations with this container when they are not optimal.

**To Reproduce**
In theory, this should be reproducible from looking at any simulation with full searches, with VarVerletLists.

The particular scenario where this was noticed, was a version of exploding liquid run on 4 MPI ranks across 2 compute nodes of [HSUper](https://www.hsu-hh.de/hpc/en/hsuper/). The exact jobscript and yaml input file are added below. The comparison can be made easier if you also look at the exact same simulation without any tuning, just VarVerletLists with vvl_as_build and AoS, and compare the iteration timings.

**Expected behaviour**
The rebuild timings during these first rebuilds should approximately match subsequent rebuilds, and alternative simulation runs with no tuning. If there are also incorrect forces being calculated, these should also be corrected.

**Screenshots**
I will add a graph demonstrating this point by 7th March 2023.

**Build environment (please complete the following information):**
* OS: Linux (HSUper)
* Compiler: GCC 12.1.0
* Build System: GNU Make 4.2.1
* Built with MPI

**explodingLiquid.yaml**
```
container                        :  [LinkedCells, VarVerletListsAsBuild, VerletClusterLists, VerletLists, VerletListsCells, PairwiseVerletLists]
traversal                        :  [lc_sliced, lc_sliced_balanced, lc_sliced_c02, lc_c01, lc_c01_combined_SoA, lc_c04, lc_c04_HCP, lc_c04_combined_SoA, lc_c08, vcl_c06, vcl_c01_balanced, vcl_sliced, vcl_sliced_c02, vl_list_iteration, vlp_sliced, vlp_sliced_c02, vlp_c08, vvl_as_built]
# data-layout                      :  [AoS, SoA]
newton3                          :  [enabled]
verlet-rebuild-frequency         :  10
verlet-skin-radius-per-timestep  :  0.01
verlet-cluster-size              :  4
[...]

PR #832, opened 2024-02-28
Make VCL no longer discouraged and make VVL discouraged

SamNewcome:
# Description

VVL has a bug which keeps on cropping up (see #721 ). As far as I am aware, there are no major problems from using this container except that the tuner might select it without it being optimal. Without a fix, this container should at the very least be discouraged.

Also, I do not see why Verlet Cluster Lists are discouraged.

### DIFF
--- a/src/autopas/options/ContainerOption.h
+++ b/src/autopas/options/ContainerOption.h
@@ -92,7 +92,7 @@ class ContainerOption : public Option<ContainerOption> {
    * @return
    */
   static std::set<ContainerOption> getDiscouragedOptions() {
-    return {Value::directSum, Value::linkedCellsReferences, Value::verletClusterLists, Value::octree};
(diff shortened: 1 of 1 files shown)


FG-TUM:
What is your reason for encouraging VCL? We discouraged it because it had significant performance issues. Are they mostly resolved?

SamNewcome:
> What is your reason for encouraging VCL? We discouraged it because it had significant performance issues. Are they mostly resolved?

I wasn't aware of performance issues. In this case I will re-add this. To avoid a repeat of what I have just done we should add a short reason why each option is discouraged. Do you have some data for VCL which I can point to or should I just write performance issues?

FG-TUM (approved):
looks like our worst performance issues with VCL are resolved.

PR #847, opened 2024-04-22
Remove redundant code in exchangeHaloParticles()

manishmishra6016:
# Description

This piece of code might be redundant

Reasons:

- `collectHaloParticlesForLeftNeighbor()` collects the particle to be sent to neighbours
-  `_haloParticles` is cleared before the loop, so it is empty?

`md-flexible` simulation for `fallingDrop.yaml` example case works fine and time measurement for `HaloParticleExchange` is 7.051s after removing this part compared to 7.728s before.

No additional tests required.


manishmishra6016:
The implementation is correct, hence closing the pull request

Issue #833, opened 2024-02-29
Intrinsics Wrappers


FG-TUM:
**Is your feature request related to a problem? Please describe.**
We have hand-crafted intrinsics versions of the LJ-Functor for AVX2, AVX512, SVE. Maintaining this is unnecessarily complicated.

**Describe the solution you'd like**
- [ ] Bin them all!
- [ ] Have only one version that uses some wrapper library which abstracts the individual implementations away. The most interesting candidate for this seems to be [google highway](https://github.com/google/highway) because, in contrast to most others, it supports SVE.

**Describe alternatives you've considered**
Keep everything as it is and have three times as much pain as needed.

**Additional context**
- [ ] Before actually deleting everything, we should compare the performance to the functors we already have.

PR #850, opened 2024-04-25
Count FLOPs and HitRate inside functor and output them to a log file

### LINKED ISSUE #695: Provide hitRate per in iterationPerformance log
Currently, the hit rate (fraction of distance calculations that lead to force contributions) is not logged per iteration. This is a handy metric in determining the performance of an algorithm and would be good also to log this.

Output to the iterationPerformance log the hit rate for that iteration.

Outputting hit rate could be an additional log option in the CMake configuration for AutoPas. I don't think this is worthwhile though, as hit rate is only one additional attribute and won't add much additional clutter to .csv file if not needed.

### LINKED ISSUE #707: The Flop Count outputs in md-flexible are misleading, and the FlopCounterFunctor does not have the functionality required for multi-site molecules
**Describe the bug**
Currently, in md-flexible,  if dontMeasureFlops is set to _true_, GFLOP, GFLOP/sec and hit-rate statistics are outputted. However, FLOP and Hit-Rate are only computed on the final time-step. GFLOP/sec is calculated using this final GFLOP count, but with the total time for the whole simulation.

Furthermore, FlopCounterFunctor does not work for kernel calls where variable numbers of FLOPs. This is a particular problem with Multi-Site molecules. This issue could be tackled together, or correct FLOP/Hit-Rate measuring for single-site molecules could be implemented first.

**Ideal solution**

- If dontMeasureFlops is set to _false_, FLOP and Hit-Rate's are computed every iteration and added to a "Total" (In the case of Hit-Rate this total is meaningless until we divide by the number of iterations.
- At the end, if dontMeasureFlops is set to false, the total GFLOPs, the average GFLOPs/sec, and the average Hit-rate are outputted.

**Alternatives and extensions**

- Switch from dontMeasureFlops to measureFlops.
- Output metrics every iteration to the iterationPerformance log, as per issue #695

### LINKED ISSUE #841: Refactor FLOP counting
So I've been looking into the `FlopCounterFunctor`. This was originally (pre-multi-site in md-flex) very tailored towards simple single-site molecules where every molecule-to-molecule interaction costs the same. I later generalised it by using the molecule's type ID to handle differently sized multi-site molecules, which results in different number of FLOPS for a molecule-to-molecule interaction. This was ultimately a bit of a "fudge" job. Now I'm looking at functors that don't have a type ID. If I extend the Flop Counter Functor to include this it gets even more messy and awkward to maintain. Similarly, in #751, an additional Flop counting class is provided for 3-body interactions. So I'm looking at better approaches:

My proposed solution is that we forget about trying to have some general case to handle this. We move the FLOP counting inside the actual functors. This would be done with template parameters to avoid having to do FLOP counting when we don't want to. 

There are three advantages to this:

- We can easily tailor how we count FLOPs to the functor. This avoids this need for awkward generalisation of the FlopCounterFunctor that is seemingly required with every new functor that is not just new vector instructions.
- It is probably more efficient to calculate FLOPs within the functor than as a separate functor (less repeated calculations + not needing to reload every pair of molecules)
- It is better for any case in which the cutoff is non-standard, e.g. multi-site molecules with site-to-site cutoffs, simulations with forces with different length cutoffs

The key disadvantage is that this provides extra work for the writer of a functor. But with examples of other functors, this is not too much more work than writing a new functor is anyway.

**Additional context**
I am not proposing forcing md-flexible to either always count FLOPs or always not count FLOPs. The functionality of md-flexible can be kept the same (so an option of counting FLOPs in the final iteration only).

Issue #707 considers two ways in which the FLOP counting in md-flexible could be used differently but not how the FLOP counting itself should be modified.

SamNewcome:
# Description

The FLOP counter functor was removed as it was too rigid. See #841 for more details on this "rigidity". The proposed alternative for counting FLOPs was for it to be handled internally by a functor.  The advantages are outlined in #841. This has been implemented.

At the same time, md-flexible's FLOP outputting mechanic was misleading. See #707 for more details. A solution to this was sketched as an idea in #695. 

This pull request

- [x] Refactors the FLOP counting interface into the functors themselves.
- [x] Implements the FLOP counting mechanic for `LJFunctor`. Implementing the FLOP counting mechanism inside the other vectorisations of this functor is not worthwhile, as their future is questionable in the face of potential vectorised wrapper variants, as discussed in #833. The `MultisiteLJFunctor` will not get an implementation in this PR, as the implementation of the functor is suboptimal and will be replaced in #810.
- [x] Refactors and extends the FLOP counting tests.
- [x] Adds an extra optional logger to AutoPas which logs FLOPs and HitRate.

## Related Pull Requests

- #PR #751 Implements a dedicated FLOP Counter Functor for 3-body functors.

## Resolved Issues

- [x] fixes #841 
- [x] fixed #707 
- [x] fixed #695 

# How Has This Been Tested?

- [x] Existing FLOP counter functor test is refactored as a test for `LJFunctor`'s FLOP counter.
- [x] Existing tests are extended to also test FLOP counting with globals
- [x] Existing tests are extended to also test data races in FLOP counting with SoA Verlet.
- [x] The FLOPs counts produced by the old functor is compared against the FLOP counts produced by the new functor for the first iteration of falling drop.
- [x] Tested that SVE functor compiles and runs on ARM cluster, with FLOP logging disabled and that it compiles with FLOP login enabled.


SamNewcome:
Results from comparing the FLOPs and Hit Rate of master (`88b0d7a8` - technically out of date but the only changes since then were to documentation) and this branch (`ea9d52f3`). Also included are the number of distance calculations and N3 & no N3 kernel calls for the new version.


<!DOCTYPE html>

<html>
<head>
	
	
	<title></title>
	<meta name="generator" content="LibreOffice 24.2.3.2 (Linux)"/>
	<style type="text/css">
		body,div,table,thead,tbody,tfoot,tr,th,td,p { font-family:"Liberation Sans"; font-size:x-small }
		a.comment-indicator:hover + comment { background:#ffd; position:absolute; display:block; border:1px solid black; padding:0.5em;  } 
		a.comment-indicator { background:red; display:inline-block; border:1px solid black; width:0.5em; height:0.5em;  } 
		comment { display:none;  } 
	</style>
	
</head>

<body>

  |   |   |   | New |   |   |   |   | Old |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
  |   |   |   | FLOPs | HitRate | Distance Calls | N3 Kernel Calls | NoN3 Kernel Calls | FLOPs | HitRate
LC | C08 | N3=ON | AoS | 34716644 | 0.299 | 2594197 | 775726 | 0 | 34716600 | 0.299
LC | C08 | N3=ON | SoA | 75444212 | 0.1009 | 7685143 | 775726 | 0 | 75444200 | 0.1009
LC | C08 | N3=OFF | AoS | 64778932 | 0.299 | 5188394 | 0 | 1551452 | 64778900 | 0.299
LC | C08 | N3=OFF | SoA | 139512164 | 0.0843 | 14963914 | 289244 | 972964 | 139512000 | 0.0843
VLC | C18 | N3=ON | AoS | 26098140 | 0.5113 | 1516884 | 775726 | 0 | 26098100 | 0.5113
VLC | C19 | N3=ON | SoA | 26098140 | 0.5113 | 1516884 | 775726 | 0 | 26098100 | 0.5113
VLC | C20 | N3=OFF | AoS | 47541924 | 0.5113 | 3033768 | 0 | 1551452 | 47541900 | 0.5113
VLC | C21 | N3=OFF | SoA | 47541924 | 0.5113 | 3033768 | 0 | 1551452 | 47541900 | 0.5113


</body>

</html>

As you can see, these match (baring the FLOP count for the old FLOP counter not having as many significant figures).

-------

Some extra musings on the FLOP counts and hit rates, for anybody interested:

For Linked Cells
- AoS has fewer distance checks (and thus fewer FLOPs and higher hit rate) than SoA because it uses a sorting threshold (have checked that AoS and SoA FLOP count matches for LC, C08, N3=ON with threshold = high (2000))
- For AoS, going from N3=ON to OFF, little under doubles FLOPs and exactly double the kernel and distance calls, as expected
- For SoA, going from N3=ON to OFF, changes the hit rate. With AoS, when you switch between N3 enabled and disabled, the number of distance/kernel calls doubles/halves exactly and the hit rate remains the same. With SoA, we always handle internal cell interactions with N3 enabled so, as you switch between the two, a small proportion of distance/kernel calls does not double/half with the rest. This results in the hit rate not entirely matching. 

For Verlet Lists, everything is as expected (N3 doubles/halves calls, hit rate same)

examples/md-flexible/src/TypeDefinitions.h
```diff
@@ -50,16 +50,33 @@ using ParticleType = mdLib::MultisiteMoleculeLJ;
...
  * Type of LJFunctorTypeAutovec used in md-flexible.
  * Switches between mdLib::LJFunctor and mdLib::LJMultisiteFunctor as determined by CMake flag
  * MD_FLEXIBLE_MODE.
  */
 #if MD_FLEXIBLE_MODE == MULTISITE
-using LJFunctorTypeAutovec = mdLib::LJMultisiteFunctor<ParticleType, true, true>;
+using LJFunctorTypeAutovec = mdLib::LJMultisiteFunctor<ParticleType, true, true, autopas::FunctorN3Modes::Both, false,
+                                                       mdFlexibleTypeDefs::countFLOPs>;
+#ifdef AUTOPAS_LOG_FLOPS
+#warning "AUTOPAS_LOG_FLOPS enabled but is not suported for compiled functor (LJMultisiteFunctor)!"
```
FG-TUM, line 75:
```suggestion
#pragma message "AUTOPAS_LOG_FLOPS enabled but is not suported for compiled functor (LJMultisiteFunctor)!"
```

In other places we used `#pragma message` over `#warning` because it seems to be supported more widely. AFAIK there is no universal compile time warning mechanism defined by the standard :/
Have you read something different recently?
If not please adapt all following `#warning` too.

src/autopas/LogicHandler.h
```diff
@@ -771,7 +773,15 @@ class LogicHandler {
...
 
+  /**
+   * Logger for configuration used and time spent breakdown of iteratePairwise.
+   */
   IterationLogger _iterationLogger;
+
+  /**
+   * Logger for FLOP count and hit rate.
+   */
+  FLOPLogger _FLOPLogger;
```
FG-TUM, line 784:
```suggestion
  FLOPLogger _flopLogger;
```

Looks a bit confusing with the all-caps

tests/testAutopas/tests/autopasInterface/AutoPasInterfaceTest.cpp
```diff
@@ -20,9 +20,8 @@
 #include "testingHelpers/commonTypedefs.h"
 
 extern template class autopas::AutoPas<Molecule>;
-using LJFunctorGlobals =
-    mdLib::LJFunctor<Molecule, /* shifting */ true, /*mixing*/ false, autopas::FunctorN3Modes::Both,
-                     /*globals*/ true>;
+using LJFunctorGlobals = LJFunctorType</* shifting */ true, /*mixing*/ false, autopas::FunctorN3Modes::Both,
+                                       /*globals*/ true>;
```
FG-TUM, line 24: maybe move this alias to `commonTypedefs`

FG-TUM (changes requested):
I like it. Here a few suggestions.

thesamriel (changes requested):

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorAVX.h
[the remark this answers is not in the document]
SamNewcome: I agree - but I would wait until we have a Google Highway version. We can then review what can and should be trimmed away / consolidated.

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorAVX.h
```diff
@@ -1012,6 +998,16 @@ class LJFunctorAVX
...
+  size_t getNumFLOPs() {
+    AutoPasLog(WARN, "LJFunctorAVX::getNumFLOPs called but is not implemented and will return 0.");
+    return 0;
+  }
+
+  double getHitRate() {
+    AutoPasLog(WARN, "LJFunctorAVX::getHitRate called but is not implemented and will return 0.");
+    return 0;
+  }
+
```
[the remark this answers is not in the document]
SamNewcome: In general, I would leave the runtime warnings, as the compiler warnings could be missed by the user and this functor could be used independently of md-flexible.

But I agree that these warnings aren't so useful if it is covered in the constructor anyway. I would like to keep debug messages here if `countFLOPs=false` (in which case, no constructor warning) but these functions are called regardless.

Happy to move messages to debug.

applicationLibrary/sph/SPHLibrary/SPHCalcDensityFunctor.h
```diff
@@ -70,6 +70,16 @@ class SPHCalcDensityFunctor : public autopas::Functor<Particle, SPHCalcDensityFu
...
+  size_t getNumFLOPs() {
+    AutoPasLog(WARN, "SPHCalcDensityFunctor::getNumFLOPs called but is not implemented and will return 0.");
+    return 0;
+  }
+
+  double getHitRate() {
+    AutoPasLog(WARN, "SPHCalcDensityFunctor::getHitRate called but is not implemented and will return 0.");
+    return 0;
+  }
+
```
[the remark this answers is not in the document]
SamNewcome: Can't move the warning to the constructor because this functor has no `countFLOPs` template parameter. But will change to DEBUG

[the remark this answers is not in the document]
SamNewcome: See `9f7c1b6f`

applicationLibrary/sph/SPHLibrary/SPHCalcHydroForceFunctor.h
```diff
@@ -531,13 +531,14 @@ class SPHCalcHydroForceFunctor : public autopas::Functor<Particle, SPHCalcHydroF
...
-    ///@todo return correct flopcount
-    return 1ul;
+  size_t getNumFLOPs() {
+    AutoPasLog(WARN, "SPHCalcHydroForceFunctor::getNumFLOPs called but is not implemented and will return 0.");
+    return 0;
+  }
+
+  double getHitRate() {
+    AutoPasLog(WARN, "SPHCalcHydroForceFunctor::getHitRate called but is not implemented and will return 0.");
+    return 0;
```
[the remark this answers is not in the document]
SamNewcome: See `9f7c1b6f`

docs/userdoc/CustomApplications.md
```diff
@@ -42,6 +42,10 @@ The critical elements to implement are:
   Indicator functions that return a `std::array` of all `AttributeNames`, which the functor needs to load from the particle to perform the calculation, as well as which fields are written.
 - `isRelevantForTuning()`:
   Indicator function to tell the tuning mechanism if iterations using this functor should be considered or not.
+- `getNumFLOPs()` and `getHitRate()`:
+  These functions return the number of FLOPs per traversal of the container and the hit-rate (the ratio of distance calculations
+  that lead to functor interactions e.g. force contributions.) These functions are only needed if `AUTOPAS_LOG_FLOPS` is
+  set to `ON`, and can be set to return 0 if you are happy without these statistics.
```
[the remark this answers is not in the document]
SamNewcome: See `7ebfcc71`

examples/md-flexible/src/Simulation.cpp
```diff
@@ -587,21 +584,6 @@ void Simulation::logMeasurements() {
         static_cast<double>(_autoPasContainer->getNumberOfParticles(autopas::IteratorBehavior::owned) * _iteration) *
         1e-6 / (static_cast<double>(forceUpdateTotal) * 1e-9);  // 1e-9 for ns to s, 1e-6 for M in MFUPs
     std::cout << "MFUPs/sec                          : " << mfups << std::endl;
-
-    if (_configuration.dontMeasureFlops.value) {
```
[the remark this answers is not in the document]
SamNewcome: Done. See `643b81d8`

tests/testAutopas/tests/autopasInterface/AutoPasInterfaceTest.cpp
SamNewcome: See `6aaea17f `

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctor.h
```diff
@@ -453,7 +502,7 @@ class LJFunctor
         const SoAFloatPrecision lj6 = lj2 * lj2 * lj2;
         const SoAFloatPrecision lj12 = lj6 * lj6;
         const SoAFloatPrecision lj12m6 = lj12 - lj6;
-        const SoAFloatPrecision fac = mask ? epsilon24 * (lj12 + lj12m6) * invdr2 * mask : 0.;
+        const SoAFloatPrecision fac = mask ? epsilon24 * (lj12 + lj12m6) * invdr2 : 0.;
```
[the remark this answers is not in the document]
SamNewcome: See `ed486a74`

src/autopas/LogicHandler.h
```diff
@@ -1176,6 +1186,10 @@ bool LogicHandler<Particle>::iteratePairwisePipeline(Functor *functor) {
                                 tuningTimer.getTotalTime(), measurements.energyPsys, measurements.energyPkg,
                                 measurements.energyRam);
 
+#ifdef AUTOPAS_LOG_FLOPS
```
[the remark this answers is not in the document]
SamNewcome: Done in `35ad4a7056369657c853a0dea09c4e40f7b049cc`

applicationLibrary/molecularDynamics/tests/templateInstatiations/AutoPasInstantiations.cpp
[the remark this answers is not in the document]
SamNewcome: No. Comparing before (191f994f) and after (83d18b80) removing them, using the Linux `time` command on my workstation (Intel(R) Core(TM) i7-10700 CPU @ 2.90GHz) with nothing else significant running:
```
time make mdLibSingleSiteTests -j10
```

Takes 4m47,685s of user time before removal and 5m9,284s afterwards. So, there was a slight slowdown, but considering that this was just one rough measurement, I am happy to assume there were no significant changes to compile time. What would be the reason against using explicit instatiations?

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctor.h
```diff
@@ -684,6 +737,100 @@ class LJFunctor
...
+   * This function is supposed to return useful FLOPs, e.g. without counting masked vector instructions or
+   * workarounds for accumulating globals on owned/halo borders twice. You could argue that, strictly speaking, we
+   * redundantly calculate forces and globals twice in the newton3 case. This function does not treat such "redundant"
+   * calculations as useless.
+   *
+   * Similarly, this function does not treat halo-halo interactions as redundant useless calculations.
+   *
+   * @return number of FLOPs since initTraversal() is called.
+   */
+  size_t getNumFLOPs() {
```
FG-TUM, line 780:
```suggestion
  size_t getNumFLOPs() override {
```
Also missing for all other instantiations of this function.

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctor.h
```diff
@@ -684,6 +737,100 @@ class LJFunctor
...
+
+      return numDistCallsAcc * numFLOPsPerDistanceCall + numKernelCallsN3Acc * numFLOPsPerN3KernelCall +
+             numKernelCallsNoN3Acc * numFLOPsPerNoN3KernelCall + numGlobalCalcsAcc * numFLOPsPerGlobalCalc;
+    } else {
+      autopas::utils::ExceptionHandler::exception("LJFunctor::getNumFLOPs called without countFLOPs enabled!");
+      return 0;
+    }
+  }
+
+  double getHitRate() {
```
FG-TUM, line 812:
```suggestion
  double getHitRate() override {
```
Also missing for all other instantiations of this function.

src/autopas/pairwiseFunctors/Functor.h
```diff
@@ -223,6 +224,18 @@ class Functor {
    */
   double getCutoff() const { return _cutoff; }
 
+  /**
+   * Get the number of FLOPs. Implementation required if FLOP logger used.
+   * @return number of FLOPs
+   */
+  virtual size_t getNumFLOPs() { return 0; }
```
FG-TUM, line 231:
```suggestion
  virtual size_t getNumFLOPs() const { return 0; }
```
We should be able to make this and `getHitRate()` both a `const` function, right?

applicationLibrary/molecularDynamics/tests/singleSiteTests/ForceCalculationTest.cpp
```diff
@@ -11,10 +11,12 @@
 #include "molecularDynamicsLibrary/LJFunctor.h"
 #include "testingHelpers/commonTypedefs.h"
 
+using forceCalculationTestLJFunctor =
+    LJFunctorType</* shifting */ false, /*mixing*/ false, autopas::FunctorN3Modes::Both, /*globals*/ false,
+                  /*relevantForTuning*/ true>;
```
FG-TUM, line 16:
```suggestion
using ForceCalculationTestLJFunctor =
    LJFunctorType</* shifting */ false, /*mixing*/ false, autopas::FunctorN3Modes::Both, /*globals*/ false,
                  /*relevantForTuning*/ true>;
```
Type -> Capital letter

Name is a bit unwieldy but ok...

applicationLibrary/molecularDynamics/tests/singleSiteTests/LJFunctorFlopCounterTest.cpp
```diff
@@ -0,0 +1,355 @@
...
+#include "molecularDynamicsLibrary/LJFunctor.h"
+#include "testingHelpers/commonTypedefs.h"
+
+extern template class autopas::AutoPas<Molecule>;
+template bool autopas::AutoPas<Molecule>::iteratePairwise(
+    mdLib::LJFunctor<Molecule, false, false, autopas::FunctorN3Modes::Both, false, true, true> *);
+template bool autopas::AutoPas<Molecule>::iteratePairwise(
+    mdLib::LJFunctor<Molecule, false, false, autopas::FunctorN3Modes::Both, true, true, true> *);
+template bool autopas::AutoPas<Molecule>::iteratePairwise(
+    mdLib::LJFunctor<Molecule, true, false, autopas::FunctorN3Modes::Both, true, true, true> *);
```
FG-TUM, line 22:
```suggestion
// AutoPas is instantiated in AutoPasInstantiations.cpp
// but iteratePairwise() versions with countFLOPs == true not,
// so they have to be explicitly instantiated here.
extern template class autopas::AutoPas<Molecule>;
template bool autopas::AutoPas<Molecule>::iteratePairwise(
    mdLib::LJFunctor<Molecule, false, false, autopas::FunctorN3Modes::Both, false, /*countFLOPs*/ true, true> *);
template bool autopas::AutoPas<Molecule>::iteratePairwise(
    mdLib::LJFunctor<Molecule, false, false, autopas::FunctorN3Modes::Both, true, /*countFLOPs*/ true, true> *);
template bool autopas::AutoPas<Molecule>::iteratePairwise(
    mdLib::LJFunctor<Molecule, true, false, autopas::FunctorN3Modes::Both, true, /*countFLOPs*/ true, true> *);
```

applicationLibrary/molecularDynamics/tests/templateInstatiations/AutoPasInstantiations.cpp
```diff
@@ -13,5 +13,5 @@
 template class autopas::AutoPas<Molecule>;
 template bool autopas::AutoPas<Molecule>::iteratePairwise(
     mdLib::LJFunctor<Molecule, /* shifting */ false, /*mixing*/ false, autopas::FunctorN3Modes::Both,
-                     /*globals*/ false, /*relevantForTuning*/ true> *);
+                     /*globals*/ false, /*countFLOPs*/ false, /*relevantForTuning*/ true> *);
```
FG-TUM, line 16:
```suggestion
// clang-format off
template bool autopas::AutoPas<Molecule>::iteratePairwise(
    mdLib::LJFunctor<Molecule,
                     /* shifting */           false,
                     /* mixing */             false,
                     autopas::FunctorN3Modes::Both,
                     /* globals */            false,
                     /* countFLOPs */         false,
                     /* relevantForTuning */  true> *);
// clang-format on
```
Just some overdue reformatting.

applicationLibrary/sph/SPHLibrary/SPHCalcDensityFunctor.h
```diff
@@ -70,6 +70,16 @@ class SPHCalcDensityFunctor : public autopas::Functor<Particle, SPHCalcDensityFu
     return flops;
   }
 
+  size_t getNumFLOPs() {
+    AutoPasLog(DEBUG, "SPHCalcDensityFunctor::getNumFLOPs called but is not implemented and will return 0.");
+    return 0;
+  }
```
FG-TUM, line 76: Why not move these Debug log warnings to the base class?

FG-TUM (changes requested):

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorAVX.h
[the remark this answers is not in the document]
thesamriel: Sorry to be so annoying about the warnings. 
But my default setup when I'm not interested in flops would use the LJ-AVX functor with `AUTOPAS_LOG_FLOPS=OFF` and log-level=warn.
This leads to a completely full terminal output :(

I'm also not so sure how to handle this if you really want a runtime warning within the functor...
Maybe the return value could be `std::numeric_limits<double>::quiet_NaN()`, to make it very clear in the output that these are no correct values?

thesamriel (changes requested):

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctor.h
```diff
@@ -710,6 +857,11 @@ class LJFunctor
     SoAFloatPrecision virialSumY = 0.;
     SoAFloatPrecision virialSumZ = 0.;
 
+    size_t numDistanceCalculationSum = 0;
+    size_t numKernelCallsN3Sum = 0;
+    size_t numKernelCallsNoN3Sum = 0;
+    size_t numGlobalCalcsSum = 0;
```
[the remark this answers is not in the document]
SamNewcome: Done in `7fa81126`

applicationLibrary/molecularDynamics/tests/templateInstatiations/AutoPasInstantiations.cpp
SamNewcome: Done in 7fa81126

applicationLibrary/molecularDynamics/tests/singleSiteTests/ForceCalculationTest.cpp
SamNewcome: Done in 7fa81126

applicationLibrary/molecularDynamics/tests/singleSiteTests/LJFunctorFlopCounterTest.cpp
SamNewcome: Done in 7fa81126

src/autopas/pairwiseFunctors/Functor.h
SamNewcome: Done in 7fa81126

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctor.h
SamNewcome: Done in 7fa81126

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctor.h
```diff
@@ -684,6 +737,100 @@ class LJFunctor
...
+  }
+
+  double getHitRate() {
+    if constexpr (countFLOPs) {
+      size_t numDistCallsAcc = std::accumulate(_aosThreadDataFLOPs.begin(), _aosThreadDataFLOPs.end(), 0ul,
+                                               [](size_t sum, const auto &data) { return sum + data.numDistCalls; });
+      size_t numKernelCallsN3Acc =
+          std::accumulate(_aosThreadDataFLOPs.begin(), _aosThreadDataFLOPs.end(), 0ul,
+                          [](size_t sum, const auto &data) { return sum + data.numKernelCallsN3; });
+      size_t numKernelCallsNoN3Acc =
```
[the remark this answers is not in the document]
SamNewcome: Done in 7fa81126

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctor.h
SamNewcome: Done in 7fa81126

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctor.h
```diff
@@ -684,6 +737,100 @@ class LJFunctor
...
+          std::accumulate(_aosThreadDataFLOPs.begin(), _aosThreadDataFLOPs.end(), 0ul,
+                          [](size_t sum, const auto &data) { return sum + data.numKernelCallsNoN3; });
+      size_t numGlobalCalcsAcc =
```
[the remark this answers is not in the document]
SamNewcome: Done in 7fa81126

examples/md-flexible/src/Simulation.cpp
```diff
@@ -145,6 +142,27 @@ Simulation::Simulation(const MDFlexConfig &configuration,
...
+#ifdef AUTOPAS_LOG_FLOPS
+  {
+    // Throw warnings if user is using a functor which doesn't support FLOP counting
+    const std::vector<MDFlexConfig::FunctorOption> functorsThatDontSupportFLOPCounting =
+#if MD_FLEXIBLE_MODE == SINGLESITE
+        {MDFlexConfig::FunctorOption::lj12_6_AVX, MDFlexConfig::FunctorOption::lj12_6_SVE};
+#else
+        {MDFlexConfig::FunctorOption::lj12_6, MDFlexConfig::FunctorOption::lj12_6_Globals,
+         MDFlexConfig::FunctorOption::lj12_6_AVX, MDFlexConfig::FunctorOption::lj12_6_SVE};
+#endif
```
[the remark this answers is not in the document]
SamNewcome: Done in 7fa81126

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctor.h
```diff
@@ -684,6 +737,100 @@ class LJFunctor
...
+      size_t numKernelCallsN3Acc =
+          std::accumulate(_aosThreadDataFLOPs.begin(), _aosThreadDataFLOPs.end(), 0ul,
+                          [](size_t sum, const auto &data) { return sum + data.numKernelCallsN3; });
+      size_t numKernelCallsNoN3Acc =
+          std::accumulate(_aosThreadDataFLOPs.begin(), _aosThreadDataFLOPs.end(), 0ul,
+                          [](size_t sum, const auto &data) { return sum + data.numKernelCallsNoN3; });
+
+      AutoPasLog(TRACE, "Number of Distance Calls           = {}", numDistCallsAcc);
+      AutoPasLog(TRACE, "Number of Newton3 Kernel Calls     = {}", numKernelCallsN3Acc);
+      AutoPasLog(TRACE, "Number of Non-Newton3 Kernel Calls = {}", numKernelCallsNoN3Acc);
```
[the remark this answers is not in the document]
FG-TUM: I'm slightly in favor of removing these trace calls. If someone is really interested in them `trace` is too verbose to be useful in a "big" simulation, so they would have to implement a dedicated logger or something like that anyways.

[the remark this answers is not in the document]
SamNewcome: My reasoning for the `TRACE` is that when debugging, it would be useful to see the "micro-metrics" that lead to the FLOP count + hit rate. In such a case, it would be worth logging these in both functions so that we such "micro-metrics" are available in the future if trying to debug either function. However, as both functions are always currently used together, this is not really needed. This is never intended to be an actual logger, and if this is wanted, a new logger should be created as you suggest.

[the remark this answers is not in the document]
FG-TUM: But if I were to debug this, wouldn't I use... the debugger? ;) Especially because first, I'd probably have to recompile the whole project to enable trace logging, so the debugger might be even quicker.
I (think I) get why you say these might be useful. However, I'm not a big fan of keeping things in the code that only "might be" useful if we don't have a clear use case. At the same time this is not a strong opinion so if you prefer it this way keep them.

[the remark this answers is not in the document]
SamNewcome: Fair point. Removed in 9321adb5

applicationLibrary/sph/SPHLibrary/SPHCalcDensityFunctor.h
SamNewcome: Not relevant after c31aa16c

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorAVX.h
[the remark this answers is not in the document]
SamNewcome: Hopefully c31aa16c solves this issue for you. I will post about it in more detail as a "main comment"

SamNewcome:
To summarise a recent change to this branch:

- @thesamriel had an issue with constant log spam due to FLOP counting metrics not being implemented in the case that `AUTOPAS_LOG_FLOPS=ON`. 
- To avoid this issue, as well as the maintainability of FLOP logging, the handling of Functors without FLOP counting implemented is moved into `FLOPLogger` itself.
- The default implementation returns negative FLOPs and hit rate.
- When the `FLOPLogger` receives negative FLOPs or hit rate, it outputs blank fields in the CSV.
- If this happens at least once, `FLOPLogger` outputs an `INFO` level log upon destruction e.g. at the end of the simulation.
- The md-flexible functors that don't implement these functions still output a `DEBUG` level log upon construction if `countFLOPs` is enabled.

src/autopas/utils/logging/FLOPLogger.cpp
```diff
@@ -0,0 +1,64 @@
...
+  spdlog::drop(_loggerName);
+#endif
+}
+
+void autopas::FLOPLogger::logIteration(size_t iteration, int numFLOPs, double hitRate) {
+#ifdef AUTOPAS_LOG_FLOPS
+  // Convert negative numbers to empty strings to represent no implementation
+  const auto numFLOPsStr = numFLOPs >= 0 ? std::to_string(numFLOPs) : "";
+  const auto hitRateStr = hitRate >= 0 ? std::to_string(hitRate) : "";
+  spdlog::get(_loggerName)->info("{},{},{}", iteration, numFLOPsStr, hitRateStr);
```
FG-TUM, line 58: What's the reason you are logging anything at all in that case and not just skip the call to the logger?

Something like:
```suggestion
  const auto numFLOPsStr = numFLOPs >= 0 ? std::to_string(numFLOPs) : "";
  const auto hitRateStr = hitRate >= 0 ? std::to_string(hitRate) : "";
  // Only write if at least one value is not empty
  if (not numFLOPsStr.empty() or not hitRateStr.empty())) {
    spdlog::get(_loggerName)->info("{},{},{}", iteration, numFLOPsStr, hitRateStr);
  }
```

src/autopas/utils/logging/FLOPLogger.cpp
```diff
@@ -0,0 +1,64 @@
...
+  auto logger = spdlog::basic_logger_mt<spdlog::async_factory>(_loggerName, outputFileName);
+  // set pattern to provide date
+  logger->set_pattern("%Y-%m-%d %T,%v");
+#endif
+}
+
+autopas::FLOPLogger::~FLOPLogger() {
+#ifdef AUTOPAS_LOG_FLOPS
+  if (_loggedEmptyFields) {
+    AutoPasLog(INFO,
```
FG-TUM, line 45: 1. I think this should be a warning. It's not spammy and if the user is annoyed by it they should disable the feature they did not implement because that is improper utilization.
2. Maybe also write this message the first time this happens otherwise a user will only read this at the end of their simulation.
3. I'm unsure if I would like this message only at the start, only the end, or both. Pro/Cons:
   - Start: Can be seen quickly after starting the simulation leading to better reaction times but easy to miss.
   - End: Harder to miss but very late.
   - Both: Best of both worlds but duplicated warning.

FG-TUM (changes requested):

src/autopas/pairwiseFunctors/Functor.h
```diff
@@ -223,6 +224,25 @@ class Functor {
...
   double getCutoff() const { return _cutoff; }
 
+  /**
+   * Get the number of FLOPs. Implementation required if FLOPLogger used.
+   *
+   * If derived class provides no implementation, the FLOPLogger interprets the default negative output as invalid and
+   * leaves a blank space in the log.
+   * @return number of FLOPs
+   */
+  [[nodiscard]] virtual int getNumFLOPs() const { return -1; }
```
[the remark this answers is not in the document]
SamNewcome: Could use `std::numeric_limits<size_t>::max()`. I think `0` is too potentially problematic.

src/autopas/utils/logging/FLOPLogger.cpp
SamNewcome: I don't want to skip the logger because it makes it clear that the functor was still called but that no metrics were provided, as opposed to that it wasn't called. Furthermore, consider the scenario that two functors are used, one of which has one function implemented only and the other has none (niche scenario, but illustrates wider point). One functor results in a log with one empty field, implying that empty fields are outputted, and the other doesn't log anything.

Ultimately, I don't think the scenario where the user has `AUTOPAS_LOG_FLOPS=ON` but is using a functor which has no implementation is that important - I just want something which works and is not going to leave the user wondering why the FLOP log makes no sense,

FG-TUM: > consider the scenario that two functors are used,

For that case you should probably also write the functor name in every line or create one log file per functor.

src/autopas/utils/logging/FLOPLogger.cpp
SamNewcome: You're a bit too premature with this comment haha - it wasn't ready for review and this doesn't really apply anymore.

FYI, this didn't actually work, I think, because the AutoPas logger gets deregistered before this object is destructed. So I opted with a much more heavy-handed approach of just logging "Not Implemented".

The Start doesn't work because we would need to force the user simulator to tell the logger which functors it will use (unless I am missing something). This could be problematic.

Option 2 could work, but might be problematic if this is in the depths of the log file and not that noticeable. As a `WARN` this might not be too bad.

I will post a comment summarising this further change.

SamNewcome:
Just to summarise the even more recent change:

FLOPLogger now logs "Not Implemented" in the CSV field if a functor does not implement `getNumFLOPs` or `getHitRate`
This is heavy-handed but
- It does not spam the regular AutoPas log at all
- If the user does not really care about FLOP logging then they aren't going to read the file
- If the user does care, they will see the problem and can do something about that.

src/autopas/utils/logging/FLOPLogger.cpp
SamNewcome: Yes, I mean this is something we should have anyway for the iteration logger (I think the 3 body branch had this? Not sure if it still does). But I would argue that this should be a separate PR simply because its really a separate issue which transcends this particular PR. It is also a lower priority issue relative to this PR (which is required for other PRs). But by keeping the functionality as it is, we future-proof for this particular problem combined with the other reasons for keeping this functionality.

src/autopas/pairwiseFunctors/Functor.h
```diff
@@ -223,6 +224,25 @@ class Functor {
...
+
+  /**
+   * Get the hit rate. Implementation required if FLOPLogger used.
+   *
+   * If derived class provides no implementation, the FLOPLogger interprets the default negative output as invalid and
+   * leaves a blank space in the log.
+   *
+   * @return (number of kernel calls) / (number of distance calculations)
+   */
+  [[nodiscard]] virtual double getHitRate() const { return -1; }
```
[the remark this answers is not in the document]
SamNewcome: see d1e26036

src/autopas/pairwiseFunctors/Functor.h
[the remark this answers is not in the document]
SamNewcome: See d1e26036

src/autopas/utils/logging/FLOPLogger.cpp
FG-TUM: Fine with me

src/autopas/pairwiseFunctors/Functor.h
```diff
@@ -223,6 +224,25 @@ class Functor {
    */
   double getCutoff() const { return _cutoff; }
 
+  /**
+   * Get the number of FLOPs. Implementation required if FLOPLogger used.
+   *
+   * If derived class provides no implementation, the FLOPLogger interprets the default numeric_limits<size_t>::max()
+   * output as invalid and leaves a blank space in the log.
```
FG-TUM, line 231: This is now incorrect, right?

src/autopas/pairwiseFunctors/Functor.h
```diff
@@ -223,6 +224,25 @@ class Functor {
...
+   * @return number of FLOPs
+   */
+  [[nodiscard]] virtual size_t getNumFLOPs() const { return std::numeric_limits<size_t>::max(); }
+
+  /**
+   * Get the hit rate. Implementation required if FLOPLogger used.
+   *
+   * If derived class provides no implementation, the FLOPLogger interprets the default NaN output as invalid and
+   * leaves a blank space in the log.
```
FG-TUM, line 240: same

FG-TUM (changes requested):
As far as I understand only some doc update is needed then it's good to merge for me.

src/autopas/pairwiseFunctors/Functor.h
SamNewcome: done in a1802f00

src/autopas/pairwiseFunctors/Functor.h
SamNewcome: done in a1802f00

thesamriel (approved):

FG-TUM (approved):

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctor.h
```diff
@@ -820,8 +820,8 @@ class LJFunctor : public autopas::Functor<Particle, LJFunctor<Particle, applyShi
       return (static_cast<double>(numKernelCallsNoN3Acc) + static_cast<double>(numKernelCallsN3Acc)) /
              (static_cast<double>(numDistCallsAcc));
     } else {
-      autopas::utils::ExceptionHandler::exception("LJFunctor::getHitRate called without countFLOPs enabled!");
-      return 0;
+      // This is needed because this function still gets called with FLOP logging disabled, just nothing is done with it
+      return std::numeric_limits<size_t>::quiet_NaN();
```
thesamriel, line 824: Is it on purpose that you chose `quiet_NaN()` here and `max()` in line 804?

SamNewcome: Yes. `quiet_NaN()` is more appropriate for an invalid number, but `size_t` returns 0 for this, which is not desired. (Actually, it is a mistake that this is `size_t` here, and is changed to `double` in a824ccee)

PR #851, opened 2024-04-25
Remove the Particle Properties Library from single-site LJ Functors

### LINKED ISSUE #815: Revisit performance of Particle Properties Library
By default, md-flexible is compiled with a mixing functionality, meaning we can have several types of particles in our simulation.
In the force functors, we retrieve the mixed coefficients from a ParticlePropertiesLibrary, essentially a look-up table.
E.g., for the Lennard-Jones functor we need `sigmaSquared` and `epsilon24`.

In our practical course MD-Lab, students reported getting worse results using such a look-up table, compared to directly calculating the mixed coefficients in the functor. Therefore, we should reevaluate if this might be true for md-flexible as well.

The mixing rules for the LJ parameters are:

$$ \sigma_{ij} = \frac{\sigma_i + \sigma_j}{2} $$

$$ \epsilon_{ij} = \sqrt{\epsilon_i \epsilon_j} $$

We probably decided for recalculated look-up tables to avoid the costly squareroot operation.
However, the additional memory access might turn out to be even more expensive. Results may vary for AoS and SoA.

**Additional idea:** Instead of storing only epsilons, we could store the square roots of epsilon to calculate $\epsilon_{ij} = \sqrt{\epsilon_i}\sqrt{ \epsilon_j}$

SamNewcome:
# Description

It has been shown that the Particle Properties Library slows down the Lennard Jones AVX512 force functor. This pull request reimplements all currently-in-master single-site functors to not use pre-computed values. It will be checked in a simulation that this is actually faster for all variants of LJ Functor on a variety of machines.

If there exists a sensible functor / machine combination where this is significantly slower, an appropriate solution will be devised to work around this before this PR is merged.

Minor Points:

- As Functor and Molecule are already so intertwined (and even more so with this PR), Molecule is removed from the LJFunctor's template parameters. This makes code development and maintenance easier as code insights can be made with the molecule class known.

## Related Pull Requests

- #PR #850 must be completed first.

## Resolved Issues

- [ ] fixes #815

# How Has This Been Tested?

- [ ] Existing tests for the functors are adapted and pass.
- [ ] mdLib tests compile and pass on ARM cluster
- [ ] md-flex compiles and runs on ARM cluster

Performance checked on a variety of machines:
- [ ] CoolMUC2 - AutoVec + AVX
- [ ] CoolMUC4 - AutoVec + Todo revisit AVX512 in another PR
- [ ] HSUper - AutoVec + Todo revisit AVX512 in another PR
- [ ] HSU Minicluster ARM machine - AutoVec + SVE


thesamriel:
@TumBeat might take care of this in a separate PR.

Issue #865, opened 2024-05-14
Make Tuning Library Dependencies Optional


SamNewcome:
**Is your feature request related to a problem? Please describe.**
Adding dependencies to AutoPas can lead to frustrating compatibility/compilation issues. We have 3/4 dependencies in AutoPas that are just used for tuning strategies, which you are not necessarily going to use:

- Active Harmony - Active Harmony tuning - as far as I am aware, nobody uses this.
- ANTLR - Rules-Based tuning
- uuid - Rules-Based Tuning (if not found on the machine) 
- Eigen - Bayesian Tuning Strategies

Furthermore, compiling these libraries can take time.

**Describe the solution you'd like**
CMake options for each library (or each tuning strategy), that disables compilation of the library. Also stops the relevant tuning strategy from being used.

**Describe alternatives you've considered**
Potentially Eigen can stay, my primary problem is with ANTLR/uuid.

I don't think anybody uses Active Harmony or intends to use it, it can alternatively be deleted completely.

**Additional context**
Frustrations arose from #864.

Issue #864, opened 2024-05-14
Add option to link against an external uuid build


SamNewcome:
**Describe the bug**
You can't compile AutoPas on the ARM minicluster. This does not appear related to it being an ARM architecture but that it does not have much software. But it is important that we can compile on the minicluster because it has ARM CPUs.

The issue appears to stem from needing to make `uuid` (a dependency of ANTLR) as it is not on the system. When compiling `uuid`, `automake-1.13` is not found causing the compilation to break. The system has `automake-1.16`.

Bizarrely, `uuid` can be compiled manually.

**To Reproduce**
Compile AutoPas on the ARM minicluster using `gcc-11`:

I get:

```bash
[ 35%] Performing build step for 'uuid_bundled'
gmake[4]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
 cd . && /bin/sh /home/sam/AutoPas/build/uuid/src/uuid_bundled/missing automake-1.13 --foreign
/home/sam/AutoPas/build/uuid/src/uuid_bundled/missing: line 81: automake-1.13: command not found
WARNING: 'automake-1.13' is missing on your system.
         You should only need it if you modified 'Makefile.am' or
         'configure.ac' or m4 files included by 'configure.ac'.
         The 'automake' program is part of the GNU Automake package:
         <http://www.gnu.org/software/automake>
         It also requires GNU Autoconf, GNU m4 and Perl in order to run:
         <http://www.gnu.org/software/autoconf>
         <http://www.gnu.org/software/m4/>
         <http://www.perl.org/>
gmake[4]: *** [Makefile:395: Makefile.in] Error 1
make[3]: *** [CMakeFiles/uuid_bundled.dir/build.make:86: uuid/src/uuid_bundled-stamp/uuid_bundled-build] Error 2
make[2]: *** [CMakeFiles/Makefile2:542: CMakeFiles/uuid_bundled.dir/all] Error 2
```

~~Potentially this could be reproducible on another machine without `uuid`.~~

**Expected behavior**
The program compiles.

**Build environment (please complete the following information):**
* OS: Linux
* Compiler: GCC 11.2.1
* Build System: CMake 3.26.3

**Potential Solutions**

1. Recheck that there is nothing incorrect being done in `autopas_antlr4cpp`. Potentially a fresh set of eyes might spot something.
2. Add an option to link against an external `libuuid`.

FG-TUM:
[...]

PR #866, opened 2024-05-16
Optional rules based tuning

SamNewcome:
# Description

Rules Based Tuning (and dependencies ANTLR and uuid) are optional.

This has two purposes:

1. Speed up compilation of fresh builds. ANTLR takes up a lot of compilation time. Whilst it only needs to be compiled, I still find it frustrating with fresh builds.
2. Compatability problems caused by compiling the bundled version of uuid. The bundled version of uuid is only used when it doesn't already exist on the machine. However, the make process specifically calls automake-1.13 which is old. 

If interested in fixing point 2: From research, this might be related to GitHub, maybe not saving timestamps, which causes an unnecessary recompile from CMake, which requires automake 1.13. Don't ask me the details - I did not want to fix this issue in this PR, so I did not verify in much detail - but the information is here to give somebody who might want to a first direction to look in.

This PR was intended to fix compilation on the ARM cluster as part of #864 and #865. In fact, it is really just a hack to get around #864 and make this a low priority. #865 shall be finished in a separate PR.

## Related Pull Requests

Needed to properly test the SVE functor for #850 and #851.

## Resolved Issues

- No issues resolved entirely.

# How Has This Been Tested?

- [x] Correct compilation with rules based tuning enabled and disabled
- [x] Specifically tested compilation on ARM cluster.

P.s. the ARM cluster compilation is still not confirmed fully fixed due to a separate unrelated issue.


src/autopas/CMakeLists.txt
```diff
@@ -31,6 +51,11 @@ target_link_libraries(
         harmony
 )
 
+# If rules-based tuning is not disabled, also link antlr4cpp
+if (AUTOPAS_ENABLE_RULES_BASED_TUNING)
+    target_link_libraries(autopas PUBLIC antlr4cpp)
+endif()
```
FG-TUM, line 57: Remove these lines and put it in the `target_link_libraries` statement from above with generator statements (for consistency)

Should be something like this (can't make a suggestion because lines are too far apart):
```cmake
        $<$<BOOL:${AUTOPAS_ENABLE_RULES_BASED_TUNING}>:antlr4cpp>
```

src/autopas/tuning/tuningStrategy/ruleBasedTuning/RuleBasedTuning.cpp
```diff
@@ -63,12 +71,52 @@ void RuleBasedTuning::reset(size_t iteration, size_t tuningPhase, std::vector<Co
...
+#endif
 }
 
 long RuleBasedTuning::getLifetimeWouldHaveSkippedTuningTime() const { return _wouldHaveSkippedTuningTimeLifetime; }
 
 long RuleBasedTuning::getLifetimeTuningTime() const { return _tuningTimeLifetime; }
 
+void RuleBasedTuning::optimizeSuggestions(std::vector<Configuration> &configQueue,
+                                          const EvidenceCollection &evidenceCollection) {
+#ifdef AUTOPAS_ENABLE_RULES_BASED_TUNING
```
FG-TUM, line 83: any reason you changed the order of the functions?

src/autopas/tuning/tuningStrategy/ruleBasedTuning/RuleBasedTuning.h
```diff
@@ -72,16 +75,26 @@ namespace autopas {
  *
  * Additionally, the class supports a so called "verify mode" where full search is performed and the given rules are
  * checked for correctness.
+ *
+ *
+ * Due to the compilation cost of ANTLR and issues with compiling the bundled dependency uuid on some machines, this
+ * tuning strategy can be disabled with AUTOPAS_ENABLE_RULES_BASED_TUNING=OFF.
```
FG-TUM, line 81:
```suggestion
* tuning strategy can be disabled with the CMake option AUTOPAS_ENABLE_RULES_BASED_TUNING=OFF.
```

src/autopas/tuning/tuningStrategy/ruleBasedTuning/RuleBasedTuning.h
```diff
@@ -72,16 +75,26 @@ namespace autopas {
...
+ *
  */
 class RuleBasedTuning : public TuningStrategyInterface {
  public:
   /**
    * A function type used to print errors found in verify mode.
    */
+
```
FG-TUM, line 89:
```suggestion
```

tools/ruleChecker/CMakeLists.txt
FG-TUM, line 1: I'm a bit confused how this actually works. If this file knows the `autopas` target shouldn't it then also already know about the `AUTOPAS_ENABLE_RULES_BASED_TUNING` option? If not maybe get the option by including our CMake module? Not sure whats the best course of action here but redefining the option feels wrond

FG-TUM (changes requested):
I trust you have checked that it actually compiles when you disable Rule Based Tuning? ;)

cmake/modules/autopas_antlr4cpp.cmake
```diff
@@ -1,101 +1,119 @@
...
+    find_package(PkgConfig)
+    pkg_check_modules(UUID QUIET uuid)
+
+    # if uuid-dev was not found on system we install it locally
+    if (NOT UUID_FOUND)
+        message(STATUS "UUID not found - using bundled version")
+        message(WARNING "Compilation of the bundled UUID is known to fail. If it does either:
+                        1) Install UUID on your machine yourself.
+                        2) Disable rules-based tuning via the CMake option -DAUTOPAS_ENABLE_RULES_BASED_TUNING=OFF.
+                        3) (Untested!) Install automake 1.13.")
```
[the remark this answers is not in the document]
SamNewcome: I agree, but - if I am correct in my understanding of the `uuid` problem - the bundled `uuid` will always fail to compile unless the machine has automake 1.13 (which is usually not the case). 

As such, this is less of a blanket warning and more of a "foretelling" of what will happen. Hence, I would see this less as blanket spam for 95% of cases and more as useful information except for 5% of cases. (Numbers are arbitrary but you get the point).

This can save the user time in debugging the error and eliminate doubt that the error was caused by them making a mistake. This is especially important for application scientists with less experience. 

1 is obvious, given that the user already understands that the bundled version is problematic. But I see no reason not to spell it out for novice HPC users and application scientists, given the two points above.

2 is not obvious. You cannot expect novice users of AutoPas to know every CMake option and what dependencies they stop from needing to be compiled when disabled, especially as the name of the option does not make it clear that this will stop `uuid` from being needed.

3 is not a random guess, as the error caused when compiling `uuid` on the ARM cluster is

~~~bash
WARNING: 'automake-1.13' is missing on your system.
~~~

which implies that installing it would fix the system. I just have not tested it because I didn't care about finding a solution to compile rules-based tuning on the ARM cluster (at least for the moment). I am happy to remove this if you still think it is not needed.

src/autopas/tuning/tuningStrategy/ruleBasedTuning/RuleBasedTuning.cpp
SamNewcome: Yes. To avoid having more `#ifdef`s than needed. So, all of the functions that are not part of the interface are completely wrapped in a single `#ifdef`. But this required moving a couple of functions about.

cmake/modules/autopas_antlr4cpp.cmake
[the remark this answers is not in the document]
SamNewcome: I agree. Fixed in `5044d8b5`. I added some information to `Building.md` to provide a little information about the matter.

src/autopas/CMakeLists.txt
```diff
@@ -11,6 +11,27 @@ file(
     "*.cu"
     "*.cuh"
 )
+set(AUTOPAS_ENABLE_RULES_BASED_TUNING
+        ON
+        CACHE
+        BOOL "Enables rules-based tuning, which, if using the bundled version, will compile ANTLR and, if uuid is not installed already, will compile uuid which can cause errors."
+        )
```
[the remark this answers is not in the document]
SamNewcome: Removed in `5044d8b5`

src/autopas/CMakeLists.txt
SamNewcome: Done in `5044d8b5`

tools/ruleChecker/CMakeLists.txt
SamNewcome: This was a mistake. Removed in `5044d8b5`

docs/userdoc/Building.md
```diff
@@ -5,10 +5,10 @@
 * a CMake generator target (`make` is tested)
 * a C++17 compiler (gcc11, clang13, and ~~icpc 2019~~ are tested)
 * OpenMP (comes with GCC, for Clang you need `libomp`)
-* For rule based tuning: `pkg-config`
 
 Optional:
 * For `tuningLogToSQL`: `libsqlite3`
+* For rule based tuning: `pkg-config`. Rule based tuning can be disabled via the CMake (see below).
```
thesamriel, line 11:
```suggestion
* For rule based tuning: `pkg-config`. Rule based tuning can be disabled via the CMake (see [below](#disabling-rule-based-tuning)).
```

as it is not directly below

cmake/modules/autopas_antlr4cpp.cmake
```diff
@@ -1,101 +1,113 @@
...
-        PREFIX ${CMAKE_CURRENT_BINARY_DIR}/uuid
-        URL ${PROJECT_SOURCE_DIR}/libs/libuuid-1.0.3.zip
-        URL_HASH MD5=9fd1e87682d24d6ca22941e0af339c8a
-        BUILD_IN_SOURCE TRUE
-        INSTALL_DIR "install"
-        CONFIGURE_COMMAND "./configure" ${UUID_CONFIG_PARAMS}
-        BUILD_COMMAND ${MAKE_EXE}
-    )
+set(AUTOPAS_ENABLE_RULES_BASED_TUNING
+        ON
```
thesamriel, line 2: Personally, I would prefer the default to be off.
My reasoning would be that rules based tuning does not really work out of the box anyway. The user has to generate or write a rule file first. Someone who goes through this effort can then also easily turn on this option.
On the other hand, most new users won't need this and when it is OFF by default they just don't have to worry about it.

src/autopas/CMakeLists.txt
```diff
@@ -12,6 +12,22 @@ file(
...
+    list(REMOVE_ITEM MY_SRC "${AUTOPAS_SOURCE_DIR}/src/autopas/tuning/tuningStrategy/ruleBasedTuning/parser_generated/autopas_generated_rule_syntax/RuleLanguageVisitor.cpp")
+    list(REMOVE_ITEM MY_SRC "${AUTOPAS_SOURCE_DIR}/src/autopas/tuning/tuningStrategy/ruleBasedTuning/parser_generated/autopas_generated_rule_syntax/RuleLanguageLexer.h")
+    list(REMOVE_ITEM MY_SRC "${AUTOPAS_SOURCE_DIR}/src/autopas/tuning/tuningStrategy/ruleBasedTuning/parser_generated/autopas_generated_rule_syntax/RuleLanguageLexer.cpp")
+    list(REMOVE_ITEM MY_SRC "${AUTOPAS_SOURCE_DIR}/src/autopas/tuning/tuningStrategy/ruleBasedTuning/parser_generated/autopas_generated_rule_syntax/RuleLanguageParser.h")
+    list(REMOVE_ITEM MY_SRC "${AUTOPAS_SOURCE_DIR}/src/autopas/tuning/tuningStrategy/ruleBasedTuning/parser_generated/autopas_generated_rule_syntax/RuleLanguageParser.cpp")
+    list(REMOVE_ITEM MY_SRC "${AUTOPAS_SOURCE_DIR}/src/autopas/tuning/tuningStrategy/ruleBasedTuning/RuleBasedProgramParser.h")
+    list(REMOVE_ITEM MY_SRC "${AUTOPAS_SOURCE_DIR}/src/autopas/tuning/tuningStrategy/ruleBasedTuning/RuleBasedProgramParser.cpp")
+endif ()
+
+
```
thesamriel, line 30: If I may suggest a slightly different approach, I would exclude the whole rules base tuning directory:

```suggestion
if (NOT AUTOPAS_ENABLE_RULES_BASED_TUNING)
    file(GLOB_RECURSE RULES_FILES "${AUTOPAS_SOURCE_DIR}/src/autopas/tuning/tuningStrategy/ruleBasedTuning/*.cpp"
            "${AUTOPAS_SOURCE_DIR}/src/autopas/tuning/tuningStrategy/ruleBasedTuning/*.h")
    # Loop over each file in RULES_FILES and remove it from MY_SRC
    foreach(FILE ${RULES_FILES})
        list(REMOVE_ITEM MY_SRC ${FILE})
    endforeach()
endif ()
```

Then you would not need all the `#ifdef` in `RuleBasedTuning.h/.cpp`, but only one `#ifdef` in `TuningStrategyFactory.cpp`, sth like:
```cpp
    case TuningStrategyOption::ruleBasedTuning: {
#ifdef AUTOPAS_ENABLE_RULES_BASED_TUNING
      tuningStrategy = std::make_unique<RuleBasedTuning>(searchSpace, /*verify mode*/ false, info.ruleFileName);
#else
utils::ExceptionHandler::exception(
    "Cannot use the TuningStrategy ruleBasedTuning without AUTOPAS_ENABLE_RULES_BASED_TUNING=OFF.");
#endif
      break;
    }
```

SamNewcome: I don't have a strong opinion either way. The advantage of not excluding `RuleBasedTuning.h/.cpp` is that the interface still exists. I vaguely remember @FG-TUM suggesting not doing this - I would be interested in his view on this.

docs/userdoc/Building.md
SamNewcome: This does not work with doxygen - my local doxygen (1.10.0) just does not include the link and the CI's doxygen build causes the CI to fail. Do you know that this should work with doxygen or do you know how to fix this?

SamNewcome: Okay issue fixed in `0529d104`. I had to use a different way of linking

src/autopas/CMakeLists.txt
FG-TUM: I'm a bit confused. I thought I already wrote something somewhere about this section and topic...

1. I'd handle the CMake code way simpler in a way that is more resistant to project structure changes:
```suggestion
if (NOT AUTOPAS_ENABLE_RULES_BASED_TUNING)
    list(FILTER MY_SRC EXCLUDE REGEX ".*parser-generated.*")
endif ()
```

2. I prefer the `ifdef` in the `RuleBasedTuning` class for several reasons:
  - You don't have to consider every include wherever in the project it happens or will happen. When you write any code you should not have to think about "do I have to treat this include special?". It should resolve this by itself. This might also affect the tests.
  - The `ifdef` is where it actually should affect the code.
  - Consistency with how we already do it for things like e.g. specialized loggers.

src/autopas/tuning/tuningStrategy/ruleBasedTuning/RuleBasedTuning.h
```diff
@@ -145,10 +158,14 @@ class RuleBasedTuning : public TuningStrategyInterface {
    */
   std::vector<rule_syntax::ConfigurationOrder> applyRules(const std::vector<Configuration> &searchSpace);
 
+  std::vector<rule_syntax::ConfigurationOrder> _lastApplicableConfigurationOrders;
```
FG-TUM, line 161:
```suggestion
  std::vector<rule_syntax::ConfigurationOrder> _lastApplicableConfigurationOrders{};
```
missing initialization
- [ ] same for `_traversalTimes` if I see this correctly

FG-TUM (changes requested):

cmake/modules/autopas_antlr4cpp.cmake
```diff
@@ -1,101 +1,113 @@
...
-        BUILD_IN_SOURCE TRUE
-        INSTALL_DIR "install"
-        CONFIGURE_COMMAND "./configure" ${UUID_CONFIG_PARAMS}
-        BUILD_COMMAND ${MAKE_EXE}
-    )
+set(AUTOPAS_ENABLE_RULES_BASED_TUNING
+        OFF
+        CACHE
+        BOOL "Enables rules-based tuning, which, if using the bundled version, will compile ANTLR and, if it is not installed uuid."
+        )
```
FG-TUM, line 5:
```suggestion
set(AUTOPAS_ENABLE_RULES_BASED_TUNING
        # Default is OFF just for faster default compilation time.
        OFF
        CACHE
        BOOL "Enables rules-based tuning, which, if using the bundled version, will compile ANTLR and, if it is not installed uuid."
        )
```

docs/userdoc/Building.md
```diff
@@ -37,6 +38,19 @@ Here is an example of a parallel compilation of the md-flexible example:
 cmake --build . --target md-flexible --parallel 12
 ```
 
+<a id="rule-based-tuning"> </a>
+
+### Enabling Rule Based Tuning
```
FG-TUM, line 43:
```suggestion
### Enabling Rule Based Tuning
<a id="rule-based-tuning"> </a>

```
I think the anchor should be after the heading.

FG-TUM (approved):
Just two minor things. Apart from that I'm fine with merging

PR #876, opened 2024-05-31
LiveInfoLogger and allow for Pausing the simulation during Tuning Phases

ManuelLerchner:
# Description

This pull request introduces two new features designed to allow accurate logging of the simulation state in `md-flexible` simulations:

1. **LiveInfoLogger**

    Currently RuleBasedTuning makes use of a `LiveInfoStruct` containing a summarized representation of the current 
    simulation state based on some numerical features. At the moment, there is no way to export this information so this 
    pull request adds a new Logger class to dump all the collected data into a `.csv` file.
    + The logger can be enabled with the  `AUTOPAS_LOG_LIVEINFO` cmake option.

2. **PauseSimulationDuringTuning**

   Currently, AutoPas doesn't pause the simulation during tuning phases. This is intended as every tested 
   configuration still contributes to the goal of finishing the simulation but comes at the cost of not having fair 
   comparisons between all tested configurations as the simulation state can change during the tuning phase.

   Since AutoPas just performs the force calculations between the particles, it is left to the specific implementation to 
   address this issue. One workaround is to just disable the position and velocity updates during such tuning phases and to 
  correctly reset the force calculations after the tuning is over. This Pull Request introduces this exact method for `md-
  flexbile`
   + The pausing can be enabled with the new `pause-simulation-during-tuning : true` setting in the input yaml files.

## Related Pull Requests

- 

## Resolved Issues

- 

# How Has This Been Tested?

+ The Log files for the LiveInfo gets created and contains all relevant data. So the Logger appears to work as expected.
+ I haven't tested the Pausing fully yet, but it seems to work as intended and doesn't interfere with the 
rest of the simulation.


src/autopas/utils/SimilarityFunctions.h
```diff
@@ -65,6 +65,12 @@ std::pair<double, double> calculateHomogeneityAndMaxDensity(const Container &con
     for (size_t i = 0; i < particleLocation.size(); i++) {
       index[i] = particleLocation[i] / cellLength;
     }
+
+    // ignore particles that are outside the domain
```
ManuelLerchner, line 69: Something like this is needed, as otherwise it is possible to access memory outside the `particlesPerCell` array. Wich causes the sph-mpi tests to fail.

Im not sure if just ignoring such particles is the right answer as it could mess with the calculated density/homogenity. 

But such cases shouldn't even happen in the first place and maybe there is something wrong in the way the index is calculated, but at least this fixes the memory access issues.

src/autopas/utils/SimilarityFunctions.h
```diff
@@ -65,6 +65,12 @@ std::pair<double, double> calculateHomogeneityAndMaxDensity(const Container &con
...
+    if (index[0] >= cellsPerDimension[0] || index[1] >= cellsPerDimension[1] || index[2] >= cellsPerDimension[2]) {
+      continue;
+    }
```
ManuelLerchner, line 72: Something like this is needed, as otherwise it is possible to access memory outside the particlesPerCell array. Wich causes the sph-mpi tests to fail.

Im not sure if just ignoring such particles is the right answer as it could mess with the calculated density/homogenity.

But such cases shouldn't even happen in the first place and maybe there is something wrong in the way the index is calculated, but at least this fixes the memory access issues.

SamNewcome: Having now read through this function, there seem to be a lot of logic mistakes going on which potentially is why you get this error. I will work on a fix for this which you can merge into here.

SamNewcome (changes requested):

SamNewcome (approved):

SamNewcome:
All looks good but wait until #895 gets merged

PR #877, opened 2024-06-04
Highway release

TumBeat:
# Description

This PR includes the first version of an implementation of the LJ-Functor using Google Highway for Vectorization.
It has to be noted that this branch already includes different vectorization patterns. Unit tests verify their correctness, however, they are currently not very high-performing.

# How Has This Been Tested?

- [x] Implementation of Test Cases oriented on Test Cases for the other functors
- [x] Performed benchmarks on AutoPasFunctorBench to check performance


FG-TUM:
Seeing all these new LJFunctors it could make sense to get at common `LJFunctor` base class and rename the current `LJFunctor` class to something like `LJFunctorAutoVec`

PR #884, opened 2024-06-07
Fix calculate homogeneity and max density

### LINKED ISSUE #881: `calculateHomogeneityAndMaxDensity` is incorrect.
**Describe the bug**
There appear to be some mistakes in `autopas/utils/SimilarityFunctions.h`'s `calculateHomogeneityAndMaxDensity`.

1. It always uses the global number of particles. When we use this within md-flexible to get the global homogeneity and max density, this is totally correct. But when we use this internally with AutoPas, to compare how similar AutoPas containers are with MPI parallelized Tuning Strategy, we want only the local number of particles. In such as case, this is not only wrong, but also involves an inefficient `MPI_Allreduce` call. This should be handled with a simple flag to switch between the cases.
2. The sph mpi test run fails, when `calculateHomogeneityAndMaxDensity` is called (such as in the CI test of #876 ) https://github.com/AutoPas/AutoPas/actions/runs/9315720544/job/25691686602. According to @ManuelLerchner, the error occurs when an out-of-bounds index is calculated. This makes sense, because we calculate which cell a particle is in, we assume our cells/the container starts at (0,0,0) - which is especially not true for MPI runs.
3. The handling of the homogeneity in md-flexible is incorrect: The homogeneity is calculated once at the start of the simulation for the entire global simulation and then in outputted at the end of the simulation, even if this has changed.

There are also cases where the documentation should be improved:

1. Make it clear in the doxygen that the function can be used for either global calculations or local calculations.
2. Don't refer to "cells" but something else, to make it clear that this isn't talking about AutoPas cells.

SamNewcome:
# Description

Fixes `calculateHomogeneityAndMaxDensity` and adds a test for it.

Previously, `calculateHomogeneityAndMaxDensity` assumed that the domain started at (0,0,0) when indexing particles into bins. Furthermore, the algorithm was sub optimal and the code was messy.

This PR:

- [x] Fixes the non-(0,0,0) domain indexing.
- [x] Improves readability, including renaming from "cells" to "bins"
- [x] Switches from perfect cube bins, with the upper boundary handled by truncating bins (messy logic + truncated bins could cause greater statistical uncertainties), to dividing the domain into identical cuboids that perfectly fit into the domain.

## Related Pull Requests

- #PR #876 highlighted the issue

## Resolved Issues

- [x] fixes #881

# How Has This Been Tested?

- [x] Existing tests
- [x] New test for this function

### DIFF
--- a/examples/md-flexible/src/Simulation.cpp
+++ b/examples/md-flexible/src/Simulation.cpp
@@ -186,10 +186,6 @@ void Simulation::finalize() {
 }
 
 void Simulation::run() {
-  _homogeneity =
-      autopas::utils::calculateHomogeneityAndMaxDensity(*_autoPasContainer, _domainDecomposition->getGlobalBoxMin(),
-                                                        _domainDecomposition->getGlobalBoxMax())
-          .first;
   _timers.simulate.start();
   while (needsMoreIterations()) {
     if (_createVtkFiles and _iteration % _configuration.vtkWriteFrequency.value == 0) {
@@ -498,18 +494,11 @@ void Simulation::logSimulationState() {
   autopas::AutoPas_MPI_Allreduce(&particleCount, &haloParticles, 1, AUTOPAS_MPI_UNSIGNED_LONG, AUTOPAS_MPI_SUM,
                                  AUTOPAS_MPI_COMM_WORLD);
 
-  double squaredHomogeneity = _homogeneity * _homogeneity;
-  double standardDeviationOfHomogeneity{};
-  autopas::AutoPas_MPI_Allreduce(&squaredHomogeneity, &standardDeviationOfHomogeneity, 1, AUTOPAS_MPI_DOUBLE,
-                                 AUTOPAS_MPI_SUM, AUTOPAS_MPI_COMM_WORLD);
-  standardDeviationOfHomogeneity = std::sqrt(standardDeviationOfHomogeneity);
-
   if (_domainDecomposition->getDomainIndex() == 0) {
     std::cout << "\n\n"
               << "Total number of particles at the end of Simulation: " << totalNumberOfParticles << "\n"
               << "Owned: " << ownedParticles << "\n"
-              << "Halo : " << haloParticles << "\n"
-              << "Standard Deviation of Homogeneity: " << standardDeviationOfHomogeneity << std::endl;
+              << "Halo : " << haloParticles << std::endl;
   }
 }
 
--- a/src/autopas/utils/SimilarityFunctions.h
+++ b/src/autopas/utils/SimilarityFunctions.h
@@ -49,6 +49,7 @@ std::pair<double, double> calculateHomogeneityAndMaxDensity(const ParticleContai
   const auto domainVolume = domainDimensions[0] * domainDimensions[1] * domainDimensions[2];
 
   // We scale the dimensions of the domain to bins with volumes which give approximately 10 particles per bin.
+  // Todo The choice of 10 is arbitrary and probably can be optimized
   const auto targetNumberOfBins = std::ceil(numberOfParticles / 10.);
   const auto targetNumberOfBinsPerDim = std::cbrt(static_cast<double>(targetNumberOfBins));
   // This is probably not an integer, so we floor to get more than 10 particles per bin than too small bins
@@ -100,7 +101,8 @@ std::pair<double, double> calculateHomogeneityAndMaxDensity(const ParticleContai
   const double homogeneity = std::sqrt(densityVariance);
   // normally between 0.0 and 1.5
   if (homogeneity < 0.0) {
-    throw std::runtime_error("homogeneity can never be smaller than 0.0, but is:" + std::to_string(homogeneity));
+    utils::ExceptionHandler::exception(
+        "calculateHomogeneityAndMaxDensity(): homogeneity can never be smaller than 0.0, but is: {}", homogeneity);
   }
   return {homogeneity, maxDensity};
 }
--- a/src/autopas/utils/SimilarityFunctions.h
+++ b/src/autopas/utils/SimilarityFunctions.h
@@ -7,7 +7,6 @@
 #pragma once
 
 #include "ThreeDimensionalMapping.h"
-#include "autopas/AutoPasDecl.h"
 #include "autopas/containers/ParticleContainerInterface.h"
 #include "autopas/utils/ArrayMath.h"
 #include "autopas/utils/WrapMPI.h"
@@ -42,7 +41,7 @@ namespace autopas::utils {
  * @return {homogeneity, maxDensity}
  */
 template <typename Particle>
-std::pair<double, double> calculateHomogeneityAndMaxDensity(const AutoPas<Particle> &container) {
+std::pair<double, double> calculateHomogeneityAndMaxDensity(const ParticleContainerInterface<Particle> &container) {
   using namespace autopas::utils::ArrayMath::literals;
 
   const auto numberOfParticles = container.getNumberOfParticles();
--- a/src/autopas/utils/SimilarityFunctions.h
+++ b/src/autopas/utils/SimilarityFunctions.h
@@ -61,7 +61,7 @@ std::pair<double, double> calculateHomogeneityAndMaxDensity(const Container &con
   const auto binVolume = domainVolume / static_cast<double>(numberOfBins);
   const auto binDimensions = domainDimensions / static_cast<double>(numberOfBinsPerDim);
 
-  // Calculate the actual number of bins per dimension. The rounding is needed in case the division slightly
+  // Calculate the actual number of bins per dimension. The rounding is needed in case the division (due to floating point errors) slightly
   // underestimates the division. The choice of dimension 0 is arbitrary.
   const int binsPerDimension = static_cast<int>(std::round(domainDimensions[0] / binDimensions[0]));
 
@@ -84,23 +84,22 @@ std::pair<double, double> calculateHomogeneityAndMaxDensity(const Container &con
   densityPerBin.reserve(numberOfBins);
   for (size_t i = 0; i < numberOfBins; i++) {
     densityPerBin[i] = static_cast<double>(particlesPerBin[i]) / binVolume;
-    if (densityPerBin[i] > maxDensity) {
-      maxDensity = densityPerBin[i];
-    }
+    maxDensity = std::max(maxDensity, densityPerBin[i]);
   }
 
   if (maxDensity < 0.0) {
-    throw std::runtime_error("maxDensity can never be smaller than 0.0, but is:" + std::to_string(maxDensity));
+    utils::ExceptionHandler::exception("calculateHomogeneityAndMaxDensity(): maxDensity can never be smaller than 0.0, but is: {}", maxDensity);
   }
   // get mean and reserve variable for densityVariance
   const double densityMean = numberOfParticles / domainVolume;
-  double densityVariance = 0.0;
 
-  // calculate densityVariance
-  for (size_t i = 0; i < numberOfBins; ++i) {
-    const double densityDifference = densityPerBin[i] - densityMean;
-    densityVariance += (densityDifference * densityDifference / numberOfBins);
-  }
+  const auto densityDifferenceSquaredSum =
+      std::transform_reduce(densityPerBin.cbegin(), densityPerBin.cend(), 0l, std::plus{},
+                            [&densityMean](auto binDensity) {
+                              const auto densityDifference = binDensity - densityMean;
+                              return densityDifference * densityDifference;
+                            });
+  const auto densityVariance = densityDifferenceSquaredSum / numberOfBins;
 
   // finally calculate standard deviation
   const double homogeneity = std::sqrt(densityVariance);
--- a/src/autopas/utils/SimilarityFunctions.h
+++ b/src/autopas/utils/SimilarityFunctions.h
@@ -77,7 +77,7 @@ std::pair<double, double> calculateHomogeneityAndMaxDensity(const ParticleContai
   std::vector<double> densityPerBin{};
   densityPerBin.reserve(numberOfBins);
   for (size_t i = 0; i < numberOfBins; i++) {
-    densityPerBin[i] = static_cast<double>(particlesPerBin[i]) / binVolume;
+    densityPerBin.push_back(static_cast<double>(particlesPerBin[i]) / binVolume);
     maxDensity = std::max(maxDensity, densityPerBin[i]);
   }
 
--- a/src/autopas/utils/SimilarityFunctions.h
+++ b/src/autopas/utils/SimilarityFunctions.h
@@ -45,7 +45,7 @@ std::pair<double, double> calculateHomogeneityAndMaxDensity(const ParticleContai
   using namespace autopas::utils::ArrayMath::literals;
 
   const auto numberOfParticles = container.getNumberOfParticles();
-  const auto domainDimensions = container.getBoxMax() - container.getBoxMax();
+  const auto domainDimensions = container.getBoxMax() - container.getBoxMin();
   const auto domainVolume = domainDimensions[0] * domainDimensions[1] * domainDimensions[2];
 
   // We scale the dimensions of the domain to bins with volumes which give approximately 10 particles per bin.
@@ -53,13 +53,10 @@ std::pair<double, double> calculateHomogeneityAndMaxDensity(const ParticleContai
   const auto targetNumberOfBinsPerDim = std::cbrt(static_cast<double>(targetNumberOfBins));
   // This is probably not an integer, so we floor to get more than 10 particles per bin than too small bins
   const auto numberOfBinsPerDim = static_cast<int>(std::floor(targetNumberOfBinsPerDim));
-  const auto numberOfBins = numberOfBinsPerDim * numberOfBinsPerDim * numberOfBinsPerDim;
-  const auto binVolume = domainVolume / static_cast<double>(numberOfBins);
   const auto binDimensions = domainDimensions / static_cast<double>(numberOfBinsPerDim);
 
-  // Calculate the actual number of bins per dimension. The rounding is needed in case the division (due to floating
-  // point errors) slightly underestimates the division. The choice of dimension 0 is arbitrary.
-  const int binsPerDimension = static_cast<int>(std::round(domainDimensions[0] / binDimensions[0]));
+  const auto numberOfBins = numberOfBinsPerDim * numberOfBinsPerDim * numberOfBinsPerDim;
+  const auto binVolume = domainVolume / static_cast<double>(numberOfBins);
 
   std::vector<size_t> particlesPerBin(numberOfBins, 0);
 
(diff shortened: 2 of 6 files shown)


src/autopas/utils/SimilarityFunctions.h
```diff
@@ -28,86 +46,68 @@ template <class Container>
...
+  const auto domainVolume = domainDimensions[0] * domainDimensions[1] * domainDimensions[2];
+
+  // We scale the dimensions of the domain to bins with volumes which give approximately 10 particles per bin.
+  const auto targetNumberOfBins = std::ceil(numberOfParticles / 10.);
+  const auto targetNumberOfBinsPerDim = std::cbrt((double)targetNumberOfBins);
+  // This is probably not an integer, so we floor to get more than 10 particles per bin than too small bins
+  const auto numberOfBinsPerDim = static_cast<int>(std::floor(targetNumberOfBinsPerDim));
+  const auto numberOfBins = numberOfBinsPerDim * numberOfBinsPerDim * numberOfBinsPerDim;
+  const auto binVolume = domainVolume / (double)numberOfBins;
+  const auto binDimensions = domainDimensions / (double)numberOfBinsPerDim;
```
FG-TUM, line 62:
```suggestion
  const auto binVolume = domainVolume / static_cast<double>(numberOfBins);
  const auto binDimensions = domainDimensions / static_cast<double>(numberOfBinsPerDim);
```

src/autopas/utils/SimilarityFunctions.h
```diff
@@ -28,86 +46,68 @@ template <class Container>
...
-        (allVolumes[i] == 0) ? 0 : (particlesPerCell[i] / allVolumes[i]);  // make sure there is no division of zero
-    if (densityPerCell[i] > maxDensity) {
-      maxDensity = densityPerCell[i];
+  std::vector<double> densityPerBin;
+  densityPerBin.reserve(numberOfBins);
+  for (size_t i = 0; i < numberOfBins; i++) {
+    densityPerBin[i] = (double)particlesPerBin[i] / binVolume;
+    if (densityPerBin[i] > maxDensity) {
+      maxDensity = densityPerBin[i];
     }
```
FG-TUM, line 89:
```suggestion
    maxDensity = std::max(maxDensity, densityPerBin[i]);
```
slightly cleaner

src/autopas/utils/SimilarityFunctions.h
```diff
@@ -28,86 +46,68 @@ template <class Container>
...
-  std::vector<double> densityPerCell(numberOfCells, 0.0);
-  for (size_t i = 0; i < particlesPerCell.size(); i++) {
-    densityPerCell[i] =
-        (allVolumes[i] == 0) ? 0 : (particlesPerCell[i] / allVolumes[i]);  // make sure there is no division of zero
-    if (densityPerCell[i] > maxDensity) {
-      maxDensity = densityPerCell[i];
+  std::vector<double> densityPerBin;
+  densityPerBin.reserve(numberOfBins);
+  for (size_t i = 0; i < numberOfBins; i++) {
+    densityPerBin[i] = (double)particlesPerBin[i] / binVolume;
```
FG-TUM, line 86:
```suggestion
    densityPerBin[i] = static_cast<double>(particlesPerBin[i]) / binVolume;
```

src/autopas/utils/SimilarityFunctions.h
```diff
@@ -28,86 +46,68 @@ template <class Container>
...
-  // calculate density for each cell
+  // calculate density for each bin and track max density
   double maxDensity{0.};
-  std::vector<double> densityPerCell(numberOfCells, 0.0);
-  for (size_t i = 0; i < particlesPerCell.size(); i++) {
-    densityPerCell[i] =
-        (allVolumes[i] == 0) ? 0 : (particlesPerCell[i] / allVolumes[i]);  // make sure there is no division of zero
-    if (densityPerCell[i] > maxDensity) {
-      maxDensity = densityPerCell[i];
+  std::vector<double> densityPerBin;
```
FG-TUM, line 83:
```suggestion
  std::vector<double> densityPerBin{};
```

src/autopas/utils/SimilarityFunctions.h
```diff
@@ -28,86 +46,68 @@ template <class Container>
...
+  for (size_t i = 0; i < numberOfBins; i++) {
+    densityPerBin[i] = (double)particlesPerBin[i] / binVolume;
+    if (densityPerBin[i] > maxDensity) {
+      maxDensity = densityPerBin[i];
     }
   }
 
-  if (maxDensity < 0.0)
+  if (maxDensity < 0.0) {
     throw std::runtime_error("maxDensity can never be smaller than 0.0, but is:" + std::to_string(maxDensity));
```
FG-TUM, line 93:
```suggestion
    utils::Exception::exception("calculateHomogeneityAndMaxDensity(): maxDensity can never be smaller than 0.0, but is: {}", maxDensity);
```

src/autopas/utils/SimilarityFunctions.h
```diff
@@ -28,86 +46,68 @@ template <class Container>
...
+    densityVariance += (densityDifference * densityDifference / numberOfBins);
   }
 
   // finally calculate standard deviation
-  const double homogeneity = sqrt(densityVariance);
+  const double homogeneity = std::sqrt(densityVariance);
   // normally between 0.0 and 1.5
-  if (homogeneity < 0.0)
+  if (homogeneity < 0.0) {
     throw std::runtime_error("homogeneity can never be smaller than 0.0, but is:" + std::to_string(homogeneity));
```
FG-TUM, line 109:
```suggestion
        utils::Exception::exception("calculateHomogeneityAndMaxDensity(): homogeneity can never be smaller than 0.0, but is: {}", homogeneity);
```

src/autopas/utils/SimilarityFunctions.h
```diff
@@ -28,86 +46,68 @@ template <class Container>
...
   double densityVariance = 0.0;
 
   // calculate densityVariance
-  for (size_t r = 0; r < densityPerCell.size(); ++r) {
-    double distance = densityPerCell[r] - densityMean;
-    densityVariance += (distance * distance / densityPerCell.size());
+  for (size_t i = 0; i < numberOfBins; ++i) {
+    const double densityDifference = densityPerBin[i] - densityMean;
+    densityVariance += (densityDifference * densityDifference / numberOfBins);
   }
```
FG-TUM, line 103: I'm not a fan of this variable + for loop pattern. Therefore, I'd replace this with some function from `algorithm.h` (`reduce` or `transform_reduce`?). But I could live with the loop if you prefer it ;)

FG-TUM (changes requested):

src/autopas/utils/SimilarityFunctions.h
```diff
@@ -16,7 +17,24 @@ namespace autopas::utils {
...
+ *
+ * These values are calculated by dividing the domain into bins of perfectly equal cuboid shapes.
+ *
+ * @note The bin shapes between different AutoPas container will not match if the dimensions of their domains don't
+ * match. Bins with greater surface area are probably more likely to feature fluctuations in the density.
+ *
+ * Not a rigorous proof, but should give an idea:
+ * - Assume homogeneous distribution.
+ * - For every particle assume it could move in any direction with equal probability (not a terrible assumption).
+ * - So every particle near a bin face has a chance to more into the neighboring bin.
```
thesamriel, line 30:
```suggestion
 * - So every particle near a bin face has a chance to move into the neighboring bin.
```

src/autopas/utils/SimilarityFunctions.h
```diff
@@ -28,86 +46,68 @@ template <class Container>
...
+  const auto targetNumberOfBins = std::ceil(numberOfParticles / 10.);
+  const auto targetNumberOfBinsPerDim = std::cbrt((double)targetNumberOfBins);
+  // This is probably not an integer, so we floor to get more than 10 particles per bin than too small bins
+  const auto numberOfBinsPerDim = static_cast<int>(std::floor(targetNumberOfBinsPerDim));
+  const auto numberOfBins = numberOfBinsPerDim * numberOfBinsPerDim * numberOfBinsPerDim;
+  const auto binVolume = domainVolume / (double)numberOfBins;
+  const auto binDimensions = domainDimensions / (double)numberOfBinsPerDim;
+
+  // Calculate the actual number of bins per dimension. The rounding is needed in case the division slightly
+  // underestimates the division. The choice of dimension 0 is arbitrary.
```
thesamriel, line 65: What do you mean with this? It's needed to correct floating point errors from the division right?

src/autopas/utils/SimilarityFunctions.h
```diff
@@ -28,86 +46,68 @@ template <class Container>
...
-      } else {
-        allVolumes[cellIndex] *= cellLength;
-      }
-    }
+
+    const auto binIndex3d = autopas::utils::ArrayMath::floorToInt((particleLocation - startCorner) / binDimensions);
+    const auto binIndex1d = autopas::utils::ThreeDimensionalMapping::threeToOneD(
+        binIndex3d, {binsPerDimension, binsPerDimension, binsPerDimension});
+
+    particlesPerBin[binIndex1d] += 1;
```
thesamriel, line 78: Should probably use a region iterator here. Wouldn't this lead to errors when the selected region is not the whole domain.

src/autopas/utils/SimilarityFunctions.h
```diff
@@ -28,86 +46,68 @@ template <class Container>
...
   }
 
-  if (maxDensity < 0.0)
+  if (maxDensity < 0.0) {
     throw std::runtime_error("maxDensity can never be smaller than 0.0, but is:" + std::to_string(maxDensity));
-
+  }
   // get mean and reserve variable for densityVariance
-  const double densityMean = numberOfParticles / volume;
+  const double densityMean = numberOfParticles / domainVolume;
```
thesamriel, line 96: I think this is also not correct for a smaller sampled region, as you divide autopas's total number of particles by the (smaller) sampled volume.

thesamriel (changes requested):
I think the function does not work if the volume spanned by `startCorner` and `endCorner` is not equal the AutoPas box volume. From the function signature and description I assume that it is supposed to be possible to sample from a subvolume.  
Preferably, this should be fixed and a test should be added for that.  
Otherwise I would change the function signature to just pass the container and get the box volume from the container.

src/autopas/utils/SimilarityFunctions.h
thesamriel: I personally don't really care about the style here, but in general, I often find the `transform_reduce` constructs harder to read than the for loop which does the computations similar to how I would do it by hand.
I would therefore be interested to know, why you(@FG-TUM ) prefer std algorithms whenever possible?  
(Maybe the readability thing is just matter of habit)

FG-TUM: Yes, I, too, think readability is more of a matter of habit. The reason why I prefer the algorithms is that it is significantly harder to sneak in suspicious behavior. For example, in an index-based loop, you can mess with the index in the loop, start or end at any number, but this might be very easy to miss in a review because you **expect** it to go from 0 to end. A std algorithm guarantees no index mess. Therefore, I prefer to use for-loops only when such shenanigans are actually necessary.

The second point is that std algorithms allow you to initialize the result variable directly and make it const, decreasing the mental complexity of the rest of the code block.

examples/md-flexible/src/Simulation.cpp
SamNewcome, line 1: The homogeneity stuff was removed from Simulation for two reasons:
- The calculation was wrong to begin with.
- We still will have a way to get the homogeneity for each container once #876 is merged so we don't need this.

This does mean we won't have a global homogeneity calculation anymore - but we don't have a correct one currently anyway.

src/autopas/utils/SimilarityFunctions.h
SamNewcome: Recall our earlier conversation. In `a00015ee`, the `startCorner` and `endCorner` parameters were removed to make it clearer that this is used only for an entire container.

src/autopas/utils/SimilarityFunctions.h
SamNewcome: Done in `c542c145`

src/autopas/utils/ThreeDimensionalMapping.h
```diff
@@ -52,12 +52,12 @@ constexpr T threeToOneD(const std::array<T, 3> &index3d, const std::array<T, 3>
...
-constexpr std::array<T, 3> oneToThreeD(T ind, const std::array<T, 3> &dims) {
+constexpr std::array<size_t, 3> oneToThreeD(T ind, const std::array<T, 3> &dims) {
   static_assert(std::is_integral<T>::value, "oneToThreeD requires integral types");
   std::array<T, 3> pos{};
-  pos[2] = ind / (dims[0] * dims[1]);
-  pos[1] = (ind - pos[2] * dims[0] * dims[1]) / dims[0];
-  pos[0] = ind - dims[0] * (pos[1] + dims[1] * pos[2]);
+  pos[2] = static_cast<size_t>(ind / (dims[0] * dims[1]));
+  pos[1] = static_cast<size_t>((ind % (dims[0] * dims[1])) / dims[0]);
+  pos[0] = static_cast<size_t>(ind % dims[0]);
```
manishmishra6016, line 60: Now that I relook at it, the existing implementation is also correct, I didn't need to change. I just had some doubt so I changed it while debugging. But the problem was elsewhere. So, I think we can revert back or keep it, both are equivalent.

src/autopas/utils/ThreeDimensionalMapping.h
```diff
@@ -41,7 +41,7 @@ constexpr T threeToOneD(T x, T y, T z, const std::array<T, 3> &dims) {
 template <typename T>
 constexpr T threeToOneD(const std::array<T, 3> &index3d, const std::array<T, 3> &dims) {
   static_assert(std::is_integral<T>::value, "threeToOneD requires integral types");
-  return (index3d[2] * dims[1] + index3d[1]) * dims[0] + index3d[0];
+  return index3d[2] * dims[1] * dims[0] + index3d[1] * dims[0] + index3d[0];
```
manishmishra6016, line 44: also equivalent.

src/autopas/utils/SimilarityFunctions.h
```diff
@@ -8,106 +8,111 @@
...
 
+  /*
+    double densityVariance = std::transform_reduce(
+          densityPerBin.begin(), densityPerBin.end(), 0.0,
+          std::plus<>(),
+          [densityMean](double density) {
+              double densityDifference = density - densityMean;
+              return densityDifference * densityDifference;
+          }
+      );
```
manishmishra6016, line 105: somehow `transform_reduce` is just returning 0., and I was unable to figure out why. So, just for now, I commented it out and went back to for-loop. But this needs to be fixed.

src/autopas/utils/SimilarityFunctions.h
```diff
@@ -8,106 +8,104 @@
...
- * @param endCorner upper right back corner of the box where the homogeneity shall be calculated.
  * @return {homogeneity, maxDensity}
  */
-template <class Container>
-std::pair<double, double> calculateHomogeneityAndMaxDensity(const Container &container,
-                                                            const std::array<double, 3> &startCorner,
-                                                            const std::array<double, 3> &endCorner) {
-  unsigned int numberOfParticles = static_cast<unsigned int>(container.getNumberOfParticles());
-  autopas::AutoPas_MPI_Allreduce(AUTOPAS_MPI_IN_PLACE, &numberOfParticles, 1, AUTOPAS_MPI_UNSIGNED_INT, AUTOPAS_MPI_SUM,
-                                 AUTOPAS_MPI_COMM_WORLD);
```
manishmishra6016, line 33: I noticed that we have removed this `MPI_AllReduce` call in this refactoring. I was looking through it and it seems like this might be needed by autotuner when MPI is on, even for md-flexible simulations, but I am not sure. 

`homogenity` and `maxDensity` is used by `MPIParallelizedStrategy` class and earlier it was getting the values calculated over the entire domain and not just individual rank (which is the case now). @FG-TUM probably can tell us better about this. We might need to adjust the function accordingly so it can be used for both cases.

src/autopas/utils/SimilarityFunctions.h
```diff
@@ -28,86 +46,68 @@ template <class Container>
...
-    cellsPerDimension[i] = ceil(domainSizePerDimension[i] / cellLength);
-  }
-  // Actual number of cells we end up with
-  numberOfCells = cellsPerDimension[0] * cellsPerDimension[1] * cellsPerDimension[2];
+  const auto numberOfParticles = container.getNumberOfParticles();
+  const auto domainDimensions = endCorner - startCorner;
+  const auto domainVolume = domainDimensions[0] * domainDimensions[1] * domainDimensions[2];
+
+  // We scale the dimensions of the domain to bins with volumes which give approximately 10 particles per bin.
+  const auto targetNumberOfBins = std::ceil(numberOfParticles / 10.);
```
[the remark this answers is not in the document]
FG-TUM: Ok makes sense but then please add a comment that this is an arbitrary value and could potentially optimized.

src/autopas/utils/SimilarityFunctions.h
FG-TUM: nope this line was not touched :P

src/autopas/utils/SimilarityFunctions.h
SamNewcome: My understanding is that this was a mistake. `MPIParallelizedStrategy` checks which subdomains are "similar" and divides the configurations to check between them. By "similar", it means which have similar _local_ homogeneities and max densities, as calculated through this - so no MPI communication needs to be done because this function should just calculate local quantities*. By including this `MPI_AllReduce` call, the function got the total numbers of particles in the entire domain. This was then used to determine bin size (more particles = smaller bins), but only local particles were considered for homogeneity and max density. - so the more MPI ranks there are, the smaller the bins. This doesn't make sense.

*there was an exception to this - in md-flexible, it appeared to be the case that the total global domain's homogeneity + max density was calculated. But this didn't work because the function was still only interesting over local particles + it was used in a misleading manner. So this was removed.

manishmishra6016: This makes sense. Thank you!

manishmishra6016 (approved):
This looks good to me :)

src/autopas/utils/SimilarityFunctions.h
SamNewcome: Done in `c542c145e99aab9fe44b88b8be3f22c83929eaea`

src/autopas/utils/SimilarityFunctions.h
SamNewcome: Not needed anymore

src/autopas/utils/SimilarityFunctions.h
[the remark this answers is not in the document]
SamNewcome: Done in `4129d542`

src/autopas/utils/SimilarityFunctions.h
SamNewcome: My bad! Done in `4129d542`

src/autopas/utils/SimilarityFunctions.h
SamNewcome: Done (as of 92553a1e)

FG-TUM: Nope you did it here: b4d74789cd1e8b259c4adc931b964198c0403dfc

FG-TUM (approved):

thesamriel (approved):

PR #895, opened 2024-06-28
Fixing out of bounds indices in SimilarityFunctions due to FP errors

SamNewcome:
# Description

Previously, it was possible for `calculateHomogeneityAndMaxDensity` to write to out-of-bounds memory when a particle was so close to a boundary that FP errors resulted in it being assigned with an out-of-bounds bin index. This PR fixes this by checking for indices that are out of bounds and, if the particle is still within the boundaries of the domain, correcting the index to be just inside the domain.

## Related Pull Requests

- #884 previously fixed some bugs in `calculateHomogeneityAndMaxDensity`, but this bug was missed.
- #876 is the PR that highlighted this problem, as before this PR, `calculateHomogeneityAndMaxDensity` was not called in the CI beyond a unit test.

## Resolved Issues

- [x] could consider this PR as properly fixing #881

# How Has This Been Tested?

- [x] CI
- [x] Testing the failing test of #876 (`sph-mpi-main.test`)
- [x] #876's CI with these changes merged


src/autopas/utils/SimilarityFunctions.h
```diff
@@ -65,10 +65,41 @@ std::pair<double, double> calculateHomogeneityAndMaxDensity(const ParticleContai
...
+    // e.g. if there are 7 bins in the x dimension, and that particle is close to the right domain boundary, the
+    // division above might result in 7.0, which gets floored to 7 corresponding to a bin index that is out of bounds!
+
+    // We therefore check for particle indices that are out of bounds and, if the particle is close to the boundary we
+    // adjust the index. We don't care about floating point errors causing incorrect indices internally in the
+    // domain as, for particles so close to a boundary, it is somewhat arbitrary which bin they fall into.
+
+    const auto binIndex3DSafe = [&]() {
+      auto newBinIndex3D = binIndex3dUnsafe;
+      for (int i = 0; i < 3; ++i) {
```
FG-TUM, line 81: I find the body of the loop a bit repetitive and confusing to read. 
Here is my suggestion how you could refactor it: 

for each dimension 
 - Check that the particle is indeed within the container and throw an exception if not. (one if, exception call is only implemented once) 
 - use `std::clamp` to force the index into the bounds.

FG-TUM (changes requested):

src/autopas/utils/SimilarityFunctions.h
SamNewcome: See a6952e7d

FG-TUM (approved):

PR #900, opened 2024-07-02
AutoTuner doesn't rebuild for trivial search space

manishmishra6016:
# Description

Currently when we run the simulation we only one configuration, AutoTuner still rebuilds the container leading to rebuild at `tuning-samples` frequency instead of `verlet-rebuild-frequency` provided in input file. And this continues until and beyond the `tuning-interval` instead of just stopping after `tuning-samples` samples of the given configuration. I didn't look into detail why this happening as there was a simple fix to the problem. 

I came across this now when I was running some simulations with single configuration on master branch for comparison with dynamic container branch. Small change here ensures that this doesn't happen.

# How Has This Been Tested?

I verified this using the Iteration Logger for the `fallingDrop` example.


FG-TUM (approved):
Nice catch, nice fix.

PR #901, opened 2024-07-03
Keeping only single (MPI) version of SPH example

manishmishra6016:
# Description

Currently, we have two separate codes `sph-main.cpp` and `sph-main-mpi.cpp`. Former is built everytime while latter is built only if MPI is available in the system. Both the codes have a large overlap. What I wanted to achieve here is to combine them into a single file which can build with/without MPI support. When it is built with MPI, it can still be executed on single-core. 

to implement this, I added an option `SPH_USE_MPI` in the cmake file.

# How Has This Been Tested?

The code builds and runs as before.


manishmishra6016:
@FG-TUM @SamNewcome @thesamriel What would be your first thoughts on this? (this is not ready for a full review)

I am not sure how strongly I want this and there is still some thought needed here. I just wanted to see if it would work and I think it is possible to combine both.

FG-TUM:
I fully support this. Imho you can go even further and combine the sph diagram generation example also into it. 

My only comment would be that I think the priority of this is rather low so only proceed on this if there are no other pressing tasks...

SamNewcome:
I fully support this!

PR #902, opened 2024-07-04
Fix MacOS Compliance

schuhmaj:
# Description 🍎

_Finally, the update nobody asked for 🙃_

## Benefits

- Compilation/ MD-Flexible/ Testing works now on macOS using Apple Clang and GCC Compilers

## Code Changes

- Update `CMakeList.txt` with a generator expression that enables  the `rt` library depending on whether you use macOS or Linux. The library is not available on macOS, but it is also not needed to access the `Time.h` header!
-  Add `<algorithm>` in `LoggerTest.h` due to the usage of `std::clamp` - some compilation problem which probably has not yet appeared before due to nondeterministic compilation, but happened on my machine
- Fixes implicit conversion for `Apple Clang` in `CellFunctor.h` and `DataLayoutConverter.h` by introducing a `static_cast` 
  - I am not certainly sure why `Apple Clang` complains about the current usage of the `DataLayoutOption` with its internal `enum` given that a *non*-explicit `constexpr` conversion operator exists. 
  - However, I assume (I could not nail it down to a concrete [C++ Standard Paper](https://en.cppreference.com/w/cpp/compiler_support))  that the integration of `constexpr` into `Apple Clang` is a bit behind, causing the issue.
- Fix the `MemoryProfiler`. A mechanism like `/proc/self/status` does not exist on macOS and led to a `segfault`. Instead, we use the available `mach.h` header to request the memory information.

## For the reviewer/ Discussable

- [X] ~Definition thing: Conversion from bytes --> kilobytes (Divide by 1024 or 1000?) - Currently 1024 is used in the `MemoryProfiler`.~  --> Since 1998, a kilobyte is officially unambiguously an increment of 1000 bytes as defined by the International Electrotechnical Commission (compare [here](http://userpage.fu-berlin.de/~ram/pub/pub_jf47ht81Ht/code_kbyte_en))
- [X] ~Shall we now support the `Apple Clang` compiler and remove the warning during the CMake Setup (or keep it for now as "unofficially" supported)~ --> We do not introduce macOS support, and keep it "unofficial" (due to additional testing requirements, i.e. a macOS CI/CD, and support for all features not yet available)

# How Has This Been Tested?

The `runTests`  && `md-flex` targets without arguments (and without MPI) have been run with the following combinations:

- macOS, GNU 14.1.0
- macOS, Apple Clang 15.0.0
- Linux ARM64 (through docker image on macOS), GNU 13.2.0
- Linux AMD64, GNU 9.4.0

### DIFF
--- a/src/autopas/pairwiseFunctors/CellFunctor.h
+++ b/src/autopas/pairwiseFunctors/CellFunctor.h
@@ -151,6 +151,7 @@ void CellFunctor<ParticleCell, ParticleFunctor, bidirectional>::processCell(Part
     return;
   }
 
+  // (Explicit) static cast required for Apple Clang (last tested version: 15.0.0)
   switch (static_cast<DataLayoutOption::Value>(_dataLayout)) {
     case DataLayoutOption::aos:
       processCellAoS(cell);
@@ -185,6 +186,7 @@ void CellFunctor<ParticleCell, ParticleFunctor, bidirectional>::processCellPair(
     return;
   }
 
+  // (Explicit) static cast required for Apple Clang (last tested version: 15.0.0)
   switch (static_cast<DataLayoutOption::Value>(_dataLayout)) {
     case DataLayoutOption::aos:
       if (_useNewton3) {
--- a/src/autopas/pairwiseFunctors/CellFunctor.h
+++ b/src/autopas/pairwiseFunctors/CellFunctor.h
@@ -151,7 +151,7 @@ void CellFunctor<ParticleCell, ParticleFunctor, bidirectional>::processCell(Part
     return;
   }
 
-  switch (_dataLayout) {
+  switch (static_cast<DataLayoutOption::Value>(_dataLayout)) {
     case DataLayoutOption::aos:
       processCellAoS(cell);
       break;
@@ -185,7 +185,7 @@ void CellFunctor<ParticleCell, ParticleFunctor, bidirectional>::processCellPair(
     return;
   }
 
-  switch (_dataLayout) {
+  switch (static_cast<DataLayoutOption::Value>(_dataLayout)) {
     case DataLayoutOption::aos:
       if (_useNewton3) {
         processCellPairAoSN3(cell1, cell2, sortingDirection);
--- a/src/autopas/CMakeLists.txt
+++ b/src/autopas/CMakeLists.txt
@@ -30,7 +30,7 @@ add_library(autopas STATIC ${MY_SRC})
 target_link_libraries(
     autopas
     PUBLIC
-        rt # required for Time.h
+        $<$<NOT:$<PLATFORM_ID:Darwin>>:rt> # required for Time.h, on macOS not needed
         ${CMAKE_THREAD_LIBS_INIT} # required for Logger and ExceptionHandler
         $<$<BOOL:${AUTOPAS_OPENMP}>:OpenMP::OpenMP_CXX>
         spdlog::spdlog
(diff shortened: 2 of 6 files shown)


FG-TUM:
> Shall we now support the Apple Clang compiler and remove the warning during the CMake Setup (or keep it for now as "unofficially" supported)

I'd only "officially" support this if we test it with our CI. There I'm not sure I want it because it means I'd have to create workarounds for a compiler I personally don't care for. But that's not a strong opinion and I'd be happy to hear everybody's opinion on this. 

--------------

Could you also check if the energy measurement stuff works? This also works with reading some MSRs (quasi files)

schuhmaj:
> I'd only "officially" support this if we test it with our CI. There I'm not sure I want it because it means I'd have to create workarounds for a compiler I personally don't care for. But that's not a strong opinion and I'd be happy to hear everybody's opinion on this.

I would definitely agree to that, if official support, then also a macOS matrix entry in the CI.

> Could you also check if the energy measurement stuff works? This also works with reading some MSRs (quasi files)

I haven't thought about that given that `AUTOPAS_ENABLE_ENERGY_MEASUREMENTS` default to `FALSE` (so it wasn't in my *WARM-UP - Get to know Autopas by making it work with macOS* 🤓
But looking at it:
**Definitely not.** `macOS` does not provide a `/sys` directory. Hence the `Linux` approach will fail.
One could include this behavior by regularly querying `sudo power metrics -n 1 --samplers cpu_power`, which yields something like this:

```
[...]

CPU Power: 248 mW
GPU Power: 8 mW
ANE Power: 0 mW
Combined Power (CPU + GPU + ANE): 256 mW
```

But I am not sure. Do you think this is good to include? I guess the energy consumption is more interesting for server-side applications, and I yet have to hear about large `Apple Mx`-based servers (even though maybe this gets a big thing as soon as the "Apple Intelligence" 👽 lands 🛸 in its private cloud compute).

I guess this also might give an answer to the first question --> No full support/ Only partial support --> Keeping the support list as-is.

FG-TUM:
> One could include this behavior by regularly querying `sudo powermetrics -n 1 --samplers cpu_power`,

Nope not an option. Simulations must be able to run without `sudo`. 

Ok but then let's ignore the energy stuff for macOS for now. @manishmishra6016 is currently looking into using a wrapper lib for this anyways, maybe this solves the problem...?

schuhmaj:
I correct myself. There **are** other options as well. However, the documentation is a bit sparse 

The `mach/task_info.h` header also contains a struct and code to access the power statistics:

* [Apple Developer](https://developer.apple.com/documentation/kernel/task_power_info_v2_data_t/1553338-cpu_energy)
* [The associated header](https://github.com/apple/darwin-xnu/blob/2ff845c2e033bd0ff64b5b6aa6063a1f8f65aa32/osfmk/mach/task_info.h#L464)

Documentation non-existent, code comments missing as well. I will give it a try.

src/autopas/pairwiseFunctors/CellFunctor.h
```diff
@@ -151,7 +151,7 @@ void CellFunctor<ParticleCell, ParticleFunctor, bidirectional>::processCell(Part
     return;
   }
 
-  switch (_dataLayout) {
+  switch (static_cast<DataLayoutOption::Value>(_dataLayout)) {
```
FG-TUM, line 154:
```suggestion
  // Static cast only needed for Apple Clang
  switch (static_cast<DataLayoutOption::Value>(_dataLayout)) {
```
Add a comment like this to all four instances and add the version with which you encountered the compiler bug.

FG-TUM (changes requested):
Just a comment why we need the unnecessary cast

src/autopas/pairwiseFunctors/CellFunctor.h
schuhmaj: Comments added ✅

PR #912, opened 2024-07-11
Feature/energy measurement with PMT

### LINKED ISSUE #849: Energy measurement with PMT
Energy measurement with [PMT](https://git.astron.nl/RD/pmt)  

**Additional context**
Currently, the energy measurements is only possible for Intel hardware through RAPL (exposed through file system). PMT offers a method to do the same for Intel (also using RAPL), AMD and Nvidia hardware.

manishmishra6016:
# Description

Using [PMT: Power measurement toolkit](https://git.astron.nl/RD/pmt/-/tree/master?ref_type=heads) for energy measurement and moving away from old implementation `RaplMeter.h`.
Added PMT library to cmake with bundled version. Created new class `EnergySensor.h` to encapsulate usage of PMT. 

When the energy measurement is enabled using `AUTOPAS_ENABLE_ENERGY_MEASURMENTS=ON`, the default sensor is set to `RAPL`. 

## Related Pull Requests

- None

## Resolved Issues

- [x] fixes #849 

# How Has This Been Tested?

- [x] Simulations with energy measurement and energy autotuning on Intel and AMD clusters
- [x] Comparison with RaplMeter implementation


cmake/modules/autopas_pmt.cmake
```diff
@@ -0,0 +1,51 @@
...
+    set_property(DIRECTORY ${pmt_SOURCE_DIR} PROPERTY EXCLUDE_FROM_ALL YES)
+endif()
+
+target_compile_options(pmt PUBLIC -DCMAKE_INSTALL_PREFIX="./build")
+
+get_target_property(propval pmt INTERFACE_INCLUDE_DIRECTORIES)
+message(" Directories: ${propval}")
+
+target_include_directories(pmt SYSTEM PUBLIC "${pmt_SOURCE_DIR}")
+#target_include_directories(pmt SYSTEM PUBLIC "${propval}")
```
SamNewcome, line 51: Don't include "dead" code. 

Is this old code that has been forgotten about? -> Delete it. 
Is this useful code in some specific scenarios? -> Add a CMake switch to toggle this on or off.
Something else -> ?

docs/userdoc/Building.md
```diff
@@ -54,12 +54,11 @@ cmake -DAUTOPAS_ENABLE_RULES_BASED_AND_FUZZY_TUNING=ON ..
 ### Energy Measurements and Tuning
 
 By default, AutoPas tunes for the best configuration according to runtime. For all Linux based systems, it is also possible to tune
-for the algorithm that consumes the least energy. This is implemented via [Intel's RAPL](https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/advisory-guidance/running-average-power-limit-energy-reporting.html) 
-(Running Average Power Limit) interface.
+for the algorithm that consumes the least energy. This is implemented via [pmt-stable](https://github.com/MaxPraus23/pmt-stable) which is a customised version tuned for performance based on [PMT](https://git.astron.nl/RD/pmt). In the current version, energy consumption can be measured via Intel's RAPL or with LIKWID.
```
SamNewcome, line 57:
```suggestion
for the algorithm that consumes the least energy. This is implemented via [pmt-stable](https://github.com/MaxPraus23/pmt-stable) which is a customized version tuned for performance based on [PMT](https://git.astron.nl/RD/pmt). In the current version, energy consumption can be measured via Intel's RAPL or with LIKWID.
```
We use American English in AutoPas :sob: :sob: :sob: :vomiting_face: :vomiting_face:

docs/userdoc/Building.md
```diff
@@ -54,12 +54,11 @@ cmake -DAUTOPAS_ENABLE_RULES_BASED_AND_FUZZY_TUNING=ON ..
...
 
-To use energy tuning, energy measurements must be enabled using the CMake option:
+To use energy tuning, the energy sensor, used for measurement, must be specified:
 ```bash
-cmake -DAUTOPAS_ENABLE_ENERGY_MEASUREMENTS=ON .. 
+cmake -DPMT_BUILD_RAPL=ON -DPMT_BUILD_LIKWID=ON .. 
 ```
```
SamNewcome, line 62: Can you expand upon this? For example, do you have to enable both of these? Or just one? What happens if nothing is enabled? These a PMT options, can you link to a documentation page which describes the options and their advantages/disadvantages? (Ideally a documentation page specific to the version you have bundled). If not, please give a brief discussion of when/where you can use these and advantages/disadvantages if relevant.

examples/md-flexible/input/AllOptions.yaml
```diff
@@ -11,8 +11,10 @@ tuning-strategies                    :  [slow-config-filter, rule-based-tuning,
 tuning-interval                      :  5000
 tuning-samples                       :  3
 tuning-max-evidence                  :  10
+tuning-metric                        :  energy
```
SamNewcome, line 14:
```suggestion
tuning-metric                        :  time # alternatively: energy
```
I know `AllOptions.yaml` is not really supposed to be "default" options (whatever that means), but I'd prefer this still.

src/autopas/utils/EnergySensor.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "autopas/options/EnergySensorOption.h"
+#include "pmt.h"
+
+namespace autopas::utils {
+
+EnergySensor::EnergySensor(EnergySensorOption sensor) : _option(sensor) {
+  if (_option != EnergySensorOption::none) {
+    _sensor = pmt::Create(sensor.to_string());
+  } else {
+    AutoPasLog(WARN, "No sensor for energy consumption measurement specified. Energy will not be measured!");
```
SamNewcome, line 18: This shouldn't be a simple `WARN`: If energy is enabled but there is no sensor, then the results used for AutoTuning will be gibberish, right? If so, I would simply throw an exception here.

src/autopas/utils/logging/IterationMeasurements.h
```diff
@@ -37,23 +37,22 @@ struct IterationMeasurements {
   bool energyMeasurementsPossible{false};
 
   /**
-   * Energy consumed by the entire system.
+   * Total energy consumed in watts.
```
SamNewcome, line 40: Shouldn't this be the average energy consumed per time in Watts? Or am I missing something?

SamNewcome (changes requested):
Some changes (but generally rather minor).

Some compatibility questions: You removed the CMake switch for energy measurements. Please confirm that all the added PMT logic is compatible on any machine where AutoPas could be compiled before. For example, can you compile AutoPas on the ARM minicluster (I don't care that I can get energy measurements on there because this isn't an existing feature of AutoPas but I still might want to be able to compile for usual time measurements)? Please speak to Manish if you don't have access to this. Furthermore, does PMT have a significant compilation time cost? If so, or if there are compatibility issues, the CMake option for energy measurements should be readded.

src/autopas/utils/EnergySensor.cpp
FG-TUM: Just to add some input on this:

If I understand the code in the `AutoTuner` correctly, you always instantiate an EnergySensor. The default sensor type is None. Therefore, by default, we would get this warning (or the exception Sam suggested). This is unpleasant.

You could now either 
- only instantiate the `EnergySensor` if the option is not `none` (e.g. wrapping the sensor in `std::optional()`), or 
- always instantiate but only sample energy if the sensor is not `none`.

Either way, I think the most important thing is that we get an exception if someone explicitly wants to measure energy but can't. If nothing is possible but also nothing is requested, I'd prefer not to get a warning.

FG-TUM (changes requested):

manishmishra6016:
@schuhmaj Jonas, can you check if this branch compiles on mac with `AUTOPAS_ENABLE_ENERGY_MEASURMENTS=ON`. I just want to confirm if keeping default option as `ON` is fine now, and PMT handles it internally. If not, we could keep default `OFF` (`ON` for the CI).

cmake/modules/autopas_pmt.cmake
```diff
@@ -1,4 +1,5 @@
 option(pmt_ForceBundled "Do not look for an installed version, always used bundled." ON)
+set(PMT_BUILD_RAPL ON CACHE BOOL "Enabling RAPL always when energy measurement is requested" FORCE)
```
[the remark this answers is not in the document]
SamNewcome: Okay:
```suggestion
set(PMT_BUILD_RAPL ON CACHE BOOL "RAPL is by default enabled when PMT is enabled" FORCE)
```

SamNewcome (changes requested):

SamNewcome (approved):

manishmishra6016:
@FG-TUM I have made the requested changes, could you have another look when you find some time?

manishmishra6016:
> ## GCC 14.2.0/ macOS 15.1/ M1 Pro
> 
> ⚠️Compilation fails due to `PMT.cpp` using `Dummy`, but not including the respective header (Hurra, compiler non-determinism) ✅If I manually add `#include "dummy/Dummy.h"` inside `<cmake-build-folder>/_deps/pmt-src/common/PMT.h`, the compilation succeeds --> One could do this magic via CMake, or we add a fixed bundled version
> ## Apple Clang 16.0.0/ macOS 15.1/ M1 Pro
> 
> ⚠️Compilation fails due to `Dummy.cpp` including `ext/alloc_traits.h`, which is a GNU-specific compiler extension. ⁉️Apparently, one can fix this by telling `clang` to use `libstdc++`, but I haven't tried that.

Thank you for checking! Then I will just keep the default `AUTOPAS_ENABLE_ENERGY_MEASUREMENTS=OFF` for now (`ON` for CI).

src/autopas/utils/EnergySensor.h
```diff
@@ -0,0 +1,97 @@
...
+   * Get watts consumed between start and end state
+   * @return double for watts consumed
+   */
+  double getWatts() const;
+
+  /**
+   * Get seconds between current start and end state
+   * @return double for seconds between time stamps
+   */
+  double getSeconds() const;
```
FG-TUM, line 62: I find the name non-intuitive. Is maybe something like `getEnergyDeltaT()` a better name?

- [ ] This also affects the corresponding column name in the iteration logger

docs/userdoc/Building.md
```diff
@@ -54,14 +54,23 @@ cmake -DAUTOPAS_ENABLE_RULES_BASED_AND_FUZZY_TUNING=ON ..
...
 
+Additionally, an energy sensor used for measurement needs to be selected. By default, RAPL sensor is enabled by setting the default for `PMT_BUILD_RAPL` to `ON`. However, other available sensors like `LIKWID` can also be used by separately compiling them using the CMake option:
+```bash
+cmake -DPMT_BUILD_LIKWID=ON .. 
+```
+Note that user must have `LIKWID` installed before enabling this option.
+Similarly, RAPL can be disabled with
+```bash
+cmake -DPMT_BUILD_RAPL=OFF ..
+```
```
FG-TUM, line 72: Nope you force this to be always ON

docs/userdoc/Building.md
```diff
@@ -54,14 +54,23 @@ cmake -DAUTOPAS_ENABLE_RULES_BASED_AND_FUZZY_TUNING=ON ..
...
-for the algorithm that consumes the least energy. This is implemented via [Intel's RAPL](https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/advisory-guidance/running-average-power-limit-energy-reporting.html) 
-(Running Average Power Limit) interface.
+for the algorithm that consumes the least energy. This is implemented via [pmt-stable](https://github.com/MaxPraus23/pmt-stable) which is a customized version tuned for performance based on [PMT](https://git.astron.nl/RD/pmt). In the current version, energy consumption can be measured via Intel's RAPL or with LIKWID.
 
 To use energy tuning, energy measurements must be enabled using the CMake option:
 ```bash
 cmake -DAUTOPAS_ENABLE_ENERGY_MEASUREMENTS=ON .. 
 ```
 
+Additionally, an energy sensor used for measurement needs to be selected. By default, RAPL sensor is enabled by setting the default for `PMT_BUILD_RAPL` to `ON`. However, other available sensors like `LIKWID` can also be used by separately compiling them using the CMake option:
```
FG-TUM, line 64: So how exactly is the user supposed to choose between the sensors? By only enabling exactly one of them? If so, why not have one CMake option that specifies the sensor? Similar to the `CMAKE_BUILD_TYPE`?

FG-TUM (changes requested):
Some observations:

- [x] When I build with Clang 17 I get the following warning that should be suppressed because it's not from us:

```cpp
/home/ga68cat/work/workspace/AutoPas/cmake-build-manually-clang/_deps/pmt-build/pmt/common/PMT.h:18:10: warning: definition of implicit copy constructor for 'State' is deprecated because it has a user-provided copy assignment operator [-Wdeprecated-copy-with-user-provided-copy]
  State &operator=(const State &state) {
         ^
/home/ga68cat/work/workspace/AutoPas/cmake-build-manually-clang/_deps/pmt-build/pmt/common/PMT.h:65:37: note: in implicit copy constructor for 'pmt::State' first required here
  virtual State GetState() { return state_previous_; };
                                    ^
```

- [x] When I enable `AUTOPAS_ENABLE_ENERGY_MEASUREMENTS`, request `--tuning-metric energy`, disable `PMT_BUILD_LIKWID` but not `PMT_BUILD_RAPL` I get the following warnings:
```bash
Detected 16 cores in 1 package
Could not open /sys/class/powercap/intel-rapl/intel-rapl:0/energy_uj
Could not open /sys/class/powercap/intel-rapl/intel-rapl:0/intel-rapl:0:0/energy_uj
Could not open /sys/class/powercap/intel-rapl/intel-rapl:0/intel-rapl:0:1/energy_uj
Could not open /sys/class/powercap/intel-rapl/intel-rapl:0/intel-rapl:0:2/energy_uj
```
The warnings are nice but I would prefer a crash because I do a energy tuning run without being able to measure energy. The logged Watt and Joule are 0.

manishmishra6016:
> When I build with Clang 17 I get the following warning that should be suppressed because it's not from us:

these warning are suppressed.

> The warnings are nice but I would prefer a crash because I do a energy tuning run without being able to measure energy. The logged Watt and Joule are 0.

I moved to a newer version of PMT, sadly this still only showed warnings, so I have now added an `std::abort` after showing the error message. It now aborts when if the sensor cannot read the energy due to sudo right issue, or if the files don't exist.

docs/userdoc/Building.md
FG-TUM: Just make sure the documentation is consistent with whatever implementation logic you choose :)

docs/userdoc/Building.md
FG-TUM: As far as I can tell, the doc does not mention how to select the sensor at runtime. Please add this or point me to where I overlooked it :D

manishmishra6016:
As I moved to a new version of PMT, I wanted to run the comparison with RaplMeter implementation again. 

### 1. Energy over a simulation
Here, I run the `SpinodalDecomposition_equilibration` example for 10000 iterations with both the implementations and compare the results. As shown in Max's thesis, PMT reports less energy than current implementation as it filters out energy from unused components (unused CPUs, GPUs, etc.). We argue this is better for comparing different configurations as it focuses more on elements that are directly involved in computing.

image

Note: the simulation was performed on HSUper with 72 threads on 1 node without MPI. 

### 2. Energy with increasing number of threads

coming soon...

docs/userdoc/Building.md
manishmishra6016: Improved the documentation in general, included details about forcing `RAPL` to `ON` always, how to compile other sensors, and select from available sensors at runtime. see [e71210b](https://github.com/AutoPas/AutoPas/pull/912/commits/e71210b8f2e7f7fad428f718a2476dfb9e78ebc0)

src/autopas/utils/EnergySensor.h
```diff
@@ -0,0 +1,97 @@
...
+   * Ending measurement. Sets _end
+   * @return returns true if _sensor available. Otherwise false
+   */
+  bool endMeasurement();
+
+  /**
+   * Get joules consumed between start and end state
+   * @return double for watts consumed
+   */
+  double getJoules() const;
```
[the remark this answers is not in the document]
manishmishra6016: done, see [e71210b](https://github.com/AutoPas/AutoPas/pull/912/commits/e71210b8f2e7f7fad428f718a2476dfb9e78ebc0)

docs/userdoc/Building.md
```diff
@@ -53,23 +53,25 @@ cmake -DAUTOPAS_ENABLE_RULES_BASED_AND_FUZZY_TUNING=ON ..
...
+In the current version, energy consumption can be measured via Intel's RAPL or with LIKWID.
+RAPL is set as the default sensor and is forced to compile always.
 
 To use energy tuning, energy measurements must be enabled using the CMake option:
 ```bash
 cmake -DAUTOPAS_ENABLE_ENERGY_MEASUREMENTS=ON .. 
 ```
 
-Additionally, an energy sensor used for measurement needs to be selected. By default, RAPL sensor is enabled by setting the default for `PMT_BUILD_RAPL` to `ON`. However, other available sensors like `LIKWID` can also be used by separately compiling them using the CMake option:
+By default, RAPL sensor is enabled by forcing the cmake option `PMT_BUILD_RAPL` to `ON`. However, other available sensors like `LIKWID` can also be used by separately compiling them using the CMake option:
```
SamNewcome, line 67: Am I correct in my understanding that RAPL cannot be disabled? If yes:
"The RAPL sensor is forcibly enabled by AutoPas forcing the cmake option `PMT_BUILD_RAPL` to `ON`."

If this just means that this option is the default:
"By default, the RAPL sensor is enabled, with the cmake option `PMT_BUILD_RAPL` set to `ON`.

cmake/modules/autopas_pmt.cmake
```diff
@@ -16,26 +16,32 @@ if (NOT ${pmt_ForceBundled})
     endif()
 endif()
 
-message(STATUS "pmt - using bundled version 1.3.1 (commit e93666bd)")
+message(STATUS "pmt - using bundled version (commit 7a56fa3a) and patch")
 
 include(FetchContent)
 
+set(FETCHCONTENT_QUIET OFF)
+set(CMAKE_VERBOSE_MAKEFILE ON)
```
SamNewcome, line 24: Why were these set?

examples/md-flexible/src/Simulation.cpp
```diff
@@ -644,7 +643,7 @@ void Simulation::logMeasurements() {
     std::cout << timerToString("    Thermostat                    ", thermostat, maximumNumberOfDigits, simulate);
     std::cout << timerToString("    Vtk                           ", vtk, maximumNumberOfDigits, simulate);
     std::cout << timerToString("    LoadBalancing                 ", loadBalancing, maximumNumberOfDigits, simulate);
-    std::cout << timerToString("One iteration                     ", simulate / static_cast<long>(_iteration),
+    std::cout << timerToString("    One iteration                 ", simulate / static_cast<long>(_iteration),
```
SamNewcome, line 646:
```suggestion
    std::cout << timerToString("  One iteration                 ", simulate / static_cast<long>(_iteration),
```
You are right that this doesn't make sense at the "root" level, but I would suggest it should be flush with simulate as it is not a "component" of simulate.

SamNewcome (changes requested):

docs/userdoc/Building.md
manishmishra6016: Yes, the first assessment is correct, we force it to be `ON`. I updated the doc as you suggested in commit [1a8e7a8](https://github.com/AutoPas/AutoPas/pull/912/commits/1a8e7a8872d2d697d979670b18332a7531d4f2bb)

cmake/modules/autopas_pmt.cmake
manishmishra6016: Ah, these were for debugging the patch errors I was getting. I removed them now, see [1a8e7a8](https://github.com/AutoPas/AutoPas/pull/912/commits/1a8e7a8872d2d697d979670b18332a7531d4f2bb)

SamNewcome (approved):
Looks good!

PR #918, opened 2024-07-23
Fixed issue with AVX functor for the mixing

TumBeat:
# Description

Fixed an issue for the AVX functor for the mixing case. In some cases, the typeIDptr always pointed at the start of the array and did not consider the current j-index.
The test cases comparing the results of the AVX functor with the usual LJFunctor were adjusted to check for the mixing case.

With this change, for some large values, the test cases did fail with this message:

_The abs_error parameter tolerance evaluates to 1e-08 which is smaller than the minimum distance between doubles for numbers of this magnitude which is 1.4901161193847656e-08, thus making this EXPECT_NEAR check equivalent to EXPECT_EQUAL. Consider using EXPECT_DOUBLE_EQ instead._

Choosing EXPECT_DOUBLE_EQ did not work for every case, instead, the tolerance was increased from 1e-08 to 2e-08 which should not really affect the correctness of the test cases.

# How Has This Been Tested?

- [x] LJFunctorAVXTest.cpp


SamNewcome (changes requested):
Looks generally good, but I'm hesitant about just changing the tests from `mixing=false` to `mixing=true`, because then we might get the a problem with no mixing. My suggestion is just template all the function with a bool for mixing, then create copies of all the actual tests for both cases.

FG-TUM:
Or better: Use the test parametrization that's already used in this test class. Just extend it by another bool for mixing and generate tests for both cases

FG-TUM (approved):

PR #923, opened 2024-07-29
Fix Configs Generator Script

FG-TUM:
# Description

The Python script that was used to generate a wide range of test cases for MD Flexible used an outdated config template.

- [x] Fix/Update template
- [x] Change defaults to something slightly more reasonable
- [ ] Still open: Estimate memory consumption and reject the last failing examples based on that.
   One way to get the available Memory without exotic Python modules:
   ```python
   # use bash to retrieve the value via free (human readable) and then cut out the value
   os.popen("free -ht | grep Mem | column -t -o ',' | cut -d ',' -f 2").read()
   ```

## ~Resolved Issues~

# How Has This Been Tested?

- [x] Ran all generated configs successfully


examples/md-flexible/scripts/generate-mdflex-configs.py
```diff
@@ -8,42 +33,46 @@
...
+# container                        :  [all]    # Currently unfeasible. See below.
+# Containers that are excluded:
+#   - DirectSum             : Takes too long
+#   - LinkedCellsReferences : Quasi equivalent to LinkedCells
+#   - Octree                : Takes too long
+#   - PairwiseVerletLists   : Uses too much memory resulting in chrashes
+container                        :  [LinkedCells, VerletLists, VarVerletLists, VerletListsCells, VerletClusterLists]
 verlet-rebuild-frequency         :  20
-verlet-skin-radius               :  0.15
+verlet-skin-radius-per-timestep  :  0.0075
```
SamNewcome, line 49: I think this is now out of date for the master

examples/md-flexible/scripts/generate-mdflex-configs.py
```diff
@@ -121,8 +144,9 @@ def generate(domainSize,
     data['box-max'] = domainSize
     data['cutoff'] = cutoff
     skin = cutoff * skinFactor
-    data['verlet-skin-radius'] = skin
-    data['verlet-rebuild-frequency'] = int(invMaxParticleSpeed * skinFactor)
+    rebuildFrequency = int(invMaxParticleSpeed * skinFactor)
+    data['verlet-skin-radius-per-timestep'] = skin / rebuildFrequency
```
SamNewcome, line 148: Similar to above comment

examples/md-flexible/scripts/generate-mdflex-configs.py
```diff
@@ -143,34 +167,52 @@ def generate(domainSize,
...
+    # If cells are larger than the domain in any dimension the scenario doesn't make sense
+    if any(item == 0 for item in cellsPerDimension):
+        return False
+    numCells = math.prod(cellsPerDimension)
+    uniformAvgParticlesPerCell = particleCounts[numParticlesName] / numCells
+    # Currently unused:
+    # volume=math.prod(domainSize)
+    # density=particleCounts[numParticlesName]/volume
+
+    # Exclusion rules. Mostly to avoid scenarios that take forever to test.
```
SamNewcome, line 202: This is a slightly awkward one. These parameters would exclude some of the current test cases which I am working on. But whilst I could suggest values which are more appropriate for my current test cases, these values could easily become as inappropriate in the future as these are for my test cases. My suggestion is to leave these as they are but add the warning:

```suggestion
    # Exclusion rules. Mostly to avoid scenarios that take forever to test. Warning: These parameters may not be appropriate for your test cases.
```

examples/md-flexible/scripts/generate-mdflex-configs.py
```diff
@@ -1,5 +1,30 @@
 #!/usr/bin/python3
+
+"""Generates a zoo of configurations for md-flexible
+
+The script generates hundreds of yaml configuration files,
+that aim to cover a wide range of possible simulation states.
+
+It is not simply the cross product of the range of parameters below,
+but also tries to exclude unfeasible combinations, e.g. huge particle
+numbers in tiny domains.
```
SamNewcome, line 10: See my comment on line 202 for my reasoning for this:

```suggestion
It is not simply the cross product of the range of parameters below,
but also tries to exclude unfeasible combinations, e.g. huge particle
numbers in tiny domains.

These exclusion criteria may not be appropriate for your test cases.
E.g. they assume a sigma of no smaller than one and a cutoff of no 
larger than 3.
```
(Potentially replace the assumption with whatever is more appropriate)

SamNewcome (changes requested):
Some changes:

- Warnings that the exclusion criteria may not be appropriate
- Changes that reflect the current dynamic rebuilding master

Related to the last point:
I think the `skinFactor` and `invMaxParticleSpeed` stuff is a bit confusing (even ignoring dynamic rebuilding). I would simply make the `skinFactor` a factor of the cutoff, which is a bit more common of a terminology. Then I would keep rebuild frequency fixed as 100. This is not so suitable but I don't know of something better to have here: with dynamic rebuilding, these "static" (dt=0) experiments become a little misleading, but this is a subject of ongoing research and I don't care about it here.

examples/md-flexible/scripts/generate-mdflex-configs.py
FG-TUM: -> 630a150

examples/md-flexible/scripts/generate-mdflex-configs.py
FG-TUM: -> 630a150

thesamriel:
Sorry, but we will just get rid of scripts in the repo entirely. See #1177

PR #925, opened 2024-08-01
Fix Object Generators

FG-TUM:
# Description

For some reason, our MD-Flex object generators did not use the generators from `autopasTools` and instead duplicated the logic.

In particular, the generator for the Gaussian distribution was buggy and generated particles outside of the specified box.

- [x] Use `autopasTools` generators in MD-Flex
  - [x] Introduce wrapper so that `std::vector` can be used as `AutoPas::ParticleContainer`
- [x] Fix Gauss Box bug
- [x] Convert all generator classes to namespaces
- [x] Use C++ random engine objects instead of `srand`
- [x] Minor fixes to `AllOptions.yaml` and printing of end-configs.

Refactoring this has uncovered a bug in `Newton3OnOffTest`: The SoA variants do not count the number of particle-particle interactions but the number of functor invocations, which do not have to be symmetrical in the case of any Verlet container.
- [x] Make `Newton3OnOffTest` count particle interactions.

## ~Resolved Issues~

# How Has This Been Tested?

- [x] Confirmed Gauss generator bug is gone by generating several scenarios and confirming that the number of particles in the system matches the input.


examples/md-flexible/src/configuration/objects/CubeUniform.h
```diff
@@ -73,20 +72,17 @@ class CubeUniform : public Object {
...
   void generate(std::vector<ParticleType> &particles) const override {
-    ParticleType particle = getDummyParticle(particles.size());
+    // Wrapper so that std::vector can be used as an AutoPas::ParticleContainer
+    auto particlesWrapper = autopasTools::PseudoContainer(particles);
+
+    using namespace autopas::utils::ArrayMath::literals;
+    const auto boxMax = _bottomLeftCorner + _boxLength;
 
-    // Set up random number generation
-    std::random_device randomDevice;
```
schuhmaj, line 79: This is less an issue with the PR, but just something I noted:
* The legacy version here used the C++ style of generating random numbers
* The now employed `RandomGenerator` uses the C-style `rand()` function.

This could be resolved by:
- [ ] Adapt `double autopasTools::generators::RandomGenerator::fRand(double fMin, double fMax)` to also use the `std::uniform_distribution` and that the function takes a seed
- [ ] Remove the `srand()` calls from the `fillWithParticles` methods

It is not necessary for this PR, but it could be flagged as `TODO` at least. So, feel free to disregard it.

examples/md-flexible/src/configuration/objects/CubeGauss.h
```diff
@@ -92,22 +94,18 @@ class CubeGauss : public Object {
...
+    // Wrapper so that std::vector can be used as an AutoPas::ParticleContainer
+    auto particlesWrapper = autopasTools::PseudoContainer(particles);
+
+    using namespace autopas::utils::ArrayMath::literals;
+    const auto boxMax = _bottomLeftCorner + _boxLength;
+
+    // dummy particle used as a template with id of the first newly generated one
+    const ParticleType dummyParticle = getDummyParticle(particles.size());
+
+    autopasTools::generators::GaussianGenerator::fillWithParticles(particlesWrapper, _bottomLeftCorner, boxMax,
```
schuhmaj, line 106: This function in `GaussianGenerator` applies some error-checking:
```cpp
if (attempts == _maxAttempts) {
  std::ostringstream errormessage;
  errormessage << "GaussianGenerator::fillWithParticles(): Could not find a valid position for particle " << i
               << "after" << _maxAttempts << "attempts. Check if your parameters make sense:" << std::endl
               << "BoxMin       = " << autopas::utils::ArrayUtils::to_string(boxMin) << std::endl
               << "BoxMax       = " << autopas::utils::ArrayUtils::to_string(boxMax) << std::endl
               << "Gauss mean   = " << autopas::utils::ArrayUtils::to_string(distributionMean) << std::endl
               << "Gauss stdDev = " << autopas::utils::ArrayUtils::to_string(distributionStdDev) << std::endl;
  throw std::runtime_error(errormessage.str());
}
```

Assuming **this logic is correct**, then the `CubeGauss` object from the example `AllOptions.yaml` file with the subsequently listed properties

```yaml
CubeGauss:
  0:
    distribution-mean              :  [2, 2, 2]
    distribution-stddeviation      :  [2, 2, 2]
    numberOfParticles              :  100
    box-length                     :  [8, 8, 8]
    bottomLeftCorner               :  [15, 15, 15]
    velocity                       :  [0, 0, 0]
    particle-type-id               :  0
```

leads to the following error

```
libc++abi: terminating due to uncaught exception of type std::runtime_error: GaussianGenerator::fillWithParticles(): Could not find a valid position for particle 0after100attempts. Check if your parameters make sense:
BoxMin       = [15, 15, 15]
BoxMax       = [23, 23, 23]
Gauss mean   = [2, 2, 2]
Gauss stdDev = [2, 2, 2]
```

The check also fails with the particle number set to `1`, it only (trivially) succeeds with particle number set to `0`.
So, either
- [ ] The object specification is wrong --> Requires a change of all `example.yaml` files using the `GaussianGenerator`

or

- [ ] The checking algorithm is somehow broken --> Requires a modification of `GaussianGenerator::fillWithParticles`

schuhmaj (changes requested):
Good changes 👍

* One optional comment about the C-style `rand` function.
* One requested change affecting either the example yaml files or the `GaussianGenerator::fillWithParticles` method

examples/md-flexible/src/configuration/objects/CubeGauss.h
FG-TUM: The mean is outside of the box so the generator works as intended. Will fix the example.

schuhmaj: Oh indeed, uff, it seems my "Brain - Department for Math Affairs" has not yet been occupied this morning.

tools/autopasTools/generators/ClosestPackingGenerator.h
```diff
@@ -12,56 +12,53 @@
...
+ * Particle properties will be used from the default particle. Particle IDs start from the default particle.
+ * @tparam Container Arbitrary container class that needs to support addParticle().
+ * @param container
+ * @param boxMin
+ * @param boxMax
+ * @param defaultParticle
+ * @param spacing Distance between all neighboring particles
+ */
+template <class Container>
+static void fillWithParticles(Container &container, const std::array<double, 3> &boxMin,
```
schuhmaj: Very picky, but the `static` here is not necessary 🙃

tools/autopasTools/generators/GaussianGenerator.h
```diff
@@ -12,45 +12,35 @@
...
-   * bad
-   */
-  constexpr static size_t _maxAttempts = 100;
-};
+namespace autopasTools::generators::GaussianGenerator {
+/**
+ * Maximum number of attempts the random generator gets to find a valid position before considering the input to be
+ * bad
+ */
+constexpr static size_t _maxAttempts = 100;
```
schuhmaj: Here in theory, as well

schuhmaj (approved):
Perfect. Simply a speechless review

PR #931, opened 2024-08-13
Fix Loading MPI Checkpoints

FG-TUM:
# Description

- [x] Refactor loading of particles from checkpoints and generators
  - [x] When in doubt or using generators, load all particles into conf of rank 0
  - [x] Each rank sends all particles it can't insert locally to all other ranks
- [x] Multiple minor fixes
  - [x] Reserve before `push_back()` in deserialization
  - [x] `const`
  - [x] Remove redundant function `getSubdomainCount()`
  - [x] rename `isNear()` to `isNearRel()` to distinguish it from new `isNearAbs()`
- [x] Behavior changes:
  - [x] Force correct `OwnershipState` upon `addParticle` via `LogicHandler` (this will be cleaned up in #821)
  - [x] When writing checkpoints, prevent positions to be rounded to backside boundaries 73ab90b
  - [x] Print number of particles upon init per rank + total
  - [x] Restrict info logs to rank 0

## Resolved Issues

- [x] Loading parallel checkpoints (`pvtu`) with different domain boundaries would result in particle loss. This is almost always the case when using mpi load balancing.

# How Has This Been Tested?

- [x] Store and load MPI parallel checkpoints
- [x] Store and load serial checkpoints
- [x] Start MPI parallel simulation with objects

### DIFF
--- a/examples/md-flexible/input/explodingLiquid.yaml
+++ b/examples/md-flexible/input/explodingLiquid.yaml
@@ -18,7 +18,7 @@ box-max                          :  [15, 60, 15]
 cell-size                        :  [1]
 deltaT                           :  0.00182367
 iterations                       :  12000
-boundary-type                    : [periodic, periodic, periodic]
+boundary-type                    :  [periodic, periodic, periodic]
 Sites:
   0:
     epsilon                      :  1.
--- a/examples/md-flexible/input/explodingLiquid.yaml
+++ b/examples/md-flexible/input/explodingLiquid.yaml
@@ -34,8 +34,9 @@ Objects:
   CubeClosestPacked:
     0:
       particle-type-id           :  0
-      box-length                 :  [15, 6, 15]
-      bottomLeftCorner           :  [0, 27, 0]
+      box-length                 :  [14, 6, 14]
+      # Move particles half spacing away from domain border to prevent automatic border extension
+      bottomLeftCorner           :  [0.5, 27, 0.5]
       particle-spacing           :  1.
       velocity                   :  [0, 0, 0]
 no-end-config                    :  true
--- a/examples/md-flexible/src/ParallelVtkWriter.cpp
+++ b/examples/md-flexible/src/ParallelVtkWriter.cpp
@@ -150,14 +150,15 @@ void ParallelVtkWriter::recordParticleStates(size_t currentIteration,
     // considered to be not part of the domain, hence such a particle would not be loaded and thus be lost.
     // This function identifies these problematic values and raises the write precision just for this value high enough
     // to be distinguishable from the boundary.
-    const auto writeWithDynamicPrecision = [&](double d, double border) {
+    const auto writeWithDynamicPrecision = [&](double position, double border) {
       const auto initialPrecision = timestepFile.precision();
       // Simple and cheap check if we even need to do anything.
-      if (border - d < 0.1) {
+      if (border - position < 0.1) {
         using autopas::utils::Math::roundFloating;
         using autopas::utils::Math::isNearAbs;
         // As long as the used precision results in the two values being indistinguishable increase the precision
-        while (isNearAbs(roundFloating(d, timestepFile.precision()), border, std::pow(10, -timestepFile.precision()))) {
+        while (isNearAbs(roundFloating(position, timestepFile.precision()), border,
+                         std::pow(10, -timestepFile.precision()))) {
           timestepFile << std::setprecision(timestepFile.precision() + 1);
           // Abort if the numbers are indistinguishable beyond machine precision
           if (timestepFile.precision() > 20) {
@@ -165,12 +166,12 @@ void ParallelVtkWriter::recordParticleStates(size_t currentIteration,
                 "ParallelVtkWriter::writeWithDynamicPrecision(): "
                 "The two given numbers are identical up to 20 digits of precision!\n"
                 "Number: " +
-                std::to_string(d) + "\n" + particle->toString());
+                std::to_string(position) + "\n" + particle->toString());
           }
         }
       }
       // Write with the new precision and then reset it
-      timestepFile << d << std::setprecision(initialPrecision);
+      timestepFile << position << std::setprecision(initialPrecision);
     };
 
     const auto pos = particle->getR();
--- a/examples/md-flexible/src/ParallelVtkWriter.cpp
+++ b/examples/md-flexible/src/ParallelVtkWriter.cpp
@@ -162,10 +162,13 @@ void ParallelVtkWriter::recordParticleStates(size_t currentIteration,
                          std::pow(10, -timestepFile.precision()))) {
           timestepFile << std::setprecision(timestepFile.precision() + 1);
           // Abort if the numbers are indistinguishable beyond machine precision
-          if (timestepFile.precision() > std::numeric_limits<double>::digits10) {
+          constexpr auto machinePrecision = std::numeric_limits<double>::digits10;
+          if (timestepFile.precision() > machinePrecision) {
             throw std::runtime_error(
                 "ParallelVtkWriter::writeWithDynamicPrecision(): "
-                "The two given numbers are identical up to 20 digits of precision!\n"
+                "The two given numbers are identical up to " +
+                std::to_string(machinePrecision) +
+                " digits of precision!\n"
                 "Number: " +
                 std::to_string(position) + "\n" + particle->toString());
           }
--- a/examples/md-flexible/src/ParallelVtkWriter.cpp
+++ b/examples/md-flexible/src/ParallelVtkWriter.cpp
@@ -78,7 +78,7 @@ void ParallelVtkWriter::recordParticleStates(size_t currentIteration,
   timestepFile
       << "        <DataArray Name=\"velocities\" NumberOfComponents=\"3\" format=\"ascii\" type=\"Float32\">\n";
   for (auto particle = autoPasContainer.begin(autopas::IteratorBehavior::owned); particle.isValid(); ++particle) {
-    auto v = particle->getV();
+    const auto v = particle->getV();
     timestepFile << "        " << v[0] << " " << v[1] << " " << v[2] << "\n";
   }
   timestepFile << "        </DataArray>\n";
@@ -86,7 +86,7 @@ void ParallelVtkWriter::recordParticleStates(size_t currentIteration,
   // print forces
   timestepFile << "        <DataArray Name=\"forces\" NumberOfComponents=\"3\" format=\"ascii\" type=\"Float32\">\n";
   for (auto particle = autoPasContainer.begin(autopas::IteratorBehavior::owned); particle.isValid(); ++particle) {
-    auto f = particle->getF();
+    const auto f = particle->getF();
     timestepFile << "        " << f[0] << " " << f[1] << " " << f[2] << "\n";
   }
   timestepFile << "        </DataArray>\n";
@@ -96,7 +96,7 @@ void ParallelVtkWriter::recordParticleStates(size_t currentIteration,
   timestepFile
       << "        <DataArray Name=\"quaternions\" NumberOfComponents=\"4\" format=\"ascii\" type=\"Float32\">\n";
   for (auto particle = autoPasContainer.begin(autopas::IteratorBehavior::owned); particle.isValid(); ++particle) {
-    auto q = particle->getQuaternion();
+    const auto q = particle->getQuaternion();
     timestepFile << "        " << q[0] << " " << q[1] << " " << q[2] << " " << q[3] << "\n";
   }
   timestepFile << "        </DataArray>\n";
@@ -105,7 +105,7 @@ void ParallelVtkWriter::recordParticleStates(size_t currentIteration,
   timestepFile
       << "        <DataArray Name=\"angularVelocities\" NumberOfComponents=\"3\" format=\"ascii\" type=\"Float32\">\n";
   for (auto particle = autoPasContainer.begin(autopas::IteratorBehavior::owned); particle.isValid(); ++particle) {
-    auto angVel = particle->getAngularVel();
+    const auto angVel = particle->getAngularVel();
     timestepFile << "        " << angVel[0] << " " << angVel[1] << " " << angVel[2] << "\n";
   }
   timestepFile << "        </DataArray>\n";
@@ -113,7 +113,7 @@ void ParallelVtkWriter::recordParticleStates(size_t currentIteration,
   // print torques
   timestepFile << "        <DataArray Name=\"torques\" NumberOfComponents=\"3\" format=\"ascii\" type=\"Float32\">\n";
   for (auto particle = autoPasContainer.begin(autopas::IteratorBehavior::owned); particle.isValid(); ++particle) {
-    auto torque = particle->getTorque();
+    const auto torque = particle->getTorque();
     timestepFile << "        " << torque[0] << " " << torque[1] << " " << torque[2] << "\n";
   }
   timestepFile << "        </DataArray>\n";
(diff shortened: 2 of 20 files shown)


examples/md-flexible/input/explodingLiquid.yaml
```diff
@@ -34,8 +34,9 @@ Objects:
   CubeClosestPacked:
     0:
       particle-type-id           :  0
-      box-length                 :  [15, 6, 15]
-      bottomLeftCorner           :  [0, 27, 0]
+      box-length                 :  [14, 6, 14]
+      # Move particles half spacing away from domain border to prevent automatic border extension
+      bottomLeftCorner           :  [0.5, 27, 0.5]
```
manishmishra6016, line 39: why is this required? and how do we communicate this to user whether or not this is required?

examples/md-flexible/src/configuration/MDFlexConfig.h
```diff
@@ -125,12 +125,6 @@ class MDFlexConfig {
    */
   void calcSimulationBox();
 
-  /**
-   * Returns the particles generated based on the provided configuration file.
-   * @return a vector containing the generated particles.
-   */
-  std::vector<ParticleType> getParticles() { return _particles; }
-
```
manishmishra6016, line 133: Was there a reason to not use the getter but instead make the vector public?

manishmishra6016 (approved):
I just have a couple of clarifications, but other than that, it looks good to me. :)

examples/md-flexible/input/explodingLiquid.yaml
FG-TUM: > Why is this required?

When you specify any object in md-flex the domain is enlarged to fit it if required. For e.g. the grid generator, the actual box size is slightly larger than the grid box (namely half the spacing in each direction) so that it is perfectly aligned with its periodic images.

Usually all of this happens automatically no action required. However if you restart form a checkpoint you have to set the box size manually. I slightly moved the generator box so that the actual domain starts at `[0,0,0]` vs `[-.5, -.5, -.5]` as before.

> Who do we communicate this?

When a simulation starts it lists the parameters with which it started. This includes the new domain box. Again there is no specific action needed just careful input.

examples/md-flexible/src/configuration/MDFlexConfig.h
FG-TUM: Usually getters should be `const` methods (so this getter was already misleading) that only provide read access. Write access usually should happen via other methods (setters, dedicated functions etc....) . However, here, I need full read and write access to the actual vector. Hence, I made it public because it is treated as such.

schuhmaj (approved):
Fine 👍

examples/md-flexible/src/Simulation.cpp
```diff
@@ -631,6 +643,62 @@ void Simulation::checkNumParticles(size_t expectedNumParticlesGlobal, size_t num
...
+  for (int senderRank = 0; senderRank < _domainDecomposition->getSubdomainCount(); ++senderRank) {
+    // don't send to ourselves
+    if (senderRank == rank) {
+      continue;
+    }
+    particleCommunicator.receiveParticles(_configuration.particles, senderRank);
+  }
+  particleCommunicator.waitForSendRequests();
+
+  // Remove duplicates. This is a safety mechanism for the odd case that two ranks loaded the same particle.
```
SamNewcome, line 689: What might be a reason for this to occur?

Furthermore, this appears to only protect against a duplicate particle coming from 2 other ranks but not if the duplicate comes from this rank and one other.

examples/md-flexible/src/configuration/MDFlexConfig.cpp
```diff
@@ -547,12 +556,18 @@ void MDFlexConfig::loadParticlesFromCheckpoint(const size_t &rank, const size_t
...
+  // If the checkpoint was written by a MPI parallel simulation the checkpoint is a pvtu file,
+  // that points to several vtu files, called pieces.
+  size_t numCheckpointPieces{getNumPiecesInCheckpoint(filename)};
+  // If the number of currently used ranks matches the number of pieces in the checkpoint,
+  // we assume that the domain layout matches and each rank only loads their own piece.
+  // If it doesn't match the overlap will be redistributed when particles are inserted into the AutoPas containers.
+  if (numRanks == numCheckpointPieces) {
+    loadParticlesFromRankRecord(filename, rank, particles);
+  } else if (rank == 0) {
+    // Otherwise, each rank 0 loads everything and distributes it later.
```
SamNewcome, line 568: I assume "each" is a typo:
```suggestion
    // Otherwise, rank 0 loads everything and distributes it later.
```

SamNewcome (changes requested):

examples/md-flexible/src/Simulation.cpp
FG-TUM: > What might be a reason for this to occur?

Honestly no idea. I don't think this can happen, hence I removed this check -> bfc0bdf2d

examples/md-flexible/src/ParallelVtkWriter.cpp
```diff
@@ -139,9 +139,47 @@ void ParallelVtkWriter::recordParticleStates(size_t currentIteration,
...
+    // we are fine with the rather low default precision.
+    // However, if a particle is very close to the domain border it can happen that the particle position is rounded
+    // exactly to the boundary position. This then causes problems when the checkpoint is loaded because boxMax is
+    // considered to be not part of the domain, hence such a particle would not be loaded and thus be lost.
+    // This function identifies these problematic values and raises the write precision just for this value high enough
+    // to be distinguishable from the boundary.
+    const auto writeWithDynamicPrecision = [&](double d, double border) {
+      const auto initialPrecision = timestepFile.precision();
+      // Simple and cheap check if we even need to do anything.
+      if (border - d < 0.1) {
```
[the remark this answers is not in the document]
FG-TUM: by default values are written with floating point precision of 6 digits to the vtk file. Thus, the "problematic delta" can be arbitrarily large and I wanted to ignore the more expensive logic further down.

Since the check is very arbitrary indeed we can also get rid of it because it has the potential to cause problems for domains that extend to ~1e6.

examples/md-flexible/src/ParallelVtkWriter.cpp
```diff
@@ -139,9 +139,47 @@ void ParallelVtkWriter::recordParticleStates(size_t currentIteration,
...
+        // As long as the used precision results in the two values being the same increase it
+        while (roundFloating(d, timestepFile.precision()) == border) {
+          timestepFile << std::setprecision(timestepFile.precision() + 1);
+          // Abort if the numbers are indistinguishable
+          if (timestepFile.precision() > 20) {
+            throw std::runtime_error(
+                "ParallelVtkWriter::writeWithDynamicPrecision(): "
+                "The two given numbers are identical up to 20 digits of precision!\n"
+                "Number: " +
+                std::to_string(d) + "\n" + particle->toString());
```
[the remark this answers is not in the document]
FG-TUM: My reasoning for throwing an error here was that, in theory, this should never happen and thus is an indicator for something worse.

If the double values are really that similar they are basically the same as `double` is only precise up to 16 digits. And this then means that something in our boundary treatment went wrong.

examples/md-flexible/src/ParallelVtkWriter.cpp
```diff
@@ -139,9 +139,47 @@ void ParallelVtkWriter::recordParticleStates(size_t currentIteration,
...
+    // considered to be not part of the domain, hence such a particle would not be loaded and thus be lost.
+    // This function identifies these problematic values and raises the write precision just for this value high enough
+    // to be distinguishable from the boundary.
+    const auto writeWithDynamicPrecision = [&](double d, double border) {
+      const auto initialPrecision = timestepFile.precision();
+      // Simple and cheap check if we even need to do anything.
+      if (border - d < 0.1) {
+        using autopas::utils::Math::roundFloating;
+        // As long as the used precision results in the two values being the same increase it
+        while (roundFloating(d, timestepFile.precision()) == border) {
```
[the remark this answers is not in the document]
FG-TUM: As far as I understand it, this can only become a problem if the border is a value that is an infinite sequence that can not be represented exactly by double like e.g. `1/3`. Finite values that can not be represented like `0.3` are fine.

~I'm not sure we can somehow come into this position for the total simulation box.~ Actually md-flex and ls1 can't (the parsers don't support this) but theoretically some other other program could set this.
For inner mpi-boundaries this doesn't matter because particles then are simply loaded into the neighbor rank.

I'll fix it.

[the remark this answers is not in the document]
FG-TUM: -> 3a165ca0d

examples/md-flexible/src/ParallelVtkWriter.cpp
```diff
@@ -139,9 +139,47 @@ void ParallelVtkWriter::recordParticleStates(size_t currentIteration,
...
+    // When we write to the file in ASCII, values are rounded to the precision of the filestream.
+    // Since a higher precision results in larger files because more characters are written,
+    // and mdflex is not intended as a perfectly precice tool for application scientists,
+    // we are fine with the rather low default precision.
+    // However, if a particle is very close to the domain border it can happen that the particle position is rounded
+    // exactly to the boundary position. This then causes problems when the checkpoint is loaded because boxMax is
+    // considered to be not part of the domain, hence such a particle would not be loaded and thus be lost.
+    // This function identifies these problematic values and raises the write precision just for this value high enough
+    // to be distinguishable from the boundary.
+    const auto writeWithDynamicPrecision = [&](double d, double border) {
```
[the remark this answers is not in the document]
FG-TUM: done 277e0de6a

examples/md-flexible/src/ParallelVtkWriter.cpp
```diff
@@ -139,9 +139,49 @@ void ParallelVtkWriter::recordParticleStates(size_t currentIteration,
...
+      // Simple and cheap check if we even need to do anything.
+      if (border - position < 0.1) {
+        using autopas::utils::Math::roundFloating;
+        using autopas::utils::Math::isNearAbs;
+        // As long as the used precision results in the two values being indistinguishable increase the precision
+        while (isNearAbs(roundFloating(position, timestepFile.precision()), border,
+                         std::pow(10, -timestepFile.precision()))) {
+          timestepFile << std::setprecision(timestepFile.precision() + 1);
+          // Abort if the numbers are indistinguishable beyond machine precision
+          if (timestepFile.precision() > 20) {
```
schuhmaj, line 164: Magic number. I would argue that `20` should be some `constexpr` constant within either this `VtkWriter`, or globally within `md-fex` context, or a globally global variable 🌍 as `EPSILON` value for comparisons in the `autopas::utils::Math`.

schuhmaj (changes requested):
Still fine 👍

But the Magic Number `20` for the precision should probably be, to some degree, a global variable 🌍/ nothing hidden in a function.

examples/md-flexible/src/ParallelVtkWriter.cpp
FG-TUM: how about 5cf81f0

schuhmaj: Even better than introducing a new constant 👍

examples/md-flexible/src/ParallelVtkWriter.cpp
```diff
@@ -139,9 +140,49 @@ void ParallelVtkWriter::recordParticleStates(size_t currentIteration,
...
+        using autopas::utils::Math::isNearAbs;
+        // As long as the used precision results in the two values being indistinguishable increase the precision
+        while (isNearAbs(roundFloating(position, timestepFile.precision()), border,
+                         std::pow(10, -timestepFile.precision()))) {
+          timestepFile << std::setprecision(timestepFile.precision() + 1);
+          // Abort if the numbers are indistinguishable beyond machine precision
+          if (timestepFile.precision() > std::numeric_limits<double>::digits10) {
+            throw std::runtime_error(
+                "ParallelVtkWriter::writeWithDynamicPrecision(): "
+                "The two given numbers are identical up to 20 digits of precision!\n"
```
schuhmaj, line 168:
```suggestion
                "The two given numbers are identical given the floating point precision of the machine!\n"
```

schuhmaj (approved):
I give this a GO 👍 Nice solution.

Optionally, but probably desired, the error message in `examples/md-flexible/src/ParallelVtkWriter.cpp` should probably adapted for consistency, too.

examples/md-flexible/src/ParallelVtkWriter.cpp
FG-TUM: done afa1a91

SamNewcome (approved):
Nice

examples/md-flexible/src/ParallelVtkWriter.cpp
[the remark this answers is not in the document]
SamNewcome: Ah I didn't realise that this is only used with `boxMax`

examples/md-flexible/src/ParallelVtkWriter.cpp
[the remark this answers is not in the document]
SamNewcome: Fair

examples/md-flexible/src/ParallelVtkWriter.cpp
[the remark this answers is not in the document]
FG-TUM: Here my code to reproduce my tests to check that this makes sense:

<details><summary><code>leCode.cpp</code></summary>
<p>

```cpp
#include <cmath>
#include <iomanip>
#include <iostream>
#include <vector>

double roundTo_math(double d, int precision) {
  const auto scale = std::pow(10, precision);
  return std::round(d * scale) / scale;
}

double roundTo_mathFloat(double d, int precision) {
  if (d == 0.0) return d;  // Special case for 0 to avoid log10 error
  const auto factor = std::pow(10, precision - std::ceil(std::log10(std::abs(d))));
  return std::round(d * factor) / factor;
}

double roundTo_ss(double d, int precision) {
  std::stringstream ss;
  ss << std::fixed << std::setprecision(precision);
  ss << d;
  double ret;
  ss >> ret;
  return ret;
}

double roundTo_ssstod(double d, int precision) {
  std::stringstream ss;
  ss << std::fixed << std::setprecision(precision);
  ss << d;
  return std::stod(ss.str());
}

double roundTo_sprintf(double d, int precision) {
  char buf[7];
  snprintf(buf, 7, "%.2f", d);
  return std::stod(buf);
}

bool isNear(double a, double b, double relativeDifference) {
  const auto greaterNumber = std::max(std::abs(a), std::abs(b));
  const auto absoluteDifference = relativeDifference * greaterNumber;
  const auto diff = std::abs(a - b);
  return diff <= absoluteDifference;
}

template<class X = int>
using vec = std::vector<X>;

int main() {
  const double border = 1/3.;
  const double d = border - 0.000000001;
  std::cout << "Naive"
            << "\n";
  std::cout << "Border            : " << border << "\n";
  std::cout << "operator<<        : " << d << "\n";
  std::cout << std::setprecision(20) << std::boolalpha;
  std::cout << "Precision 20"
            << "\n";
  std::cout << "Border            : " << border << "\n";
  std::cout << "operator<<        : " << d << "\n";

  for (int precision = 6; precision < 25; ++precision) {
    std::cout << std::setprecision(precision);
    std::cout << "Precision : " << std::setw(2) << std::right << precision
              << " | roundTo_mathFloat : " << std::setw(26) << std::left << roundTo_mathFloat(d, precision)
              << " | is equal to border: " << std::setw(5) << std::right<< (roundTo_mathFloat(d, precision) == border)
              << " | is near border: " << std::setw(5) << std::right<< (std::abs(roundTo_mathFloat(d, precision) - border) < std::pow(10, -precision))
              << " | diff: " << std::setw(30) << std::right<< std::abs(roundTo_mathFloat(d, precision) - border)
              << " | max tolerance: " << std::setw(30) << std::right<< std::pow(10, -precision - 1) << "\n";
  }
}

```

</p>
</details>

PR #935, opened 2024-08-22
Verlet Cluster Lists Bugfix

### LINKED ISSUE #917: Verlet Cluster Lists appear to result in incorrect forces
**Describe the bug**
When running VCL with exploding liquid, there are explosions (as in problematic ones.

This scenario is the standard exploding liquid, with a verlet skin per timestep of 0.025. This was used, because in earlier runs with LInked Cells and Verlet List Cells, it was found frequently that particles might move ~0.011 per timestep. Running Linked Cells/Verlet List Cells with this new timestep resulted in no fast particle throws. 

Now, when Verlet Cluster Lists are used with this step-size, this verlet skin per timestep is no longer appropriate. The first error is
~~~bash
[2024-07-22 13:41:53.195] [AutoPasLog] [warning] [ParticleBase.h:171] Particle 10325: Distance between old and new position is larger than expected: 0.0007136447901825499 > 0.00015625000000000003
A particle moved farther than verletSkinPerTimestep/2: 0.0267141 > 0.025/2 = 0.0125
Particle
ID      : 10325
Position: [3.98776, 83.9955, 17.821]
Velocity: [3.46025, 4.87561, -4.74658]
Force   : [0, 0, 0]
~~~
meaning a particle was able to move significantly faster than expected without prior "moving faster"s. This indicates an error with VCL rather than the chosen verlet skin per timestep.

**To Reproduce**
Run current exploding liquid with only Verlet Cluster Lists C06 Aos (or SoA), newton3=enabled and a larger domain size (see yaml file)

**Expected behavior**
No warnings.

* OS: Linux
* Compiler: gcc11
* Build System: Make, CMake
* CoolMUC2

**Additional context**
Full yaml file:

~~~yaml
# This yaml file is for single-site molecular simulation. Uncomment the Molecules option to run this experiment using
# md-flexible compiled for multi-site molecules.
# container                        :  [LinkedCells, VarVerletListsAsBuild, VerletClusterLists, VerletLists, VerletListsCells, PairwiseVerletLists]
traversal                        :  [vcl_c06]
data-layout                      :  [AoS]
verlet-rebuild-frequency         :  10
verlet-skin-radius-per-timestep  :  0.025
selector-strategy                :  Fastest-Mean-Value
tuning-strategies                :  []
tuning-interval                  :  6000
tuning-samples                   :  10
functor                          :  Lennard-Jones
cutoff                           :  2
box-max                          :  [30, 120, 30]
deltaT                           :  0.00182367
iterations                       :  12000
boundary-type                    : [periodic, periodic, periodic]
Sites:
  0:
# Uncomment below to run a multi-site simulation.
#Molecules:
#  0:
#    site-types                   :  [ 0 ]
#    relative-site-positions      :  [ [0, 0, 0 ] ]
#    moment-of-inertia            :  [ 1., 1., 1. ]
Objects:
  CubeClosestPacked:
    0:
      box-length                 :  [30, 12, 30]
      bottomLeftCorner           :  [0, 54, 0]
      particle-spacing           :  1.
no-progress-bar                  :  true


~~~

dmartin13:
# Description

This PR is intended to fix a bug with the Verlet Cluster Lists. 
Particles were not sorted into the correct towers when recreating the neighborlists. The problem was that the order in which out-of-bound particles of the tower were collected and the resizing of the towerblock was wrong. The out-of-bound particles were collected first, then the tower block (and thus also the bounding boxes of the tower) were recalculated. However, the out-of-bounds particles should be collected only after the new tower bounding boxes have been calculated.

## Resolved Issues

- [x] fixes #917 

# How Has This Been Tested?

The scenario from #917 with a single thread was used. The results for VCL and linked cells were compared (particle interactions and resulting forces). A python script was used that calculates possible differences in the particle interactions from the log output of the calculation with both containers. After the bugfix, there are no more differences between the two containers. The same particle interactions and the same forces are calculated.

### DIFF
--- a/src/autopas/containers/verletClusterLists/VerletClusterListsRebuilder.h
+++ b/src/autopas/containers/verletClusterLists/VerletClusterListsRebuilder.h
@@ -86,8 +86,8 @@ class VerletClusterListsRebuilder {
     const auto numTowersNew = numTowersPerDim[0] * numTowersPerDim[1];
 
     // collect all particles that are now not in the right tower anymore
-    auto invalidParticles = collectOutOfBoundsParticlesFromTowers();
-    // collect all remaining particles that are not yet assigned to towers
+    std::vector<std::vector<Particle>> invalidParticles;
+    //  collect all particles that are not yet assigned to towers
     invalidParticles.push_back(std::move(_particlesToAdd));
     _particlesToAdd.clear();
     // if we have less towers than before, collect all particles from the unused towers.
@@ -98,6 +98,11 @@ class VerletClusterListsRebuilder {
     // resize to number of towers.
     // Attention! This uses the dummy constructor so we still need to set the desired cluster size.
     _towerBlock.resize(towerSideLength, numTowersPerDim);
+
+    // after resizing the towers we collect all the particles that are out of bounds
+    const auto collectedParticlesFromTowers = collectOutOfBoundsParticlesFromTowers();
+    invalidParticles.insert(invalidParticles.end(), collectedParticlesFromTowers.begin(),
+                            collectedParticlesFromTowers.end());
 
     // create more towers if needed and make an estimate for how many particles memory needs to be allocated
     // Factor is more or less a random guess.
--- a/src/autopas/containers/verletClusterLists/VerletClusterListsRebuilder.h
+++ b/src/autopas/containers/verletClusterLists/VerletClusterListsRebuilder.h
@@ -86,8 +86,14 @@ class VerletClusterListsRebuilder {
     const auto numTowersNew = numTowersPerDim[0] * numTowersPerDim[1];
 
     // collect all particles that are now not in the right tower anymore
-    std::vector<std::vector<Particle>> invalidParticles;
-    //  collect all particles that are not yet assigned to towers
+    std::vector<std::vector<Particle>> invalidParticles{};
+    // Reserve for:
+    //   - _particlesToAdd (+1)
+    //   - surplus towers (+max(0, numTowersNew - numTowersOld))
+    //   - particles out of bounds of new towers (+numTowersNew)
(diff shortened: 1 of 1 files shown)


src/autopas/containers/verletClusterLists/VerletClusterListsRebuilder.h
```diff
@@ -86,8 +86,8 @@ class VerletClusterListsRebuilder {
     const auto numTowersNew = numTowersPerDim[0] * numTowersPerDim[1];
 
     // collect all particles that are now not in the right tower anymore
-    auto invalidParticles = collectOutOfBoundsParticlesFromTowers();
-    // collect all remaining particles that are not yet assigned to towers
+    std::vector<std::vector<Particle>> invalidParticles;
+    //  collect all particles that are not yet assigned to towers
     invalidParticles.push_back(std::move(_particlesToAdd));
```
FG-TUM, line 91:
```suggestion
    std::vector<std::vector<Particle>> invalidParticles{};
    // Reserve for:
    //   - _particlesToAdd (+1)
    //   - surplus towers (+max(0, numTowersNew - numTowersOld))
    //   - particles out of bounds of new towers (+numTowersNew)
    invalidParticles.reserve(1 + std::max(0, static_cast<int>(numTowersOld) - static_cast<int>(numTowersNew)) + numTowersNew)
    // collect all particles that are not yet assigned to towers
    invalidParticles.push_back(std::move(_particlesToAdd));
```

- [x] formatting
- [x] reserve before `push_back` / `insert`
- [x] casts before substracting unsigned integers

...should be right I think?

FG-TUM (changes requested):
Very nice catch! While we are at it we should improve the memory allocation.

FG-TUM (approved):

PR #937, opened 2024-08-26
Initial Force Calculation for MD-Flexible

### LINKED ISSUE #816: Add initial force calculation to md-flexible
**Describe the bug**
There should be an initial force calculation before starting into the main simulation time loop. Currently, the old forces used for the first position and velocity updates are still 0.

dmartin13:
# Description

As described in #816, md-flexible lacks an initial force calculation for the Störmer Verlet algorithm. 
For the reasons given in the discussion in #816, an initial force calculation has again been omitted. In this PR, only a comment describing the reasons is inserted.

## Resolved Issues

- [x] fixes #816 

# Open Questions
~After introducing the initial force calculation, we have an additional tuning iteration. For example, if you run `./md-flexible` without any arguments, we end up with 11/10 tuning iterations. The question is, how do we want to handle this?
At the moment I have a bool (`_initialTuningIterationRecorded`) which indicates if the first force calculation was a tuning iteration. This is then logged separate from the other tuning iterations. But I'm not so happy with it. Better ideas are welcome.~

# How Has This Been Tested?

- only a comment has been added

### DIFF
--- a/docs/userdoc/SimulationLoop.md
+++ b/docs/userdoc/SimulationLoop.md
@@ -10,11 +10,6 @@ AutoPas assumes that any (short-range) particle simulation boils down to three h
 ## Code Example
 A conceptual example of a primary simulation loop:
 ```c++
-
-// Initial force calculation
-YourFunctor functor();                       // User code
-autopas->iteratePairwise(&functor);          // Tuning and parallelization happens here
-
 // main simulation loop
 while (needsMoreIterations()) {
    // 1. Particles movement as a result of interactions or simulation forces.
--- a/docs/userdoc/SimulationLoop.md
+++ b/docs/userdoc/SimulationLoop.md
@@ -10,6 +10,11 @@ AutoPas assumes that any (short-range) particle simulation boils down to three h
 ## Code Example
 A conceptual example of a primary simulation loop:
 ```c++
+
+// Initial force calculation
+YourFunctor functor();                       // User code
+autopas->iteratePairwise(&functor);          // Tuning and parallelization happens here
+
 // main simulation loop
 while (needsMoreIterations()) {
    // 1. Particles movement as a result of interactions or simulation forces.
--- a/examples/md-flexible/src/Simulation.cpp
+++ b/examples/md-flexible/src/Simulation.cpp
@@ -422,7 +422,11 @@ void Simulation::updateInteractionForces() {
   // count time spent for tuning
   if (_currentIterationIsTuningIteration) {
     _timers.forceUpdateTuning.addTime(timeIteration);
-    ++_numTuningIterations;
+    if (_iteration == 0 and (not _initialTuningIterationRecorded)) {
+      _initialTuningIterationRecorded = true;
+    } else {
+      ++_numTuningIterations;
+    }
   } else {
     _timers.forceUpdateNonTuning.addTime(timeIteration);
     // if the previous iteration was a tuning iteration and the current one is not
@@ -595,7 +599,7 @@ void Simulation::logMeasurements() {
 
     std::cout << "Tuning iterations                  : " << _numTuningIterations << " / " << _iteration << " = "
               << (static_cast<double>(_numTuningIterations) / static_cast<double>(_iteration) * 100.) << "%"
-              << "\n";
+              << (_initialTuningIterationRecorded ? " (+1 initial tuning)" : "") << "\n";
 
     auto mfups =
         static_cast<double>(_autoPasContainer->getNumberOfParticles(autopas::IteratorBehavior::owned) * _iteration) *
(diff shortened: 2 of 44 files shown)


docs/userdoc/SimulationLoop.md
```diff
@@ -10,6 +10,11 @@ AutoPas assumes that any (short-range) particle simulation boils down to three h
 ## Code Example
 A conceptual example of a primary simulation loop:
 ```c++
+
+// Initial force calculation
+YourFunctor functor();                       // User code
+autopas->iteratePairwise(&functor);          // Tuning and parallelization happens here
+
```
FG-TUM, line 17: I'd not add this here. 
1. The Functor you create here would shadow persist for the whole simulation ;)
2. Whether this is necessary or not depends on the kind of simulation and this document is intended to serve as a generic example of the necessary steps and where AutoPas is involved. Thus I'd keep everything else to a minimum to retain focus.

examples/md-flexible/src/Simulation.cpp
```diff
@@ -415,7 +422,11 @@ void Simulation::updateInteractionForces() {
   // count time spent for tuning
   if (_currentIterationIsTuningIteration) {
     _timers.forceUpdateTuning.addTime(timeIteration);
-    ++_numTuningIterations;
+    if (_iteration == 0 and (not _initialTuningIterationRecorded)) {
```
FG-TUM, line 425:
```suggestion
    if (_iteration == 0 and not _initialTuningIterationRecorded) {
```

FG-TUM (changes requested):
The initial force update also has to consider potential global forces.

SamNewcome (commented):
An alternative to your handling of the first force calculation:

Move initial force calculation to initialisation (+ included in initialisation time etc.)

Run `iteratePairwise` with a not relevant for tuning version of the force functor, and don't include this in any iteration statistics. Two caveats to this:

- `relevantForTuining` is a template parameter which leads us to template hell. But I also don't see any reason why this needs to be a template parameter so it could be removed.
- I have no idea how AutoPas behaves in the case that tuning is not trivial but the first tuning phase is relevant for tuning. Something somewhat sensible should happen here (I don't really care about optimal)


This is more work, but in my opinion much cleaner. I'm happy to see if others (@FG-TUM @manishmishra6016 @thesamriel @schuhmaj ) feel if this is worth it or not, as I'm not too opinionated on this.

examples/md-flexible/src/Simulation.cpp
```diff
@@ -178,6 +178,22 @@ void Simulation::finalize() {
...
+// The initial force calculation is omitted here for the following reasons:
+// 1. md-flexible is a test application to demonstrate how AutoPas can be used and not a software for physically
+// correct simulation.
+// 2. md-flexible is often used during development for performance measurements or memory analysis. An additional
+// force calculation step would unnecessarily complicate the simulation loop and distort the statistics.
+// 3. We also do not need an initial force calculation or the oldForce for checkpoint loading. Reason: In the last
+// simulation step before saving the checkpoint, F was calculated and a velocity update was performed with this F. In
+// the first iteration after loading the checkpoint file, the position update is calculated with this F (loaded from the
+// checkpoint file) from the last simulation step, then F_old is set to F and the new force is calculated. This
+// automatically continues the Störmer Verlet algorithm seamlessly.
```
FG-TUM, line 196:
```suggestion
// Note: An initial force calculation should actually take place before the first position update, according to
// Störmer Verlet the literature (Griebel et al., Numerical simulation in molecular dynamics: numerics, algorithms,
// parallelization, applications, 2007). In this implementation, the first position update is calculated with zero
// forces, if none are given initially, and only the velocity is used. This means that, essentially, the molecules do not start the simulation at
// the given positions but are already moved by one timestep in the direction of their initial velocity. If you want
// to observe the exact movement of particles, you should bear this in mind.
// The initial force calculation is omitted here for the following reasons:
// 1. The primary focus for md-flexible is to be a test application and demonstrator of how AutoPas can be used. Not to be a perfect simulator.
// 2. md-flexible is often used during development for performance measurements or memory analysis. An additional
// force calculation step would unnecessarily complicate the simulation loop and distort the statistics.
// 3. We also do not need an initial force calculation or the oldForce for checkpoint loading. Reason: In the last
// simulation step before saving the checkpoint, F was calculated, and a velocity update was performed with this F. In
// the first iteration after loading the checkpoint file, the position update is calculated with this F (loaded from the
// checkpoint file) from the last simulation step, then F_old is set to F and the new force is calculated. This
// automatically continues the Störmer Verlet algorithm seamlessly.
```

(might need reformatting after my suggestions)

I'd prefer if this were moved to the doxygen doc block in `Simulation.h` as a `@note`. Please make sure that the whole block is then also rendered as a note and not only the first sentence by building it locally.

FG-TUM (changes requested):

examples/md-flexible/src/Simulation.cpp
dmartin13: Do you want the comment directly above the `class Simulation` as note, or in the doxygen doc of `void run()`?

FG-TUM: Good question. I'd put it in the doc for `run()`

FG-TUM (approved):

PR #943, opened 2024-09-03
Fixed Tuning Intervals

### LINKED ISSUE #942: Non-Tuning-Phase length behavior
Currently, the number of iterations between two tuning phases, which we call the tuning frequency, is a fixed number of iterations. These iterations are counted after the last tuning iteration.
This means that when a tuning strategy is employed where the number of tuning iterations is not known in advance, it is hard to estimate when the next tuning phase will start.

The number of tuning iterations is subtracted from the non-tuning iterations. Meaning if we specify a tuning frequency of e.g. 1000 tuning will always start at iteration 0, 1000, 2000, ...

- [ ] ~Option A: Leave everything as it is. (meh)~
- [x] Option B: Implement the suggested behavior.
- [ ] ~Option C: Implement the suggested behavior without replacing the old one and provide an option to switch between the two.~

**Additional context**
This behavior would help to synchronize tuning and MPI load balancing.

dmartin13:
# Description
So far, after each tuning phase, a predefined number of iterations was waited until the next tuning phase started. This makes it difficult to predict when the next tuning phase will start if the number of iterations in a tuning phase is not known.
In this PR this is changed so that a tuning phase always starts after a fixed number of iterations. (i.e. whenever `iteration % tuningInteval == 0`).

For this purpose, the counting does not start from the end of each tuning phase until the next one, but from the beginning of each tuning phase.

If a tuning phase takes more iterations than the specified `tuningInterval`, an exception is thrown.

## Resolved Issues

- [x] fixes #942 

# How Has This Been Tested?

- [x] Existing tests
- [x] New testcase `testTuningIntervalIsFixed` that checks if tuning always start at fixed intervals.
- [x] New testcase `testTuningPhaseLongerThanTuningInterval` that checks the behavior of tuning phases which are longer than the `tuningInterval`
- [ ] The coverage report shows all relevant lines are tested.


src/autopas/tuning/AutoTuner.cpp
```diff
@@ -146,23 +149,23 @@ bool AutoTuner::tuneConfiguration() {
 
   // CASE: End of a tuning phase. This is not exclusive to the other cases!
   if (_configQueue.empty()) {
-    // if the queue is empty we are done tuning.
-    _iterationsSinceTuning = 0;
+    //  if the queue is empty we are done tuning.
```
FG-TUM, line 152:
```suggestion
    // If the queue is empty, we are done tuning.
```

src/autopas/tuning/AutoTuner.cpp
```diff
@@ -146,23 +149,23 @@ bool AutoTuner::tuneConfiguration() {
...
 
-  return stillTuning;
+  return _isTuning;
 }
 
 const Configuration &AutoTuner::getCurrentConfig() const { return _configQueue.back(); }
 
 std::tuple<Configuration, bool> AutoTuner::getNextConfig() {
-  // If we are not (yet) tuning or there is nothing to tune return immediately.
+  //  If we are not (yet) tuning or there is nothing to tune return immediately.
```
FG-TUM, line 168:
```suggestion
  // If we are not (yet) tuning or there is nothing to tune, return immediately.
```
Are the double spaces intentional?

src/autopas/tuning/AutoTuner.cpp
```diff
@@ -280,6 +283,14 @@ void AutoTuner::addMeasurement(long sample, bool neighborListRebuilt) {
 void AutoTuner::bumpIterationCounters() {
   ++_iteration;
   ++_iterationsSinceTuning;
+  if (inTuningPhase()) {
+    ++_iterationsInLastTuningPhase;
+    if (_iterationsInLastTuningPhase > _tuningInterval) {
+      throw autopas::utils::ExceptionHandler::AutoPasException(
+          "Tuning needs more iterations than the tuning interval. Please use a higher value for the tuning interval.");
```
FG-TUM, line 290:
```suggestion
      utils::ExceptionHandler::exception(
          "Tuning needs more iterations than the tuning interval ({}). Please use a higher value for the tuning interval.", _tuningInterval);
```
The wrapper function provides formatting.

tests/testAutopas/tests/tuning/AutoTunerTest.cpp
```diff
@@ -169,6 +169,48 @@ TEST_F(AutoTunerTest, testAllConfigurations) {
...
+  size_t iterations = 0;
+  size_t tuningIterations = 0;
+  while (iterations < numIterations) {
+    stillTuning = logicHandler.iteratePairwisePipeline(&functor);
+    if (not lastWasTuningInteration and stillTuning) {
+      EXPECT_TRUE(iterations % tuningInterval == 0) << "Tuning phase does at start at fixed iteration number.";
+    }
+    ++iterations;
+    lastWasTuningInteration = stillTuning ? true : false;
+  }
```
FG-TUM, line 211: please also add a container update because @thesamriel might move the iteration counting logic from `iteratePairwise()` to `updateContainer()` at some point.

src/autopas/tuning/AutoTuner.h
```diff
@@ -382,5 +382,21 @@ class AutoTuner {
...
   TuningDataLogger _tuningDataLogger;
+
+  /**
+   * Is set to true during a tuning phase.
+   */
+  bool _isTuning{false};
+
+  /**
+   * Is only set to true in the iteration where a tuning phase ends. This is done in tuneConfiguration() and set back to
+   * false in bumpIterationCounters()
```
FG-TUM, line 393: This means that from the moment the optimal configuration until the next bump of counters this is true?
Then I suggest this formulation:
```suggestion
   * Is only set to true for the last iteration of a tuning phase ends.
   * Specifically, from when tuneConfiguration() selects the optimum until it is reset to false in bumpIterationCounters().
```

src/autopas/tuning/AutoTuner.cpp
```diff
@@ -61,8 +61,8 @@ void AutoTuner::addHomogeneityAndMaxDensity(double homogeneity, double maxDensit
 
 void AutoTuner::logIteration(const Configuration &conf, bool tuningIteration, long tuningTime) {
   // only log if we are at the end of a tuning phase
-  if (not tuningIteration and _iterationsSinceTuning == 0) {
-    // This string is part of several older scripts, hence it is not recommended to change it.
+  if (_endOfTuningPhase) {
+    //  This string is part of several older scripts, hence it is not recommended to change it.
```
FG-TUM, line 65:
```suggestion
    // This string is part of several older scripts, hence it is not recommended to change it.
```

FG-TUM (changes requested):
I'm not fully through the logic yet but here are already some thoughts

src/autopas/tuning/AutoTuner.cpp
thesamriel, line 290: Would it be possible to handle this case with a warning instead of an error? Or would this mess with all the counters too much?

src/autopas/tuning/AutoTuner.h
```diff
@@ -382,5 +382,21 @@ class AutoTuner {
...
+   */
+  bool _endOfTuningPhase{false};
+
+  /**
+   * Stores the number of iterations the previous tuning phase used. Is reset at the beginning of a new tuning phase.
+   */
+  size_t _iterationsInLastTuningPhase{0};
```
thesamriel, line 400: Personally, I would like to rename this variable to be a bit more concise.
```suggestion
  size_t _lastTuningIterations{0};
```
Some others are all out for verbose variable names, so I think both is fine ;)

src/autopas/tuning/AutoTuner.cpp
FG-TUM: And how should AutoPas behave in such a case? 
- Restart the tuning? -> This means we'll never converge or test everything.
- Continue the tuning? -> When should then the next tuning phase start?

In my opinion, this might, in theory, be a recoverable failure, but in reality, this is completely undesirable, and I would want to start over with the simulation.

thesamriel: I would want continue the tuning.
The relevant case would be a tuning strategy, where the first tuning phase takes significantly more iterations than the subsequent phases.
Say `tuningInterval = 500`, tuning phase 1 takes 510 iterations due to a crazy large search space. Search space for the following tuning phases is reduced such that tuning phase 2 takes only 100 iterations, and tuning phase 3 only 20.

If this is not trivial to handle or totally undesirable, I'm fine to leave it

thesamriel: I would be fine with both options really, whatever would be easier to implement. 
Potentially, I like the skipping one tuning phase option more.

My reason to have it as recoverable 'failure', would be to make it more robust to (unexperienced) user input . If you are particularly interested in the tuning process, then this can lead to confusion of course. But then, you should also be able to use sensible values for the tuning interval imo.

FG-TUM: I think this touches upon the general question how we should treat the case that a tuning phase doesn't start when we expect it. There is also `forceRetune()` with which you can start a tuning phase at any point. If tuning phases are strictly bound to specific iteration intervals then these manually triggered tuning phases can easily run into them.

So let's go for the following:
- Tuning phases always start at `iteration % tuningInterval`, except if we are already in a tuning phase.
- If we are tuning and a new phase is triggered 
  - by the tuning interval simply continue tuning. We can consider this as "merging" two tuning phases.
  - manually via `forceRetune()` abort the current phase and start a new tuning phase. The rationale here is that there must be a reason why we want to start over, e.g. drastic change in the simulation.
- All these cases should throw a meaningful warning.

FG-TUM:
The doc string of (at least) `AutoTuner::_tuningInterval` has to be updated :)

src/autopas/tuning/AutoTuner.cpp
dmartin13: should be implemented now in [01e8856](https://github.com/AutoPas/AutoPas/pull/943/commits/01e8856c60f0e56e2bd2f71935a8f83e2da79c1f)

src/autopas/tuning/AutoTuner.cpp
```diff
@@ -103,8 +107,11 @@ bool AutoTuner::tuneConfiguration() {
   };
 
   // Determine where in a tuning phase we are
-  if (_iterationsSinceTuning == _tuningInterval) {
-    // CASE: Start of a tuning phase
+  if (_iterationsSinceTuning == _tuningInterval and not(_lastTuningIterations >= _tuningInterval)) {
+    // CASE: Start of a tuning phase (if _lastTuningIterations > _tuningInterval the current tuning phase takes more
+    // iterations tahn the tuning interval -> continue tuning)
```
FG-TUM, line 112:
```suggestion
  // if _lastTuningIterations >= _tuningInterval the current tuning phase takes more iterations than the tuning interval
  // -> continue tuning
  if (_iterationsSinceTuning == _tuningInterval and not(_lastTuningIterations >= _tuningInterval)) {
    // CASE: Start of a tuning phase
```
Just some reformatting

src/autopas/tuning/AutoTuner.h
FG-TUM: Correct me if I'm wrong, but to me, it looks like this counter tracks the number of iterations from the most recent start of a tuning phase until the most recent end. Thus, this is not necessarily the last tuning phase but, depending on the situation the current one.

I'd maybe rename it to something like:
- `_iterationsInTuningPhase`
- `_iterationsInCurrentTuningPhase`
- `_iterationsInMostRecentTuningPhase`

(I'd prefer it to start with `_iterations` so it is more consistent with the other iteration counters)
Yes, those are longer names, but they capture better the meaning.

Also: We need to update the doc of the variable.

src/autopas/tuning/AutoTuner.cpp
```diff
@@ -290,7 +307,13 @@ void AutoTuner::bumpIterationCounters() {
 bool AutoTuner::willRebuildNeighborLists() const {
   const bool inTuningPhase = this->inTuningPhase();
   // How many iterations ago did the rhythm of rebuilds change?
-  const auto iterationBaseline = inTuningPhase ? (_iterationsSinceTuning - _tuningInterval) : _iterationsSinceTuning;
+  // If we are in a tuning phase and in the first tuning iteration but tuneConfiguration has not been called yet
+  // (_iterationsSinceTuning == _tuningInterval) we have 0 as baseline, otherwise the baseline are the iterations we
+  // already did in this tuning phase. If we are not in a tuning phase we subtract the the number of iterations from the
+  // previous tuning phase to get the baseline
```
FG-TUM, line 313:
```suggestion
  // If we are in a tuning phase and in the first tuning iteration but tuneConfiguration has not been called yet
  // (_iterationsSinceTuning == _tuningInterval) we have 0 as the baseline, otherwise, the baseline is the iterations we
  // already did in this tuning phase. If we are not in a tuning phase, we subtract the number of iterations from the
  // previous tuning phase to get the baseline
```

src/autopas/tuning/AutoTuner.cpp
```diff
@@ -280,6 +287,16 @@ void AutoTuner::addMeasurement(long sample, bool neighborListRebuilt) {
 void AutoTuner::bumpIterationCounters() {
   ++_iteration;
   ++_iterationsSinceTuning;
+  if (inTuningPhase()) {
+    ++_lastTuningIterations;
+    if (_lastTuningIterations == _tuningInterval + 1) {
+      // The tuning interval triggered a new tuning phase during a running tuning phase -> continue tuning ("merge" of
+      // two tuning phases)
```
FG-TUM, line 294:
```suggestion
      // The tuning interval triggered a new tuning phase during a running tuning phase.
      // -> continue tuning ("merge" of two tuning phases)
```

FG-TUM (changes requested):

src/autopas/tuning/AutoTuner.h
dmartin13: I decided to go with `_iterationsInMostRecentTuningPhase`. Done in [24bbc12](https://github.com/AutoPas/AutoPas/pull/943/commits/24bbc12ae21f2cf9b58ccb144927c54eaa09c53e)

src/autopas/tuning/AutoTuner.cpp
dmartin13: done in [b9e4111](https://github.com/AutoPas/AutoPas/pull/943/commits/b9e4111c93ce42af332b045755214316b102b040)

src/autopas/tuning/AutoTuner.cpp
dmartin13: done in [b9e4111](https://github.com/AutoPas/AutoPas/pull/943/commits/b9e4111c93ce42af332b045755214316b102b040)

src/autopas/tuning/AutoTuner.cpp
dmartin13: done in [b9e4111](https://github.com/AutoPas/AutoPas/pull/943/commits/b9e4111c93ce42af332b045755214316b102b040)

FG-TUM (changes requested):

FG-TUM (approved):

PR #949, opened 2024-09-17
removing distance check for zero skin case

manishmishra6016:
# Description

This PR contains two main changes:
- Currently the distance check in `ParticleBase.h` in `setRDistanceCheck()` doesn't consider the case of skin=0. In that case, it always gives out a WARN statement (which affects performance if compiled with logger, even when logger is turned off at runtime). Zero skin condition is handled here.
- In `LogicHandler.h`, we already add a haloParticle to the buffer, so adding ownership again after it has been added is not needed. And so, this is removed.

## Related Pull Requests

- #931 ownership set in LogicHandler

# How Has This Been Tested?

Verified by running cases with zero skin.

### DIFF
--- a/src/autopas/particles/ParticleBase.h
+++ b/src/autopas/particles/ParticleBase.h
@@ -169,7 +169,8 @@ class ParticleBase {
     const auto distanceVec = r - _r;
     const double distanceSquared = utils::ArrayMath::dot(distanceVec, distanceVec);
     setR(r);
-    const bool distanceIsFine = distanceSquared < maxDistSquared or maxDistSquared == 0;
+    const bool distanceIsFine =
+        distanceSquared < maxDistSquared or autopas::utils::Math::isNearAbs(maxDistSquared, 0., 1e-12);
     if (not distanceIsFine) {
       AutoPasLog(WARN, "Particle {}: Distance between old and new position is larger than expected: {} > {}", _id,
                  distanceSquared, maxDistSquared);
--- a/src/autopas/particles/ParticleBase.h
+++ b/src/autopas/particles/ParticleBase.h
@@ -169,7 +169,7 @@ class ParticleBase {
     const auto distanceVec = r - _r;
     const double distanceSquared = utils::ArrayMath::dot(distanceVec, distanceVec);
     setR(r);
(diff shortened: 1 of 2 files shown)


src/autopas/particles/ParticleBase.h
```diff
@@ -169,7 +169,7 @@ class ParticleBase {
     const auto distanceVec = r - _r;
     const double distanceSquared = utils::ArrayMath::dot(distanceVec, distanceVec);
     setR(r);
-    const bool distanceIsFine = distanceSquared < maxDistSquared;
+    const bool distanceIsFine = distanceSquared < maxDistSquared or maxDistSquared == 0;
```
SamNewcome, line 172: I'm concerned about the use of `==`. As `maxDistSquared` is a calculated value in md-flexible (as opposed to just being assigned to zero), with might not exactly be zero even if it is supposed to be. Even if we could forcibly assign this as zero in the case of md-flexible, we couldn't do this for every code. I would use `isNearAbs` or, if you are worried about the cost, a simple `maxDistSquared < 1e-12` is probably enough(?).

FG-TUM (commented):
I agree with Sam's comment

src/autopas/particles/ParticleBase.h
manishmishra6016: @SamNewcome @FG-TUM 
I want to go with 'maxDistSquared < 1e-12' because I am comparing to zero, I am confident it is positive (square) and it would be slightly faster if number of particles in simulation is large. Is this fine?

I could use `isNearAbs()` alternatively.

SamNewcome: I'm fine with using a simple `maxDistSquared < 1e-12`. I'm slightly concerned that 1e-12 is not necessarily always going to be suitable. E.g. if the simulation was using units/scales such that 1e-12 was actually an expected amount to move in a timestep. I am however somewhat okay with this in the context that we aren't going to do anything like this in the short term and in the long term we hope to replace this mechanic.

manishmishra6016: Ok, fair point :) done!

FG-TUM (approved):

PR #950, opened 2024-09-19
Limit the Number of Spacial Locks

FG-TUM:
# Description

## Problem
`LogicHandler` sets up a bunch of locks for the parallelized remainder traversal between buffers and the container to avoid race conditions. The number of these locks depends on the domain size and the interaction length.

In a perfect world, we could calculate a virtual 3D grid with mesh width `interactionLength`, have one lock per grid cell and be happy. Alas, the world is shit.
If the domain is very large and/or the interaction length very small the number of locks skyrockets and allocating space for a few billion locks just is not good.

## Solution
This PR introduces a mechanism to limit the number of locks in a geometry-aware way.
The limit is set to `1e6` so our usual simulations should not be affected at all. E.g. the limit is _just_ reached if your domain is `300x300x300` with a cutoff of `2.5` and a skin of `.5`.

- [x] Implement limiting mechanism for locks
- [x] Implement new way to choose responsible lock for a region

## ~Resolved Issues~

# How Has This Been Tested?

- [x] TSAN in CI

### DIFF
--- a/src/autopas/LogicHandler.h
+++ b/src/autopas/LogicHandler.h
@@ -18,7 +18,6 @@
 #include "autopas/containers/TraversalInterface.h"
 #include "autopas/iterators/ContainerIterator.h"
 #include "autopas/options/IteratorBehavior.h"
-#include "autopas/particles/Particle.h"
 #include "autopas/tuning/AutoTuner.h"
 #include "autopas/tuning/Configuration.h"
 #include "autopas/tuning/selectors/ContainerSelector.h"
@@ -42,7 +41,7 @@ namespace autopas {
 /**
  * The LogicHandler takes care of the containers s.t. they are all in the same valid state.
  * This is mainly done by incorporating a global container rebuild frequency, which defines when containers and their
- * neighbor lists will be rebuild.
+ * neighbor lists will be rebuilt.
  */
 template <typename Particle>
 class LogicHandler {
@@ -90,7 +89,7 @@ class LogicHandler {
    * Returns a non-const reference to the currently selected particle container.
    * @return Non-const reference to the container.
    */
-  inline autopas::ParticleContainerInterface<Particle> &getContainer() {
+  autopas::ParticleContainerInterface<Particle> &getContainer() {
     return _containerSelector.getCurrentContainer();
   }
 
@@ -127,7 +126,7 @@ class LogicHandler {
             // Fast remove of particle, i.e., swap with last entry && pop.
             std::swap(p, buffer.back());
             buffer.pop_back();
-            // Do not increment the iter afterwards!
+            // Do not increment the iter afterward!
           };
           if (p.isDummy()) {
             // We remove dummies!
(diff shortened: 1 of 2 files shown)


src/autopas/LogicHandler.h
```diff
@@ -560,16 +557,56 @@ class LogicHandler {
...
+        const auto locksInFirstDimFloat = std::floor(std::cbrt(maxNumSpacialLocks * prodProportions));
+        // From this and the proportions relative to the first dimension, we can calculate the remaining number of locks
+        const std::array<size_t, 3> locksPerDimLimited = {
+            static_cast<size_t>(locksInFirstDimFloat),  // omitted div by 1
+            static_cast<size_t>(locksInFirstDimFloat / boxSideProportions[1]),
+            static_cast<size_t>(locksInFirstDimFloat / boxSideProportions[2]),
+        };
+        return locksPerDimLimited;
+      }
+    }();
```
schuhmaj, line 609: Given that this lambda is called once and directly, it should probably be highlighted in some way/ or not be a lambda anymore.

Three suggestions to refactor to avoid confusion:
1) Make it a proper static private function of the class calculating the `limitedLocksPerDim`
2) Unfold the lambda, place the body directly into the surrounding function
3) Change the initializer from `const auto locksPerDim` to `const std::array<size_t, 3> locksPerDim`. This way, a developer sees in the **first** line of the declaration that this is not a lambda but just a complex initialization.

I would go with either 1) or 3)

src/autopas/LogicHandler.h
```diff
@@ -560,16 +557,56 @@ class LogicHandler {
...
     using namespace autopas::utils::ArrayMath::literals;
     using autopas::utils::ArrayMath::ceil;
     using autopas::utils::ArrayUtils::static_cast_copy_array;
 
-    // one lock per interaction length + one for each halo region
-    const auto locksPerDim = static_cast_copy_array<size_t>(ceil(boxLength * interactionLengthInv) + 2.);
+    // The maximum number of spatial locks is capped at 1e6.
+    // Without this cap, very large domains (or tiny cutoffs) would generate an insane number of locks,
+    // that could blow up the memory.
+    constexpr size_t maxNumSpacialLocks{1000000};
```
schuhmaj, line 575: I think this limitation should be documented at the member variable as well, probably together with a reference to the function:

```cpp
  /**
   * Locks for regions in the domain. Used for buffer <-> container interaction.
   * <something like capped to 1e6 in initialization function xy>
   */
  std::vector<std::vector<std::vector<std::unique_ptr<std::mutex>>>> _spacialLocks;
```

If I understand correctly, `1e6` is an educated magic number. If so, it should also probably be stated somewhere.

schuhmaj (changes requested):
Good changes 👍. Minor suggestion to refactor/ comment a bit on the `1e6` number.

src/autopas/LogicHandler.h
FG-TUM: Mentioned limit at declaration of `_spacialLocks` -> 5e7347d
"Justified" the choice of `1e6` -> 4818bd2

schuhmaj (approved):
👍

PR #955, opened 2024-10-11
Change VTK Rank Output to UnstructuredGrid

FG-TUM:
# Description

- [x] VTK files for Rank information is now in `UnstructuredGrid` (before `StructuredGrid`)
- [x] Add tags to the filenames to better indicate their contents (`Particles`, `Ranks`)
- [x] `const` (almost) all functions in the `ParallelVTKWriter`.
- [x] Converter script from the old file format to the new.

## Resolved Issues 

The issue was that Paraview would not render surfaces of rank boxes that aren't at the outside of the full domain box (or short: inner surfaces are invisible). It turns out this is a "feature" of the `StructuredGrid`. For us, this is highly undesirable because we want to visualize all volumes. With `UnstructuredGrid` Paraview renders everything.

# How Has This Been Tested?

1. Create some output:
   ```bash
   mpirun -n 4 md-flexible --particles-per-dimension 20 --vtk-filename foo --vtk-write-frequency 1
   ```
2. Check output/foo
3. Load everything into Paraview
4. ???
5. Profit

### DIFF
--- a/examples/md-flexible/README.md
+++ b/examples/md-flexible/README.md
@@ -120,8 +120,8 @@ For examples how to define and configure each object see [`input/AllOptions.yaml
   the domain decomposition.
 The cells contain additional information about the configuration of the AutoPas
   container responsible for simulating the cell.
-To visualize the particle records load the .pvtu files in ParaView. To visualize
-  the cells of the decomposition load the .pvts files in ParaView.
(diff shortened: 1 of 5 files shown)


manishmishra6016 (approved):
This is really nice, now we can see individual domains too. I wonder why they hide this information for structured grids!? seems like something useful always. Anyways, thank you :)

PR #956, opened 2024-10-11
Fix `FuzzyTuningTest::testParseRuleFile()`

FG-TUM:
# Description

The test writes a rule file and then tries to parse it. The `ofstream` was not closed before parsing thus, it could happen that the file was not written yet. Happend [here](https://github.com/AutoPas/AutoPas/actions/runs/11295866924/job/31419527790?pr=955).

## Resolved Issues

# How Has This Been Tested?

- [x] CI


manishmishra6016 (approved):
I think this looks good to merge. This should solve the pipeline fail in unstructured VTK PR too. :)

PR #965, opened 2024-11-11
Fixing md-flexible's calculation of Globals

### LINKED ISSUE #964: md-flexible incorrectly compiles Globals version of AVX and SVE functors
**Describe the bug**
md-flexible incorrectly compiles Globals version of AVX and SVE functors. This is extra misleading as the CMake descriptions say that it doesn't do this.

SamNewcome:
# Description

md-flexible currently compiles the calculate globals version of the AVX and SVE functors, despite the CMake variable description suggesting otherwise. Furthermore, it is not suitable that global calculations cannot be switched on and off for the AVX and SVE functors anyway.

This PR creates a separate CMake variable for switching on and off globals calculations and removes the `MD_FLEXIBLE_FUNCTOR_AUTOVEC_GLOBALS` CMake variable as it is now redundant. 

Shifting is always applied when globals are calculated. The reasoning for this is

- This PR provides a high-priority fix, so I don't want to add any unnecessary features.
- Even with the existing AutoVec globals functors, we always applied shift, so this never was a feature in md-flexible.
- Applying shift is not expected to impose a significant performance cost.

## Resolved Issues

- [x] fixes #964 

# How Has This Been Tested?

- [ ] Current CI
- [x] Testing compiling md-flexible with globals on and off for both the AutoVec and AVX functors

The SVE functor has not been tested because there is a separate bug there. As the AVX fix is a high priority, and the SVE fix is a low priority, I do not wish to wait for that separate fix to be merged.

### DIFF
--- a/.github/workflows/TestSuites.yaml
+++ b/.github/workflows/TestSuites.yaml
@@ -119,7 +119,6 @@ jobs:
             -DMD_FLEXIBLE_USE_MPI=${UseMPI} \
             -DMD_FLEXIBLE_MODE=${{ matrix.md-flex-mode }} \
             -DMD_FLEXIBLE_FUNCTOR_AUTOVEC=ON \
-            -DMD_FLEXIBLE_FUNCTOR_AUTOVEC_GLOBALS=OFF \
             -DMD_FLEXIBLE_FUNCTOR_AVX=${{ matrix.config[8] }} \
             -DMD_FLEXIBLE_FUNCTOR_SVE=OFF \
             -DMD_FLEXIBLE_ENABLE_ALLLBL=${UseMPI} \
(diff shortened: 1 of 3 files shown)


manishmishra6016 (approved):

thesamriel (approved):
> Even with the existing AutoVec globals functors, we always applied shift, so this never was a feature in md-flexible.

It seems, like shift is enabled for every defined LJ functor for md-flexible. Good reminder and smth to keep in mind :D

thesamriel:
I found multiple leftover usages of the deleted CMake flag `MD_FLEXIBLE_FUNCTOR_AUTOVEC_GLOBALS` afterwards.
I will delete them for #751

PR #968, opened 2024-11-12
Smoothening fix

SamNewcome:
# Description

The Smoothening test, `SmoothingTest::loessLastPoint`, would fail on the ARM minicluster. 

In this Pull Request, I cleaned up the smoothening functionality, adding comments and some other refactoring. I explicitly used type conversion. I also used much larger `yvals` in the test, which reflected actual `yvals` (i.e. timings) used in reality. This resulted in the `SmoothingTest::loessLastPoint` test passing on the ARM cluster (not sure exactly what fixed the issue but I suspect the type conversions). 

In addition, smoothening is made optional.

# How Has This Been Tested?

- [x] CI
- [x] Running `SmoothingTest::loessLastPoint` on ARM cluster and checking it passes.


tests/testAutopas/tests/tuning/utils/SmoothingTest.cpp
```diff
@@ -21,7 +22,7 @@ TEST(SmoothingTest, lowessLastPoint) {
 
   // YS values with F = .25, NSTEPS = 0, DELTA = 0.0
   {
-    auto out = autopas::smoothing::smoothLastPoint(obs, (.25 * obs.size()));
-    EXPECT_EQ(out, 6);
+    auto out = autopas::smoothing::smoothLastPoint(obs, (size_t)(.25 * (double)obs.size()));
```
schuhmaj, line 25: Is there are reason you're using the C-style cast instead of C++'s `static_cast`?

schuhmaj (approved):
Minor comment to `casting`, but overall, great changes: The smoothing test was always the one that failed on my MacBook - no more 👍👍👍!

--------

Everything works fine with GCC 14.2.0.

However, I noticed a strange thing, which is related to Apple Clang 16 and Google Test. (**Not associated with this Pull Request!** This holds for the `master` branch as well)

When compiling with Apple Clang 16, the linker complains about missing `arm` symbols of the Google Mock library. One can fix this problem by bumping `autopas_gtest.cmake` and the bundled library to Google Test `1.15.2`:

```cmake
FetchContent_Declare(
        autopas_gtest
        URL ${AUTOPAS_SOURCE_DIR}/libs/googletest-1.15.2.zip
        URL_HASH MD5=eb1c5c237d13ed12bf492d3997ca6b0d
)
```

@FG-TUM @SamNewcome  Should I open an issue to update the bundled Google Test version? It would be rather a tiny PR. Or is the version `1.11.0` a strict requirement?

manishmishra6016 (commented):
Looks good in general, just left a couple of comments for clarification :+1:

src/autopas/tuning/AutoTuner.cpp
```diff
@@ -243,9 +249,12 @@ void AutoTuner::addMeasurement(long sample, bool neighborListRebuilt) {
     const long reducedValue = estimateRuntimeFromSamples();
     _evidenceCollection.addEvidence(currentConfig, {_iteration, _tuningPhase, reducedValue});
 
-    // smooth evidence to remove high outliers. If smoothing results in a higher value use the original value.
+    // If LOESS-based smoothening is enabled, use it to smooth evidence to remove high outliers. If smoothing results in
+    // a higher value or if LOESS-based smoothening is disabled, use the original value.
     const auto smoothedValue =
-        std::min(reducedValue, smoothing::smoothLastPoint(*_evidenceCollection.getEvidence(currentConfig), 5));
+        _useLOESSSmoothening
+            ? std::min(reducedValue, smoothing::smoothLastPoint(*_evidenceCollection.getEvidence(currentConfig), 5))
```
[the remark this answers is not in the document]
SamNewcome: This is my understanding. Regarding exactly why we do this, I refer to @FG-TUM

src/autopas/tuning/utils/Smoothing.cpp
```diff
@@ -9,61 +9,45 @@
...
+      // tri-cube function
+      weights[j] = autopas::utils::Math::pow<3>(
+          1.0 -
+          autopas::utils::Math::pow<3>(static_cast<double>(residual) / static_cast<double>(maxDistFromIntervalEdge)));
+    } else if (residual == 0) {
+      // the tri-cube function tends to inf as residual tends to 0. weight=1 matches residual = 1. Using weight=1 for
+      // residual = 0 is not appropriate, however LOESS is not suitable in the residual=0 case anyway. However, we do
+      // not currently in AutoPas require the residual=0 case (this would imply multiple evidence per configuration).
+      weights[j] = 1.0;
     }
```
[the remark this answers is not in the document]
SamNewcome: Yes you are totally correct. I've removed this. See 55e96a8e.



> does floating point error in weights matters here?

What do you mean by this?

FG-TUM:
> @FG-TUM @SamNewcome Should I open an issue to update the bundled Google Test version? It would be rather a tiny PR. Or is the version `1.11.0` a strict requirement?

@schuhmaj Please fix it in a separate PR.

SamNewcome:
> @FG-TUM @SamNewcome Should I open an issue to update the bundled Google Test version? It would be rather a tiny PR. Or is the version `1.11.0` a strict requirement?

I have no issues with updating this.

tests/testAutopas/tests/tuning/utils/SmoothingTest.cpp
SamNewcome: Nope, I just forgot to use it here. Updated in 7b40e0d3

src/autopas/tuning/utils/Smoothing.cpp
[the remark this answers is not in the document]
manishmishra6016: Ah, nothing, I was just thinking-out-loud on why if-else might have been used. No worries with the changes now! :+1:

src/autopas/tuning/AutoTuner.cpp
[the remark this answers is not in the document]
FG-TUM: Smoothing should only counteract outliers that are slower. The assumption is that if an iteration is suddenly faster, it is not due to something going wrong.

schuhmaj:
GoogleTest Update, as discussed in #970 👍

src/autopas/tuning/AutoTuner.cpp
```diff
@@ -45,6 +46,11 @@ AutoTuner::AutoTuner(TuningStrategiesListType &tuningStrategies, const SearchSpa
     autopas::utils::ExceptionHandler::exception("AutoTuner: Passed tuning strategy has an empty search space.");
   }
   AutoPasLog(DEBUG, "Points in search space: {}", _searchSpace.size());
+  if (_useLOESSSmoothening) {
+    AutoPasLog(DEBUG, "AutoTuner constructed with LOESS Smoothening enabled.");
+  } else {
+    AutoPasLog(DEBUG, "AutoTuner constructed with LOESS Smoothening disabled.");
+  }
```
FG-TUM, line 53:
```suggestion
  AutoPasLog(DEBUG, "AutoTuner constructed with LOESS Smoothening {}.", _useLOESSSmoothening ? "enabled" : "disabled");
```

src/autopas/tuning/utils/Smoothing.h
```diff
@@ -14,6 +14,35 @@
...
+/**
+ * Calculates the weights for the k-neighbors of the last point in the vector using the tri-cube function.
+ * If the sum of weights is <= 0 all neighbors are at the same x coordinate as the point and smoothing will not change
+ * the point. This case is indicated by the bool in the return tuple.
+ *
+ * @param points Sorted collection of observations.
+ * @param pointsPerEstimation Number of neighbors to consider for smoothing.
+ * @param maxDistFromIntervalEdge Largest distance between the point that shall be fitted and its neighbors.
+ * @return Tuple of a vector containing the weights for the neighbors
+ * and a bool indicating whether fitting is unnecessary.
```
FG-TUM, line 26: Can you clarify the wording here a bit? `true` means unnecessary?

src/autopas/tuning/utils/Smoothing.h
FG-TUM, line 1: As discussed let's move the declarations back to the cpp to make them more private.

src/autopas/tuning/utils/Smoothing.cpp
```diff
@@ -9,61 +9,40 @@
...
 
   double sumOfWeights = 0.0;
 
-  // compute all weighs that can be assumed to be non-zero
-  for (size_t j = 0; j < weights.size(); ++j) {
-    auto xj = points[j + firstIndex].iteration;
+  // compute all weights that can be assumed to be non-zero
+  for (size_t j = 0; j < pointsPerEstimation; ++j) {
+    // set weight to zero
+    weights[j] = 0.;
```
FG-TUM, line 28: This seems redundant

src/autopas/tuning/utils/Smoothing.cpp
```diff
@@ -9,61 +9,40 @@
...
+    const auto xj = static_cast<long>(points[j + firstIndex].iteration);
     // residual = |xi - xj|
-    const size_t residual = xj > xi ? xj - xi : xi - xj;
-
-    // if x are too far apart ...
-    if (residual <= maxDistFromIntervalEdgeHigh) {
-      if (residual > maxDistFromIntervalEdgeLow) {
```
FG-TUM, line 49: can you explain why you removed / replaced this? (incl the else)

src/autopas/tuning/utils/Smoothing.cpp
```diff
@@ -119,42 +87,39 @@ double calculateYFitSimple(const std::vector<autopas::Evidence> &points, size_t
...
-  size_t indexToFit = points.size() - 1;
-  // find neighborhood
+  // get the index of the point to be smoothed (the last point)
+  const auto indexToFit = points.size() - 1;
+  // get the index of the first point in the neighborhood around indexToFit used for smoothening
   const auto firstIndex = indexToFit - pointsPerEstimation + 1;
 
-  // maxDistFromIntervalEdge = max(xi - xFirst, xLast - xi)
-  auto maxDistFromIntervalEdge = std::max(points[indexToFit].iteration - points[firstIndex].iteration, 0ul);
+  // maxDistFromIntervalEdge = xi - xFirst. By distance, we refer to the distance in iteration number.
```
FG-TUM, line 114:
```suggestion
  // maxDistFromIntervalEdge = xi - xFirst.
  // By distance, we refer to the distance in iteration number.
```
Just a readability suggestion

FG-TUM (changes requested):

src/autopas/tuning/utils/Smoothing.cpp
SamNewcome: `residual > maxDistFromIntervalEdgeLow` holds always unless `residual==0`. In the old code, the `residual==0` case is handled by the `else` clause (see original lines 53-55). This sets the weight to `1.0`. Except the formula provided in the `if` clause will result in `1.0` as well if evaluated with `residual=0`. Therefore, there is no need to handle this case separately.

Except, of course, for performance reasons. But I doubt the `pow`s here are causing any noticeable performance cost so I would rather simplify the code.

src/autopas/tuning/utils/Smoothing.h
SamNewcome: `true` means the fit is "ok". I'll add this when I move the declarations back.

SamNewcome: See b3e45a23

src/autopas/tuning/utils/Smoothing.cpp
SamNewcome: Removed in 61bc56ac

manishmishra6016 (approved):
looks good :+1:

PR #971, opened 2024-11-15
Remove references to old MD_FLEXIBLE_FUNCTOR_AUTOVEC_GLOBALS CMake flag

thesamriel:
# Description

In #965, we removed the CMake variable `MD_FLEXIBLE_FUNCTOR_AUTOVEC_GLOBALS` in favor for `MD_FLEXIBLE_CALC_GLOBALS`, which enables the calculation of globals for any LJ-Functor (autovec, avx, sve) or AT-Functor.

This PR removes any leftover usages of the deleted CMake variable.
Additionally, the `MDFlexConfig::FunctorOption::lj12_6_globals` was removed, as it is no longer needed with the reworked globals switch.

# How Has This Been Tested?

- [x] Test Pipeline runs fine
- [x] Manually tested that enabling/disabling globals for LJ functor works


manishmishra6016 (approved):
looks good :+1:

SamNewcome (approved):
Sorry about the mistakes! Thanks for spotting and correcting them!

Issue #796, opened 2023-10-27
Cleaner handling of functors in md-flexible


SamNewcome:
**Is your feature request related to a problem? Please describe.**
At the moment we have a rather spaghetti-based situation whereby the user specifies which functor they would like compiled. At the moment this includes every possible vectorisation. Then depending on whether AutoPas is in multi-site mode or not, a particular version is compiled. I see this as getting rather spaghetti-like with more and more functors - both in terms on vectorisation and the actual force model.

**Describe the solution you'd like**
A compile option `AUTOPAS_APPLICATION_LIBRARY_PREFERRED_VECTORIZATION` (or potentially a short alternative). The user sets a preferred vectorisation choice, e.g. `AUTOVEC`, `AVX512`, `SVE`. If the user's preferred choice is not available for a particular force model (LJ, Mie + Singlesite/Multisite), then a warning is thrown and a fallback option is used. For now the fallback option could be the AutoVec version, but equally, it could also be a vector wrapper version. We remove the need for vectorisation to be specified in the input.

For functors based on different force models (e.g. Lennard-Jones, Mie, (Axilrod-Teller-Muto?)), we keep the current compile time system + specification in the input. 

My suggestion is based on the assumption that users don't need to compile and use multiple different vectorisations of one force calculation to be compiled at any one time. I think that this is a fair assumption - you generally just want the best available option for your architecture. The exception to this is when you are actually comparing implementations of the same functor. But this can be easily handled with separate build folders. As this exception is pretty much just for developers of functors, I think this solution is suitable.

**Describe alternatives you've considered**
The default option is `AUTOMATIC`, which determines the best vectorisation based on architecture and what is available for a particular force model.

FG-TUM:
We already offer the user not to compile versions of functors
https://github.com/AutoPas/AutoPas/blob/b292345510f218132b353ed775ba9c8274c1ce0c/examples/md-flexible/CMakeLists.txt#L19-22

This could (and probably should) be extended to every functor we offer in our application library.

SamNewcome:
No, I fully understand this but my suggestion was a way of cleaning this up as new force models and vectorisation methods are developed. If this isn't clear, please let me know and I will reformulate the first comment.

The reason for the feature is that I don't see a reason why a user would want to compile multiple vectorisation types for one particular force model. In general, people just want the "best". 

[...]

PR #972, opened 2024-11-17
Highway clone based

TumBeat:
# Description

This PR introduces the Google-Highway Lennard Jones Functor and a different particle interaction orders (vectorization patterns) for the SoAPair and SoASingle Functor methods.

Fixes 

- #833 
- #796 (As good default option provided)

# How Has This Been Tested?

- [x] Implemented Unit Tests for Highway vs. AVX
- [x] Implemented Unit Tests for the various patterns vs. AVX


applicationLibrary/molecularDynamics/tests/singleSiteTests/LJFunctorAVXTest.cpp
```diff
@@ -129,6 +129,7 @@ void LJFunctorAVXTest::testLJFunctorVSLJFunctorAVXTwoCells(bool newton3, bool do
     if constexpr (mixing) {
       particle.setTypeId(particle.getID() % 5);
     }
+    particle.setTypeId(particle.getID() % 5);
```
thesamriel: Is this line supposed to be here? If so, the  `if constexpr (mixing)` can be removed.

examples/md-flexible/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ target_link_libraries(md-flexible PUBLIC autopas autopasTools molecularDynamicsL
 option(MD_FLEXIBLE_FUNCTOR_AUTOVEC "Compile AutoVec Functor." OFF)
 option(MD_FLEXIBLE_FUNCTOR_AVX "If instruction set is available, compile AVX Functor." ON)
 option(MD_FLEXIBLE_FUNCTOR_SVE "If instruction set is available, compile SVE Functor." ON)
+option(MD_FLEXIBLE_FUNCTOR_HWY "If instruction set is available, compile Highway Functor." ON)
```
thesamriel: I think if no instruction set is available, highway falls back to normal scalar instructions.
```suggestion
option(MD_FLEXIBLE_FUNCTOR_HWY "Compile Highway Functor, which uses the best available SIMD instruction set supported by google highway." ON)
```

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1377 @@
...
+ * @date 27.03.2024
+ * @author Luis Gall
+ */
+
+#pragma once
+
+#include <hwy/highway.h>
+
+#include <algorithm>
+
```
thesamriel:
```suggestion
```

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1377 @@
...
+    const auto *const __restrict typeIDptr = soa.template begin<Particle_T::AttributeNames::typeId>();
+
+    // initialize and declare vector variables
+    auto virialSumX = _zeroDouble;
+    auto virialSumY = _zeroDouble;
+    auto virialSumZ = _zeroDouble;
+    auto uPotSum = _zeroDouble;
+
+    size_t i = soa.size() - 1;
+    for (; checkFirstLoopCondition<true, vecPattern>(i, 0); decrementFirstLoop<vecPattern>(i)) {
```
thesamriel: I noticed smth interesting here.
`i` is of type `size_t` and gets decremented every iteration until it reaches 0. In the next call to `decrementFirstLoop`, `i` is decremented again, which leads to a wrap around and `i` being the maximum unsigned long number. Then, in `checkFirstLoopCondition` it is implicitly converted to a `signed long` value, which in most implementations leads to a negative number and therefore to `i >= 0`  being false. However, this is actually undefined behavior, and we should solve this differently. 
Easiest option would be to make `i` a signed integer. Otherwise, the loop conditions would need to be handled differently.

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1377 @@
...
+/** Type for a Double vector register */
+using VectorDouble = decltype(highway::Zero(tag_double));
+/** Type for a Long vector register */
+using VectorLong = decltype(highway::Zero(tag_long));
+/** Highway tag for a half-filled double register */
+constexpr highway::Half<highway::DFromV<VectorDouble>> tag_double_half;
+/** Type for a Double Mask */
+using MaskDouble = decltype(highway::FirstN(tag_double, 1));
+/** Type for a Long Mask */
+using MaskLong = decltype(highway::FirstN(tag_long, 2));
```
SamNewcome: It would be good if we had some kind of "highway utils" file, so this can be shared. Plus, I feel a little uncomfortable about the `namespace highway = hwy::HWY_NAMESPACE;` getting included everywhere this header get included - I am fine if this is rather upfront i.e. every file in the application library/example simulator is okay with being "contaminated" by this namespace alias.

I don't have a strong opinion right now about whether this should be in AutoPas or in the application library. Imo, the argument for putting this in AutoPas, is that it would be really nice if we could avoid users have to write such ugly functors with effectively lots of boilerplate code by providing a class (i.e. in AutoPas src) which implements all the boilerplate stuff such that the specific functors only need to implement the kernel. But right now, there's no real need for this.

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1377 @@
...
+ * @tparam relevantForTuning Whether the auto-tuner should consider this functor.
+ * @tparam countFLOPs counts FLOPs and hitrate. Not implemented for this functor. Please use the AutoVec functor.*/
+template <class Particle_T, bool applyShift = false, bool useMixing = false,
+          autopas::FunctorN3Modes useNewton3 = autopas::FunctorN3Modes::Both, bool calculateGlobals = false,
+          bool countFLOPs = false, bool relevantForTuning = true>
+
+class LJFunctorHWY
+    : public autopas::PairwiseFunctor<Particle_T, LJFunctorHWY<Particle_T, applyShift, useMixing, useNewton3,
+                                                               calculateGlobals, countFLOPs, relevantForTuning>> {
+  using SoAArraysType = typename Particle_T::SoAArraysType;
```
SamNewcome:
```suggestion
  using SoAArraysType = Particle_T::SoAArraysType;
```

thesamriel (changes requested):
I had a couple of small things in the functor file, where I took the liberty to edit these directly.
There's a couple more minor things.

Lastly, we might want to discuss if we really merge all input yaml files or if we keep them somewhere separate.

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
dmartin13: Good catch! I have modified this slightly and now use a `std::ptrdiff_t`, which can also be negative.

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1366 @@
...
+      if constexpr (!remainder) {
+        const auto upperFxAcc = highway::UpperHalf(tag_double_half, fxAcc);
+        const auto upperFyAcc = highway::UpperHalf(tag_double_half, fyAcc);
+        const auto upperFzAcc = highway::UpperHalf(tag_double_half, fzAcc);
+
+        const auto upperFxAccExt = highway::ZeroExtendVector(tag_double, upperFxAcc);
+        const auto upperFyAccExt = highway::ZeroExtendVector(tag_double, upperFyAcc);
+        const auto upperFzAccExt = highway::ZeroExtendVector(tag_double, upperFzAcc);
+
+        int index = reversed ? i - 1 : i + 1;
```
SamNewcome:
```suggestion
        const auto index = reversed ? i - 1 : i + 1;
```

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1366 @@
...
+
+      const auto lowerFxAccExt = highway::ZeroExtendVector(tag_double, lowerFxAcc);
+      const auto lowerFyAccExt = highway::ZeroExtendVector(tag_double, lowerFyAcc);
+      const auto lowerFzAccExt = highway::ZeroExtendVector(tag_double, lowerFzAcc);
+
+      fxPtr[i] += highway::ReduceSum(tag_double, lowerFxAccExt);
+      fyPtr[i] += highway::ReduceSum(tag_double, lowerFyAccExt);
+      fzPtr[i] += highway::ReduceSum(tag_double, lowerFzAccExt);
+
+      if constexpr (!remainder) {
```
SamNewcome:
```suggestion
      if constexpr (not remainder) {
```

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1366 @@
...
+      remainder ? highway::StoreN(fzNew, tag_double, &fzPtr[i], restI) : highway::StoreU(fzNew, tag_double, &fzPtr[i]);
+    }
+  }
+
+  template <bool newton3>
+  inline void computeGlobals(const VectorDouble &virialSumX, const VectorDouble &virialSumY,
+                             const VectorDouble &virialSumZ, const VectorDouble &uPotSum) {
+    const int threadnum = autopas::autopas_get_thread_num();
+
+    double globals[4]{highway::ReduceSum(tag_double, virialSumX), highway::ReduceSum(tag_double, virialSumY),
```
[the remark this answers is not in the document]
SamNewcome: And `const`

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1366 @@
...
+                      highway::ReduceSum(tag_double, virialSumZ), highway::ReduceSum(tag_double, uPotSum)};
+
+    double factor = 1.;
+    if constexpr (newton3) {
+      factor = 0.5;
+    }
```
SamNewcome:
```suggestion
    const double factor = newton3 ? 0.5 : 1.;
```

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1366 @@
...
+          epsilons[index] = _PPLibrary->getMixing24Epsilon(*typeID1, *(typeID2Ptr + k));
+          sigmas[index] = _PPLibrary->getMixingSigmaSquared(*typeID1, *(typeID2Ptr + k));
+
+          if constexpr (applyShift) {
+            shifts[index] = _PPLibrary->getMixingShift6(*typeID1, *(typeID2Ptr + k));
+          }
+        }
+      }
+    } else if constexpr (vecPattern == VectorizationPattern::pVecx1) {
+      for (int n = 0; n < (remainderI ? restI : _vecLengthDouble); ++n) {
```
SamNewcome: `n` -> `i`

applicationLibrary/molecularDynamics/tests/singleSiteTests/LJFunctorTestHWY.h
```diff
@@ -0,0 +1,56 @@
...
+#include "testingHelpers/commonTypedefs.h"
+
+using VectorizationPattern = autopas::VectorizationPatternOption::Value;
+
+using LJFunctorHWYTestingTuple =
+    std::tuple<bool /*mixing*/, bool /*newton3*/, bool /*doDeleteSomeParticles*/, VectorizationPattern>;
+
+class LJFunctorTestHWY : public AutoPasTestBase, public ::testing::WithParamInterface<LJFunctorHWYTestingTuple> {
+ public:
+  LJFunctorTestHWY() : AutoPasTestBase() {}
```
SamNewcome:
```suggestion
  LJFunctorTestHWY() {}
```

applicationLibrary/molecularDynamics/tests/singleSiteTests/LJFunctorTestHWY.h
SamNewcome: Add documentation

applicationLibrary/molecularDynamics/tests/singleSiteTests/LJFunctorTestHWY.h
```diff
@@ -0,0 +1,56 @@
+/**
+ * @file LJFunctorTestHWY.h
+ * @author Luis Gall
+ * @date 04/23/24
```
SamNewcome: -> DD/MM/YYYY

applicationLibrary/molecularDynamics/tests/singleSiteTests/LJFunctorTestHWY.cpp
```diff
@@ -0,0 +1,584 @@
...
+  ljFunctorHWY.SoAExtractor(cell1NoHWY, cell1NoHWY._particleSoABuffer, 0);
+  ljFunctorHWY.SoAExtractor(cell2NoHWY, cell2NoHWY._particleSoABuffer, 0);
+
+  ASSERT_TRUE(AoSParticlesEqual(cell1HWY, cell1NoHWY)) << "Cells 1 not equal after extracting.";
+  ASSERT_TRUE(AoSParticlesEqual(cell2HWY, cell2NoHWY)) << "Cells 2 not equal after extracting.";
+
+  ljFunctorHWY.endTraversal(newton3);
+  ljFunctorAVX.endTraversal(newton3);
+
+  double tolerance = 1e-8;
```
SamNewcome: Make the `tolerance` `constexpr`. Also, I try a lower tolerance - I'm a little suspicious of it being too high.

applicationLibrary/molecularDynamics/tests/singleSiteTests/LJFunctorTestHWY.cpp
```diff
@@ -0,0 +1,584 @@
...
+
+  ASSERT_TRUE(AoSParticlesEqual(cellAVX, cellHWY)) << "Cells not equal after extracting.";
+
+  ljFunctorAVX.endTraversal(newton3);
+  ljFunctorHWY.endTraversal(newton3);
+
+  double tolerance = 1e-8;
+  // TODO : uncomment before release
+  // EXPECT_NEAR(ljFunctorAVX.getPotentialEnergy(), ljFunctorHWY.getUpot(), tolerance) << "global uPot";
+  // EXPECT_NEAR(ljFunctorAVX.getVirial(), ljFunctorHWY.getVirial(), tolerance) << "global virial";
```
SamNewcome: Similar

applicationLibrary/molecularDynamics/tests/singleSiteTests/LJFunctorTestHWY.cpp
SamNewcome: Add brief comment descirptions above each test

src/autopas/baseFunctors/Functor.h
```diff
@@ -167,6 +168,12 @@ class Functor {
    */
   [[nodiscard]] double getCutoff() const { return _cutoff; }
 
+  /**
+   * Setter for the vectorization pattern to choose
```
SamNewcome:
```suggestion
   * Setter for the vectorization pattern to be used.
```

src/autopas/options/VectorizationPatternOption.h
```diff
@@ -0,0 +1,84 @@
...
+ */
+class VectorizationPatternOption : public Option<VectorizationPatternOption> {
+ public:
+  /**
+   * Possible choices for the vector patterns
+   */
+  enum Value {
+    /**
+     * Interact one particle from the first list with the full vector length from the second list
+     */
```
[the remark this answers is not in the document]
SamNewcome: And similar for the other patterns

src/autopas/options/VectorizationPatternOption.h
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Provides a way to iterate over the possible choices of Vectorization Patterns
+   * @return map option -> string representation
+   */
+  static std::map<VectorizationPatternOption, std::string> getOptionNames() {
+    return {
+        {VectorizationPatternOption::p1xVec, "1xVectorLength"},
+        {VectorizationPatternOption::p2xVecDiv2, "2xVectorLengthDiv2"},
+        {VectorizationPatternOption::pVecDiv2x2, "VectorLengthDiv2x2"},
+        {VectorizationPatternOption::pVecx1, "VectorLengthx1"},
```
SamNewcome:
```suggestion
        {p1xVec, "1xVectorLength"},
        {p2xVecDiv2, "2xVectorLengthDiv2"},
        {pVecDiv2x2, "VectorLengthDiv2x2"},
        {pVecx1, "VectorLengthx1"},
```

src/autopas/tuning/Configuration.cpp
```diff
@@ -88,6 +88,10 @@ bool autopas::Configuration::hasCompatibleValues() const {
       return false;
     }
   }
+  if (vecPattern != VectorizationPatternOption::p1xVec) {
+    // @todo if not HWY Functor, this should return false
+    // Consequently, we need some sort of access to the functor
```
SamNewcome: Do this ToDo.

Fyi, I wouldn't actually do this here in `Configuration` because you don't have access to the functor. There is another possibility to do this in `LogicHandler::isConfigurationApplicable`, where you have the functor - this is where we exclude a configuration e.g. if `newton3` is not allowed by the functor.

Potentially we should have the base `Functor` have some `allowedVecPatterns` method which allows you to check is a pattern is allowed. By default, for `PairwiseFunctor` this can be `p1xVec` so we don't have to make changes.

For triwise, potentially, we need a separate `TriwiseVectorizationPatternOption` which for now just includes a `3B_default` enum, which again the `TriwiseFunctor` has set by default. The idea would be to set this up either so that it is clear that triwise functors do not have different patterns (and to not confuse matters that giving them as a placeholder a pairwise pattern like `p1xVec` which doesn't make sense in the case that we have three lists), and to provide a possibility for adding this functionality to triwise functors in the future. I'm not sure whether a separate option class or some distinction within the existing option class makes more sense.

You may then run into issue #1032 (see my comment referenced in the issue where I provide more details).

It may therefore be that we need a fix to this issue before we can merge this branch in. Check by running e.g. the AVX functor with additional patterns than just `p1xVec`. If this is an issue, maybe do this in a separate PR.

examples/md-flexible/src/configuration/MDFlexConfig.h
```diff
@@ -515,7 +522,8 @@ class MDFlexConfig {
   MDFlexOption<FunctorOption, __LINE__> functorOption{// Default is a dummy option
                                                       FunctorOption::none, "functor", true,
                                                       "Pairwise force functor to use. Possible Values: (lennard-jones "
-                                                      "lennard-jones-AVX lennard-jones-SVE lennard-jones-globals)"};
+                                                      "lennard-jones-AVX lennard-jones-SVE lennard-jones-globals "
```
SamNewcome: Remove globals

SamNewcome (changes requested):
There are some performance related suggestions. Imo, these aren't critical and don't have to be done this PR. We have our AutoPas internal meeting on Friday so I might discuss this with @thesamriel then, whether we want you to investigate this. For now, it would be good if you at least comment with your views on if you think they are good ideas, or meh, you think they are not worth investigating, or you don't know - don't spend too much time researching, just if your previous experience suggests something.

SamNewcome (changes requested):
There are some performance related suggestions. Imo, these aren't critical and don't have to be done this PR. We have our AutoPas internal meeting on Friday so I might discuss this with @thesamriel then, whether we want you to investigate this. For now, it would be good if you at least comment with your views on if you think they are good ideas, or meh, you think they are not worth investigating, or you don't know - don't spend too much time researching, just if your previous experience suggests something.

src/autopas/options/VectorizationPatternOption.h
```diff
@@ -0,0 +1,84 @@
...
+
+#include <set>
+
+#include "autopas/options/Option.h"
+
+namespace autopas {
+inline namespace options {
+
+/**
+ * Class representing the choices of possible vectorization patterns for the Pairwise Functors
```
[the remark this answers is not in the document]
SamNewcome: Here's the DOI: https://doi.org/10.48550/arXiv.2512.03565

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1377 @@
...
+      i += 2;
+    } else if constexpr (vecPattern == VectorizationPattern::pVecDiv2x2) {
+      i += _vecLengthDouble / 2;
+    } else if constexpr (vecPattern == VectorizationPattern::pVecx1) {
+      i += _vecLengthDouble;
+    }
+  }
+
+  template <bool reversed>
+  constexpr int obtainFirstLoopRest(const int i, const int stop) {
```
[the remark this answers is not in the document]
dmartin13: I use `obtainILoopRemainderLength` now.

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1377 @@
...
+   * @param cutoff
+   * @param particlePropertiesLibrary
+   */
+  explicit LJFunctorHWY(double cutoff, ParticlePropertiesLibrary<double, size_t> &particlePropertiesLibrary)
+      : LJFunctorHWY(cutoff, nullptr) {
+    static_assert(useMixing,
+                  "Not using Mixing but using a ParticlePropertiesLibrary is not allowed! Use a different constructor "
+                  "or set mixing to true.");
+    _PPLibrary = &particlePropertiesLibrary;
+  }
```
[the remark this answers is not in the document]
SamNewcome: Thanks. It would be good if you could also do this for the ATM HWY functor if you haven't already, but I wouldn't worry about AutoVec, AVX, and SVE.

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1377 @@
...
+    } else {
+      return -1;
+    }
+  }
+
+  template <bool remainder, bool reversed, VectorizationPattern vecPattern>
+  inline void fillIRegisters(const size_t i, const double *const __restrict xPtr, const double *const __restrict yPtr,
+                             const double *const __restrict zPtr,
+                             const autopas::OwnershipState *const __restrict ownedStatePtr, VectorDouble &x1,
+                             VectorDouble &y1, VectorDouble &z1, MaskDouble &ownedMaskI, const size_t restI) {
```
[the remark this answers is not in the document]
dmartin13: The problem here is that we later do a logical and with the `ownedMaskI` and the `cutoffMask` (double mask). With highway, logical operations can only be performed with the same mask types.
So we have two options:
1) Create double masks directly, as we do now. This is not so nice because it's not very intuitive why this is being done.
2) Create integer masks. Then we would have to cast the `cutoffMask` in the kernel. I don't know if this would have a negative impact on performance. Probably only a minor one.

applicationLibrary/molecularDynamics/tests/singleSiteTests/LJFunctorTestHWY.cpp
```diff
@@ -0,0 +1,584 @@
...
+    PPL.addLJParametersToSite(0, 1., 1.);
+    PPL.addSiteType(1, 1.5);
+    PPL.addLJParametersToSite(1, 2., 1.);
+    PPL.addSiteType(2, 2.);
+    PPL.addLJParametersToSite(2, 1., 1.);
+    PPL.addSiteType(3, 2.5);
+    PPL.addLJParametersToSite(3, 2., 1.);
+    PPL.addSiteType(4, 3.);
+    PPL.addLJParametersToSite(4, 1., 1.);
+    PPL.calculateMixingCoefficients();
```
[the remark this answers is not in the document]
dmartin13: After doing that, I had to reduce the tolerance again. I'm a little concerned that there might be a small bug in the functor somewhere. (problematic only with newton3)

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1366 @@
...
+    if constexpr (applyShift) {
+      _shift6 = highway::Set(tag_double, ParticlePropertiesLibrary<double, size_t>::calcShift6(
+                                             epsilon24, sigmaSquare, highway::GetLane(_cutoffSquared)));
+    } else {
+      _shift6 = _zeroDouble;
+    }
+    _epsilon24AoS = epsilon24;
+    _sigmaSquareAoS = sigmaSquare;
+    if constexpr (applyShift) {
+      _shift6AoS = ParticlePropertiesLibrary<double, size_t>::calcShift6(epsilon24, sigmaSquare, _cutoffSquareAoS);
```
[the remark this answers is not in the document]
SamNewcome: Hmm okay. Then just leave them as they are

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1405 @@
...
+      fyPtr[i] += highway::ReduceSum(tag_double, fyAcc);
+      fzPtr[i] += highway::ReduceSum(tag_double, fzAcc);
+    } else if constexpr (vecPattern == VectorizationPattern::p2xVecDiv2) {
+      const auto lowerFxAcc = highway::LowerHalf(fxAcc);
+      const auto lowerFyAcc = highway::LowerHalf(fyAcc);
+      const auto lowerFzAcc = highway::LowerHalf(fzAcc);
+
+      const auto lowerFxAccExt = highway::ZeroExtendVector(tag_double, lowerFxAcc);
+      const auto lowerFyAccExt = highway::ZeroExtendVector(tag_double, lowerFyAcc);
+      const auto lowerFzAccExt = highway::ZeroExtendVector(tag_double, lowerFzAcc);
```
thesamriel: Similar as above, I'm not sure why we need `ZeroExtendVector`

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1405 @@
...
+    const auto drX2 = drX * drX;
+    const auto drY2 = drY * drY;
+    const auto drZ2 = drZ * drZ;
+
+    const auto dr2 = drX2 + drY2 + drZ2;
+
+    const auto cutoffMask = highway::Le(dr2, _cutoffSquared);
+    const auto dummyMask =
+        highway::And(ownedMaskI, highway::Ne(highway::ConvertTo(tag_double, ownedStateJLong), _ownedStateDummy));
+    const auto cutoffDummyMask = highway::And(cutoffMask, dummyMask);
```
thesamriel:
```suggestion
    const auto dummyMask =
        highway::And(ownedMaskI, highway::MaskFromVec(highway::ConvertTo(tag_double, ownedStateJLong)));
    const auto cutoffDummyMask = highway::MaskedLe(dummyMask, dr2, _cutoffSquared);
```

I think this should do the same thing and be slightly better performing.

thesamriel (changes requested):
I had some time to look a bit into highway itself, so there are some more comments

cmake/modules/autopas_highway.cmake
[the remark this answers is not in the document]
dmartin13: Okay, those are all the ones I found. To be honest, I don't think it's necessary to expose them to the user (maybe HWY_CMAKE_SSE2, HWY_CMAKE_ARM7, ...).

```
HWY_CMAKE_ARM7          OFF       "Set copts for Armv7 with NEON (requires vfpv4)?"
HWY_CMAKE_LSX           ON        "Add -mlsx flag?"
HWY_CMAKE_LASX          ON        "Add -mlasx flag?"
HWY_CMAKE_SSE2          OFF       "Set SSE2 as baseline for 32-bit x86?"
HWY_CMAKE_RVV           ON        "Set copts for RISCV with RVV?"
HWY_WARNINGS_ARE_ERRORS OFF       "Add -Werror flag?"
HWY_CMAKE_HEADER_ONLY   OFF       "Change to header-only?"
HWY_ENABLE_CONTRIB      ON        "Include contrib/"
HWY_ENABLE_EXAMPLES     ON        "Build examples"
HWY_ENABLE_INSTALL      ON        "Install library"
HWY_ENABLE_TESTS        ON        "Enable HWY tests"
HWY_TEST_STANDALONE     OFF       "Disable use of googletest"
HWY_CXX_STD_TGT_COMPILE_FEATURE   [cxx_std_17, cxx_std_14, cxx_std_11]
HWY_THREAD_FLAGS                  "-pthread"
HWY_THREAD_LIBS                   ${CMAKE_THREAD_LIBS_INIT})
HWY_SOURCES
HWY_TEST_SOURCES
HWY_FLAGS
HWY_LIBRARY_TYPE                  ["SHARED", "STATIC"]
HWY_LIBRARY_VERSION               "${CMAKE_PROJECT_VERSION}"
HWY_PC_FILES                      libhwy.pc
HWY_PC_DISABLE_FUTEX_CFLAGS       "-DHWY_DISABLE_FUTEX"
HWY_PC_WIN32_SYNCHRONIZATION_LIBS "-lsynchronization"
HWY_PC_HWY_TEST_CFLAGS            "-DHWY_TEST_STANDALONE=1"
HWY_PC_HWY_TEST_REQUIRES          "gtest"
HWY_SYSTEM_GTEST        OFF       "Use pre-installed googletest?"
HWY_TEST_FILES
HWY_TEST_LIBS                     hwy hwy_test
HWY_GTEST_LIBS
```

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1405 @@
...
+   */
+  LJFunctorHWY() = delete;
+
+  /**
+   * Constructor for Functor with mixing enabled/disabled. When using this functor without particlePropertiesLibrary it
+   * is necessary to call setParticleProperties() to set internal constants.
+   * @param cutoff
+   * @param particlePropertiesLibrary
+   */
+  explicit LJFunctorHWY(double cutoff, ParticlePropertiesLibrary<double, size_t> *particlePropertiesLibrary = nullptr)
```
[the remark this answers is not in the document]
dmartin13: I have implemented this. The only consequence is that we now have to use `return f(LJFunctorTypeHWY{cutoff, std::ref(particlePropertiesLibrary)});` instead of `return f(LJFunctorTypeHWY{cutoff, particlePropertiesLibrary});`, but I think that's okay.

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1405 @@
...
+      const auto upperFy = highway::UpperHalf(tag_double_half, fy);
+      const auto upperFz = highway::UpperHalf(tag_double_half, fz);
+
+      const auto fxCombined = lowerFx + upperFx;
+      const auto fyCombined = lowerFy + upperFy;
+      const auto fzCombined = lowerFz + upperFz;
+
+      const auto fxCombinedExt = highway::ZeroExtendVector(tag_double, fxCombined);
+      const auto fyCombinedExt = highway::ZeroExtendVector(tag_double, fyCombined);
+      const auto fzCombinedExt = highway::ZeroExtendVector(tag_double, fzCombined);
```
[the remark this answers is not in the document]
dmartin13: I modified it like this and it works.

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1366 @@
...
+          const auto mask = overlapMasks[remainder ? rest : _vecLengthDouble / 2];
+          lowerFxExt = highway::IfThenElseZero(mask, lowerFxExt);
+          lowerFyExt = highway::IfThenElseZero(mask, lowerFyExt);
+          lowerFzExt = highway::IfThenElseZero(mask, lowerFzExt);
+        }
+      }
+
+      fx2Ptr[j] -= highway::ReduceSum(tag_double, lowerFxExt);
+      fy2Ptr[j] -= highway::ReduceSum(tag_double, lowerFyExt);
+      fz2Ptr[j] -= highway::ReduceSum(tag_double, lowerFzExt);
```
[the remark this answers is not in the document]
dmartin13: The `ZeroExtendVector` operations have now been omitted, as well as the middle part:

```
if constexpr (reversed) {
	if (j > i - _vecLengthDouble / 2) {
	  const auto mask = overlapMasks[remainder ? rest : _vecLengthDouble / 2];
	  lowerFxExt = highway::IfThenElseZero(mask, lowerFxExt);
	  lowerFyExt = highway::IfThenElseZero(mask, lowerFyExt);
	  lowerFzExt = highway::IfThenElseZero(mask, lowerFzExt);
	}
}
```

(see: [https://github.com/AutoPas/AutoPas/pull/972#discussion_r2616424024]())

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1366 @@
...
+            const auto mask = overlapMasks[_vecLengthDouble / 2];
+            upperFxExt = highway::IfThenElseZero(mask, upperFxExt);
+            upperFyExt = highway::IfThenElseZero(mask, upperFyExt);
+            upperFzExt = highway::IfThenElseZero(mask, upperFzExt);
+          }
+        }
+
+        fx2Ptr[j + 1] -= highway::ReduceSum(tag_double, upperFxExt);
+        fy2Ptr[j + 1] -= highway::ReduceSum(tag_double, upperFyExt);
+        fz2Ptr[j + 1] -= highway::ReduceSum(tag_double, upperFzExt);
```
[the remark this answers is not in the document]
dmartin13: similar (see: [https://github.com/AutoPas/AutoPas/pull/972#discussion_r2617283776]())

applicationLibrary/molecularDynamics/tests/singleSiteTests/LJFunctorTestHWY.cpp
```diff
@@ -0,0 +1,584 @@
...
```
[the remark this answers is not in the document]
dmartin13: I adjusted the test parameters slightly. On the one hand, I made the LJ parameters more varied, and on the other hand, I made the simulation box slightly larger. This prevents particles with higher sigma that are tightly packed together from generating high forces. Now I can use `1e-8` in all tests, as with AVX. I also went through the tests systematically again and looked at some code sections, but I couldn't find any bugs. I suspect it's due to FP inaccuracies when very large forces are generated.

applicationLibrary/molecularDynamics/tests/singleSiteTests/LJFunctorTestHWY.cpp
```diff
@@ -0,0 +1,584 @@
...
+  double *const __restrict fyptr1 = soa1.template begin<Molecule::AttributeNames::forceY>();
+  double *const __restrict fzptr1 = soa1.template begin<Molecule::AttributeNames::forceZ>();
+  double *const __restrict fxptr2 = soa2.template begin<Molecule::AttributeNames::forceX>();
+  double *const __restrict fyptr2 = soa2.template begin<Molecule::AttributeNames::forceY>();
+  double *const __restrict fzptr2 = soa2.template begin<Molecule::AttributeNames::forceZ>();
+
+  for (size_t i = 0; i < soa1.size(); ++i) {
+    EXPECT_EQ(idptr1[i], idptr2[i]);
+
+    double tolerance = 2e-8;
```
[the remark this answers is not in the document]
dmartin13: see [https://github.com/AutoPas/AutoPas/pull/972#discussion_r2617302046]()

applicationLibrary/molecularDynamics/tests/singleSiteTests/LJFunctorTestHWY.cpp
dmartin13: see [https://github.com/AutoPas/AutoPas/pull/972#discussion_r2617302046]()

applicationLibrary/molecularDynamics/tests/singleSiteTests/LJFunctorTestHWY.cpp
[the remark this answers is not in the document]
dmartin13: see [https://github.com/AutoPas/AutoPas/pull/972#discussion_r2617302046]()

applicationLibrary/molecularDynamics/tests/singleSiteTests/LJFunctorTestHWY.cpp
```diff
@@ -0,0 +1,584 @@
...
+    PPL.addSiteType(1, 1.5);
+    PPL.addLJParametersToSite(1, 2., 1.);
+    PPL.addSiteType(2, 2.);
+    PPL.addLJParametersToSite(2, 1., 1.);
+    PPL.addSiteType(3, 2.5);
+    PPL.addLJParametersToSite(3, 2., 1.);
+    PPL.addSiteType(4, 3.);
+    PPL.addLJParametersToSite(4, 1., 1.);
+    PPL.calculateMixingCoefficients();
+  }
```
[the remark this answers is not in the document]
dmartin13: see [https://github.com/AutoPas/AutoPas/pull/972#discussion_r2617302046]()

SamNewcome (changes requested):
Looks good. Only really very minor things left.

tests/testAutopas/tests/tuning/AutoTunerTest.cpp
```diff
@@ -81,52 +82,52 @@ TEST_F(AutoTunerTest, testAllConfigurations) {
...
-  configsPerContainer[autopas::ContainerOption::octree] = 4;
+      configsPerContainer[autopas::ContainerOption::linkedCells] - 16;
+  // VerletLists:           vl_list_iteration           (AoS <=> SoA, noNewton3, 4 vecPattern) = 8
+  configsPerContainer[autopas::ContainerOption::verletLists] = 8;
+  // VerletListsCells:      vlc_sliced                  (AoS <=> SoA, newton3 <=> noNewton3, 4 vecPattern) = 16
+  //                        vlc_sliced_balanced         (AoS <=> SoA, newton3 <=> noNewton3, 3 heuristics, 4 vecPattern)
+  //                        = 48 vlc_sliced_colored          (AoS <=> SoA, newton3 <=> noNewton3, 4 vecPattern) = 16
+  //                        vlc_c18                     (AoS <=> SoA, newton3 <=> noNewton3, 4 vecPattern) = 16 vlc_c01
+  //                        (AoS <=> SoA, noNewton3, 4 vecPattern)                             = 8 vlc_c08 (AoS <=> SoA,
+  //                        newton3 <=> noNewton3, 4 vecPattern)                 = 16
```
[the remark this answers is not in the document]
dmartin13: Good catch! That was actually still a bug in the code because the autotuner generated too many valid configurations. I have now introduced `CompatibleVectorizationPattern`.

src/autopas/containers/CompatibleVectorizationPattern.h
```diff
@@ -0,0 +1,40 @@
...
+    case ContainerOption::verletListsCells:
+    case ContainerOption::pairwiseVerletLists:
+    case ContainerOption::varVerletListsAsBuild: {
+      return std::set<autopas::VectorizationPatternOption>{VectorizationPatternOption::p1xVec};
+    }
+    default: {
+      return std::set<autopas::VectorizationPatternOption>{autopas::VectorizationPatternOption::getAllOptions()};
+    }
+  }
+}
```
SamNewcome: My concern with this is that this gets forgotten and somebody adds a new container that can only use `p1xVec`.

I would suggest just adding to `CONTRIBUTING.md` under "Adding a new Container" something like
```
* Check that `CompatibleVectorizationPattern::allCompatibleVectorizationPattern` is correctly set.
```
I think that this is good enough - at the end of the day, if this is incorrectly set, the only real issue is all the vectorisation pattern options get trailed, but under-the-hood simply do the same thing.

SamNewcome (changes requested):
All good. Only some minor comment suggestions.

src/autopas/tuning/tuningStrategy/BayesianClusterSearch.cpp
```diff
@@ -15,10 +15,12 @@ autopas::BayesianClusterSearch::BayesianClusterSearch(
...
     const std::set<LoadEstimatorOption> &allowedLoadEstimatorOptions,
     const std::set<DataLayoutOption> &allowedDataLayoutOptions, const std::set<Newton3Option> &allowedNewton3Options,
-    size_t maxEvidence, AcquisitionFunctionOption predAcqFunction, const std::string &outputSuffix,
-    size_t predNumLHSamples, unsigned long seed)
+    const std::set<VectorizationPatternOption> &allowedVecPatternOptions, size_t maxEvidence,
+    AcquisitionFunctionOption predAcqFunction, const std::string &outputSuffix, size_t predNumLHSamples,
+    unsigned long seed)
     : _interactionType(interactionType),
       _containerOptionsSet(allowedContainerOptions),
+      _vecPatternOptions(allowedVecPatternOptions.begin(), allowedVecPatternOptions.end()),
```
thesamriel: I think this should go beneath the N3 initialization.
(There's a compiler warning on the initialization order here)

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1434 @@
...
+    MaskLong ownedMaskJLong = highway::Ne(ownedStateJLong, highway::Zero(tag_long));
+
+    // convert to a double mask since we perform logical operations with other double masks in the kernel.
+    ownedMaskJ = highway::RebindMask(tag_double, ownedMaskJLong);
+  }
+
+  template <bool remainderI, bool remainderJ, bool reversed, VectorizationPattern vecPattern>
+  inline void fillPhysicsRegisters(const size_t *const typeID1Ptr, const size_t *const typeID2Ptr,
+                                   VectorDouble &epsilon24s, VectorDouble &sigmaSquareds, VectorDouble &shift6s,
+                                   const unsigned int j, const unsigned int restI, const unsigned int restJ) {
```
thesamriel: `j`, the function parameter seems to not be used inside the function. All usages of j seem to be redefinitions.

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1434 @@
...
+   */
+  explicit LJFunctorHWY(double cutoff, std::optional<std::reference_wrapper<ParticlePropertiesLibrary<double, size_t>>>
+                                           particlePropertiesLibrary = std::nullopt)
+      : autopas::PairwiseFunctor<Particle_T, LJFunctorHWY>(cutoff),
+        _cutoffSquareAoS{cutoff * cutoff},
+        _potentialEnergySum{0.},
+        _virialSum{0., 0., 0.},
+        _aosThreadData{},
+        _postProcessed{false},
+        _PPLibrary{particlePropertiesLibrary} {
```
thesamriel: The ordering is not quite correct.
Also, `_potentialEnergySum` and `_aosThreadData` are already zero initialized at their declaration.
For consistency, I would probably zero initialize all of these at declaration and remove them from the constructor

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1434 @@
...
+
+  /**
+   * Specifies whether the functor is capable of using the specified Vectorization Pattern in the SoA functor.
+   * Thi functor can handle p1xVec, p2xVecDiv2, pVecDiv2x2, pVecx1
+   *
+   * @param vecPattern
+   * @return whether the functor is capable of using the specified Vectorization Pattern
+   */
+  bool isVecPatternAllowed(const VectorizationPattern vecPattern) override final {
+    return std::find(_vecPatternsAllowed.begin(), _vecPatternsAllowed.end(), vecPattern) != _vecPatternsAllowed.end();
```
thesamriel:
```suggestion
    return std::ranges::find(_vecPatternsAllowed, vecPattern) != _vecPatternsAllowed.end();
```
I like ranges :D

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1434 @@
...
+  }
+
+  /**
+   * Specifies whether the functor is capable of using the specified Vectorization Pattern in the SoA functor.
+   * Thi functor can handle p1xVec, p2xVecDiv2, pVecDiv2x2, pVecx1
+   *
+   * @param vecPattern
+   * @return whether the functor is capable of using the specified Vectorization Pattern
+   */
+  bool isVecPatternAllowed(const VectorizationPattern vecPattern) override final {
```
thesamriel:
```suggestion
  bool isVecPatternAllowed(const VectorizationPattern vecPattern) final {
```

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1434 @@
...
+      const VectorDouble fyNew = highway::Add(oldFy, fyAcc);
+      const VectorDouble fzNew = highway::Add(oldFz, fzAcc);
+
+      remainder ? highway::StoreN(fxNew, tag_double, &fxPtr[i], restI) : highway::StoreU(fxNew, tag_double, &fxPtr[i]);
+      remainder ? highway::StoreN(fyNew, tag_double, &fyPtr[i], restI) : highway::StoreU(fyNew, tag_double, &fyPtr[i]);
+      remainder ? highway::StoreN(fzNew, tag_double, &fzPtr[i], restI) : highway::StoreU(fzNew, tag_double, &fzPtr[i]);
+    }
+  }
+
+  template <bool newton3>
```
thesamriel:
```suggestion
```
doesn't use `newton3`

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1434 @@
...
+      }
+    }
+
+    MaskLong ownedMaskILong = highway::Ne(ownedStateILong, highway::Zero(tag_long));
+
+    // convert to a double mask since we perform logical operations with other double masks in the kernel.
+    ownedMaskI = highway::RebindMask(tag_double, ownedMaskILong);
+  }
+
+  template <bool remainder, bool reversed, VectorizationPattern vecPattern>
```
thesamriel:
```suggestion
  template <bool remainder, VectorizationPattern vecPattern>
```

thesamriel (changes requested):
Looks good, mostly comments on tidyness.
It might be nice to mention somewhere that only the LJ highway functor supports the vector patterns and LJFunctorAVX/SVE don't. Not sure where the  best place for this is, though.

SamNewcome:
> It might be nice to mention somewhere that only the LJ highway functor supports the vector patterns and LJFunctorAVX/SVE don't. Not sure where the best place for this is, though.

Done in 7e64ebd8

I added something about this to `AllOptions.yaml` and all the functor classes. It is not tragic if other patterns are provided with functors that don't support them - these configurations are simply rejected.

thesamriel (approved):

SamNewcome:
I have also in this branch now fixed #1008

.github/CONTRIBUTING.md
```diff
@@ -131,6 +131,10 @@ Possible log levels are:`trace`, `debug`, `info`, `warn`, `err`, `critical`, `of
 * Create a new set of compatible traversals in [`CompatibleTraversals`](/src/autopas/containers/CompatibleTraversals.h).
 * Create a new `case` statement in [`StaticContainerSelector`](/src/autopas/utils/StaticContainerSelector.h).
 * Add a case for the new container in [`ContainerSelector::generateContainer()`](/src/autopas/tuning/selectors/ContainerSelector.h).
+* Check [`CompatibleVectorizationPattern::allCompatibleVectorizationPattern()`](...) is correct.
```
thesamriel: This is missing the link I think

examples/md-flexible/CMakeLists.txt
```diff
@@ -30,6 +31,7 @@ target_compile_definitions(
         $<$<BOOL:${MD_FLEXIBLE_FUNCTOR_SVE}>:MD_FLEXIBLE_FUNCTOR_SVE>
         $<$<BOOL:${MD_FLEXIBLE_FUNCTOR_ATM_AUTOVEC}>:MD_FLEXIBLE_FUNCTOR_ATM_AUTOVEC>
         $<$<BOOL:${MD_FLEXIBLE_CALC_GLOBALS}>:MD_FLEXIBLE_CALC_GLOBALS>
+        $<$<BOOL:${MD_FLEXIBLE_FUNCTOR_HWY}>:MD_FLEXIBLE_FUNCTOR_HWY>
```
thesamriel: Can you move it 2 lines up for consistency?

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1462 @@
...
+                        const VectorDouble &z1, const double *const __restrict x2Ptr,
+                        const double *const __restrict y2Ptr, const double *const __restrict z2Ptr,
+                        double *const __restrict fx2Ptr, double *const __restrict fy2Ptr,
+                        double *const __restrict fz2Ptr, const size_t *const typeID1Ptr, const size_t *const typeID2Ptr,
+                        VectorDouble &fxAcc, VectorDouble &fyAcc, VectorDouble &fzAcc, VectorDouble &virialSumX,
+                        VectorDouble &virialSumY, VectorDouble &virialSumZ, VectorDouble &uPotSum,
+                        const unsigned int restI, const unsigned int restJ) {
+    VectorDouble epsilon24s = highway::Undefined(tag_double);
+    VectorDouble sigmaSquareds = highway::Undefined(tag_double);
+    VectorDouble shift6s = highway::Undefined(tag_double);
```
thesamriel, line 1031: In the case of `applyShift` being false, `shift6s` stays undefined, but is still added to the potential energy later. initialize it to 0 instead?

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1462 @@
...
+    // We overestimate the array size. This should be a tight/perfect upper bound on x86, and should never be large
+    // enough on e.g. ARM/RISC-V to cause issues.
+    HWY_ALIGN std::array<double, _maxVecLengthDouble> x2Tmp{};
+    HWY_ALIGN std::array<double, _maxVecLengthDouble> y2Tmp{};
+    HWY_ALIGN std::array<double, _maxVecLengthDouble> z2Tmp{};
+    HWY_ALIGN std::array<double, _maxVecLengthDouble> fx2Tmp{};
+    HWY_ALIGN std::array<double, _maxVecLengthDouble> fy2Tmp{};
+    HWY_ALIGN std::array<double, _maxVecLengthDouble> fz2Tmp{};
+    HWY_ALIGN std::array<size_t, _maxVecLengthDouble> typeID2Tmp{};
+    // ownedStates2Tmp is int64_t because we can directly use static_case on the individual elements below, unlike other
```
thesamriel: `static_cast`

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1462 @@
...
+    // functors where we have to resort to unsafe reinterpret_cast
```
thesamriel, line 1182: Why do we have to use `reinterpret_cast` in the other functors?

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1462 @@
...
+
+    reduceAccumulatedForce<reversed, remainderI, vecPattern>(i, fxPtr1, fyPtr1, fzPtr1, fxAcc, fyAcc, fzAcc, restI);
+  }
+
+  /**
+   * Templatized version of SoAFunctorSingle
+   * @tparam newton3 Whether to use newton3
+   * @param soa
+   */
+  template <bool newton3>
```
thesamriel: Can remove the newton3 template here as it is not used

src/autopas/options/VectorizationPatternOption.h
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * Interact two particles from the first list with half the vector length from the second list
+     *
+     * ---------------------------------
+     * | i | i | i | i |i+1|i+1|i+1|i+1|
+     * ---------------------------------
+     * | j |j+1|j+2|j+3| j |j+1|j+2|j+3|
+     *  --------------------------------
+     */
+    p2xVecDiv2,
```
thesamriel, line 51: What happens if you compile HWY functor on a machine without vector instructions?
Would highway give a compiler error? Or would it be compiled for a 'single-lane vector', which would be problematic with these vector patterns?

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1462 @@
...
+    MaskLong ownedMaskILong = highway::Ne(ownedStateILong, highway::Zero(tag_long));
+
+    // convert to a double mask since we perform logical operations with other double masks in the kernel.
+    ownedMaskI = highway::RebindMask(tag_double, ownedMaskILong);
+  }
+
+  template <bool remainder, VectorizationPattern vecPattern>
+  static void handleNewton3Reduction(const VectorDouble &fx, const VectorDouble &fy, const VectorDouble &fz,
+                                     double *const __restrict fx2Ptr, double *const __restrict fy2Ptr,
+                                     double *const __restrict fz2Ptr, const size_t i, const size_t j,
```
thesamriel: `i` is not used and could be removed

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -0,0 +1,1462 @@
...
+
+  inline void computeGlobals(const VectorDouble &virialSumX, const VectorDouble &virialSumY,
+                             const VectorDouble &virialSumZ, const VectorDouble &uPotSum) {
+    const int threadnum = autopas::autopas_get_thread_num();
+
+    const std::array<double, 4> globals{
+        highway::ReduceSum(tag_double, virialSumX), highway::ReduceSum(tag_double, virialSumY),
+        highway::ReduceSum(tag_double, virialSumZ), highway::ReduceSum(tag_double, uPotSum)};
+
+    _aosThreadData[threadnum].virialSum[0] += globals[0];
```
thesamriel: Why have this intermediate `std::array` here?
Could just do `_aosThreadData[threadnum].virialSum[0] += highway::ReduceSum(tag_double, virialSumX);` I'd say.

examples/md-flexible/src/Simulation.cpp
```diff
@@ -141,6 +150,8 @@ Simulation::Simulation(const MDFlexConfig &configuration,
                                               autopas::InteractionTypeOption::pairwise);
   _autoPasContainer->setAllowedTraversals(_configuration.traversalOptions.value,
                                           autopas::InteractionTypeOption::pairwise);
+  _autoPasContainer->setAllowedVecPatterns(_configuration.vecPatternOptions.value,
```
thesamriel, line 153: set it also for triwise, even if it's not yet supported, because it's already added to the search space later

applicationLibrary/molecularDynamics/tests/singleSiteTests/LJFunctorTestHWY.cpp
```diff
@@ -0,0 +1,628 @@
...
+}
+
+/**
+ * Checks that two non empty SoAs' particles are equal
+ *
+ * @tparam SoAType
+ * @param soa1
+ * @param soa2
+ * @return true if the two SoAs' particles are equal
+ */
```
thesamriel: All functions are documented twice, in the header and cpp. One place should be enough

thesamriel (changes requested):

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
SamNewcome: Because we have to reinterpret ownership state as an integer

src/autopas/options/VectorizationPatternOption.h
SamNewcome: In theory HWY compiles on non vector instructions. I don't know how this would behave in that case. But I'm not aware of any machine which we might want to run on not supporting at least two lanes. And anwsering this would be some work and as this PR had already been approved, I really just want to focus on anything thrown up between then and now (i.e. the recent few commits changing CI and CMake).

examples/md-flexible/src/Simulation.cpp
SamNewcome: I'd rather not implement this because it wasn't a feature in the already approved version of this PR. The only reason this PR wasn't merged back then was because there were updates to the master and resulting issues with the CI and CMake.

This would also not be so trivial. The current vector patterns don't make much sense for 3B and you would probably need new 3B patterns. I believe when David was working on the ATM HWY branch we discussed using 1xVec as a palceholder for now.

.github/CONTRIBUTING.md
SamNewcome: Done in 305cbeec

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
SamNewcome: Done in 305cbeec

examples/md-flexible/CMakeLists.txt
SamNewcome: Done in 305cbeec

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
SamNewcome: Done in 305cbeec

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
SamNewcome: Done in 305cbeec

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
SamNewcome: Done in 305cbeec

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
SamNewcome: Done in 305cbeec

applicationLibrary/molecularDynamics/tests/singleSiteTests/LJFunctorTestHWY.cpp
SamNewcome: Done in 305cbeec

src/autopas/options/VectorizationPatternOption.h
thesamriel: Yeah, this was more of a question to understand, no need to assure of this now.

examples/md-flexible/src/Simulation.cpp
thesamriel: I didn't mean to implement it for triwise. It just felt odd, that later we create the triwise search space for all vector patterns.
I guess the other configs are thrown out later during the simulation when doing the `functor.supportsVecPattern` check, which is not a huge overhead. 
All good then

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
thesamriel: Okay, apparently because when casting the pointers we need reinterpret, got it!

thesamriel (approved):

examples/md-flexible/src/Simulation.cpp
SamNewcome: Yeah, I get what you mean. I guess longer term we would have a different list of vec pattern options for 3B. I'd probably say that once we have that we make a separate setter.

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
SamNewcome: Yeah, sorry I didn't explain well. This is imo, something quite easy to break because you are not doing any actual conversion but simply pretending that it is a integer. I think in most compilers this should work, but there are no guarantees and if we were to change the underlying type of the ownership state to e.g. a different integer type with less bytes, it would probably break. Because here we are casting the actual values and not pointers, we don't have to deal with this.

SamNewcome (approved):

PR #843, opened 2024-04-04
3Body Verlet List Traversals AoS

Alexander-Haberl-TUM:
# Description

### Adding 3-Body AoS traversals to the VerletLists container:
- ListIterationTraversal3B: iterates over 2-Body neighbor lists in a double for loop to find all triplets
- ListIntersectionTraversalSorted3B: finds all triplets by intersecting the 2-Body neighbor lists of all pairs of neighboring particles to find their common neighbors, intersection is done through std::set_intersection() which requires the neighborlists to be sorted
- ListIntersectionTraversalHashing3B: finds triplets like ListIntersectionTraversalSorted3B, but intersection is done by creating a hash set of one neighbor list and then checking if the entries of the other neighborlist are in there
- PairListIterationTraversal3B: relies on 3-Body neighbor lists which store pairs of neighbors such that all three distances between the particles are less than cutoff + skin, iterates over these lists once

### Added a C08 based traversal for building 2-Body neighborlists for use by 3-Body traversals:
The ListIntersectionTraversals and the newton3 version of the ListIterationTraversal3B require full neighbor lists of Halo particles, therefore we need a Linked Cells traversal that computes all interactions between Halo particles for neighbor list building.
- new classes: CellFunctorNeighborListBuild3B, C08BasedNeighborListBuilding3B, LCC08CellHandlerNeighborList3B, LCC08NeighborListBuilding3B

### Changes to VerletLists.h:
- added iterateTriwise()
- added rebuildNeighborLists(TriwiseTraversal)
- updateNeighborLitstsAoS3B(newton3)
- updatePairwiseNeighborListsAoS3B(newton3): updates the 3-Body neighborLists
- generatePairwiseAoSNeighborLists(): generates empty 3-Body neighborlists for all particles

# How Has This Been Tested?

- ran TraversalComparison3B() to check force calculation against other 3-Body traversals

PR #981, opened 2024-12-04
Mini rework of neighborlist validity control

thesamriel:
# Description

## Problem

During the preparation of #843, I ran into problems with neighbor lists not rebuilding in the case where I use LinkedCells for pairwise traversals and Verlet Lists for triwise interactions.
The reason for this is the logic behind `_neighborListsAreValid` in `LogicHandler.h`. The old logic with the newly added triwise interactions was as follows:
In `LogicHandler::updateContainer()`, which is called once per timestep, we check whether neighbor lists will be rebuilt based on current iteration counters. If so, `_neighborListsAreValid` is set to false. During the next call to `computeInteractions()`, the neighbor lists will be built and `_neighborListsAreValid` is set back to true.
If we now have another call to `computeInteractions()` in the same timestep with a different Container (e.g. pairwise then triwise), the neighbor lists for the new container will not be built since `_neighborListsAreValid` is still true.

## Solution

I tried to make the logic clearer and safer.
- The method `bool neighborListsAreValid()` is refactored to `void setNeighborListsStatus()`, as the methods main objective is setting the state member variable `_neighborListsAreValid`. It is supposed to be called once per timestep (e.g. in `updateContainer()`) to set the status of neighbor lists depending on current iteration counters.
- The status for neighbor lists is determined in another method than the actual rebuild. This creates windows where a user might (accidentally) destroy the neighbor lists in between, which would go unnoticed by AutoPas. Therefore, particles are *always* added to the LogicHandler buffers first and later moved to the container at the next rebuild step. This can be avoided by setting `forceToContainer` to true, but should be used with caution.

## Alternatives

Alternatively, I considered adding separate state variables for neighbor lists for pairwise and triwise traversals.
I like the implemented approach more, as it provides additional safety and simplicity. It comes with the potential downside of unnecessary particle copies during rebuild iterations, where particles are added to the buffers and immediately afterwards moved to the container.

# How Has This Been Tested?

- All existing tests
- New test `tests/logicHandler/RebuildNeighborListsTest` from branch 3bodyVerletList


src/autopas/LogicHandler.h
```diff
@@ -158,21 +159,22 @@ class LogicHandler {
...
-    bool doDataStructureUpdate = not neighborListsAreValid();
+    setNeighborListsStatus();
+    const bool doDataStructureUpdate = not _neighborListsAreValid.load(std::memory_order_relaxed);
+
+    // We will do a rebuild in this timestep
+    if (doDataStructureUpdate) {
+      _stepsSinceLastListRebuild = 0;
+    }
+    ++_stepsSinceLastListRebuild;
+    _containerSelector.getCurrentContainer().setStepsSinceLastRebuild(_stepsSinceLastListRebuild);
```
SamNewcome, line 170: Why was this removed from the `if (_functorCalls > 0)` branch?

src/autopas/LogicHandler.h
```diff
@@ -317,7 +319,7 @@ class LogicHandler {
     }
     Particle particleCopy = p;
     particleCopy.setOwnershipState(OwnershipState::owned);
-    if (not _neighborListsAreValid.load(std::memory_order_relaxed)) {
+    if (forceToContainer) {
       // Container has to (about to) be invalid to be able to add Particles!
```
SamNewcome, line 323: Update comment

src/autopas/LogicHandler.h
```diff
@@ -344,7 +346,7 @@ class LogicHandler {
     }
     Particle haloParticleCopy = haloParticle;
     haloParticleCopy.setOwnershipState(OwnershipState::halo);
-    if (not _neighborListsAreValid.load(std::memory_order_relaxed)) {
+    if (forceToContainer) {
       // If the neighbor lists are not valid, we can add the particle.
```
SamNewcome, line 350: Update comment

tests/testAutopas/tests/logicHandler/RebuildNeighborListsTest.cpp
```diff
@@ -0,0 +1,162 @@
...
+                                                            _cellSizeFactor,
+                                                            autopas::TraversalOption::vl_pair_list_iteration_3b,
+                                                            autopas::LoadEstimatorOption::none,
+                                                            autopas::DataLayoutOption::aos,
+                                                            autopas::Newton3Option::disabled,
+                                                            autopas::InteractionTypeOption::triwise};
+  return {_confLc_c01_noN3_3B, _confVl_list_it_noN3_3B, _confVl_pairlist_noN3_3B};
+}
+
+TEST_P(RebuildNeighborListsTest, testRebuildDifferentContainerPairwiseTriwise) {
```
SamNewcome, line 59: Add a short comment explaining the test

SamNewcome (changes requested):
Can you check the performance of this compared to master? I don't expect any significant performance cost of adding particles not directly to the container.

It appears that the `forceToContainer` mechanic is only useful for tests. I'm reluctant to add this dangerous mechanic when it is only useful there. Could you come up with an alternative mechanic? Or alternatively, use `updateContainer` in the tests to put the particles inside the container.

src/autopas/LogicHandler.h
thesamriel: I simply deleted it. I think the new code is self-explanatory.

src/autopas/LogicHandler.h
thesamriel: I simply deleted it. I think the new code is self-explanatory.

thesamriel:
> Can you check the performance of this compared to master? I don't expect any significant performance cost of adding particles not directly to the container.

Sure, I'll add it as soon as I have access to HSUper again. (I might merge it before that though (not in master), since I need this in the `3bodyVerletList` branch.)

> It appears that the `forceToContainer` mechanic is only useful for tests. I'm reluctant to add this dangerous mechanic when it is only useful there. Could you come up with an alternative mechanic? Or alternatively, use `updateContainer` in the tests to put the particles inside the container.

I also wanted to keep this option for the user, but tbh., I also can't think of a very reasonable scenario where it might be needed. So I would not be totally against removing this option.

SamNewcome:
> > Can you check the performance of this compared to master? I don't expect any significant performance cost of adding particles not directly to the container.
> 
> Sure, I'll add it as soon as I have access to HSUper again. (I might merge it before that though (not in master), since I need this in the `3bodyVerletList` branch.)
> 

CoolMUC4 would also be fine if you wanted to do this beforehand.

thesamriel:
bro, why is this still up

PR #982, opened 2024-12-06
Add box padding warning

thesamriel:
# Description

When particles are generated on or close to the simulation box boundaries, we add some padding to avoid particles being too close to the boundary.
This can lead to confusion since the simulation has a different density than expected. I had problems with this myself and many students as well. That's why I think a warning is adequate.

## How was this tested?
- [x] Check manually.
- [x] CI

### DIFF
--- a/examples/md-flexible/src/configuration/MDFlexConfig.cpp
+++ b/examples/md-flexible/src/configuration/MDFlexConfig.cpp
@@ -453,6 +453,8 @@ std::string MDFlexConfig::to_string() const {
 
 void MDFlexConfig::calcSimulationBox() {
   const double interactionLength = cutoff.value + verletSkinRadiusPerTimestep.value * verletRebuildFrequency.value;
+  const auto preBoxMin = boxMin.value;
+  const auto preBoxMax = boxMax.value;
 
   // helper function so that we can do the same for every object collection
   // resizes the domain to the maximal extents of all objects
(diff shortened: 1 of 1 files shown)


manishmishra6016 (approved):
I think it's a good idea :+1:

examples/md-flexible/src/configuration/MDFlexConfig.cpp
```diff
@@ -477,6 +479,10 @@ void MDFlexConfig::calcSimulationBox() {
   resizeToObjectLimits(sphereObjects);
   resizeToObjectLimits(cubeClosestPackedObjects);
 
+  if (boxMin.value != preBoxMin or boxMax.value != preBoxMax) {
+    std::cout << "WARNING: Simulation box increased due to particles being too close to the boundaries." << std::endl;
+  }
```
FG-TUM, line 484: If you want, you could add the information from which value it changed. You could also omit this because the user could deduce this by comparing the input with the log output. But since you are targeting users who are not aware of this, they might not think about doing this.

FG-TUM (approved):

examples/md-flexible/src/configuration/MDFlexConfig.cpp
thesamriel: The configuration log output is pretty much directly beneath the warning, which is why I would omit it for now.

PR #984, opened 2024-12-10
Mixed precision

TheWhale2307:
# Description

An attempt at adding mixed precision for certain parts of the code for AutoPas

## ~Resolved Issues~

# How Has This Been Tested?

All currently existing tests and possibly own additional ones in the future


thesamriel:
Unfortunately has become out of date and would need to be redone.

PR #985, opened 2024-12-11
Refactoring of liveInfo

SamNewcome:
# Description

This PR includes a number of fixes, tidys, and extensions to LiveInfo.

Fixes:
- The number of owned particles was determined by getting the number of owned particles and then subtracting the number of halo particles. This is now fixed.
- Live Info previously only considered particles in the container and not particles in the buffer. These particles are now included in the statistics. This is important because  (a) these particles may enter the container upon rebuild and so are important for choosing the container and (b) I want the live info to be configuration independent to avoid current configurations influencing the live info which influences tuning strategies. Technically, currently, the configuration does not influence the size of the buffer (I think?), but we are looking at changes where it will.
- `relativeParticlesPerCellStdDev` (originally `particlesPerCellStdDev` (the metric was always relative, just renamed to make this clearer) is now correctly calculated (the division by the number of cells is now inside the square root)

Tidys:
- The `SimilarityFunctions` statistics (homogeneity and maxDensity) were removed from `LiveInfo`, as I found the statistics to be not very suitable, and decided the overhead was not worth it. I kept the functions to be used for their original purpose in `MPIParallelizedTuning`, because I didn't want to spend time showing that it is beneficial to change these.
- This also means that the frule files, which used "homogeneity", don't work anymore. This info was simply removed from the files and a warning was added.
- Functor statistics are removed, removing the need for live info to be templated to one specific functor.
- A new class, `ParticleBinStructure` is added, which tidies all binning statistics into this class, reducing the code duplication between the same statistics for multiple types of bins. 
- The bins in live info now benefit from the safe indexing originally used in the `SimilarityFunctions` bins. 
- Some info names were made more descriptive.
- The "cells" used for statistics match those of Linked Cells with cell size factor 1. Specifically, the base cell length is `interactionLength` not `cutoff` and they are stretched to fit the domain.
- The calculation of `estimatedNumNeighborInteractions` is explained in comments and is generalised to work with any cell dimensions as a result of the above.
- Unit tests are actually provided for LiveInfo now, so this should hopefully not be a problem in the future.

Extensions:
- Added some statistics to `liveInfo`. These are the median and upper and lower quartile particle counts per bin. 

Misc:
- The `floorToInt` and `ceilToInt` ArrayMath functions are generalised to cast to any specified type.

# How Has This Been Tested?

- [x] CI
- [x] Checking live info output


src/autopas/utils/ParticleBinStructure.h
```diff
@@ -0,0 +1,357 @@
...
+   * @param boxMin Lower left corner of region considered for particle binning.
+   * @param boxMax Upper right corner of region considered for particle binning.
+   * @param cutoff Not needed for particle counting, but needed for estimatedNumberOfNeighborInteractions statistic.
+   */
+  ParticleBinStructure(size_t numBinsPerDim, std::array<double, 3> binLength, std::array<double, 3> boxMin,
+                       std::array<double, 3> boxMax, double cutoff)
+      : ParticleBinStructure({numBinsPerDim, numBinsPerDim, numBinsPerDim}, binLength, boxMin, boxMax, cutoff){};
+
+  /**
+   * Adds to the counter for the bin the particle falls into.
```
manishmishra6016, line 62:
```suggestion
   * Determines the appropriate bin for the particle based on its position, and
   * increments the corresponding bin counter to reflect one more particle in that bin.
```

src/autopas/utils/ArrayMath.h
```diff
@@ -322,33 +322,37 @@ template <class T, std::size_t SIZE>
...
- * Floors all array elements and converts them to integers.
- * @tparam T floating point type
+ * Floors all array elements and converts them to a different type. Useful to floor array elements and cast them to
+ * some integer type.
+ * @tparam target_T target type. By default int.
+ * @tparam float_T floating point type
  * @tparam SIZE size of the array
  * @param a input array
- * @return New array with floored elements of new type int.
+ * @return New array with floored elements of new type.
```
manishmishra6016, line 331: if I understand correctly, this just means that user can specify precision and type of int (unsigned, signed, long) here?

src/autopas/utils/ArrayMath.h
```diff
@@ -322,33 +322,37 @@ template <class T, std::size_t SIZE>
...
  */
-template <class T, std::size_t SIZE>
-[[nodiscard]] constexpr std::array<int, SIZE> floorToInt(const std::array<T, SIZE> &a) {
-  std::array<int, SIZE> result{};
+template <typename target_T = int, typename float_T, std::size_t SIZE>
```
manishmishra6016, line 333: how do we ensure that `target_T` is only an integer type? Does it automatically throw error when user tries to cast to, let's say `char`?

manishmishra6016 (commented):
Generally looks really good! It would be really useful. 

Just another point: If we are not sure about how well fuzzy tuning will perform now, maybe it is worth documenting somewhere the last commit where it performed as expected. Is my assessment correct that it might not perform like before now? (with homogeneity being removed)

SamNewcome:
> Just another point: If we are not sure about how well fuzzy tuning will perform now, maybe it is worth documenting somewhere the last commit where it performed as expected. Is my assessment correct that it might not perform like before now? (with homogeneity being removed)

Yes, this is a good point. But I would change the phrasing that emphasises that the commit was the last one with all of the input parameters considered in training _not_ that this was the last commit where it performed well. Ultimately, these input files were trained from CoolMUC2 data, so really anybody seriously using fuzzy-logic with decision trees should create them anew regardless of what I do in this commit.

examples/md-flexible/input/fuzzyRulesComponents.frule
```diff
@@ -161,66 +161,56 @@ OutputMapping:
 
 ## Container_DataLayout
 
-if ("homogeneity" == "lower than 0.049") && ("threadCount" == "lower than 18.0")
+if ("particlesPerCellStdDev" == "lower than 0.014") && ("meanParticlesPerCell" == "lower than 1.553") && ("threadCount" == "lower than 18.0")
```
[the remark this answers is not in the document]
SamNewcome: They aren't - this is just git. All I did was remove lines with "homogeneity". I do not expect the new rules to work well (but I have never tested how they perform). 

My reasoning: I wanted to remove "homogeneity", as in my experiments, it was a bit of a mess of a statistic and didn't really contribute to whatever algorithm was optimal. I think my original suggestion to Manuel to include this in live info really came just from a lack of understanding about the metric. 

As the `frule` file is probably unsuitable anyway for any "power user" (because it is based on performance data from a cluster that no longer exists), the importance of this file is more to provide a reference. Hence, I wanted to keep it but didn't want to spend time seriously updating the rule file, so I simply removed what would no longer work.

src/autopas/tuning/AutoTuner.cpp
```diff
@@ -371,41 +371,32 @@ long AutoTuner::estimateRuntimeFromSamples() const {
   return (reducedValueBuilding + (_rebuildFrequency - 1) * reducedValueNotBuilding) / _rebuildFrequency;
 }
 
-bool AutoTuner::prepareIteration() {
+void AutoTuner::sendDomainSimilarityStatisticsAtStartOfTuningPhase() {
   // Flag if this is the first iteration in a new tuning phase
   const bool startOfTuningPhase = _iteration % _tuningInterval == 0 and not _isTuning;
```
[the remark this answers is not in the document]
SamNewcome: The idea was that the old `prepareIteration` did two things: send domain similarity statistics at the start of a tuning phase, and return if live info needs to be gathered. I chose to split this into two seperate functions: `sendDomainSimilarityStatisticsAtStartOfTuningPhase()` and `needsLiveInfo()`, because it seemed cleaner to me. But this is not something I feel strongly about.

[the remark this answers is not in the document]
SamNewcome: This was actually implemented slightly incorrectly and wouldn't have sent domain similarity statistics. This was corrected in 28f2278.

src/autopas/tuning/AutoTuner.h
```diff
@@ -92,13 +92,17 @@ class AutoTuner {
   bool needsHomogeneityAndMaxDensityBeforePrepare() const;
```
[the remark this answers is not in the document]
SamNewcome: Done in 28f2278 (documentation) and 133f064 (function name)

src/autopas/utils/ArrayMath.h
SamNewcome: Yes exactly

src/autopas/utils/ArrayMath.h
SamNewcome: The code fails to compile. It's messy, but we can't really solve this without upgrading to C++20.

SamNewcome: Something will go wrong. Not totally sure if a compiler error will occur or if this will be a runtime error. We can handle this nicer if we upgrade to C++20, but this is outside the scope of this PR.

src/autopas/tuning/tuningStrategy/LiveInfo.h
```diff
@@ -377,6 +395,13 @@ class LiveInfo {
     return var;
   }
 
+  template <class Variant, size_t... Idx>
+  static Variant g(std::istream &in, size_t idx, std::index_sequence<Idx...>) {
```
thesamriel, line 399: Is this missing documentation?

src/autopas/tuning/tuningStrategy/LiveInfo.h
```diff
@@ -33,6 +34,53 @@ class LiveInfo {
...
+  /**
+   * Returns a bin structure the domain is divided into 3x3x3 uniform bins.
+   *
+   * @param domainSize size of (sub)domain
+   * @param boxMin Lower left corner of (sub)domain.
+   * @param boxMax Upper right corner of (sub)domain.
+   * @param cutoff Cutoff.
+   * @return
+   */
+  static utils::ParticleBinStructure buildBlurredBinStructure(const std::array<double, 3> &domainSize,
```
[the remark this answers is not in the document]
SamNewcome: Not totally sure - this was the original terminology before my refactor, I guess the idea is that it is more coarse-grained than the cells-bins. Happy to change this if you have a better suggestion.

src/autopas/utils/ArrayMath.h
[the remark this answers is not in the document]
SamNewcome: So that you can floor/cast to something other than `int`.

src/autopas/tuning/tuningStrategy/LiveInfo.h
SamNewcome: Yes. I'm not sure if we require documentation for `private` functions? Fyi, this isn't a function I added, but already existed, and I would honestly need a little time to come up with good documentation for this.

SamNewcome: Ignore what I previously said - it appears I accidentally copy and pasted the already existing `readIndex` function with new function name `g` for some reason. I've removed it in d14d8a89

manishmishra6016 (approved):
Looks good to me to merge. :muscle: 

Perhaps we can still document the last commit with frule file working (i.e. homogeneity and maxDensity still available, [0243834](https://github.com/AutoPas/AutoPas/commit/024383466a66f434c40b64fe04546a24d48e35fb)). I understand the rules were generated on CoolMUC-2 cluster, which is no longer available. But we have never shown (?) how bad rules get when you change hardware, as an example when I initially tested them on my laptop, they still performed well, which is essentially a different platform. So, removing these parameters perhaps have greater impact on their performance.

We can also decide if we want to keep them or not. If yes, it might be nice to update the rule for CoolMUC-4/HSUper where we mostly use them. Of course, this would be a separate PR.

src/autopas/utils/ArrayMath.h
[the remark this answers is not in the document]
thesamriel: What do you think about renaming to `floorAndCast`?
"Casted" sounds kind of wrong to me: https://gramomat.com/en/english/conjugation/cast 😅

SamNewcome:
> Perhaps we can still document the last commit with frule file working (i.e. homogeneity and maxDensity still available, [0243834](https://github.com/AutoPas/AutoPas/commit/024383466a66f434c40b64fe04546a24d48e35fb)). I understand the rules were generated on CoolMUC-2 cluster, which is no longer available. But we have never shown (?) how bad rules get when you change hardware, as an example when I initially tested them on my laptop, they still performed well, which is essentially a different platform. So, removing these parameters perhaps have greater impact on their performance.
> 
> We can also decide if we want to keep them or not. If yes, it might be nice to update the rule for CoolMUC-4/HSUper where we mostly use them. Of course, this would be a separate PR.

No, you're right - all I have is some preliminary data from my 2023 paper, which suggests that there are at least differences between an Intel workstation and an ARM-based cluster - and even then, the preliminary data only suggests some suboptimality but not complete changes in what is optimal - it would be naive to make these rules redundant.

Part of my reasoning from removing these was:

1. It discourages the use of these metrics (by not making it possible to use them)
2. LiveInfo does have a not-insignificant performance cost. Only using it in occasional tuning phases is fine, but this is noticeable if you have the live info logger enabled. Removing the extra binning technique should save some time. This aids in data gathering, where we want the logger enabled, as well as some scenarios where we might want to do frequent tuning.

One option would be to re-add homogeneity and max density to LiveInfo. The version used in my alg. sel. paper does include them, albeit under different names so this wouldn't require coding from scratch as I can reuse some older commits. This means that the `frule` files can remain the same aside from name changes.

What we could do (but probably best done in a later PR), is create some `LiveInfoRequest` struct, which the tuning strategy passes to the `AutoTuner`, which in turn passes to `LiveInfo`. It could contain a bool for each variable and skip its calculation if disabled. This way, we only gather live info that is requested, meaning we still save on performance.

src/autopas/utils/ArrayMath.h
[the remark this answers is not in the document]
SamNewcome: Yeah, this also sounds good. Done in 0d38c6f3

thesamriel (approved):

manishmishra6016:
> What we could do (but probably best done in a later PR), is create some `LiveInfoRequest` struct, which the tuning strategy passes to the `AutoTuner`, which in turn passes to `LiveInfo`. It could contain a bool for each variable and skip its calculation if disabled. This way, we only gather live info that is requested, meaning we still save on performance.

I think this sounds reasonable, we can leave it for a future PR. :smile:

SamNewcome:
### Refactoring of SimilarityFunctions

These were changes which I had originally made, but then removed as I removed `homogeneity` and `maxDensity` from `LiveInfo`. Manish made a good point about keeping these functions, so I re-added my refactoring work on these functions.

**Recap of the master:**
`SimilarityFunctions` calculated `homogeneity` and `maxDensity` statistics, originally for use in `MPIParallelizedTuning` but later added to `LiveInfo`. To be clear: `homogeneity` is simply just a particle bin standard deviation metric and is in no way superior than other homogeneity metrics.

**What I did**
- `SimilarityFunctions` is removed, as its binning logic is made redundant by `ParticleBinStructure`, giving "`homogenity`" and "`maxDensity`" all of the corrections made to the other particle bins with this function and there is now less redundant code.
- `ParticleBinStructure` now also calculates particle density, to produce "`homogeneity`" and "`maxDensity`"
- The bins previously used in `SimilarityFunctions` are now called `particleDependentBins`, as their main differentiating attribute is that their size depends on the number of particles. `homogeneity` is then renamed to `particleDependentBinDensityStdDev` and `maxDensity` to `particleDependentBinMaxDensity`. This also makes it clearer what `homogeneity` really was.

src/autopas/LogicHandler.h
```diff
@@ -1850,20 +1849,26 @@ std::tuple<Configuration, std::unique_ptr<TraversalInterface>, bool> LogicHandle
           }
         });
   } else {
-    if (autoTuner.needsHomogeneityAndMaxDensityBeforePrepare()) {
+    const auto needsDensityStatistics = autoTuner.needsDomainSimilarityStatistics();
```
thesamriel, line 1852: I find `DensityStatistics` easier to understand. Is there a reason for the different naming here? Or could it also be `autoTuner.needsDensityStatistics`?

src/autopas/LogicHandler.h
```diff
@@ -1850,20 +1849,26 @@ std::tuple<Configuration, std::unique_ptr<TraversalInterface>, bool> LogicHandle
...
+    const auto needsLiveInfo = autoTuner.needsLiveInfo();
+    if (needsDensityStatistics or needsLiveInfo) {
+      // Gather Live Info
       utils::Timer timerCalculateHomogeneity;
       timerCalculateHomogeneity.start();
-      const auto &container = _containerSelector.getCurrentContainer();
-      const auto [homogeneity, maxDensity] = autopas::utils::calculateHomogeneityAndMaxDensity(container);
+      auto particleIter = this->begin(IteratorBehavior::ownedOrHalo);
+      info.gather(particleIter, _neighborListRebuildFrequency, getNumberOfParticlesOwned(), _logicHandlerInfo.boxMin,
+                  _logicHandlerInfo.boxMax, _logicHandlerInfo.cutoff, _logicHandlerInfo.verletSkin);
```
thesamriel, line 1860: `_logicHandlerInfo` is only used during the construction of the LogicHandler I think. Might want to get the recent values from the container instead

src/autopas/LogicHandler.h
```diff
@@ -1850,20 +1849,26 @@ std::tuple<Configuration, std::unique_ptr<TraversalInterface>, bool> LogicHandle
...
+      if (needsDensityStatistics) {
+        const auto particleDependentBinDensityStdDev = info.get<double>("particleDependentBinDensityStdDev");
+        const auto particleDependentBinMaxDensity = info.get<double>("particleDependentBinMaxDensity");
+        autoTuner.addDomainSimilarityStatistics(particleDependentBinDensityStdDev, particleDependentBinMaxDensity,
+                                                timerCalculateHomogeneity.getTotalTime());
+        autoTuner.sendDomainSimilarityStatisticsAtStartOfTuningPhase();
+      }
+      if (needsLiveInfo) {
+        autoTuner.receiveLiveInfo(info);
+      }
```
thesamriel, line 1871: If the autotuner needs Density statistics, we anyways gather the full live info, right?
I feel like we could just do
```suggestion
      if (needsDensityStatistics or needsLiveInfo) {
        autoTuner.receiveLiveInfo(info);
      }
```
And then set handle the density stuff in the autotuner instead of having all this logic in here.

src/autopas/tuning/tuningStrategy/LiveInfo.h
```diff
@@ -41,179 +118,161 @@ class LiveInfo {
...
+   * - meanParticlesPerCell: The average number of particles per cell-bin.
+   * - particlesPerCellStdDev: The standard deviation in the number of particles per cell-bin.
+   * - relativeParticlesPerCellStdDev: The, relative to the mean, standard deviation in the number of particles per
+   * cell-bin.
+   * - estimatedNumNeighborInteractions: Rough estimation of number of neighbor interactions. Assumes that neighboring
+   * cell-bins contain roughly the same number of particles. Estimation does not work well if this is not the case.
+   * ---- Particle Dependent Bin Statistics ----
+   * - particleDependentBinMaxDensity: The maximum density of particles in a particle-dependent bin.
+   * - particleDependentBinDensityStdDev: The standard deviation in the density of particles in a particle-dependent
+   * bin. In earlier versions of AutoPas, this was referred to as "Homogeneity".
```
thesamriel, line 166:
```suggestion
   * - particleDependentBinDensityStdDev: The standard deviation in the density of particles per particle-dependent
   * bin. In earlier versions of AutoPas, this was referred to as "Homogeneity".
```

src/autopas/tuning/tuningStrategy/MPIParallelizedStrategy.h
```diff
@@ -82,16 +82,16 @@ class MPIParallelizedStrategy : public TuningStrategyInterface {
...
+   * The smoothed particle-dependent bin standard deviation density of the whole simulation.
+   * See LiveInfo::gather().
    */
-  double _smoothedHomogeneity{-1.};
+  double _smoothedPDBinStdDevDensity{-1.};
 
   /**
-   * The maximal density throughout the whole domain.
-   * See SimilarityFunctions::calculateHomogeneityAndMaxDensity().
+   * The smoothed maximum particle-dependent bin density throughout the whole domain.
```
thesamriel, line 91:
```suggestion
   * The smoothed maximum of the particle-dependent bin density throughout the whole simulation.
```

src/autopas/tuning/tuningStrategy/MPIParallelizedStrategy.h
```diff
@@ -82,16 +82,16 @@ class MPIParallelizedStrategy : public TuningStrategyInterface {
 
  private:
   /**
-   * The smoothed homogeneity of the whole simulation.
-   * See SimilarityFunctions::calculateHomogeneityAndMaxDensity().
+   * The smoothed particle-dependent bin standard deviation density of the whole simulation.
```
thesamriel, line 85:
```suggestion
   * The smoothed standard deviation of particle-dependent bin densities throughout the whole simulation.
```

src/autopas/LogicHandler.h
SamNewcome: The `_logicHandlerInfo` values _should_ be up to date, as the getters and setters for AutoPas use them. But we should probably have an issue about getting rid of these statistics from the container, because the duplication of these variables and the possibility that they might not be synced is a bit concerning.

src/autopas/LogicHandler.h
SamNewcome: Nice Idea. Done in c3c7699d. One bit of functionality I changed was that we no longer have a `DEBUG` level log of the cost of getting the domain similarity statistics. But I added a `DEBUG` level log of the cost of gathering live info - which means it can also be used with rule/fuzzy/decision tree tuning, which could be used to determine the cost of getting domain similarity statistics.

src/autopas/tuning/tuningStrategy/MPIParallelizedStrategy.h
SamNewcome: Similar to above for the last part. See 943ad321

src/autopas/LogicHandler.h
thesamriel: Okay, I see!

manishmishra6016 (commented):
Looks good! I can approve it once the CI pipeline is fixed. :)

manishmishra6016 (approved):

thesamriel (approved):

PR #989, opened 2025-01-07
Feature/additional-metrics

JannikHoog:
# Description

All preliminary changes made by Jannik Hoog for my Bachelor Thesis "Comparison of energy performance metrics for autotuning molecular simulations with Autopas" supervised my Manish

## Current Issues

at maybe half of the iterations the EnergyDelayProduct is negative.

### DIFF
--- a/examples/md-flexible/input/AllOptions.yaml
+++ b/examples/md-flexible/input/AllOptions.yaml
@@ -16,7 +16,7 @@ verlet-rebuild-frequency             :  20
 verlet-skin-radius-per-timestep      :  0.01
 verlet-cluster-size                  :  4
 selector-strategy                    :  Fastest-Absolute-Value
-tuning-metric                        :  time
+tuning-metric                        :  time, energy, energyPerFLOP, energyDelayProduct
 tuning-strategies                    :  [slow-config-filter, rule-based-tuning, predictive-tuning]
 tuning-interval                      :  5000
 tuning-samples                       :  3
--- a/src/autopas/LogicHandler.h
+++ b/src/autopas/LogicHandler.h
@@ -1085,7 +1085,10 @@ IterationMeasurements LogicHandler<Particle>::computeInteractions(Functor &funct
 
   functor.endTraversal(newton3);
   const auto [energyPsys, energyPkg, energyRam, energyTotal] = autoTuner.sampleEnergy();
+  const auto numFLOP = functor.getNumFLOPs();
+  const auto energyPerFLOP = energyTotal / numFLOP;
   timerTotal.stop();
+  const auto energyDelayProduct = energyTotal * timerTotal.getTotalTime();
 
   constexpr auto nanD = std::numeric_limits<double>::quiet_NaN();
   constexpr auto nanL = std::numeric_limits<long>::quiet_NaN();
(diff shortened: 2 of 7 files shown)


examples/md-flexible/input/AllOptions.yaml
```diff
@@ -16,7 +16,7 @@ verlet-rebuild-frequency             :  20
 verlet-skin-radius-per-timestep      :  0.01
 verlet-cluster-size                  :  4
 selector-strategy                    :  Fastest-Absolute-Value
-tuning-metric                        :  time
+tuning-metric                        :  time, energy, energyPerFLOP, energyDelayProduct
```
manishmishra6016, line 19:
```suggestion
tuning-metric                        :  time #alternatively: energy, energyPerFLOP, energyDelayProduct
```
Show other available options as a comment, as only one can be chosen at a time.

src/autopas/utils/logging/IterationLogger.cpp
```diff
@@ -36,7 +36,10 @@ autopas::IterationLogger::IterationLogger(const std::string &outputSuffix, bool
     csvHeader.append(
         ",energyPsys[J],"
         "energyPkg[J],"
-        "energyRam[J]");
+        "energyRam[J],"
+        "numFLOPs,"
+        "energyPerFLOP[J/#Flops],"
```
manishmishra6016, line 41: I think it might be nice to only have the metrics related to FLOP when the flop logger is turned ON. User might not want to use them if they have a performance overhead anyway.

src/autopas/utils/logging/IterationLogger.cpp
```diff
@@ -57,15 +60,15 @@ autopas::IterationLogger::~IterationLogger() {
 
 void autopas::IterationLogger::logIteration(const autopas::Configuration &configuration, size_t iteration,
                                             const std::string &functorName, bool inTuningPhase, long timeTuning,
-                                            const IterationMeasurements &measurements) const {
+                                            const IterationMeasurements &measurements, long numFLOPs) const {
```
manishmishra6016, line 63: isn't `numFLOPs` already part of the measurement? Why do you have it as a separate argument again?

src/autopas/LogicHandler.h
```diff
@@ -1641,7 +1647,7 @@ bool LogicHandler<Particle>::computeInteractionsPipeline(Functor *functor,
                measurements.energyPkg, measurements.energyRam);
   }
   _iterationLogger.logIteration(configuration, _iteration, functor->getName(), stillTuning, tuningTimer.getTotalTime(),
-                                measurements);
+                                measurements, functor->getNumFLOPs());
```
manishmishra6016, line 1650: You already have numFLOPs in measurement, this is probably not needed. Or am I missing something?

src/autopas/LogicHandler.h
```diff
@@ -1085,7 +1085,10 @@ IterationMeasurements LogicHandler<Particle>::computeInteractions(Functor &funct
 
   functor.endTraversal(newton3);
   const auto [energyPsys, energyPkg, energyRam, energyTotal] = autoTuner.sampleEnergy();
+  const auto numFLOP = functor.getNumFLOPs();
+  const auto energyPerFLOP = energyTotal / numFLOP;
   timerTotal.stop();
+  const auto energyDelayProduct = energyTotal * timerTotal.getTotalTime();
```
manishmishra6016, line 1091: I don't see what might be wrong here honestly. But you should be able to see in logger whether both `energyTotal` and `timeTotal` is positive. If both are positive, `energyDelayProduct` must be positive too.

manishmishra6016 (commented):
Just some minor comments.

PR #990, opened 2025-01-14
Feature/additional metrics

JannikHoog:
## Description
All preliminary changes made by Jannik Hoog for my Bachelor Thesis "Comparison of energy performance metrics for autotuning molecular simulations with Autopas" supervised my Manish

## Current Issues

EDP values are still a bit odd

### DIFF
--- a/src/autopas/LogicHandler.h
+++ b/src/autopas/LogicHandler.h
@@ -1086,7 +1086,7 @@ IterationMeasurements LogicHandler<Particle>::computeInteractions(Functor &funct
   functor.endTraversal(newton3);
   const auto [energyPsys, energyPkg, energyRam, energyTotal] = autoTuner.sampleEnergy();
   const auto numFLOP = functor.getNumFLOPs();
-  const auto energyPerFLOP = energyTotal / numFLOP;
+  double energyPerFLOP = static_cast<double>(energyTotal) / static_cast<double>(numFLOP);
   timerTotal.stop();
   double energyDelayProduct = static_cast<double>(energyTotal) * static_cast<double>(timerTotal.getTotalTime());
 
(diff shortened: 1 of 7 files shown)


src/autopas/LogicHandler.h
```diff
@@ -1656,6 +1662,10 @@ bool LogicHandler<Particle>::computeInteractionsPipeline(Functor *functor,
             return measurements.timeTotal;
           case TuningMetricOption::energy:
             return measurements.energyTotal;
+          case TuningMetricOption::energyPerFLOP:
+            return static_cast<long>(measurements.energyPerFLOP);
+          case TuningMetricOption::energyDelayProduct:
+            return static_cast<long>(measurements.energyDelayProduct);
```
manishmishra6016, line 1668: So, the issue I found out was that your normalize heavily the calculations of `energyPerFLOP` and `energyDelayProduct` and when you cast them to long here, they become 0.

src/autopas/LogicHandler.h
```diff
@@ -1085,7 +1085,10 @@ IterationMeasurements LogicHandler<Particle>::computeInteractions(Functor &funct
 
   functor.endTraversal(newton3);
   const auto [energyPsys, energyPkg, energyRam, energyTotal] = autoTuner.sampleEnergy();
+  long numFLOP = static_cast<long>(functor.getNumFLOPs());
+  double energyPerFLOP = static_cast<double>(energyTotal / 1e6) / static_cast<double>(numFLOP);
```
manishmishra6016, line 1089:
```suggestion
  double energyPerFLOP = static_cast<double>(energyTotal) / numFLOP;
```
this should fix your tuning results with energyPerFLOP

manishmishra6016 (commented):
With the changes mentioned here, it worked for me. So, I would suggest making these changes and generating your plots. I hope this makes sense and works. Let me know if there is some issue still.

manishmishra6016:
'energyPerFLOP' is probably not required for AutoPas as it behaves similarly to just `energy` (and leads to wrong results --> configurations with newton3 option disabled). `energyDelayProduct` is potentially interesting, but needs to be looked into in more detail. Closing this PR without merging for now.

PR #994, opened 2025-01-28
AutoTuning User Doc

### LINKED ISSUE #870: User documentation about tuning
**The Problem**
From what I can see, we actually don't have any user documentation about tuning and tuning strategies.

I think we should add a file to the [userdocs](https://github.com/AutoPas/AutoPas/tree/master/docs/userdoc) about tuning. This should explain how tuning works in general and present the available tuning strategies.

FG-TUM:
# Description

A first version of documentation about the AutoTuning process.

@thesamriel Maybe you should add some lines on how the 3 Body tuning interacts with this.

## Resolved Issues

- [x] fixes #870 

# ~How Has This Been Tested?~

### DIFF
--- a/docs/userdoc/AutoTuning.md
+++ b/docs/userdoc/AutoTuning.md
@@ -1,0 +1,46 @@
+# AutoTuning
+
+One of the primary features of AutoPas is the dynamic automated algorithm selection for the force calculation.
+Since AutoPas follows the philosophy of [Convention over Configuration](https://en.wikipedia.org/wiki/Convention_over_configuration), it will automatically tune itself to find the optimum of all its internal algorithm configurations.
+However, this naive tuning is just an exhaustive search by trialling all algorithms over successive iterations.
+While this is guaranteed to find the optimum, it is very costly and inefficient.
+Therefore, AutoPas offers a bunch of [Tuning Strategies](https://github.com/AutoPas/AutoPas/blob/master/src/autopas/options/TuningStrategyOption.h) that can be applied and combined for an optimal experience.
+
+## Tuning Loop
+To better understand what the strategies do, we first take a look at the way AutoPas does its tuning.
+
+The basic idea is that the AutoTuner class determines a configuration, which the LogicHandler then assembles and applies to the ParticleContainer.
+In order to determine the optimal configuration, the AutoTuner has a queue of candidates that it tries out by using it for one interaction computation in order to measure their performance.
+Those iterations while candidate configurations are evaluated are referred to as tuning phase.
+AutoPas works on the assumption that particle simulations evolve very slow and thus the simulation state of subsequent iterations is sufficiently similar to compare the configurations' performances.
+When the queue is fully processed, the configuration with the optimal measurement is determined to be the current algorithm optimum and is used until the start of the next tuning phase.
+
+### Tuning Interval
+The interval between the start of two tuning phases can be configured by the user.
+Iterations for sampling the configurations do count against this interval so if an tuning interval of 1000 is set tuning will start at iterations 0, 1000, 2000.
+Should the tuning take longer than one interval any starts that fall into an active tuning phase are skipped.
+
+### Tuning Strategies
+By default, the AutoTuner configuration queue is filled with all applicable configurations and processed sequentially.
+Tuning strategies can filter, modify, and reorder this queue.
+After each sampling of a configuration, they receive the measurement, called evidence, and some live information about the current state of the domain.
+Multiple strategies can be activated at the same time.
+They are all applied in every decision process in the order they are specified in the vector of `ConfigurationOption`s passed to AutoPas.
+
+### Sampling, Selecting, and Smoothing
+For each evidence, AutoPas combines multiple subsequent measurements, called samples.
+This is necessary for two reasons.
+First, Verlet based algorithms require iterations with rebuild steps that take significantly longer than those without.
+Thus at least two iterations are necessary to calculate a weighted average.
+Second, especially measurements from iterations which take significantly less time than one second tend to be noisy.
+This is tackled by reducing all measurements of a type (rebuilding vs. not-rebuilding) for one configuration with a [SelectorStrategy](https://github.com/AutoPas/AutoPas/blob/master/src/autopas/options/SelectorStrategyOption.h).
+Further, when data from at least two previous tuning phases already exists, [LOESS](https://en.wikipedia.org/wiki/Local_regression) can be applied to mitigate positive outliers.
+
+## Tuning Metric
+AutoPas not only offers tuning for minimal runtime.
+The tuning metric can also be switched to minimize energy consumption.
+See [Building](https://github.com/AutoPas/AutoPas/blob/master/docs/userdoc/Building.md) for instructions how to compile support for this.
+
+## Related Files and Folders
+- AutoTuner.h
+- TuningStrategyInterface.h


docs/userdoc/AutoTuning.md
```diff
@@ -0,0 +1,46 @@
...
+They are all applied in every decision process in the order they are specified in the vector of `ConfigurationOption`s passed to AutoPas.
+
+### Sampling, Selecting, and Smoothing
+For each evidence, AutoPas combines multiple subsequent measurements, called samples.
+This is necessary for two reasons.
+First, Verlet based algorithms require iterations with rebuild steps that take significantly longer than those without.
+Thus at least two iterations are necessary to calculate a weighted average.
+Second, especially measurements from iterations which take significantly less time than one second tend to be noisy.
+This is tackled by reducing all measurements of a type (rebuilding vs. not-rebuilding) for one configuration with a [SelectorStrategy](https://github.com/AutoPas/AutoPas/blob/master/src/autopas/options/SelectorStrategyOption.h).
+Further, when data from at least two previous tuning phases already exists, [LOESS](https://en.wikipedia.org/wiki/Local_regression) can be applied to mitigate positive outliers.
```
SamNewcome, line 37: I'm super uncomfortable with this for reasons we've already discussed. Personally, I wouldn't include this at all. It's not that I think this issue doesn't require a solution, which I don't currently have, but I would generally recommend increasing the number of tuning samples over using this as the temporary "good enough for now" solution, especially once #995 gets merged to mitigate costs associated with requiring a high number of tuning samples).
If you still want to keep this in, I would suggest
```suggestion
Whilst gathering many samples can help reduce the noise, it can also be costly due to having to gather many samples for bad configurations.
To alleviate this, when data from at least two previous tuning phases already exists, a variation of [LOESS](https://en.wikipedia.org/wiki/Local_regression) can be applied to mitigate positive outliers. 
This should be used with caution, as it can lead to very bad configurations being selected.
```

SamNewcome (changes requested):
Some changes. I would also suggest that @AlphaGergedan also reviews this, as he will be working on this.

(Also, could you wait until #998 gets merged before merging this?)

docs/userdoc/AutoTuning.md
```diff
@@ -0,0 +1,46 @@
...
+
+The basic idea is that the AutoTuner class determines a configuration, which the LogicHandler then assembles and applies to the ParticleContainer.
+In order to determine the optimal configuration, the AutoTuner has a queue of candidates that it tries out by using it for one interaction computation in order to measure their performance.
+Those iterations while candidate configurations are evaluated are referred to as tuning phase.
+AutoPas works on the assumption that particle simulations evolve very slow and thus the simulation state of subsequent iterations is sufficiently similar to compare the configurations' performances.
+When the queue is fully processed, the configuration with the optimal measurement is determined to be the current algorithm optimum and is used until the start of the next tuning phase.
+
+### Tuning Interval
+The interval between the start of two tuning phases can be configured by the user.
+Iterations for sampling the configurations do count against this interval so if an tuning interval of 1000 is set tuning will start at iterations 0, 1000, 2000.
```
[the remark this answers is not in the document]
AlphaGergedan: Oh I didn't know this. I think might be important? Does this mean, that the iterations spent on sampling configurations during a tuning phase are included in the overall iterations? So their results are not thrown away, but used in the actual simulation?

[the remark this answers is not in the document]
SamNewcome: Yes exactly

docs/userdoc/AutoTuning.md
```diff
@@ -0,0 +1,46 @@
...
+One of the primary features of AutoPas is the dynamic automated algorithm selection for the force calculation.
+Since AutoPas follows the philosophy of [Convention over Configuration](https://en.wikipedia.org/wiki/Convention_over_configuration), it will automatically tune itself to find the optimum of all its internal algorithm configurations.
+However, this naive tuning is just an exhaustive search by trialling all algorithms over successive iterations.
+While this is guaranteed to find the optimum, it is very costly and inefficient.
+Therefore, AutoPas offers a bunch of [Tuning Strategies](https://github.com/AutoPas/AutoPas/blob/master/src/autopas/options/TuningStrategyOption.h) that can be applied and combined for an optimal experience.
+
+## Tuning Loop
+To better understand what the strategies do, we first take a look at the way AutoPas does its tuning.
+
+The basic idea is that the AutoTuner class determines a configuration, which the LogicHandler then assembles and applies to the ParticleContainer.
```
AlphaGergedan, line 12: Adding links to important handlers / classes at the start of documentation might be useful for newcomers who would like to see their implementations :)

```suggestion
The basic idea is that the [AutoTuner](https://github.com/AutoPas/AutoPas/blob/master/src/autopas/tuning/AutoTuner.h) class determines a configuration, which the [LogicHandler](https://github.com/AutoPas/AutoPas/blob/master/src/autopas/LogicHandler.h) then assembles and applies to the [ParticleContainer](https://github.com/AutoPas/AutoPas/blob/master/src/autopas/containers/ParticleContainerInterface.h).
```

docs/userdoc/AutoTuning.md
```diff
@@ -0,0 +1,46 @@
...
+The interval between the start of two tuning phases can be configured by the user.
+Iterations for sampling the configurations do count against this interval so if an tuning interval of 1000 is set tuning will start at iterations 0, 1000, 2000.
+Should the tuning take longer than one interval any starts that fall into an active tuning phase are skipped.
+
+### Tuning Strategies
+By default, the AutoTuner configuration queue is filled with all applicable configurations and processed sequentially.
+Tuning strategies can filter, modify, and reorder this queue.
+After each sampling of a configuration, they receive the measurement, called evidence, and some live information about the current state of the domain.
+Multiple strategies can be activated at the same time.
+They are all applied in every decision process in the order they are specified in the vector of `ConfigurationOption`s passed to AutoPas.
```
AlphaGergedan, line 28: What is `ConfigurationOption` passed to AutoPas? Do you mean which strategies are enabled when starting the simulation?

docs/userdoc/AutoTuning.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Sampling, Selecting, and Smoothing
+For each evidence, AutoPas combines multiple subsequent measurements, called samples.
```
AlphaGergedan, line 31: I am a little confused here. For the need of an average: The first argument is that some configs may need longer, therefore, may have variance of higher magnitudes compares to configs that are quicker? But then the second argument is the configs that are 'very' fast tend to be noisy. So is what I understand correct: Some configs we have to trial might be noisy for some reason, and if another config is not noisy, then we might select the wrong strategy (due to very large noise) if we only trial both once.

And also, is the suggested solution fixed (two iterations), or can it be configured as well? 

And final question, then is there a reason for having `Fastest-Absolute-Value` for the `tuning-strategy` in the [FallingDrop](https://github.com/AutoPas/AutoPas/blob/master/examples/md-flexible/input/fallingDrop.yaml) example in md-flexible?

AlphaGergedan (approved):
Nice to read! Added some comments on the parts I got confused a little.

docs/userdoc/AutoTuning.md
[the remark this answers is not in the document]
FG-TUM: That's why I wanted to state it so explicitly. 
@AlphaGergedan as the newest to the club what formulation is the clearest for you?

docs/userdoc/AutoTuning.md
```diff
@@ -0,0 +1,46 @@
...
+When the queue is fully processed, the configuration with the optimal measurement is determined to be the current algorithm optimum and is used until the start of the next tuning phase.
+
+### Tuning Interval
+The interval between the start of two tuning phases can be configured by the user.
+Iterations for sampling the configurations do count against this interval so if an tuning interval of 1000 is set tuning will start at iterations 0, 1000, 2000.
+Should the tuning take longer than one interval any starts that fall into an active tuning phase are skipped.
+
+### Tuning Strategies
+By default, the AutoTuner configuration queue is filled with all applicable configurations and processed sequentially.
+Tuning strategies can filter, modify, and reorder this queue.
```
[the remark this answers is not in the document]
FG-TUM: Then let's drop the "and/". I want to say modify because tuning strategies can technically do whatever they want with the queue. Even inserting stuff. 

```suggestion
Tuning strategies can filter, reorder, or otherwise modify this queue.
```

docs/userdoc/AutoTuning.md
FG-TUM: I want to keep it in the doc as long as it is in the code. Otherwise this is too easily forgotten. Imho your last sentence is too harsh but I get what you mean. Give me a sec to come up with a suggestion.

FG-TUM:
```suggestion
Whilst gathering many samples can help reduce the noise, it can also be costly due to having to gather many samples for sub optimal configurations.
To alleviate this, when data from at least two previous tuning phases already exists, a variation of [LOESS](https://en.wikipedia.org/wiki/Local_regression) can be applied to mitigate positive outliers. 
Use this feature with caution since it can sometimes lead to undesirable tuning outcomes. 
When in doubt compare the reduced vs smoothed values from the `TuningDataLogger`. 
```

docs/userdoc/AutoTuning.md
FG-TUM: We could also enhance other doc files with more links.

docs/userdoc/AutoTuning.md
SamNewcome: Yeah this looks fine!

docs/userdoc/AutoTuning.md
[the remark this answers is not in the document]
AlphaGergedan: Your formulation is more informative but a bit confusing I think. Maybe it would be clearer to state this more explicitly like
- `The iterations used for sampling configurations during a tuning phase are part of the overall simulation and are not discarded`. Or
- `The results from these sampled configurations contribute to the simulation’s progress`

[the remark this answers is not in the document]
FG-TUM: Ah what you touch upon is something slightly different.
- My point here is that the tuning interval is the number of iterations between two tuning phase starts as opposed to the end of one phase until the start of the next phase.
- Your point is that the simulation is progressing during the tuning.

Is that correct? This second point is mentioned in the paragraph above (Tuning Loop) but I made it more explicit (-> 06879c2bc23336f7559bb52ae09badfa8019bba9). Although advancing the simulation is not part of AutoPas but the simulator around it.

docs/userdoc/AutoTuning.md
FG-TUM: My bad it was meant to be like this:
```suggestion
They are all applied in every decision process in the order they are specified in the vector of `TuningStrategyOption`s passed to AutoPas.
```

docs/userdoc/AutoTuning.md
FG-TUM: About your points:
1. Yes any short measurement is inherently noisy thus we want more samples to get the strategy that is better "on average". With only one sample of each config the random noise might lead us to pick the wrong config.
2. Configs that are based on Verlet lists have iterations where they rebuild their lists. This takes significantly longer because of memory movement etc.. However, these iterations only occur once in a while (-> Verlet rebuild interval or frequency). Thus we need to know at least how fast an iteration with and how fast an iteration without a rebuild is.
3. Yes the amount of samples can be configured (`AutoPas::setNumSamples()`).
4. You can choose how to arrive at a value from the gathered samples. This is the selector-strategy. Since we are dealing with outliers, regular averaging might not be desirable because it is highly susceptible to them. You can also use the median. However, the question is do we really want to include any impact of noise in our measurements? If the answer is no, then an argument can be made that we should only look at the best sample because this has the least noise in it.

docs/userdoc/AutoTuning.md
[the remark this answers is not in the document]
AlphaGergedan: Yes. I specifically thought about progressing during the tuning when reading this, but now it's clear. I now think if you say: `[...] 1000 is set, tuning will start at iterations 0, 1000, 2000.`, it's clear that the tuning interval is the num. iterations between two tuning-starts if you also know that the simulation may progress depending on the solver.

AlphaGergedan (approved):

Issue #991, opened 2025-01-14
Remove the _rebuildInterval vector used for dynamic containers


manishmishra6016:
**Is your feature request related to a problem? Please describe.**
`_rebuildInterval` vector can grow quite large for big simulations. 

**Describe the solution you'd like**
As suggested by @FG-TUM in #821 `_rebuildInterval` vector can be replaced by using `_stepsSinceLastRebuild`, `_iterations` (and possibly `_currentNumberOfRebuilds`) to calculate the mean rebuild frequency. 

**Describe alternatives you've considered**
Leave it as it is.

**Additional context**
This can be handled together with Issue #954

SamNewcome:
What was the main purpose of storing every rebuild interval instead of Fabio's suggestion?

manishmishra6016:
> What was the main purpose of storing every rebuild interval instead of Fabio's suggestion?

There was no strong reason except for the ease of it. It made sense as you can easily retrieve min, max, and mean of rebuild intervals. Of course, all of this can always be obtained from the results of the iteration logger. I totally agree with Fabio's suggestion that this is unnecessary for calculating mean RF over the simulation and can be removed. 

The reason why I didn't want to fix it already was that I thought we might still need some form of vector to store the rebuild intervals during the simulation phase. So I thought I will just handle it with #954.

Issue #783, opened 2023-09-14
Rethink `AutoTuner::willRebuildNeighborLists()`


FG-TUM:
**Is your feature request related to a problem? Please describe.**
https://github.com/AutoPas/AutoPas/blob/3053eeca87758a9e89670c62fdb53a83cd8ea2e8/src/autopas/tuning/AutoTuner.cpp#L268-L275

`AutoTuner::willRebuildNeighborLists()` currently works by counting iterations and, from there, estimating if neighbor lists should be rebuilt because of the rebuild frequency or container changes due to tuning. The problem with this is, that if we ever change our logic, such that we rebuild earlier, e.g. by collecting fewer samples than `_maxSamples`, this logic breaks down producing silent errors.

**Describe the solution you'd like**
A more robust estimation system that is not based on (too much) state.

**Describe alternatives you've considered**
Sam's suggestion is to give the AutoTuner a flag for if we are switching configurations. This can easily be set within AutoTuner::getNextConfig().
We can remove all the rebuild-naturally-after-_rebuildFrequency-iterations logic from AutoTuner::willRebuildNeighborLists(). Then, all the function does is determine if the neighbour lists should be rebuilt due to changing configuration (and so should get renamed e.g. AutoTuner::willRebuildNeighborListsDueToChangingConfiguration()). And so all the function needs to do is return true or false depending on if the flag is set and, if so, reset the flag as well.
_Originally posted by @SamNewcome in https://github.com/AutoPas/AutoPas/pull/759#discussion_r1318276202_

**Additional context**
This logic was introduced after the AutoTuner lost access to the container in the process of creating #759

thesamriel:
This was refactored in #1063 . The LogicHandler is the sole truth of iteration numbers.

PR #995, opened 2025-01-29
(Naive) Early Stopping

ManuelLerchner:
# (Naive) Early Stopping

## Problem

The Autotuning process can be very costly at times, as many "bad" configurations can end up being evaluated.
Many of those bad configurations are orders of magnitude slower than the optimal one (see [Tobias Humig Project Report](https://mediatum.ub.tum.de/doc/1730215/qwnftyj83fo5se9xzl0lokq4a.pdf), [Manuel Lerchner Thesis](https://mediatum.ub.tum.de/doc/1752754/j9j37ciousd48fg4av11hu4s1.pdf)), which can drastically increase the total runtime of the simulation.

## Change

This PR introduces an improvement to this problem by preventing such bad configurations from being resampled multiple times during the tuning phases.
It works by detecting `long-running` samples and simply skipping to the next configuration in the configuration queue. (See next graphic).

image

The current detection of `long-running` samples is quite naive. It works by comparing the slowdown ($$slowdown = \frac{time_{curr\ sample}}{time_{best\ sample\ so\ far}}$$) to a user-defined threshold called $allowedSlowdownFactor$.

## Benchmark: Exploding Liquid + PredictiveTuning (CoolMUC2)

Early tests suggest that a good threshold is around $$allowedSlowdownFactor=5$$ (see next chart), although I suspect the value is not universal and might change depending on the hardware/scenario/tuning-strategy (More Benchmarks needed).

For the Exploding Liquid benchmark with PredictiveTuning, the total simulation time can be reduced from 28.62s to 23.23s at $$allowedSlowdownFactor ≈ 5$$, corresponding to a 18.9% speedup in total simulation time compared to not using Early Stopping. The time spent in tuning phases is reduced by ≈60% in this case.

total_time_average_full_scale

# How Has This Been Tested?

+ Benchmarks


SamNewcome (changes requested):
Generally very nice. A couple of minor points:

- I'm concerned that the parameter name is too interchangeable with the relative blacklist range used with the Slow Config filter strategy, especially if you don't look too closely at the doxygen description. I would try to make it clearer that these are separate. One option could be to call this the early stopping factor (or something like that). Alternatively, you could rename the slow config filter's parameter.
- I'm concerned that this factor could be too exclusionary of the Verlet-based methods with expensive first iterations. This probably doesn't show in the current results because Verlet-based methods appear to have implementational issues, so they are (in general) unlikely to be optimal. I would ideally want the comparison to be made on the weighted average. For the sake of simplicity, I would suggest simply requiring two iterations and then using the weighted average. This could, however, potentially be improved. I would also be okay with leaving this for this PR, but adding this as an issue to be addressed in the future.

src/autopas/tuning/AutoTuner.cpp
```diff
@@ -444,4 +462,22 @@ bool AutoTuner::inTuningPhase() const {
 const EvidenceCollection &AutoTuner::getEvidenceCollection() const { return _evidenceCollection; }
 
 bool AutoTuner::canMeasureEnergy() const { return _energyMeasurementPossible; }
+
+void AutoTuner::evaluateConfigurationPerformance(long performance) {
+  _fastestIteratePairwiseTime = std::min(_fastestIteratePairwiseTime, performance);
+
+  double slowdownFactor = static_cast<double>(performance) / static_cast<double>(_fastestIteratePairwiseTime);
```
manishmishra6016, line 469: I have the same concern as Sam here. Because you compare only the first iteration, you bias towards the configuration that have low `rebuildIteration` time. These might have overall larger time due to a more expensive `nonRebuildIteration` time. Based on current implementations, the current approach biases towards `LinkedCells` over VL-based containers.

manishmishra6016 (commented):
I was trying to run the simulations with different `max-allowed-slowdown-factor` but I end up with the segmentation fault for most values (works fine when the factor is ~5 or higher, but fails for 4 or lower in case of fallingDrop example). I didn't look into it in any detail, but did you encounter this as well? Or am I doing something wrong?

ManuelLerchner:
> I was trying to run the simulations with different `max-allowed-slowdown-factor` but I end up with the segmentation fault for most values (works fine when the factor is ~5 or higher, but fails for 4 or lower in case of fallingDrop example). I didn't look into it in any detail, but did you encounter this as well? Or am I doing something wrong?

I also encountered this occasionally for certain values of `max-allowed-slowdown-factor`. 
However, I'm not sure if this is a direct result of my changes. Sam mentioned something about Verletlists being broken, so maybe it is caused by that.

Referring to our conversation on Elements, some of the crashes seem to be caused by `ClusterList.getClusterThreadPartition()` in `VCLC01BalancedTraversal.h`, as the function somehow returns invalid references, in some executions. However, I did not yet look deeper into why this is the case.

SamNewcome:
> Referring to our conversation on Elements, some of the crashes seem to be caused by `ClusterList.getClusterThreadPartition()` in `VCLC01BalancedTraversal.h`, as the function somehow returns invalid references in some executions. However, I did not yet look deeper into why this is the case.

I have also independently had issues with `VCLC01BalancedTraversal.h`, but didn't get a chance to look into it.

ManuelLerchner:
I tried a new way of calculating the slowdown factor. Now it works with `estimateRuntimeFromSamples()` in combination with `_evidenceCollection`. 
I changed it to require at least two samples, before considering an Early-Stop, and `estimateRuntimeFromSamples()` does the `weighted average` internally.

Since this method is less affected by noise, you can get away with way lower `allowed-slowdown-factors`, and somehow the "U-Shape is gone".

And I fixed the `segmentation faults`.

total_time_average

src/autopas/tuning/AutoTuner.cpp
```diff
@@ -463,21 +459,33 @@ const EvidenceCollection &AutoTuner::getEvidenceCollection() const { return _evi
...
     AutoPasLog(DEBUG,
                "Configuration is {} times slower than the current fastest traversal time. This is higher than the "
-               "maximum allowed slowdown factor of {}. Further evaluation of this configuration will be skipped.",
-               slowdownFactor, _maxAllowedSlowdownFactor);
+               "earlyStoppingFactor factor of {}. Further samples of this configuration will be skipped.",
+               slowdownFactor, _earlyStoppingFactor);
     _earlyStoppingOfResampling = true;
+
+    // Pretend that we are in the last iteration of this sample. This is required for the periodic rebuilding
+    _iterationBaseline += (_maxSamples - (_iterationBaseline % _maxSamples)) - 1;
```
manishmishra6016, line 488: Maybe explain the formula in the comment a bit (it works fine for what I tried). But somethings are not clear, for eg. shouldn't `_iterationBaseline % _maxSamples` always be just `_iterationBaseline`?

src/autopas/tuning/AutoTuner.cpp
```diff
@@ -463,21 +459,33 @@ const EvidenceCollection &AutoTuner::getEvidenceCollection() const { return _evi
...
 bool AutoTuner::canMeasureEnergy() const { return _energyMeasurementPossible; }
 
-void AutoTuner::evaluateConfigurationPerformance(long performance) {
-  _fastestIteratePairwiseTime = std::min(_fastestIteratePairwiseTime, performance);
-
-  double slowdownFactor = static_cast<double>(performance) / static_cast<double>(_fastestIteratePairwiseTime);
+void AutoTuner::checkEarlyStoppingCondition() {
+  if (_samplesNotRebuildingNeighborLists.empty()) {
+    // Wait until a sample without expensive rebuilding occurred to make it fairer for Verlet-based methods.
+    // This should generally happen in the second sample
```
manishmishra6016, line 465:
```suggestion
    // Wait until a sample without the expensive rebuilding has occurred to compare the configurations based on estimated runtime for the non-tuning phase.
    // This should generally happen in the second sample.
```
Maybe better?

manishmishra6016 (approved):

examples/md-flexible/src/configuration/MDFlexConfig.h
```diff
@@ -368,12 +368,13 @@ class MDFlexConfig {
...
    */
-  MDFlexOption<double, __LINE__> maxAllowedSlowdownFactor{
-      std::numeric_limits<double>::infinity(), "max-allowed-slowdown-factor", false,
-      "Maximum allowed slowdown factor for the auto-tuner. A configuration performing worse than the current optimum "
-      "by this factor will be blacklisted from the current tuning phase"};
+  MDFlexOption<double, __LINE__> earlyStoppingFactor{
+      std::numeric_limits<double>::infinity(), "early-stopping-factor", false,
+      "EarlyStoppingFactor for the auto-tuner. A configuration seeming to perform worse than the "
+      "previously best configuration "
+      "by this factor will not be sampled again"};
```
SamNewcome, line 377:
```suggestion
      "by this factor will not be sampled again during that tuning phase."};
```

src/autopas/AutoPasDecl.h
```diff
@@ -687,14 +687,11 @@ class AutoPas {
   void setNumSamples(unsigned int numSamples) { _autoTunerInfo.maxSamples = numSamples; }
 
   /**
-   * Set the maximum allowed slowdown factor for the auto tuner. If a configuration is slower than the optimum
-   * configuration found so far by more than this factor, it is blacklisted an will not be tested again until the next
-   * tuning phase.
-   * @param maxAllowedSlowdownFactor
+   * Set the earlyStoppingFactor for the auto tuner. If a configuration seems to be slower than the optimum
+   * configuration found so far by more than this factor, it will not be sampled again
```
SamNewcome, line 691:
```suggestion
   * configuration found so far by more than this factor, it will not be sampled again during that tuning phase.
```

src/autopas/tuning/AutoTuner.h
```diff
@@ -229,12 +229,12 @@ class AutoTuner {
   bool canMeasureEnergy() const;
 
   /**
-   * Notify the tuner about the iteratePairwise runtime of the current configuration.
-   * Keeps track of the fastest runtime so far and triggers early stopping of the sampling phase if the current runtime
-   * is worse than the fastest runtime by more than the maximum allowed slowdown factor.
-   * @param performance The runtime of the current configuration.
+   * Checks whether the current configuration seems to perform so poorly that it shouldn't be resampled further
```
SamNewcome, line 232:
```suggestion
   * Checks whether the current configuration performs so poorly that it shouldn't be resampled further within this tuning phase.
```

src/autopas/tuning/AutoTuner.h
```diff
@@ -229,12 +229,12 @@ class AutoTuner {
...
+   * If the current configuration seems worse than the previosuly best configuration by more than the
```
SamNewcome, line 233:
```suggestion
   * If the currently sampled configuration is worse than the current best configuration by more than the
```

src/autopas/tuning/AutoTuner.h
```diff
@@ -229,12 +229,12 @@ class AutoTuner {
...
+   * earlyStoppingFactor factor, it will not be sampled again. Uses the _estimateRuntimeFromSamples() function to
```
SamNewcome, line 234:
```suggestion
   * earlyStoppingFactor factor, it will not be sampled again this tuning phase. Uses the _estimateRuntimeFromSamples() function to
```

src/autopas/tuning/AutoTuner.h
```diff
@@ -325,15 +325,10 @@ class AutoTuner {
   size_t _maxSamples;
 
   /**
-   * Maximum allowed slowdown factor for the auto-tuner. A configuration performing worse to the currently fastest
-   * iteratePairwise run by more than this factor is not considered for the next tuning phase.
+   * EarlyStoppingFactor for the auto-tuner. A configuration performing worse than the currently best configuration
+   * by more than this factor will not be sampled again
```
SamNewcome, line 329:
```suggestion
   * by more than this factor will not be sampled again this tuning phase.
```

SamNewcome (changes requested):
This is not explicitly part of your PR, but the documentation for `_iterationBaseline` isn't sufficient, which meant I was confused by some of your changes. Could you add to the documentation for this:
"Used only for triggering rebuilds when configurations switch during tuning phases, which occurs when _iterationBaseline % _maxSamples == 0. _iterationBaseline may therefore be modified to "skip" iterations e.g. when early stopping is used."
(Technically, in the master, it is also used to trigger rebuilds during the non-tuning phase, but this trigger is redundant and will be removed in #991 so I don't see the point of adding that the the doxygen just for Manish to remove it)

ManuelLerchner:
> One question: the total time values for previous benchmark and current one seem to be different (current one being ~12-25 while earlier it was 23-26, so are we looking at different problems? This might affect the shape too?

It is still the same benchmark (Exploding Liquid + Predictive Tuning).
+ I've just looked at the logs again, and the old plots were using 14-threads for the simulations, while the new one just uses 1. So maybe some overhead got removed.
+ And apparently they shut down CoolMUC2 in December and since then the new `serial-cluster` seems to be based on CoolMUC4 ([See here](https://doku.lrz.de/job-processing-on-the-linux-cluster-10745970.html)). Maybe this also explains some of the performance gains.

I've uploaded the raw data to my Repository, if you want to compare the logs and the used .yaml files ([AutoPas-AutoTuning-Seminar-Paper](https://github.com/ManuelLerchner/AutoPas-AutoTuning-Seminar-Paper/tree/main/data/explodingLiquid/cluster/predictiveTuningEvidenceBased)).

src/autopas/tuning/AutoTuner.cpp
ManuelLerchner: `_iterationBaseline` is not required to be less than `_maxSamples`. It starts at 0 at the start of the tuning phase and gets incremented every iteration.

`_iterationBaseline` is primarily used in:
```cpp
bool AutoTuner::willRebuildNeighborLists() const {
  // What is the rebuild rhythm?
  const auto iterationsPerRebuild = this->inTuningPhase() ? _maxSamples : _rebuildFrequency;
  // _iterationBaseLine + 1 since we want to look ahead to the next iteration
  const auto iterationBaselineNextStep = _forceRetune ? _iterationBaseline : _iterationBaseline + 1;
  return (iterationBaselineNextStep % iterationsPerRebuild) == 0;
}
```

to trigger a rebuild whenever a new configuration is evaluated.

With Early-Stopping we want to force a rebuild in the next iteration, as a new configuration will be evaluated.

The formula simply adds the number of samples that we skipped to `_iterationBaseline`, to pretend we are at the end of the sampling phase. I will add this in the comments.

ManuelLerchner:
Maybe rewriting the `AutoTuner::willRebuildNeighborLists()` function as suggested in #783 may be appropriate now. The function is getting harder and harder to maintain, and was the source of some very annoying bugs during the development of this PR.

SamNewcome (approved):
Looks good!

Issue #999, opened 2025-02-07
The `TranslationVisitor` used by (fuzzy) rules based tuning matches "literals" to "properties" via checking if the type matches which is error prone


SamNewcome:
The `TranslationVisitor` is used by (fuzzy) rules based tuning to match values (e.g. "LinkedCells") of any properties (e.g. "Container") of the configuration. This is used to create patterns which are subsets of the search space. Individual configurations match against this pattern if all elements of the configuration are included in the pattern - in other words, if the configuration is in the pattern's subset. If a pattern features no elements for a property, all values for that property are allowed. So a pattern could include Linked Cells and Verlet Lists as possible containers but no elements for traversal, meaning that any configuration with Linked Cells or Verlet Lists as the container (regardless of traversal) matches the pattern. Alternatively, if that pattern included LC C01 as the only traversal value, it would not match with any configuration that did not have this traversal, even if that configuration had Verlet Lists as the container.

`ConfigurationPattern::add(const RuleVM::MemoryCell &value)` adds `value` to the pattern. It determines which property `value` is from its type. `_containers`, `_traversals`, `_loadEstimators`, `_dataLayouts` and `_newton3Options` are all enums, so the value Linked Cells will not get added to the `_traversals` list of traversals in the pattern. `_cellSizeFactors` is the only part of the pattern that uses the type `double` currently.

There are two issues with this:

1. If we add another `double` property to the configuration, there is no way to distinguish between these.
2. If there is a property-type mismatch, it is just ignored. E.g. as is currently the case, `makeLiteral` maps cell size factor to a `size_t`, which then does not match the type of `_cellSizeFactors`. No warnings are printed.

My suggestions are, for the problems respectively:

1. We parse the property when we add the value to the pattern. 
2. We throw an exception if the types mismatch.

Issue #996, opened 2025-01-29
When a tuning strategy has not implemented `rejectConfiguration`, `LogicHandler` can get stuck in an infinte loop


SamNewcome:
**Describe the bug**
In `LogicHandler::selectConfiguration`, there is a `while (true)` loop. This checks if the current configuration is applicable and, if not, tells the autotuner to reject the configuration (`autoTuner::rejectConfig`), which in turn calls for each tuning strategy `rejectConfiguration`. Some tuning strategies (Fuzzy Tuning, Rule Based Tuning, Random Search, Slow Config Filter, Sort By Name) have not implemented this. 

After calling `<tuningStrategy>::rejectConfiguration`, `AutoTuner::rejectConfig` then gets all tuning strategies to optimise suggestions as normal, and returns the next configuration. However, if the calls to `<tuningStrategy>::rejectConfiguration` don't change the optimised suggestion (e.g. because it is not implemented), `AutoTuner::rejectConfig` returns the same configuration that was just rejected. Therefore, the loop repeats infinitely.

**To Reproduce**
Run the attached input file with the given jobscript and fuzzy rule file on HSUper with a 5x8x9 setup (I do not have an easier way of reproducing this at this time)
Due to GitHub rejecting "exotic" file types, the extensions should be `.yaml` and `.frule`.

[rayleighTaylor.txt](https://github.com/user-attachments/files/18588131/rayleighTaylor.txt)
[fuzzyRulesSuitability.txt](https://github.com/user-attachments/files/18588138/fuzzyRulesSuitability.txt)

**Expected behavior**
It is totally fine that this should lead to failure - if this occurs, something has really gone wrong. The problem is the infinite loop making the failure much more significant. E.g. when I ran this with TRACE, I got a stupidly large output file because every loop was logging messages. Even with info logging, this leads to seg faults, which aren't helpful in finding the error.

I would suggest that if `autoTuner.rejectConfig` returns the original configuration, simply throw an exception. Alternatively, we could just remove the configuration from the search space - but this was a feature that was actively disabled. E.g. there is a `bool` `rejectIndefinitely` set to `false` which stops this from happening (fyi, I think the bool name is misleading: this causes the configuration to be rejected indefinitely) @FG-TUM Was there any reason for this?

I would also not have the `while (true)` loop in general, as I think this is dangerous. Even if the loop stopped after 10000 iterations, it would still be much better.

PR #1000, opened 2025-02-07
Fixing type of cellSizeFactor in TranslationVisitor::makeLiteral

SamNewcome:
# Description

`TranslationVisitor::makeLiteral` is responsible for (fuzzy) rules-based tuning for creating "literals" (basically just values that correspond to an element of a configuration). This was casting cell size factor as a `size_t` which does not match the pattern's allowed cell-size factors type (see #999), meaning (fuzzy) rules-based tuning did not handle cell size factors correctly.

`TranslationVisitor::makeLiteral` now correctly makes this literal a `double`, fixing the issue.

# How Has This Been Tested?

- [x] CI
- [x] Code runs using an artificial fuzzy tuning rules file that used to fail (see #996)

This PR is really just a hotfix. #999 would make this problem less likely to occur in the future.


manishmishra6016 (approved):
Good catch :+1:

PR #1002, opened 2025-02-18
Fast Particle Buffer

xhulia028:
This PR is related to my Bachelor Thesis: "Special Handling of Fast Particles in AutoPas to Reduce Rebuild Time".


tests/testAutopas/tests/autopasInterface/AutoPasInterfaceTest.cpp
```diff
@@ -346,7 +346,9 @@ void testSimulationLoop(const autopas::Configuration &conf) {
   // do first simulation loop
   doSimulationLoop(autoPas, &functor);
   doAssertions(autoPas, &functor, numParticles, __LINE__);
-  moveParticlesAndResetF({autoPas.getVerletSkin() / 6, 0., 0.});
+
+  // forcing the particle to move inwards
+  moveParticlesAndResetF({-autoPas.getVerletSkin() / 6, 0., 0.});
```
SamNewcome, line 351: Note to whoever picks up this PR: This was done because there is bad handling if a particle gets moved outside the container and a rebuild occurs (which is the case here due to the buffer particle threshold stuff). This does not appear to be directly related to this PR but pre-existing bad behaviour.

SamNewcome (changes requested):
The key issues which should be resolved by whoever picks up this PR are:

- Fast Particle Buffer should not replace the existing dynamic rebuild criteria. It should be a togglable option through the AutoPas interface. (Long term potentially tunable but not this PR)
- The buffer rebuild threshold should be selectable through the AutoPas interface (again, potentially tunable in the long term)
- The feature needs unit tests.

thesamriel:
Closed in favor of #1170

PR #1007, opened 2025-03-20
Particle Alias/Typename fix

### LINKED ISSUE #1006: Verlet Lists' Neighbour Lists include pointers to `ParticleBase`.
**Describe the bug**
If you run the default exploding liquid with only `vl_list_iteration`, `AoS`, and N3L `disabled`, you get a segmentation fault due to trying to get mixing parameters where one particle is `ParticleBase`. I am not sure exactly where this occurs, but I suspect this is due to the class name `Particle` being both used as a template type name (where it would be parameterised with `MoleculeLJ`) and an AutoPas scoped alias for `ParticleBase` with double precision.

SamNewcome:
# Description

"`Particle`" is used in AutoPas both as a template parameter (i.e. the particle type AutoPas is templated with) and an alias for `ParticleBase` with double precision. This could easily lead to name conflicts like suspected in #1006. 

In this PR we deal with this in two ways:

- The `Particle` alias is removed as it is too generic a name for what is really a base particle. Instead, the alias `ParticleFP64` (renamed to `ParticleBaseFP64`) is used as it is more verbose. In the tests, the alias `Particle` is replaced with `ParticleFP64` (Why not `ParticleBaseFP64`? No strong reason, other than it is used in the tests as a generic FP64 Particle whereas the AutoPas alias is used as an alias for the base particle class - in short, the emphasis in the tests is not really that it is the base class)
- Template parameters `Particle` are renamed `ParticleT`. The contributing guidelines state that all template parameter `typename`s/`class`es should have a `T` postfix. I do not change this for all template parameters and this task would be simply an issue for the future.

After making these changes, I am not totally sure about if the postfix `T` is necessary - the key issue really was that the `Particle` alias shouldn't have existed - and it could be perceived as ugly. Changing all the template parameters to `T` would be a real pain. I would be interested in your opinions.

## Resolved Issues

- [x] fixes #1006

# How Has This Been Tested?

- [ ] CI
- [x] Running `CTest` on ARM cluster
- [x] Checking that this fixes #1006

### DIFF
--- a/src/autopas/particles/ParticleDefinitions.h
+++ b/src/autopas/particles/ParticleDefinitions.h
@@ -10,11 +10,6 @@
 #include "autopas/particles/ParticleBase.h"
 
 namespace autopas {
-
-/**
- * Particle with all variables in 32 bit precision
- */
-using ParticleBaseFP32 = ParticleBase<float, unsigned long>;
 /**
  * Particle with all variables in 64 bit precision
  */
--- a/src/autopas/particles/ParticleDefinitions.h
+++ b/src/autopas/particles/ParticleDefinitions.h
@@ -1,5 +1,5 @@
 /**
- * @file Particle.h
+ * @file ParticleDefinitions.h
  *
  * @date 17 Jan 2018
  * @author tchipevn
--- a/src/autopas/particles/ParticleDefinitions.h
+++ b/src/autopas/particles/ParticleDefinitions.h
@@ -14,14 +14,10 @@ namespace autopas {
 /**
  * Particle with all variables in 32 bit precision
  */
-using ParticleFP32 = ParticleBase<float, unsigned long>;
+using ParticleBaseFP32 = ParticleBase<float, unsigned long>;
 /**
  * Particle with all variables in 64 bit precision
  */
-using ParticleFP64 = ParticleBase<double, unsigned long>;
-/**
- * Alias for Particle with all variables in 64 bit precision
- */
-using Particle = ParticleFP64;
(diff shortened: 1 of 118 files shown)


manishmishra6016:
The change related to `ParticleDefinitions.h` is very nice. I also like `_T` for template but with the other change, it may not be absolutely necessary and can be skipped if everyone is of opinion that changing everything else would be too much work. I personally don't mind if we have this just for particle template and not include it in guidelines just yet.

src/autopas/particles/ParticleDefinitions.h
```diff
@@ -14,14 +14,10 @@ namespace autopas {
 /**
  * Particle with all variables in 32 bit precision
  */
-using ParticleFP32 = ParticleBase<float, unsigned long>;
+using ParticleBaseFP32 = ParticleBase<float, unsigned long>;
```
thesamriel, line 17: I feel like we could also remove this, as running autopas with floats here doesn't work anyways :D

thesamriel (commented):
I generally also support this idea. It makes it clearer what are typenames, what are actual class names etc...
Of course it would be nice to have this consistently, but I see that it is quite some tedious work...

SamNewcome:
Okay, I will keep the postfix (but will use `_T` as suggested by @manishmishra6016)

src/autopas/particles/ParticleDefinitions.h
SamNewcome: Done fe39d538

manishmishra6016 (approved):
Looks good to me!

PR #1012, opened 2025-04-09
Fix macOS Compilation Workflow

schuhmaj:
# Changelog

## Overall changes

This PR fixes compilation issues on macOS 15.4 using Apple Clang 17.0.0 and GNU GCC 14.2.

Further, I've added a new job which ensures compliance with the two macOS compilation workflows: GNU GCC and Apple Clang.
This will help to sustain macOS/ ARM compliance.

## Details

- Add explicit static cast where necessary ([contextual implicit conversion](https://stackoverflow.com/questions/25143860/implicit-conversion-from-class-to-enumeration-type-in-switch-conditional) from class to integral enum value in `switch` statements seems to be broken in Apple Clang)
- Remove the unnecessary `template` keyword in `VLCC08Traversal.h:80` (confusing the compiler - this actually seems to be standard compliant, so no Apple Clang issue, but rather a GCC issue)
- Add very specific `add_definitions(-D_Alignof=alignof)` in the case of macOS and GCC compiler. There appears to be an issue with the newest system headers (macOS 15.4 onwards) when using GCC
- If energy measurements are disabled, then don't link `pmt`
- Add new workflow to test compilation of `md-flexible` on macOS ARM64 using GNU `gcc@14` and `Apple Clang`
- Given these changes, I added `Apple Clang` into the supported compilers list


SamNewcome (approved):
Looks good.

manishmishra6016 (approved):
Looks good! thanks for doing this. :D

PR #1013, opened 2025-04-14
Make pmt download optional

thesamriel:
# Description

If Energy Measurements are disabled, we don't need pmt. Therefore, we don't need to bother about it during CMake configuration. Saves a little bit of time and diskspace ;)


manishmishra6016 (approved):
Looks good to me!

schuhmaj (approved):
👍

PR #1014, opened 2025-04-22
Hierarchical Grid container

atacann:
# Description

Adds hierarchical grid(Hgrid) container and some traversals that use it. Also adds a scaling cutoff option (MD_FLEXIBLE_SCALING_CUTOFF) for md-flexible. Using scaling cutoff, the cutoff between 2  particles is scalingCutoffMultiplier * average of their sigma values.

The basic idea of hierarchical grid is that there are a LinkedCells container in each level of the hierarchy with different cutoffs. Particles with different sizes (sigmas for MoleculeLJs) get assigned to different levels. Then both intra and inter-level interactions are computed. Inter-level interactions are more efficient in HGrids instead of using a single LinkedCells container as HGrid calls functor between less amount of cells resulting in less distance calculations (Number of force calculations is the same as linked cells). 

# How Has This Been Tested?
All usual container tests also run with HGrid but HGrid only uses a single level so behaves like a LinkedCells in those tests.
I added HGrid specific tests in tests/tests/containers/linkedCells that tests iterators and force calculations with scaling cutoff and multiple levels.

# Notes
- Enable MD_FLEXIBLE_SCALING_CUTOFF cmake option to use it.
- Scaling cutoff is supported only with LJFunctor and LJFunctorAVX currently
- If scaling cutoff mode is enabled, cutoff in MDFlexConfig is assumed as scaling cutoff multiplier. It is prob confusing and ther e is a better way. Also if a cutoffs vector is defined then all other containers than HGrid uses the largest cutoff in cutoffs as their cutoff.


src/autopas/containers/linkedCells/HierarchicalGrid.h
```diff
@@ -0,0 +1,557 @@
...
+    maxLength -= _skin;
+    // fix cutoffs if they are bigger than maxLength
+    for (auto &cutoff : _cutoffs) {
+      if (cutoff > maxLength) {
+        AutoPasLog(ERROR,
+                   "Cutoff {} is greater than the maximum length according to Box size - skin {}, lowering cutoff",
+                   cutoff, maxLength);
+        cutoff = maxLength;
+      }
+    }
```
SamNewcome, line 90: AutoPas shouldn't be used anyway with such a large cutoff compared to the box size. I would simply throw an exception - the simulation makes no sense and the cutoffs shouldn't be changed.

SamNewcome (changes requested):
Some initial review comments, but I didn't get too far through the code.

SamNewcome (changes requested):
Finished the code review. Generally very nice. I don't think that there is too much work to do. Most of the suggestions are either minor or relate to one of the major points which I'll summarise here:

### Renaming `cutoffs` to `hGridMinCellSizesPerLevel`

I know that `cutoffs` was the placeholder name since early in this project, but looking over all the code, I find `cutoffs` is a little misleading. The cutoffs provided do not necessarily reflect any actual cutoffs but are really the minimum cell sizes for each level (before scaling to perfectly fit the box). The distinction between cutoffs and cutoff is also potentially misleading.

I would simply rename this variable, either to my suggestion or something similar.

### Enabling scaling cutoff

It would be good to enable scaling cutoff without having to set `cutoffs`/`hGridMinCellSizesPerLevel` as it makes it clearer that this feature is activated or not.

For the purposes of AutoPas, this can simply be a regular boolean flag. This would enable hierarchical grids (and any other features which develop for this mode in the future). The assumption would be that, with this mode activated, we assume that the user correctly implements scaling cutoffs in their functors.

I would keep the `MD_FLEXIBLE_SCALING_CUTOFF` CMake variable and have this enable the AutoPas flag in addition to compiling different functors.


### Add User and Developer Documenation

This mode should be described in the user documentation (i.e. in `docs/userdoc`)

- What it is used for
- How to enable it
- What is expected from the user e.g. that they should implement `getSize()` and how it is used, that they need to correctly apply the scaled cutoff in their functor (including for two particles of different sizes)

From the developer documentation (Doxygen) viewpoint, simply add what the CI tells you is missing.


### Tests

Merge new test files into old to reduce code duplication.

Also missing are unit tests for `LJFunctor` with scaled cutoff enabled. Similarly, these don't need new tests probably but extensions of existing tests.


### Misc

There are various other minor points. These include a couple of optimisations where I would like to see performance results as I am concerned that they overcomplicate the code unnecessarily.


Let me know if you have any questions or any disagreements.

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -0,0 +1,557 @@
...
+            cellIndex = prevCells + cellCount;
+          }
+          particleIndex = 0;
+        }
+        prevCells += cellCount;
+        ++currentLevel;
+      } else {
+        return {std::get<0>(ret), std::get<1>(ret) + prevCells, std::get<2>(ret)};
+      }
+    }
```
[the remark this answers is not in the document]
AlexanderGlas: I think this behavior seems to match the documentation of ParticleContainerInterface::getParticle:
"If a index pair is given that does not exist but is also not beyond the last cell, the next fitting particle shall be returned. 
Example: If [4,2] does not exist, [5,1] shall be returned
(or whatever is the next particle that fulfills the iterator requirements).
If there is no next fitting particle {nullptr, 0, 0} is returned."

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -0,0 +1,557 @@
...
+                                                                                   iteratorBehavior, boxMin, boxMax);
+      if (std::get<0>(ret) == nullptr) {
+        // not found, check next level
+        if (cellIndex - prevCells < cellCount) {
+          // if current level was just finished, we need to increase cellIndex here
+          if (iteratorBehavior & IteratorBehavior::forceSequential) {
+            ++cellIndex;
+          } else {
+            cellIndex = prevCells + cellCount;
+          }
```
SamNewcome, line 374: This doesn't make sense in my mind. Why typically skip a level but only advance a cell if `IteratorBehavior::forceSequential`

SamNewcome: In other containers, `IteratorBehavior::forceSequential` is used to force a stride of 1, whereas otherwise the parallelism occurs in a round-robin static load balancing. I think this should therefore be incorrect, or at least insufficiently explained.

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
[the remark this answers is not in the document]
SamNewcome: Yes, you are right.

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctor.h
```diff
@@ -32,11 +32,11 @@ namespace mdLib {
  * @tparam calculateGlobals Defines whether the global values are to be calculated (energy, virial).
  * @tparam countFLOPs counts FLOPs and hitrate
  * @tparam relevantForTuning Whether or not the auto-tuner should consider this functor.
- * @tparam scalingCutoff_T If set to true, the cutoff will be scaled by the sigma of the particles.
+ * @tparam scalingCutoff If set to true, the cutoff will be scaled by the sigma of the particles.
```
SamNewcome, line 35: Keep the `_T`

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctor.h
```diff
@@ -912,6 +924,7 @@ class LJFunctor
         for (size_t j = 0; j < vecsize; j++) {
           if constexpr (useMixing) {
             sigmaSquared = sigmaSquareds[j];
+            cutoffSquared = _cutoffSquared * sigmaSquared;
```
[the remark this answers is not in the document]
SamNewcome: +1 to copilot

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorAVX.h
```diff
@@ -38,11 +38,11 @@ namespace mdLib {
  * @tparam calculateGlobals Defines whether the global values are to be calculated (energy, virial).
  * @tparam relevantForTuning Whether or not the auto-tuner should consider this functor.
  * @tparam countFLOPs counts FLOPs and hitrate. Not implemented for this functor. Please use the AutoVec functor.
- * @tparam scalingCutoff_T If set to true, the cutoff will be scaled by the sigma of the particles.
+ * @tparam scalingCutoff If set to true, the cutoff will be scaled by the sigma of the particles.
```
SamNewcome, line 41: Keep the `_T`

examples/md-flexible/src/configuration/MDFlexConfig.h
```diff
@@ -513,8 +513,8 @@ class MDFlexConfig {
   /**
    * cutoffs for Hierarchical Grid container levels
```
SamNewcome, line 514: Change also the documentation

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/AxilrodTellerMutoFunctor.h
```diff
@@ -592,7 +592,7 @@ class AxilrodTellerMutoFunctor
   static_assert(sizeof(AoSThreadDataGlobals) % 64 == 0, "AoSThreadDataGlobals has wrong size");
   static_assert(sizeof(AoSThreadDataFLOPs) % 64 == 0, "AoSThreadDataFLOPs has wrong size");
 
-  const double _cutoffSquared;
+  double _cutoffSquared;
```
SamNewcome, line 595: Leave the `const`

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctor.h
```diff
@@ -1195,7 +1208,7 @@ class LJFunctor
   static_assert(sizeof(AoSThreadDataGlobals) % 64 == 0, "AoSThreadDataGlobals has wrong size");
   static_assert(sizeof(AoSThreadDataFLOPs) % 64 == 0, "AoSThreadDataFLOPs has wrong size");
 
-  const double _cutoffSquared;
+  double _cutoffSquared;
```
SamNewcome, line 1211: Leave the `const`

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorAVX.h
```diff
@@ -1005,13 +1015,13 @@ class LJFunctorAVX
   };
   const __m256i _ownedStateDummyMM256i{0x0};
   const __m256i _ownedStateOwnedMM256i{_mm256_set1_epi64x(static_cast<int64_t>(autopas::OwnershipState::owned))};
-  const __m256d _cutoffSquared{};
+  __m256d _cutoffSquared{};
```
SamNewcome, line 1018: Leave the `const`

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorAVX.h
```diff
@@ -1005,13 +1015,13 @@ class LJFunctorAVX
...
   __m256d _shift6 = _mm256_setzero_pd();
   __m256d _epsilon24{};
   __m256d _sigmaSquared{};
 #endif
 
-  const double _cutoffSquaredAoS = 0;
+  double _cutoffSquaredAoS = 0;
```
SamNewcome, line 1024: Leave the `const`

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorSVE.h
```diff
@@ -885,7 +887,7 @@ class LJFunctorSVE
   double _sigmaSquared{0.};
 #endif
 
-  const double _cutoffSquaredAoS;
```
SamNewcome, line 888: Leave the `const`

src/autopas/baseFunctors/Functor.h
```diff
@@ -264,6 +264,9 @@ class Functor {
   }
 
  protected:
+  /**
+   * Current interaction cutoff used by the functor.
+   */
```
SamNewcome, line 269: I think this `_cutoff` was originally modified during the simulation to use different cutoffs. But this was not done in the final version.

You can remove the `protected` and the word "Current"

src/autopas/containers/hierarchicalGrid/traversals/HGBlockTraversal.h
```diff
@@ -28,16 +28,27 @@ namespace autopas {
...
 
-  explicit HGBlockTraversal(Functor_T *functor, DataLayoutOption dataLayout, bool useNewton3, int blockMultiplier)
-      : HGTraversalBase<ParticleCell_T>(dataLayout, useNewton3), _blockMultiplier(blockMultiplier), _functor(functor) {}
+  /**
+   * Constructor.
+   * @param functor Pairwise functor used for interactions.
+   * @param numLevels Number of levels in the hierarchical grid.
+   * @param dataLayout Data layout used by the traversal.
+   * @param useNewton3 Whether Newton3 optimization is enabled.
+   * @param blockMultiplier Multiplier controlling target blocks per color.
```
SamNewcome, line 42: Add more description

SamNewcome: Also, in the class documentation, it specifies this as either 4 or 8. Either that documentation should be generalised or there needs to be some error checking here.

src/autopas/containers/hierarchicalGrid/traversals/HGBlockTraversal.h
```diff
@@ -28,16 +28,27 @@ namespace autopas {
...
+   */
+  explicit HGBlockTraversal(Functor_T *functor, size_t numLevels, DataLayoutOption dataLayout, bool useNewton3,
+                            int blockMultiplier)
+      : HGTraversalBase<ParticleCell_T>(numLevels, dataLayout, useNewton3),
+        _blockMultiplier(blockMultiplier),
+        _functor(*functor) {}
```
SamNewcome, line 48: I like the `functor` has been switched from a pointer to a reference. But here you also don't need to use pointers, you can take a reference to the `functor` as an argument of the constructor.

src/autopas/containers/hierarchicalGrid/traversals/HGBlockTraversal.h
```diff
@@ -0,0 +1,210 @@
...
+      }
+      const size_t stride_x = stride[0], stride_y = stride[1], stride_z = stride[2];
+
+      // do the colored traversal
+      const size_t numColors = stride_x * stride_y * stride_z;
+      AUTOPAS_OPENMP(parallel)
+      for (size_t col = 0; col < numColors; ++col) {
+        const std::array<size_t, 3> startIndex(utils::ThreeDimensionalMapping::oneToThreeD(col, stride));
+
+        AUTOPAS_OPENMP(for schedule(dynamic, 1) collapse(3))
```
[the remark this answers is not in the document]
SamNewcome: Don't do this

src/autopas/containers/hierarchicalGrid/traversals/HGBlockTraversal.h
```diff
@@ -0,0 +1,210 @@
...
+   * @return A new traversal that is applicable to a specific LinkedCells level
+   */
+  std::unique_ptr<TraversalInterface> generateNewTraversal(const size_t level) override {
+    const auto traversalInfo = this->getTraversalSelectorInfo(level);
+    return std::make_unique<LCC08Traversal<ParticleCell_T, Functor_T>>(
+        traversalInfo.cellsPerDim, &_functor, traversalInfo.interactionLength, traversalInfo.cellLength,
+        this->_dataLayout, this->_useNewton3);
+  }
+
+  [[nodiscard]] bool isApplicable() const override { return true; }
```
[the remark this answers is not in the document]
SamNewcome: +1 to Copilot's suggestion -> this is really important

src/autopas/containers/hierarchicalGrid/traversals/HGBlockTraversal.h
```diff
@@ -131,23 +142,69 @@ class HGBlockTraversal : public HGTraversalBase<ParticleCell_T>, public HGTraver
...
     return std::make_unique<LCC08Traversal<ParticleCell_T, Functor_T>>(
-        traversalInfo.cellsPerDim, _functor, traversalInfo.interactionLength, traversalInfo.cellLength,
+        traversalInfo.cellsPerDim, &_functor, traversalInfo.interactionLength, traversalInfo.cellLength,
         this->_dataLayout, this->_useNewton3);
   }
+
+  [[nodiscard]] bool isApplicable() const override { return true; }
+
+  /**
+   * Finds the best group size for a given target number of blocks per color.
```
SamNewcome, line 170: Needs to be clearer what "group size" is

src/autopas/containers/hierarchicalGrid/traversals/HGBlockTraversal.h
```diff
@@ -28,16 +28,27 @@ namespace autopas {
...
+   * @param useNewton3 Whether Newton3 optimization is enabled.
+   * @param blockMultiplier Multiplier controlling target blocks per color.
+   */
+  explicit HGBlockTraversal(Functor_T *functor, size_t numLevels, DataLayoutOption dataLayout, bool useNewton3,
+                            int blockMultiplier)
+      : HGTraversalBase<ParticleCell_T>(numLevels, dataLayout, useNewton3),
+        _blockMultiplier(blockMultiplier),
+        _functor(*functor) {}
 
   void traverseParticles() override {
```
SamNewcome, line 50: IMO, this needs some work as it is quite confusing.

src/autopas/containers/hierarchicalGrid/traversals/HGTraversalBase.h
```diff
@@ -17,53 +18,66 @@ namespace autopas {
...
-   * @param maxDisplacement maximum displacement of any particle in the container, ignored if dynamic containers is not
-   * used
-   * @param stepsSinceLastRebuild number of time-steps since last rebuild
-   * @param rebuildFrequency frequency of rebuild
    */
-  void setLevels(std::vector<std::unique_ptr<LinkedCells<Particle>>> *levels, std::vector<double> &cutoffs, double skin,
-                 double maxDisplacement, unsigned int stepsSinceLastRebuild, const unsigned int rebuildFrequency) {
-    _numLevels = cutoffs.size();
-    _cutoffs = cutoffs;
+  void setLevels(std::vector<std::unique_ptr<LinkedCells<ParticleType>>> *levels,
```
SamNewcome, line 41: This is not an appropriate use of `unique_ptr`, as in no way are `levels` owned by this object. I would either find a way to pass this as a reference in the constructor, or just use raw pointers (less nice, but at least honest)

src/autopas/containers/hierarchicalGrid/traversals/HGTraversalBase.h
```diff
@@ -17,53 +18,66 @@ namespace autopas {
...
   double _skin;
-  double _maxDisplacement;
-  unsigned int _stepsSinceLastRebuild;
-  unsigned int _rebuildFrequency;
-  // the ratio between cutoffs of levels from which point distance should be checked before checking distance of
-  // particles can do particle to level specific instead of level to level specific in the future
+  /**
+   * Cutoff ratio threshold for enabling cell distance checks.
+   * At this ratio of max Cutoffs between two levels, distance checks are employed to sort out unnecessary interactions.
+   * Particles can do particle to level specific instead of level to level specific in the future.
```
SamNewcome, line 69: This is still unclear

src/autopas/containers/hierarchicalGrid/traversals/HGTraversalBase.h
```diff
@@ -74,45 +88,35 @@ class HGTraversalBase : public TraversalInterface {
...
-   * Compute the color for the given level.
-   * Computed the needed stride by checking all lower levels and takes the maximum for each dimension.
+   * Compute the needed stride to avoid race conditions.
+   * For SoA consider that two cells interacting on the same lower level cell is problematic, as two threads can operate
+   * on the same buffer.
    * @param level The hierarchy level to compute the stride for.
+   * @param topDown If true compute stride for top-down traversal, otherwise bottom-up.
    * @return The stride for each dimension.
    */
   std::array<size_t, 3> computeStride(const size_t level, const bool topDown = true) {
```
SamNewcome, line 122: This function is only relevant to the Block traversal it seems -> move to block traversal. If you use it in C08, keep it here.

src/autopas/containers/hierarchicalGrid/traversals/HGTraversalBase.h
```diff
@@ -231,21 +201,17 @@ class HGTraversalBase : public TraversalInterface {
...
     }
   }
 
   void initTraversal() override {
     this->_traversals.resize(this->_numLevels);
-    for (size_t i = 0; i < this->_numLevels; ++i) {
-      this->_traversals[i] = this->generateNewTraversal(i);
+    for (size_t level = 0; level < this->_numLevels; ++level) {
+      this->_traversals[level] = this->generateNewTraversal(level);
+      TraversalInterface *intraLevelTraversal = this->_traversals[level].get();
```
SamNewcome, line 212:
```suggestion
      // prepareTraversal requires a raw pointer
      this->_traversals[level] = this->generateNewTraversal(level);
      TraversalInterface *intraLevelTraversal = this->_traversals[level].get();
```

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -22,46 +22,43 @@ namespace autopas {
...
+   * @param loadEstimator the load estimation algorithm for balanced traversals.
+
    */
   HierarchicalGrid(const std::array<double, 3> &boxMin, const std::array<double, 3> &boxMax,
-                   const std::vector<double> &cutoffs, const double skin, const unsigned int rebuildFrequency,
-                   const double cellSizeFactor = 1.0)
-      : ParticleContainerInterface<Particle>(skin),
+                   const std::vector<double> &hGridMaxCutoffPerLevel, const double skin,
+                   const double cellSizeFactor = 1.0, const size_t sortingThreshold = 8,
+                   LoadEstimatorOption loadEstimator = LoadEstimatorOption::squaredParticlesPerCell)
```
SamNewcome, line 48: I would remove this. The Load Estimator is only relevant with (sliced) balanced traversals, and I do not care about these with HGrids as I doubt we ever get homogeneous enough that these become worthwhile.

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -70,46 +67,47 @@ class HierarchicalGrid : public ParticleContainerInterface<Particle> {
     _currentLevel.resize(autopas_get_max_threads() * _cacheOffset, 0);
     _prevCells.resize(autopas_get_max_threads() * _cacheOffset, 0);
 
-    if (_cutoffs.empty()) {
-      utils::ExceptionHandler::exception("Error: Hierarchical Grid cutoffs vector is empty.");
-    }
+    // In case cutoffs are empty, the traversals will return not applicable, as it makes no sense to use HG, instead of
+    // LC
+
```
SamNewcome, line 72: I would remove this comment. You could add to the class documentation that the objects of the class can still be constructed, with the assumption that in cases where there are no `hGridMaxCutoffPerLevel`, this configuration is skipped at `Traversal::isApplicable`.

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -70,46 +67,47 @@ class HierarchicalGrid : public ParticleContainerInterface<Particle> {
...
     // largest interaction length
-    const double interactionLengthLargest = _cutoffs.back() + _skin;
+    const double interactionLengthLargest = _maxCutoffPerLevel.back() + _skin;
     // generate LinkedCells for each hierarchy, with different cutoffs
     _levels.reserve(_numLevels);
     for (size_t i = 0; i < _numLevels; ++i) {
-      // here, LinkedCells with smaller cutoffs need to have bigger halo regions, because particles on level above
+      // here, LinkedCells with smaller cellSizes need to have bigger halo regions, because particles on level above
       // can potentially interact with those, even though these halo particles cannot interact with other particles
       // inside its own level a hacky way: make all LinkedCells interactionLength equal to the biggest one, adjust
```
SamNewcome, line 103: Make "a hacky way" a new line

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -70,46 +67,47 @@ class HierarchicalGrid : public ParticleContainerInterface<Particle> {
...
       // cellSizeFactor for each LinkedCells so that the cellLength is equal to actual interaction length of that level
-      const double interactionLengthLevel = _cutoffs[i] + _skin;
+      // For traversals, we also choose the proper, smaller interaction length
```
SamNewcome, line 105:
```suggestion
      // For traversals, we still use the proper, smaller interaction length
```

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -307,18 +304,20 @@ class HierarchicalGrid : public ParticleContainerInterface<Particle> {
 
   [[nodiscard]] TraversalSelectorInfo getTraversalSelectorInfo() const override {
     // return traversal info of hierarchy with biggest interactionLength for autoPas container
-    return _levels.back()->getTraversalSelectorInfo();
+    auto infoHighest = _levels.back()->getTraversalSelectorInfo();
+    return TraversalSelectorInfo(infoHighest.cellsPerDim, infoHighest.interactionLength, infoHighest.cellLength, 0,
+                                 _numLevels);
   }
```
SamNewcome, line 310: Bit concerned that in the future developers might not realise that this info is potentially misleading for HGrids. 

Add to the `TraversalSelectorInfo` class, for the documentation of `cellsPerDim`, `interactionLength`, and `cellLength` the caveat: "In the case of a Hierarchical Grid container, this is the cellsPerDim/interactionLength/cellLength of the coarsest grid."

Add here to the function documentation that it returns the traversal info of the coarsest grid in the hierarchy.

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -475,63 +473,46 @@ class HierarchicalGrid : public ParticleContainerInterface<Particle> {
...
+  /**
+   * Per-thread index of the hierarchy level currently used by getParticleImpl().
+   */
   mutable std::vector<size_t> _currentLevel;
+  /**
+   * Per-thread number of cells already skipped in earlier hierarchy levels during getParticleImpl().
+   */
   mutable std::vector<size_t> _prevCells;
+  /**
+   * Cache-line padding factor for _currentLevel and _prevCells to reduce false sharing between threads.
```
SamNewcome, line 491: Is "factor" the correct term here? I would expect this to mean something that is multiplied with something.

Maybe "Offset to provide cache-line padding for..."

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -475,63 +473,46 @@ class HierarchicalGrid : public ParticleContainerInterface<Particle> {
...
+   */
   const size_t _cacheOffset;
-  const unsigned int _rebuildFrequency;
-  std::vector<std::unique_ptr<LinkedCells<Particle>>> _levels;
-  std::vector<double> _cutoffs;
+
+  std::vector<std::unique_ptr<LinkedCells<Particle_T>>> _levels;
+  std::vector<double> _maxCutoffPerLevel;
```
SamNewcome, line 496: Add documentation

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -475,63 +473,46 @@ class HierarchicalGrid : public ParticleContainerInterface<Particle> {
...
-  size_t getHierarchyLevel(const ParticleType &p) const {
+  size_t getHierarchyLevel(const Particle_T &p) const {
     const double cutoff = p.getSize();
     for (size_t i = 0; i < _numLevels; ++i) {
-      if (_cutoffs[i] >= cutoff) {
+      if (_maxCutoffPerLevel[i] >= cutoff) {
         return i;
       }
     }
     AutoPasLog(ERROR,
```
SamNewcome, line 512: Use exception instead of Error log

src/autopas/AutoPasDecl.h
```diff
@@ -604,20 +604,28 @@ class AutoPas {
...
-  [[nodiscard]] std::vector<double> getCutoffs() const { return _logicHandlerInfo.cutoffs; }
+  [[nodiscard]] std::vector<double> getHGridMaxCutoffPerLevel() const {
+    return _logicHandlerInfo.hGridMaxCutoffPerLevel;
+  }
 
   /**
-   * Set cutoffs for each level of Hierarchical Grid container
-   * @param cutoffs
+   * Set the maximum Cutoffs for each level of the Hierarchical Grid container. If this container is used, particles
+   * will be sorted into the grid level with the smallest minimum cell size that is still larger than that particle's
```
SamNewcome, line 623: "smallest maximum cutoff that is still" -> I think this better reflects your terminology.

src/autopas/AutoPasDecl.h
```diff
@@ -603,6 +603,31 @@ class AutoPas {
...
+   * size scaled by the cutoff. The actual cell sizes could be larger, due to potentially adding a skin to each cell or
+   * increasing cell sizes to perfectly fit the box. If scaled cutoffs are not used, this does nothing.
+   * @param hGridMaxCutoffPerLevel
+   */
+  void setHGridMaxCutoffPerLevel(const std::vector<double> hGridMaxCutoffPerLevel) {
+    if (std::any_of(hGridMaxCutoffPerLevel.begin(), hGridMaxCutoffPerLevel.end(),
+                    [](double cellSize) { return cellSize <= 0.0; })) {
+      utils::ExceptionHandler::exception(
+          "Error: The maximum cutoff sizes used by the HGrid container have to be positive: {} <= 0.0!");
```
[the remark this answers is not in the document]
SamNewcome: +1 to copilot's feedback

src/autopas/LogicHandlerInfo.h
```diff
@@ -44,5 +45,10 @@ class LogicHandlerInfo {
    * This is currently used in rebuild frequency estimation for dynamic containers.
    */
   double deltaT{0};
+  /**
+   * The (minimum) cell size for each level of a HierarchicalGrid.
+   * If not set (i.e. empty), HierarchicalGrid will not be considered as a valid option for tuning.
```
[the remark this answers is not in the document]
SamNewcome: Ignore copilot -> HGrid should get excluded in the single level case e.g. in `Traversal::isApplicable`

tests/testAutopas/tests/containers/TraversalComparison.cpp
```diff
@@ -10,10 +10,14 @@
 
 #include "autopas/tuning/selectors/ContainerSelector.h"
 #include "autopas/tuning/selectors/TraversalSelector.h"
-#include "autopas/utils/StaticCellSelector.h"
 #include "autopas/utils/StringUtils.h"
 #include "autopasTools/generators/UniformGenerator.h"
 
+/*
+ * @todo: Integrate HgridTraversalComparison tests according to
+ * https://github.com/AutoPas/AutoPas/pull/1014/changes/BASE..ad9eb3091137eb141f23c2a82bbb51186f0c0cb5#r2171213075
```
SamNewcome, line 18: This does not seem to link to a comment

SamNewcome: I guess this refers to https://github.com/AutoPas/AutoPas/pull/1014#discussion_r2171213075

I would undo deleting the HGridTraversalComparison tests. Otherwise you've simply removed the thing to merge into this file, without actually doing the merge.

src/autopas/containers/linkedCells/HierarchicalGrid.h
```diff
@@ -0,0 +1,557 @@
...
+    // make sure cutoffs are sorted
+    std::sort(_cutoffs.begin(), _cutoffs.end());
+    // make all cutoffs unique
+    _cutoffs.resize(std::unique(_cutoffs.begin(), _cutoffs.end()) - _cutoffs.begin());
+    if (_cutoffs.size() != _numLevels) {
+      AutoPasLog(ERROR, "Cutoffs should all be unique, adjusting cutoffs size from {} to {}", _numLevels,
```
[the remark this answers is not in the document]
SamNewcome: Changed to `WARN`

src/autopas/containers/linkedCells/traversals/HGTraversalBase.h
```diff
@@ -0,0 +1,458 @@
...
+    _levels = levels;
+    _skin = skin;
+    _maxDisplacement = maxDisplacement;
+    _stepsSinceLastRebuild = stepsSinceLastRebuild;
+    _rebuildFrequency = rebuildFrequency;
+  }
+
+ protected:
+  size_t _numLevels;
+  std::vector<std::unique_ptr<LinkedCells<Particle>>> *_levels;
```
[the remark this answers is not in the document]
SamNewcome: This was marked resolved but not actually resolved.

src/autopas/containers/hierarchicalGrid/traversals/HGTraversalBase.h
```diff
@@ -74,45 +88,35 @@ class HGTraversalBase : public TraversalInterface {
...
-    const double cutoff = (this->_cutoffs[upperLevel] + this->_cutoffs[lowerLevel]) / 2;
-    return cutoff + currentSkin();
+    const double cutoff = (this->_maxCutoffPerLevel[upperLevel] + this->_maxCutoffPerLevel[lowerLevel]) / 2;
+    return cutoff + _skin;
   }
 
   /**
-   * Check if ratio between two levels is small enough to consider checking distance of each cell to early break
-   * and not call functors.
+   * Check if the ratio of level grid sizes is small enough to that the distance between each cell should be checked
```
SamNewcome, line 102:
```suggestion
   * Check if the ratio of level grid sizes is small enough that the distance between each cell should be checked
```

SamNewcome (changes requested):
### Review for @AlexanderGlas 's changes to this branch.

Here are my suggestions.

Overall, there's a lot of really nice work here making the baseline HGrid code much cleaner and addressing some issues. Most issues with your changes are relatively minor.

What I have not done is thoroughly gone through the resolved comments from my original Atacan review. I spotted at least one resolved comment that hadn't actually been resolved, and at least one other where it was unclear how/if it was resolved. Typically, only resolve comments where it is clearly resolved and leave a comment unless it is really obvious that it is resolved suitably. A nice simple example of this is #1068: except for the applied suggested change, Markus responded to my comments with what was done including linking to a commit ID.

Please go through the resolved comments, double-check that they have actually been done, and write comments if it is not clear how/if they have been resolved. I'm not super bothered if you can refer to a commit ID (although this would be ideal in the future e.g. for changes in the C08 branch)

src/autopas/containers/linkedCells/traversals/HGTraversalBase.h
```diff
@@ -0,0 +1,458 @@
...
+  }
+
+  /**
+   * Compute intra-level interactions for all levels.
+   */
+  void computeIntraLevelInteractions() {
+    for (size_t level = 0; level < this->_numLevels; level++) {
+      // We do not simply call computeInteractions() here as we want to store SoA after inter-level traversals are
+      // computed. They will be loaded in HGTraversalBase::endTraversal().
+      TraversalInterface *temp = this->_traversals[level].get();
```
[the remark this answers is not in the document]
AlexanderGlas: Changed the name, however, not sure what is meant by changing the pointer to an address/not sure how to implement this elegantly.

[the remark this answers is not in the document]
SamNewcome: Yeah, it's a bit awkward looking at it in more detail. I'd just leave it.

src/autopas/containers/linkedCells/traversals/HGTraversalBase.h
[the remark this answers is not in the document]
AlexanderGlas: Sorry, because of the question mark, I thought this was optional/a matter of taste.

src/autopas/containers/hierarchicalGrid/traversals/HGBlockTraversal.h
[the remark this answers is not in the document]
AlexanderGlas: Yes, I wrote this in the HedgeDoc. The return true is a temporary solution, as otherwise the CI pipeline would currently fail. Some test expects a compatible traversal for some Hgrid configuration, and I was not sure how to/ did not have time to properly adapt the tests.

src/autopas/LogicHandler.h
```diff
@@ -1161,13 +1162,24 @@ void LogicHandler<Particle_T>::checkNeighborListsInvalidDoDynamicRebuild() {
...
   }
+  if (_neighborListInvalidDoDynamicRebuild) {
+    // will be rebuilt, set max displacement to 0
+    getContainer().setMaxDisplacement(0);
+  } else {
+    // set maxDisplacement to the maximum over all particles
+    // trim if displacement is larger than skin/2 (can cause problems in some tests otherwise)
+    getContainer().setMaxDisplacement(std::sqrt(maxDistanceSquare) + 1e-15);
+  }
+
```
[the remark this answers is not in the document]
AlexanderGlas: This was simply removed, without testing for benefits.

src/autopas/containers/TraversalInterface.h
```diff
@@ -68,6 +68,13 @@ class TraversalInterface {
    */
   [[nodiscard]] DataLayoutOption getDataLayout() const { return _dataLayout; }
 
+  /**
+   * This function os used in case halo region length is different from interactionLength, i.e. for Hierarchical Grid's
+   * lower levels.
+   * @param haloRegionLength length of the halo region, it is usually cutoff + skin
+   */
+  virtual void setHaloRegionLength(const double haloRegionLength) {}
+
```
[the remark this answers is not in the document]
AlexanderGlas: All this logic was simply removed; however, in the HgridC08 branch, the intra-level traversal is performed using HGC08, which does respect potentially larger halo regions.

src/autopas/containers/hierarchicalGrid/traversals/HGBlockTraversal.h
AlexanderGlas: Can do this, of course; however, I think all the other traversals use a pointer, so it looks odd in the TraversalSelector when all traversals take a pointer, except for HGBlockTraversal.

tests/testAutopas/tests/containers/TraversalComparison.cpp
AlexanderGlas: Makes sense. I removed it because it did not compile after master was merged into Hgrid due to some interface changes, and I did not want to spend time fixing it, since it was meant to be removed anyway.

SamNewcome: Okay, I'm happy to keep it removed. It's still there in the git history.

src/autopas/containers/hierarchicalGrid/traversals/HGBlockTraversal.h
SamNewcome: I don't mind too much if it is simply that other traversals use a pointer -> they should really use a reference anyway. But if doing this for `HGBlockTraversal` would require changing the interface for all already, then just leave things as they are.

src/autopas/containers/hierarchicalGrid/traversals/HGBlockTraversal.h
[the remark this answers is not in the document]
SamNewcome: Was the test `AutoPasInterfaceTest::ConfighasCompatibleValuesVSTraversalIsApplicable`? If so, implement it as copilot suggests and just leave that test failing. This test is flawed (#1031) and will get removed in PR #1071 anyway.

src/autopas/containers/linkedCells/traversals/HGTraversalBase.h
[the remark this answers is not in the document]
SamNewcome: No that's fair, but it would be good to have your reason for keeping it here. If this is simply, you wanted to focus on C08 so didn't address everything in this branch, that's totally fine, but then I would leave this unresolved. If there's a reason for keeping this as a unique pointer, then write down why.

src/autopas/containers/hierarchicalGrid/traversals/HGTraversalBase.h
```diff
@@ -0,0 +1,381 @@
...
+   */
+  void computeIntraLevelInteractions() {
+    for (size_t level = 0; level < this->_numLevels; level++) {
+      // We do not simply call computeInteractions() here as we want to store SoA after inter-level traversals are
+      // computed. They will be loaded in HGTraversalBase::endTraversal().
+      this->_traversals[level]->traverseParticles();
+    }
+  }
+
+  void initTraversal() override {
```
SamNewcome, line 208: This should be `public` visibility.

src/autopas/containers/hierarchicalGrid/traversals/HGTraversalBase.h
```diff
@@ -0,0 +1,381 @@
...
+    this->_traversals.resize(this->_numLevels);
+    for (size_t level = 0; level < this->_numLevels; ++level) {
+      this->_traversals[level] = this->generateNewTraversal(level);
+      TraversalInterface *intraLevelTraversal = this->_traversals[level].get();
+      this->_levels->at(level)->prepareTraversal(intraLevelTraversal);
+      this->_traversals[level]->initTraversal();
+    }
+  }
+
+  void endTraversal() override {
```
SamNewcome, line 218: This should be `public` visibility.

src/autopas/containers/hierarchicalGrid/traversals/HGTraversalBase.h
```diff
@@ -0,0 +1,381 @@
...
+    // stores SoA if SoA is used
+    for (size_t level = 0; level < this->_numLevels; level++) {
+      this->_traversals[level]->endTraversal();
+    }
+  }
+
+  // HG traversals only make sense if there are multiple levels, otherwise just use LinkedCells
+  [[nodiscard]] bool isApplicable() const override { return _numLevels > 1; }
```
SamNewcome, line 226: This should be `public` visibility.

SamNewcome (commented):
Some extra things I noticed during review of C08 branch

src/autopas/containers/hierarchicalGrid/traversals/HGTraversalBase.h
SamNewcome, line 1: This class is basically the "Interface" class for HG traversals. Meanwhile the actual `HGTraversalInterface` is kinda just a formality. I would merge the two together so there is just one `HGTraversalInterface` class.

PR #1016, opened 2025-05-07
Fix `AutoTunerTest::testBuildNotBuildTimeEstimation`'s non-determinism

SamNewcome:
# Description

`AutoTunerTest::testBuildNotBuildTimeEstimation` tests the weighted averaging between build and non build iterations. It currently does this by computing interactions between two particles with a mock functor that simply sleeps using [`std::this_thread::sleep_for`](https://en.cppreference.com/w/cpp/thread/sleep_for). This function "[b]locks the execution of the current thread for **at least** the specified `sleep_duration`". Therefore, it is not deterministic and as such, the test can fail due to sleeps longer than  `sleep_duration`.

This PR makes this test deterministic by, instead of mocking an interaction taking a certain amount of time, directly adding "measurements" to the autoTuner.

(I also changed the `autoTunerInfo` default for `useLOESSSmoothening` to `false` in line with e.g. the default in md-flexible)

# How Has This Been Tested?

- [ ] CI

### DIFF
--- a/src/autopas/tuning/utils/AutoTunerInfo.h
+++ b/src/autopas/tuning/utils/AutoTunerInfo.h
@@ -43,6 +43,6 @@ struct AutoTunerInfo {
   /**
    * Flag for whether LOESS Smoothening is used to smoothen the tuning results.
    */
-  bool useLOESSSmoothening{true};
+  bool useLOESSSmoothening{false};
 };
 }  // namespace autopas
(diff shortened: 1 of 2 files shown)


manishmishra6016 (approved):
I think this looks good :+1: 
Manually adding the measurements is a nice idea!

thesamriel (approved):
Good change! I had some trouble with this before as well.

PR #1017, opened 2025-05-08
Update dependencies and CMake

SamNewcome:
# Description

This PR includes a few updates so that AutoPas can be used error- and warning-free with CMake 4.0.

More specifically:

- yaml-cpp is updated to the current master branch (commit `2f86d13`), which sets the minimum CMake version for compilation of yaml-cpp to 3.5. This is required because CMake 4.0 is incompatible with CMake < 3.5, leading to errors. Unfortunately, the latest release (0.8.0) doesn't include this fix.
- `FetchContent_Populate` is deprecated, and so we now use the recommended (and cleaner) `FetchContent_MakeAvailable` with yaml-cpp and gtest.
- We still require the deprecated `FetchContent_Populate` with Eigen, as we do not want the implicit `add_subdirectory` done by `FetchContent_MakeAvailable`. This command is "wrapped" by commands that set policy `CM0169` to `OLD` and then back to `NEW`, suppressing the warning produced.
- (There is a solution that allows us to use `FetchContent_MakeAvailable` without the implicit `add_subdirectory` but it is equally as hacky, so I didn't switch to it - but I added how it would be done, in case `FetchContent_Populate` is removed)
- ANTLR and its dependencies were updated. This meant that new parser-generated files needed to be generated. I do not believe that the generated files in the `autopas_generated_fuzzy_rule_syntax` and `autopas_generated_rule_syntax` directories need to be checked carefully. As ANTLR migrated from their custom `antlr4cpp::Any` to `std::any`, some syntax needed to be changed. Finally, to aid in future maintainability, I included the website where ANTLR C++ runtimes can be found and the commands to generate new files.

# How Has This Been Tested?

- [x] CI (Checks compilation with minimum CMake version)
- [x] Compilation with CMake 4.0.


schuhmaj (approved):
Looks all good to me 👍. Thank you for the update.

And good to know that hacky alternative if `FetchContent_Populate` gets removed, I also have an application somewhere else for this :)

PR #1019, opened 2025-05-16
Fixing Optional PMT

SamNewcome:
# Description

The CMake option `AUTOPAS_ENABLE_ENERGY_MEASUREMENTS` is first declared in `src/autopas/CMakeLists.txt` _after_ the `autopas_pmt.cmake` module has been run. This CMake module is wrapped in an `if (AUTOPAS_ENABLE_ENERGY_MEASUREMENTS)` so is skipped as this option is not yet added. Somehow (for CMake reasons I do not understand), CMake would still build pmt, it just would not know of the dependency of autopas on pmt, and begin building autopas before pmt was complete. This meant compilation could fail unless this option was directly passed in in the CMake command, e.g. as in the CI.

This PR moves the CMake option to the top of `autopas_pmt.cmake`, so it is already declared, fixing this issue.

# How Has This Been Tested?

- [x] CI
- [x] Local compilation with `cmake ..` followed by `make md-flexible -j6`, which fails on the `master` branch. 
- [x] Checking local compilation like above with `AUTOPAS_ENABLE_ENERGY_MEASUREMENTS = OFF`


manishmishra6016 (approved):
Looks good, thank you :+1:
I had kept the option in `src/autopas/CMakeLists.txt` because I thought we use it while calling `target_link_libraries()` (line 42/37) and I somehow thought it should be declared before this, but I see this is not the case. :sweat_smile: Options (eg.  `AUTOPAS_OPENMP`) are declared in respective `.cmake` file under modules.

I am still a bit unsure about exact order of these modules running and tend to feel like variable should have been set still before running `autopas_pmt.cmake` but I will clear that up in person. And this PR can be merged irrespective of it as it sticks to how we went about other options related to other libraries.

SamNewcome:
> Looks good, thank you 👍 I had kept the option in `src/autopas/CMakeLists.txt` because I thought we use it while calling `target_link_libraries()` (line 42/37) and I somehow thought it should be declared before this, but I see this is not the case. 😅 Options (eg. `AUTOPAS_OPENMP`) are declared in respective `.cmake` file under modules.

We do still declare this before the `target_link_libraries()`. If you look at the root `CMakeLists.txt`, you see that `add_subdirectory(libs)` (which in turn calls `include(autopas_pmt)`) is before `add_subdirectory(src)` (which in turn calls `add_subdirectory(autopas)`)

> 
> I am still a bit unsure about exact order of these modules running and tend to feel like variable should have been set still before running `autopas_pmt.cmake` but I will clear that up in person. And this PR can be merged irrespective of it as it sticks to how we went about other options related to other libraries.

Sounds good

schuhmaj (approved):

PR #1024, opened 2025-06-03
Make AutoPas use C++20

santiagonar1:
# Description

As part of an internal collaboration with Sam, we decided that it would be a good idea to bump the C++ standard version used by AutoPas from C++17 to C++20. This PR contains the minimum set of changes to make that happen. In particular, I have updated the documentation to indicate that C++20 is used by default and fixed, with the help of Sam, some minor bugs that were causing compilation errors. 

As a side note, inside the project, there are several `@todo c++20` that are low-hanging fruit, but I would recommend that those be addressed in separate pull requests. 

## Related Pull Requests

- None so far

## Resolved Issues

- Not any I could find

# How Has This Been Tested?

So far, I only checked that it compiled and that several tests from the CI are passing locally. I tried to run a ctest on my Mac, but (1) there are too many tests to make a full local check viable, and (2) some of them are timing out both on master and on this branch, so I prefer to check what your CI produces.

If you provide a subset of tests I should run locally, I would of course be happy to do so 😉.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -6,7 +6,7 @@ Please keep in mind the following notes while working.
 
 ## C++
 ### General Notes
-* Cpp standard: C++17. If there is a piece of code, which could be done better using a newer standard, please add a comment like `@todo C++20` including the alternative version of the code.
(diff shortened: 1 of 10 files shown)


SamNewcome (approved):
Looks good! Thanks!

PR #1040, opened 2025-10-27
Total energy measurement in MD-Flexible

manishmishra6016:
# Description

This PR adds a measurement for total energy consumed in MD-flexible. We use the same sensor specified by the user in the input file, which is currently used by iteration-level measurements in AutoPas. 

Caution while running MPI simulations with energy measurements on! RAPL files require `sudo`  access, but MPI strongly discourages running an MPI application with `sudo`. The way to circumvent this issue is to change the access to rapl files for all users using: ```sudo chmod -R a+r /sys/class/powercap/intel-rapl*```, before running the simulation (which will no longer need `sudo`). This is not required for runs on the Linux Cluster.

## Related Pull Requests

- #849 

# How Has This Been Tested?

- [x] Measurements with `AUTOPAS_ENABLE_ENERGY_MEASUREMENTS` as `ON`/`OFF` with and without MPI for fallingDrop scenario. Checking correlation with overall runtime.

### DIFF
--- a/docs/userdoc/Building.md
+++ b/docs/userdoc/Building.md
@@ -73,6 +73,22 @@ Additionally, an energy sensor used for measurement needs to be selected during 
 The default is set to `rapl` and will be used when no option is specified. 
 However, when multiple sensors are compiled, user can choose any of the available sensors by specifying the `energy-sensor` option.
 
+#### RAPL Permissions
+Energy measurements using Intel RAPL require read access to the system powercap files located at:
+```bash
+/sys/class/powercap/intel-rapl*
+```
+By default, these files are readable only by the root user and simulation need to be run with `sudo`. This is not ideal.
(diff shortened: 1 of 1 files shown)


SamNewcome (approved):

manishmishra6016:
> The code looks fine. Is the MPI warning you mention in the PR description also somewhere else? Or could we add it as a warning in md-flexible when energy measurements and MPI are both enabled? (Not entirely sure how problematic this really is)

`mpirun` already gives the warning before executing the code (screenshot below). They seem to strongly not recommend it. So, it seems like we don't need to give a warning in 'md-flexible' anything here? But maybe I could add the "solution" to the documentation, what do you think? I didn't add it because there are other ways to solve it as well, and I thought that just documenting it in PR would be enough.

image

SamNewcome (approved):

PR #1044, opened 2025-11-10
LifeInfo Parametrized Constructor

PatrickMetscher:
# LiveInfo Parametrized Constructor

This change added a parameterized LiveInfo constructor. This is required for external tools developed along with AutoPas. In my case this is a tool for estimating the runtime of a program based on previously collected tuning data.

# How Has This Been Tested?

- [X] Detailed Test for the parameter initialization
- [X] Simulations are unchanged (Compared main to this, output must be "identical", but some minor noise due to multi-threading)

### DIFF
--- a/src/autopas/AutoPasImpl.h
+++ b/src/autopas/AutoPasImpl.h
@@ -107,7 +107,7 @@ void AutoPas<Particle_T>::init() {
     tuningStrategies.reserve(_tuningStrategyOptions.size());
     for (const auto &strategy : _tuningStrategyOptions) {
       tuningStrategies.emplace_back(TuningStrategyFactory::generateTuningStrategy(
-          searchSpace, strategy, _tuningStrategyFactoryInfo, _outputSuffix));
+          searchSpace, strategy, _tuningStrategyFactoryInfo, interactionType, _outputSuffix));
     }
     if (_useTuningStrategyLoggerProxy) {
       tuningStrategies.emplace_back(std::make_unique<TuningStrategyLogger>(_outputSuffix));
--- a/src/autopas/tuning/tuningStrategy/TuningStrategyFactory.cpp
+++ b/src/autopas/tuning/tuningStrategy/TuningStrategyFactory.cpp
@@ -40,8 +40,9 @@ SearchSpaceGenerators::OptionSpace inferOptionDimensions(const std::set<Configur
 }
 
 std::unique_ptr<TuningStrategyInterface> generateTuningStrategy(const std::set<Configuration> &searchSpace,
-                                                                TuningStrategyOption tuningStrategyOption,
+                                                                const TuningStrategyOption tuningStrategyOption,
                                                                 const TuningStrategyFactoryInfo &info,
+                                                                const InteractionTypeOption interactionType,
                                                                 const std::string &outputSuffix) {
   std::unique_ptr<TuningStrategyInterface> tuningStrategy = nullptr;
   switch (static_cast<TuningStrategyOption>(tuningStrategyOption)) {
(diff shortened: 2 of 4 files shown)


tests/testAutopas/tests/tuning/tuningStrategy/LiveInfoTest.cpp
```diff
@@ -9,9 +9,200 @@
...
+      parameters = {
+          {10, 5,   2.0, 0.1,  10.0, 10.0, 10.0, 64, 4, 20, 8, 2, 1,   3,   2,   1,
+           3,  2.0, 0.5, 0.25, 100,  0.8,  0.05, 4,  1, 2,  1, 3, 2.0, 0.3, 0.15},
+          {50, 25,  3.0, 0.2, 20.0, 20.0, 20.0, 128, 8, 30, 16, 4, 2,   6,   4,  2,
+           6,  3.0, 1.0, 0.5, 200,  1.6,  0.1,  8,   2, 4,  2,  6, 3.0, 0.6, 0.3},
+          {25, 8,   0.750, 0.15, 15.0, 15.0, 15.0, 96, 6, 25, 12, 3, 1,   4,    3,  2,
+           4,  1.5, 0.75,  0.35, 150,  1.2,  0.08, 6,  1, 3,  1,  4, 1.5, 0.45, 0.2},
+          // Constructor with the lowest allowed values
+          {1, 0,       epsilon, 0.0, epsilon, epsilon, epsilon, 1, 1, 1, 1, 0, 0,       0,   0,  0,
+           0, epsilon, 0.0,     0.0, 1,       epsilon, 0.0,     1, 0, 0, 0, 0, epsilon, 0.0, 0.0},
```
SamNewcome: I don't think `epsilon` is the lowest positive number. But also, you tend to allow `0.0` in your checks. So maybe replace `epsilon` with `0.0`. I'm not totally sure of the purpose of this test, so check what I am saying.

tests/testAutopas/tests/tuning/tuningStrategy/LiveInfoTest.cpp
SamNewcome: Can you also add a unit test that checks that the number of infos in the `infos` map matches when `gather` is called versus your constructor? This ensures that if an extra statistic is added, it is not left out of your constructor.

tests/testAutopas/tests/tuning/tuningStrategy/LiveInfoTest.cpp
```diff
@@ -104,103 +99,26 @@ TEST_F(LiveInfoTest, ParametrizedConstructorCreatesCorrectInfo) {
 }
 
 /**
- * Tests that the parametrized constructor throws when given invalid input.
+ * Tests that the constructor initializes all required infos.
```
SamNewcome:
```suggestion
 * Tests that the constructor that takes the live infos as arguments initializes all required infos.
```

tests/testAutopas/tests/tuning/tuningStrategy/LiveInfoTest.cpp
```diff
@@ -104,103 +99,26 @@ TEST_F(LiveInfoTest, ParametrizedConstructorCreatesCorrectInfo) {
...
+ *
+ * This test verifies that no arguments are collected by the gather function wich are not gathered by the constructor.
```
SamNewcome:
```suggestion
 * This test verifies that all live infos collected by the gather function are also set by the constructor which takes them as arguments.
```

SamNewcome (changes requested):

SamNewcome (approved):
Looks good. Pull the updates from the master and wait until the CI goes green before merging (but I doubt there will be any issues)

PR #1049, opened 2025-11-27
Fix clang compilation with omp and address sanitizer

thesamriel:
# Description

I bumped the clang compiler version in our Docker image to clang-14, which enables more C++ 20 support.

For some reasons, the newer Clang version(s) have problems with the address sanitizer enabled, as it cannot find the correct OpenMP library for C++, after the initialization of the sanitizer.

The easy fix for this is placing the omp module before the sanitizers module in `CMakeLists.txt`


thesamriel:
Apparently, no longer problematic.

PR #1055, opened 2026-01-10
Feature/pseudo verlet lists

LarsD-UNI:
# Description
Implementation of a new container. Pseudo Verlet lists store 13 SortedCellViews for each cell, each of which projects the particles onto the cell connection axis to a neighboring cell. By sorting the particles, the loop can be terminated earlier without needing to check each particle to see if it is within the cutoff. By introducing a skin, the sorted cell views do not have to be updated at each calculation step. So far, pseudo Verlet lists are only available with AoS.

A new container, PseudoVerletLists, has been added that supports three traversals: c01, c08, and c18. In addition, a new cell functor, CellfunctorPsVL, has been created, which is used by the traversals.

# How Has This Been Tested?

The implementation was tested using the existing AllContainersTest and TraversalComparison tests. In addition, various MD-flexible examples, such as fallingDrop and explodingLiquid, were successfully executed for each traversal. New tests were also added to PseudoVerletListsTest.

- [x] AllContainersTest
- [x] TraversalComparison
- [x] PseudoVerletListsTest


SamNewcome (changes requested):
Nice. So the good is generally looking nice and I'm quite happy with the tests. There are no issues which must be addressed before collecting results.

The main issue is the code duplication, which we already alluded to earlier in the project. The actual container I don't really have any issues with, but I do with the cell functor and the traversals. I've suggested some solutions -  you are welcome to propose alternatives but I'd recommend that you comment these under where I've discussed solutions just so I can see the idea before you implement them, in case they aren't suitable.

The other issues relate are

- You don't correctly return `false` in `isApplicable` when using PsVL with cell size factor < 1. And in some other parts of the code, you hardcode CSF >= 1 too strongly. It is fine to only implement the CSF >= 1 case, but we should minimise the work that needs to be done to implement the CSF < 1 case, and we should also make it clear where parts of the code assume CSF >= 1.
- The documentation is sometimes missing details/unclear variable names/similar issues.

For all comments:

- If you have clearly resolved an issue, you can mark it as resolved. If you are in doubt, leave the comment as unresolved and I can do it in a later code review if I think it is resolved.
- Unless your changes to the code obviously fix an issue, comment what you have done to fix it.
- If you disagree with an issue, write it in a comment and include why. This is totally fine: sometimes I've made a mistake and sometimes what I've suggested won't work.

LarsD-UNI (commented):
I have addressed all your comments.

src/autopas/baseFunctors/CellFunctor.h
```diff
@@ -165,6 +262,18 @@ void CellFunctor<ParticleCell, ParticleFunctor, bidirectional>::processCell(Part
...
 }
 
+template <class ParticleCell, class ParticleFunctor, bool bidirectional>
+void CellFunctor<ParticleCell, ParticleFunctor, bidirectional>::processCell(unsigned long cellIndex) {
+  switch (static_cast<DataLayoutOption::Value>(_dataLayout)) {
+    case DataLayoutOption::aos:
+      processCellAoS(cellIndex);
+      break;
+    case DataLayoutOption::soa:
+      AutoPasLog(ERROR, "PsVL don't support soa!");
```
SamNewcome: Use an exception instead of an `ERROR`-log. (The error log is a bit of a niche use case - if there is an error we generally want the program to actually stop instead of just log something)

src/autopas/baseFunctors/CellFunctor.h
```diff
@@ -204,35 +313,61 @@ void CellFunctor<ParticleCell, ParticleFunctor, bidirectional>::processCellPair(
...
-      _functor->AoSFunctor(p1, p2, true);
-    } else {
-      if (not p1.isHalo()) {
-        _functor->AoSFunctor(p1, p2, false);
+void CellFunctor<ParticleCell, ParticleFunctor, bidirectional>::processCellPair(
+    unsigned long cell1Index, unsigned long cell2Index, const std::array<double, 3> &sortingDirection) {
+  signed long directionIndex = getDirectionIndex(sortingDirection);
+
+  // (Explicit) static cast required for Apple Clang (last tested version: 15.0.0)
+  switch (static_cast<DataLayoutOption::Value>(_dataLayout)) {
```
SamNewcome: Similar

src/autopas/baseFunctors/CellFunctor.h
```diff
@@ -100,6 +128,15 @@ class CellFunctor {
...
 
+  /**
+   * processCellPairsAoSN3 for PseudoVerletLists.
+   * @param cell1Index
+   * @param cell2Index
+   * @param directionIndex Normalized vector connecting centers of cell1 and cell2.
+   */
+  void processCellPairAoSN3(unsigned long cell1Index, unsigned long cell2Index, signed long directionIndex);
+
+  void processCellPairAoSNoN3Util(SortedCellView<ParticleCell> &cell1Sorted, SortedCellView<ParticleCell> &cell2Sorted);
```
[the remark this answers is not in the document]
SamNewcome: Also, use a more descriptive name that Util. This is really something like a processSortedCellPairAoSNoN3

src/autopas/baseFunctors/CellFunctor.h
```diff
@@ -117,6 +162,26 @@ class CellFunctor {
 
   void processCellSoANoN3(ParticleCell &cell);
 
+  void interactParticlesNoN3(auto &p1, auto &p2);
+
+  void interactParticles(auto &p1, auto &p2);
+
```
SamNewcome: Add documentation

src/autopas/baseFunctors/CellFunctor.h
SamNewcome: Similar

src/autopas/baseFunctors/CellFunctor.h
```diff
@@ -117,6 +162,26 @@ class CellFunctor {
 
   void processCellSoANoN3(ParticleCell &cell);
 
+  void interactParticlesNoN3(auto &p1, auto &p2);
+
+  void interactParticles(auto &p1, auto &p2);
```
SamNewcome: Add documentation

src/autopas/containers/linkedCells/traversals/LCC08CellHandler.h
```diff
@@ -61,6 +66,11 @@ class LCC08CellHandler {
    */
   void processBaseCell(std::vector<ParticleCell> &cells, unsigned long baseIndex);
 
+  /**
+   * @param baseIndex Index respective to which box is constructed.
+   */
+  void processBaseCell(unsigned long baseIndex);
```
SamNewcome: Add documentation -> Every function should have documentation saying what it does

src/autopas/containers/pseudoVerletLists/traversals/PsVLC01Traversal.h
```diff
@@ -35,28 +33,24 @@ class PsVLC01Traversal : public C01BasedTraversal<ParticleCell, PairwiseFunctor,
...
   void traverseParticles() override;
 
   /**
    * C01 traversals are only usable if useNewton3 is disabled.
    * @return
    */
   [[nodiscard]] bool isApplicable() const override {
-    if (this->_useNewton3 == false && this->_dataLayout == DataLayoutOption::aos) {
+    if (this->_useNewton3 == false && this->_dataLayout == DataLayoutOption::aos && this->_overlap[0] == 1 &&
+        this->_overlap[1] == 1 && this->_overlap[2] == 1) {
```
SamNewcome: Replace `&&` with `and`.

You need to update the documentation.

src/autopas/containers/pseudoVerletLists/PseudoVerletLists.h
```diff
@@ -116,21 +118,22 @@ class PseudoVerletLists : public VerletListsLinkedBase<Particle_T> {
...
-   * neighboring cell.
+   * Orientation List: or each cell, for each direction, a sortedCellView is stored, sorted along a
+   * connecting vector between this cell center and the center of a neighboring cell.
+   * Only the sortedCellView with directions to neighboring cells with a greater Index than the base cell are stored.
    */
   std::vector<std::vector<SortedCellView<ParticleCellType>>> _orientationList;
 
   /**
-   * Stores the normalized directions to the neighboring cells.
+   * Stores the normalized directions to the neighboring cells. With an index greater than the base cell.
```
SamNewcome:
```suggestion
   * Stores the normalized directions to the neighboring cells with an index greater than the base cell.
```

SamNewcome (changes requested):
Looks much nicer and a lot less code duplication than before. The main issue now is the `CellFunctor`. I've added a comment on the file itself.

src/autopas/baseFunctors/CellFunctor.h
```diff
@@ -109,61 +116,94 @@ class CellFunctor {
...
-  void processCellPairAoSN3Util(SortedCellView<ParticleCell> &cell1Sorted, SortedCellView<ParticleCell> &cell2Sorted);
+  void processCellAoSPsVL(unsigned long cellIndex);
 
   /**
    * Applies the functor to all particle pairs between cell1 and cell2
    * exploiting newtons third law of motion.
    * @param cell1
    * @param cell2
    * @param sortingDirection Normalized vector connecting centers of cell1 and cell2.
+   * @tparam newton3 determines if Newton3 is used or not. Matches _useNewton3.
```
SamNewcome:
```suggestion
   * @tparam newton3 determines if Newton3 is used or not. The version of this function actually run must match _useNewton3.
```
Technically, both versions are still compiled regardless of `_useNewton3`. This makes this "matching" clearer.

src/autopas/baseFunctors/CellFunctor.h
```diff
@@ -109,61 +116,94 @@ class CellFunctor {
...
+  template <bool newton3>
+  void processCellPairAoSPsVL(unsigned long cell1Index, unsigned long cell2Index, signed long directionIndex);
 
-  void processCellPairAoSNoN3Util(SortedCellView<ParticleCell> &cell1Sorted, SortedCellView<ParticleCell> &cell2Sorted);
   /**
-   * Applies the functor to all particle pairs between cell1 and cell2
-   * without exploiting newtons third law of motion.
+   * Utilizes the SortedCellViews to reduce the calculations needed by processCellPairAoS in certain cases.
+   * Used for PseudoVerletLists or if the number of Particles in the cells are above the sortingThreshold.
+   * @tparam newton3 determines if Newton3 is used or not. Matches _useNewton3.
```
SamNewcome:
```suggestion
   * @tparam newton3 determines if Newton3 is used or not. The version of this function actually run must match _useNewton3.
```

src/autopas/baseFunctors/CellFunctor.h
```diff
@@ -109,61 +116,94 @@ class CellFunctor {
...
-  void processCellSoANoN3(ParticleCell &cell);
+  /**
+   * SoA version of processCell if newton3 is not used.
+   * @param cell
+   */
+  void processCellSoANoN3(ParticleCell_T &cell);
 
-  void interactParticlesNoN3(auto &p1, auto &p2);
+  /**
+   * Calculates the interatctions of two particles in different cells.
```
SamNewcome:
```suggestion
   * Calculates the interactions of two particles in different cells.
```

SamNewcome (changes requested):
Looking really nice. Especially `CellFunctor` is so much clearer now. The one remaining thing is I would've expected only one function for `interactParticles` - I think this is possible but in case there is a good reason not to do this, the documentation and function names should make clear the distinction and why this distinction exists.

Aside from that, it is just a bunch of really rather minor documentation/similar changes.

The other thing still to do is move PsVL to the Verlet List Linked Based. Like what you did for the C08 Cell Handler, do the move in a single commit and use `git mv`.

PR #1063, opened 2026-02-07
Move AutoTuner(s) to separate TunerManager

thesamriel:
# Description

This PR refactors the LogicHandler-AutoTuner interaction by introducing a `TuningManager` between them.
This `TuningManager` holds a map of AutoTuners for different interaction types (pairwise, triwise) and can therefore orchestrate between them if necessary.

The AutoTuner itself is simplified to simply care about finding the next config without knowing about the outer time stepping of the simulation.
Additional iteration counters have been removed to stick with a single source of truth.
The responsibilities are therefore:

- `LogicHandler`:
  - Tracks the iteration number via `updateContainer()`
  - Asks the `TuningManager` for configurations during `computeInteractions()`

- `TuningManager`:
  - Holds a map with the `AutoTuner`s
  - Tracks tuning phases
  - Orchestrates between `AutoTuner`s by finding the best *global* configuration after a tuning phase

- `AutoTuner`:
  - Has a search space, tuning strategies and a config queue
  - Returns the next configuration for the respective interaction type based on the queue, strategies and number of samples

# How Has This Been Tested?

- All existing `AutoTuner` tests pass. (Some have been moved to `TuningManagerTest.cpp`
- Plus some new tests for the `TuningManager` and `AutoTuner`


src/autopas/tuning/searchSpace/Evidence.h
```diff
@@ -23,9 +23,17 @@ class Evidence {
...
-  long value;
+  long reducedValue;
+  /**
+   * Value of the measurement for only the rebuilding step.
+   */
+  long rebuildValue;
+  /**
+   * Value of the measurement for the actual traversal (without potential rebuild).
+   */
+  long traversalValue;
```
manishmishra6016: Why do we need these individual "evidence" (rebuild only or traverse only)?

src/autopas/tuning/searchSpace/EvidenceCollection.h
```diff
@@ -6,20 +6,34 @@
...
+  /**
+   * Helper enum to switch between evidence measurements modes.
+   */
+  enum EvidenceMode {
+    // Evidence of rebuild and traversal measurement combined taking into account the rebuild frequency.
+    REDUCED,
+    // Evidence of only the traversal measurement.
+    TRAVERSAL,
+    // Evidence of the rebuild + traversal measurement.
+    TOTAL
```
manishmishra6016: Is this just needed because you need to rebuild in every step? and would it be better to just identify this case and set rebuild_frequency as 1 and use reducedValue?

src/autopas/LogicHandler.h
```diff
@@ -160,31 +162,22 @@ class LogicHandler {
...
-  [[nodiscard]] std::vector<Particle_T> updateContainer() {
+  std::vector<Particle_T> updateContainer() {
+    ++_iteration;
+
 #ifdef AUTOPAS_ENABLE_DYNAMIC_CONTAINERS
     this->checkNeighborListsInvalidDoDynamicRebuild();
 #endif
     bool doDataStructureUpdate = not neighborListsAreValid();
 
-    if (_functorCalls > 0) {
```
manishmishra6016, line 169: I am glad you got rid of this. I wasn't sure what the purpose of this if-statement was. But I didn't want to change, as I was wondering whether it might affect logging; maybe you could verify that? And ignore this if you have already verified.

src/autopas/LogicHandler.h
```diff
@@ -1408,6 +1373,7 @@ std::tuple<std::unique_ptr<TraversalInterface>, bool> LogicHandler<Particle_T>::
   if (traversalPtr and generateNewContainer) {
     _currentContainerSelectorInfo = containerInfo;
     setCurrentContainer(std::move(containerPtr));
+    _neighborListsAreValid.store(false, std::memory_order_relaxed);
```
manishmishra6016: Why is this now needed? 
Wouldn't tuningManager already mark the current iteration as a "rebuild iteration" if it were going to select a new configuration?

src/autopas/tuning/AutoTuner.cpp
```diff
@@ -431,14 +406,17 @@ const TuningMetricOption &AutoTuner::getTuningMetric() const { return _tuningMet
...
-bool AutoTuner::inFirstTuningIteration() const { return (_iteration % _tuningInterval == 0); }
+// bool AutoTuner::inFirstTuningIteration() const { return (_iteration % _tuningInterval == 0); }
 
 bool AutoTuner::inLastTuningIteration() const { return _endOfTuningPhase; }
 
-bool AutoTuner::inFirstConfigurationLastSample() const { return (_iteration % _tuningInterval == _maxSamples - 1); }
+bool AutoTuner::inFirstConfigurationLastSample() const {
+  // todo: first config should be determined different in case a strategy filters out configs
+  return (_configQueue.size() == _searchSpace.size()) and (getCurrentNumSamples() == _maxSamples - 1);
+}
```
manishmishra6016: So, this works only when the first config is not rejected by the tuning strategy? 
Would it be okay to move it to the tuning manager class and keep it the same as before (which would always work even if ugly)? It also fits better in the tuning manager as it doesn't depend on the interaction type.

manishmishra6016 (commented):
Hi Markus,
The changes look quite good overall and would really help in reducing the clutter. 👏 
The only major comment/question is about different evidence modes; it wasn't very clear to me if they are absolutely needed. And there are some other comments you can have a look at. Let me know if I misunderstood something or if my comments don't make sense.

src/autopas/tuning/searchSpace/EvidenceCollection.cpp
```diff
@@ -22,30 +23,63 @@ const std::vector<Evidence> *EvidenceCollection::getEvidence(const Configuration
...
 
-std::tuple<Configuration, Evidence> EvidenceCollection::getOptimalConfiguration(size_t tuningPhase) const {
+std::tuple<Configuration, Evidence> EvidenceCollection::getBestConfigForContainer(
+    ContainerOption containerOption) const {
+  return getOptimalConfiguration(_latestTuningPhase, EvidenceMode::REDUCED, containerOption);
+}
+
+std::tuple<Configuration, Evidence> EvidenceCollection::getBestConfigNotReduced() const {
+  return getOptimalConfiguration(_latestTuningPhase, EvidenceMode::TOTAL, std::nullopt);
+}
```
SamNewcome: I'd change the function name somewhat. Maybe "NotReduced" -> "DuringRebuild" (or similar). 

Just makes in clearer imo that this is not some "default" function but has a specific purpose.

(I'd also be fine with directly calling `getOptimalConfiguration`).

src/autopas/tuning/searchSpace/EvidenceCollection.cpp
```diff
@@ -22,30 +23,63 @@ const std::vector<Evidence> *EvidenceCollection::getEvidence(const Configuration
...
+  // Helper lambda to fetch the metric as determined by the EvidenceMode
+  auto getValue = [mode](const Evidence &e) -> long {
+    switch (mode) {
+      case EvidenceMode::REDUCED:
+        return e.reducedValue;
+      case EvidenceMode::TRAVERSAL:
+        return e.traversalValue;
+      case EvidenceMode::TOTAL:
+        return e.rebuildValue + e.traversalValue;
+    }
```
SamNewcome, line 67: Add an exception default clause

src/autopas/tuning/AutoTuner.h
```diff
@@ -273,69 +250,72 @@ class AutoTuner {
...
+   * @param optimalConfig The configuration to be used from now on until the next tuning phase.
+   */
+  void forceOptimalConfiguration(const Configuration &optimalConfig);
+
+  /**
+   * Get the set of all container types present in the search space.
+   * Used by the Coordinator to calculate the intersection.
+   * @return set of all container types present in the search space.
+   */
+  std::set<ContainerOption> getSearchSpaceContainers() const;
```
SamNewcome: You want not just the container option, but also the cell size factor. (It maybe is a good idea to create a separate `ContainerConfiguration`, to collect parts of the configuration related to building the container - but can be done in the future)

src/autopas/tuning/AutoTuner.h
```diff
@@ -273,69 +250,72 @@ class AutoTuner {
...
    * If it is the end of the tuning phase, determine the optimal configuration and set this as the configuration to be
    * used until the next tuning phase, as well as setting other relevant class members (_endOfTuningPhase, _isTuning,
    * _samplesRebuildingNeighborLists, _iterationBaseline)
    */
-  void handleEndOfTuningPhaseIfRelevant();
+  void handleEndOfTuningPhase(size_t tuningPhase);
 
   /**
-   * Strategy how to reduce the sampled values to one value.
+   * Selects the best configuration for the current container from the evidence collection.
```
SamNewcome:
```suggestion
   * Selects the best configuration, given the current container, from the evidence collection.
```
(Imo a bit clearer)

src/autopas/tuning/AutoTuner.h
```diff
@@ -273,69 +250,72 @@ class AutoTuner {
...
-  std::vector<std::unique_ptr<TuningStrategyInterface>> _tuningStrategies;
+  size_t getCurrentNumSamples() const;
 
   /**
-   * Counter for the current simulation iteration.
-   * The first iteration has number 0.
+   * Estimate the runtime from the current samples according to the SelectorStrategy and rebuild frequency.
+   * Samples are weighted so that we normalize to the expected number of (non-)rebuild iterations and then divide by the
+   * rebuild frequency.
+   * @return estimate time for one iteration
```
SamNewcome: What are the three components of the tuple?

src/autopas/tuning/searchSpace/EvidenceCollection.h
```diff
@@ -6,20 +6,34 @@
...
  * Class to manage all evidence.
  */
 class EvidenceCollection {
  public:
+  /**
+   * Helper enum to switch between evidence measurements modes.
+   */
+  enum EvidenceMode {
+    // Evidence of rebuild and traversal measurement combined taking into account the rebuild frequency.
+    REDUCED,
```
SamNewcome: I'd be careful with this term. The `TRAVERSAL` time is still reduced, but with the selector strategy only. What you have here is then a second reduction.

The confusing this is that I think we use the term "reduced" the same thing you mean in the loggers. 

I'm happy with you keeping this as it is. But this is maybe something to be aware of in comments/documentation. E.g. here, you can specify exactly what you mean be reduction and maybe with `TRAVERSAL` that this is still reduced according to the chosen selector strategy from all the samples.

src/autopas/tuning/AutoTuner.h
```diff
@@ -273,69 +250,72 @@ class AutoTuner {
...
    */
-  size_t _iteration{0};
+  [[nodiscard]] std::tuple<long, long, long> estimateRuntimeFromSamples() const;
 
   /**
-   * The number of the current tuning phase.
-   * If we are currently between phases this is the number of the last phase.
-   * The first tuning phase has number 0.
-   * See bumpIterationCounters() for more details.
+   * Strategy how to reduce the sampled values to one value.
```
SamNewcome: Make this clearer in the context of reduced values from weighting with RF.

src/autopas/tuning/AutoTuner.cpp
```diff
@@ -149,26 +165,44 @@ bool AutoTuner::tuneConfiguration() {
...
+
+  _samplesRebuildingNeighborLists.resize(_maxSamples);
+  _samplesTraverseInteractions.resize(_maxSamples);
+}
+
+void AutoTuner::forceOptimalConfiguration(const Configuration &optimalConfig) {
+  _configQueue.clear();
+  _configQueue.push_back(optimalConfig);
+
+  // Fill up sample buffers to indicate we are not collecting samples anymore
```
SamNewcome: I don't get why this is needed. Maybe write a brief comment.

src/autopas/tuning/AutoTuner.cpp
```diff
@@ -179,71 +213,48 @@ const Configuration &AutoTuner::getCurrentConfig() const {
...
 
-  // let all strategies optimize the queue in the order they are defined.
-  // If any is still tuning consider the tuning phase still ongoing.
-  std::for_each(_tuningStrategies.begin(), _tuningStrategies.end(), [&](auto &tuningStrategy) {
-    tuningStrategy->optimizeSuggestions(_configQueue, _evidenceCollection);
-    AutoPasLog(DEBUG, "ConfigQueue after applying {}::optimizeSuggestions(): (Size={}) {}",
-               tuningStrategy->getOptionType().to_string(), _configQueue.size(),
-               utils::ArrayUtils::to_string(_configQueue, ", ", {"[", "]"},
-                                            [](const auto &conf) { return conf.toShortString(false); }));
-  });
```
SamNewcome, line 229: Why is this removed from `rejectConfig`?

src/autopas/tuning/AutoTuner.cpp
```diff
@@ -305,51 +317,21 @@ void AutoTuner::addMeasurement(long sampleRebuild, long sampleTraverseParticles,
...
-
-  // AutoTuner only triggers rebuild during the tuning phase
-  const auto iterationsPerConfig = this->inTuningPhase() ? _maxSamples : std::numeric_limits<unsigned int>::max();
-  // _iterationBaseLine + 1 since we want to look ahead to the next iteration
-  const auto iterationBaselineNextStep = _forceRetune ? _iterationBaseline : _iterationBaseline + 1;
-  return (iterationBaselineNextStep % iterationsPerConfig) == 0;
+  // This function should return true if the next call to tuneConfiguration() will lead to a new configuration being
+  // selected. A new configuration will be selected if:
+  //  The current configuration has finished collecting samples (either _maxSamples reached or early stopping).
+  //  We prepare for a forced retuning.
```
SamNewcome:
```suggestion
  // - The current configuration has finished collecting samples (either _maxSamples reached or early stopping).
  // - We prepare for a forced retuning.
```

src/autopas/tuning/AutoTuner.cpp
```diff
@@ -431,14 +406,17 @@ const TuningMetricOption &AutoTuner::getTuningMetric() const { return _tuningMet
 bool AutoTuner::inTuningPhase() const {
   // If _iteration % _tuningInterval == 0 we are in the first tuning iteration but tuneConfiguration has not
   // been called yet.
-  return (_iteration % _tuningInterval == 0 or _isTuning or _forceRetune) and not searchSpaceIsTrivial();
+  return (_isTuning or _forceRetune) and not searchSpaceIsTrivial();
 }
 
-bool AutoTuner::inFirstTuningIteration() const { return (_iteration % _tuningInterval == 0); }
+// bool AutoTuner::inFirstTuningIteration() const { return (_iteration % _tuningInterval == 0); }
```
SamNewcome: Deadcode

src/autopas/tuning/AutoTuner.cpp
```diff
@@ -431,14 +406,17 @@ const TuningMetricOption &AutoTuner::getTuningMetric() const { return _tuningMet
...
 
 bool AutoTuner::inLastTuningIteration() const { return _endOfTuningPhase; }
 
-bool AutoTuner::inFirstConfigurationLastSample() const { return (_iteration % _tuningInterval == _maxSamples - 1); }
+bool AutoTuner::inFirstConfigurationLastSample() const {
+  // todo: first config should be determined different in case a strategy filters out configs
```
SamNewcome: Imo, this needs to be dealt with this PR. Otherwise, the function will not work with pretty much any strategy, if I understand correctly.

src/autopas/tuning/AutoTuner.cpp
```diff
@@ -447,27 +425,32 @@ bool AutoTuner::canMeasureEnergy() const { return _energyMeasurementPossible; }
 void AutoTuner::setRebuildFrequency(double rebuildFrequency) { _rebuildFrequency = rebuildFrequency; }
 
 void AutoTuner::checkEarlyStoppingCondition() {
-  if (_iterationBaseline < _maxSamples) {
+  if (_evidenceCollection.empty()) {
```
SamNewcome: I'm not sure this is correct. If I understand this correctly, this will only skip if there is no prior evidence at all, whereas we should skip only if there is no prior evidence yet this tuning phase.

src/autopas/LogicHandler.h
thesamriel: The main reason was that the `_iteration` counter should be 0 for the first time step. But because `updateContainer()` is  called before the first force calculations, it would have been incremented to 1 already.
`_iteration` is now initialized as `size_t _iteration{std::numeric_limits<size_t>::max()};` and the first icrement in `updateContainer()` will set it to 0.

src/autopas/LogicHandler.h
thesamriel: Yes, you're totally right. I removed it.

src/autopas/tuning/searchSpace/EvidenceCollection.h
thesamriel: Yeah, that's for the case where pairwise and triwise use different containers and we would need to rebuild between every single functor call.
In theory that's equivalent to a rebuild_frequency of 1. However, currently we store the 'reduced' values only during the tuning phase (rebuild / rebuild_freq + traversal). And from just this value I cannot deduce the rebuild + traversal time at the end from the tuning phase.
If we wanted to reduce the evidence footprint, we could only store rebuild and traversal times and compute the 'reduced' values on the fly when needed. I left it in though, to not have to change everything else that used the 'reduced' value.

src/autopas/tuning/searchSpace/Evidence.h
thesamriel: see other comment

src/autopas/tuning/AutoTuner.cpp
thesamriel: Yes, your're right! Moved it to the TuningManager in [c16a705](https://github.com/AutoPas/AutoPas/pull/1063/commits/c16a70570b8168cfaa194f3eeeeb52aa56effc1e)
Thanks for pointing this out, this todo was meant to be resolved before the PR, but I forgot.

src/autopas/tuning/TuningManager.h
```diff
@@ -0,0 +1,219 @@
...
+
+  /**
+   * Add an AutoTuner to the TuningManager, which will take over ownership.
+   * @param interactionType
+   * @param tuner A unique_ptr to the new tuner.
+   */
+  void addAutoTuner(std::unique_ptr<AutoTuner> tuner, InteractionTypeOption::Value interactionType);
+
+  /**
+   *
```
SamNewcome: Add a brief description.

src/autopas/tuning/TuningManager.h
```diff
@@ -0,0 +1,219 @@
...
+   * To be called at the end of a tuning phase.
+   * Compares evidence of all active AutoTuners and sets the best configurations leading to an overall optimal time step
+   * performance.
+   */
+  void setOptimalConfigurations();
+
+  /**
+   * Find all container options that are part of all currently managed AutoTuner instances.
+   */
+  std::set<ContainerOption> setCommonContainerOption();
```
SamNewcome:
```suggestion
  std::set<ContainerOption> getCommonContainerOptions();
```
(I guess you meant set as `std::set`, but it looks like a setter otherwise.)

src/autopas/tuning/TuningManager.cpp
```diff
@@ -0,0 +1,223 @@
...
+  return isStart or std::ranges::any_of(_autoTuners, [&](const auto &tunerEntry) {
+           return tunerEntry.second->willRebuildNeighborLists();
+         });
+}
+
+bool TuningManager::needsLiveInfo(const size_t currentIteration) {
+  const bool isStart = isStartOfTuningPhase(currentIteration);
+  const bool aboutToBegin = tuningPhaseAboutToBegin(currentIteration);
+  return (isStart or aboutToBegin) and
+         std::ranges::any_of(_autoTuners, [&](const auto &tuner) { return tuner.second->needsLiveInfo(); });
```
SamNewcome: In the original handling or this, liveInfo is gathered & received by tuning strategies if
(isStart and tuning strategy needs LiveInfo) or (aboutToBegin and tuning strategy needs domain similarity statistics)

What you do here will then gather live info (which can get expensive) during the `aboutToBegin` period, even if the tuning strategy only needs live info (i.e. at the start of the tuning phase). Change this to avoid this.

src/autopas/tuning/TuningManager.h
```diff
@@ -0,0 +1,219 @@
...
+   */
+  bool needsLiveInfo(size_t currentIteration);
+
+  /**
+   * @return True, if the search space of all AutoTuners consists of only one valid configuration.
+   */
+  bool allSearchSpacesAreTrivial() const;
+
+  /**
+   * @return True, if the tuning phase just finished for all AutoTuners.
```
SamNewcome, line 117: Maybe briefly write about the `and (not _transitionToOptimalConfigurations)`

src/autopas/tuning/AutoTuner.h
```diff
@@ -273,69 +243,72 @@ class AutoTuner {
...
-  bool tuneConfiguration();
+  bool tuneConfiguration(size_t currentIteration, size_t tuningPhase, bool isStartOfTuningPhase);
+
+  /**
+   * Pushes the provided config into the _configQueue and ends tuning immediately.
+   * Used by the TuningManager, when we want a different configuration than the one determined by one AutoTuner in
+   * isolation.
+   * @param optimalConfig The configuration to be used from now on until the next tuning phase.
+   */
+  void forceOptimalConfiguration(const Configuration &optimalConfig);
```
SamNewcome: Maybe use something less aggresive than "force". As far as I can see, this is the standard way to set the optimal configuration and is not something forced over the standard way. Maybe then `setOptimalConfiguration`?

src/autopas/LogicHandler.h
```diff
@@ -873,22 +877,18 @@ class LogicHandler {
...
 
   /**
    * The current iteration number.
+   * Initialized as max such that ++_iteration == 0 for the first iteration.
    */
-  unsigned int _iteration{0};
+  size_t _iteration{std::numeric_limits<size_t>::max()};
 
   /**
    * The iteration number at the end of last tuning phase.
```
SamNewcome: Update this

tests/testAutopas/tests/tuning/TuningManagerTest.cpp
```diff
@@ -0,0 +1,976 @@
...
+  tunerManager->addAutoTuner(std::move(autoTuner), autopas::InteractionTypeOption::pairwise);
+
+  // Iteration 0: Start of Phase -> Should need info
+  EXPECT_TRUE(tunerManager->needsLiveInfo(0)) << "Manager should request LiveInfo at the start of a phase.";
+
+  // Iteration 50: Middle of Phase -> Should NOT need info
+  EXPECT_FALSE(tunerManager->needsLiveInfo(50)) << "Manager should block LiveInfo mid-phase.";
+
+  // Iteration 95: Phase about to begin (within 10 steps of interval 100) -> Should need info
+  EXPECT_TRUE(tunerManager->needsLiveInfo(95)) << "Manager should request LiveInfo when a phase is about to begin.";
```
SamNewcome: This is incorrect (see [here](https://github.com/AutoPas/AutoPas/pull/1063/changes#r3168497539)). My suggestion is to modify this test such that:

- You have two dummy tuning strategies. One should have `needsLiveInfo` return `true`, and the other should have `needsDomainSimilarityStatistics` return `false`.
- Loop over all permutations of AutoTuner including/excluding each strategy (i.e. 4 permutations)
- Do this test, with the expected TRUE/FALSE modified as nessecary.

SamNewcome (changes requested):
Mostly minor comments, different variable/function names, or documentation suggestions.

The two more significant issues are:

- handling of early stopping (https://github.com/AutoPas/AutoPas/pull/1063/changes#r3154965563)
- Unnecessary gathering of live info when we want live info but not domain similarity statistics (https://github.com/AutoPas/AutoPas/pull/1063/changes#r3168497539, https://github.com/AutoPas/AutoPas/pull/1063/changes#r3169133249). Tbf, this was somewhat confusing to begin with so I understand the confusion. We could maybe make this clearer.

src/autopas/tuning/AutoTuner.cpp
SamNewcome: Outdated when I finished review -> ignore

src/autopas/tuning/AutoTuner.cpp
SamNewcome: Outdated when I finished review -> ignore

src/autopas/tuning/searchSpace/EvidenceCollection.h
manishmishra6016: Okay, yes, that's true, you need both at the end of the tuning phase. Would be interesting to see if these combinations that require rebuilding every step are optimal choices in any scenario!  

Another thing: the values logged in the tuningData and tuningResults loggers would still be the reduced values and might not reflect the values used for the actual decision. This should be noted somewhere (and can hopefully be fixed in the future).

src/autopas/tuning/searchSpace/EvidenceCollection.cpp
thesamriel: Changed the name to `getBestConfigWithRebuild` in 25bc407dc4183dad4996eb0b99d012e16e399df8

src/autopas/tuning/searchSpace/EvidenceCollection.cpp
thesamriel: added here: 25bc407dc4183dad4996eb0b99d012e16e399df8

src/autopas/tuning/AutoTuner.h
thesamriel: Yes that's true, I kind of just looked away at this.
I'm handling it by passing both containerOption and cellSizeFactor now. Could be turned into a struct later.
Done in a2f064c60854f6bfe5e7a3c53718153fbf270361

src/autopas/tuning/searchSpace/EvidenceCollection.h
thesamriel: You're right. I tried avoiding the term 'reduced' now.
This enum value I changed to `EFFECTIVE`. 
AI also suggested `AMORTIZED`, but this sounded a bit too complicated to me. Let me know if you prefer something else.

src/autopas/tuning/AutoTuner.cpp
thesamriel: Actually, it was not needed. Deleted now.

src/autopas/tuning/AutoTuner.cpp
thesamriel: No reason, good catch! Brought it back ;)

src/autopas/tuning/AutoTuner.cpp
thesamriel: Yeah, you're right. I modified the `empty()` method to optionally take in a tuning phase number.
See cdeb78b0d78681f125b741b535bdfd3a11e7f18d

src/autopas/tuning/TuningManager.cpp
thesamriel: Changed this here: cdeb78b0d78681f125b741b535bdfd3a11e7f18d

I didn't realize the difference tbh. Good catch.

src/autopas/LogicHandler.h
thesamriel: Is this at the correct line? Not sure what you're referring to.

tests/testAutopas/tests/tuning/TuningManagerTest.cpp
thesamriel: I see now. Adapted the test here:
[742db84](https://github.com/AutoPas/AutoPas/pull/1063/commits/742db84694c6c83d31bdc5905752205e71fb71ce)

src/autopas/tuning/TuningManager.h
thesamriel: Actually, the `tuningPhaseJustFinished()` logic was a bit unnecessarily complicated.
Changed `_transitionToOptimalConfigurations` to `_tuningFinishedLastIteration`, which should be clearer imo.
See [742db84](https://github.com/AutoPas/AutoPas/pull/1063/commits/742db84694c6c83d31bdc5905752205e71fb71ce)

src/autopas/LogicHandler.h
SamNewcome: Yes - but sorry I meant the documentation specifically. As in "The iteration number after the last tuning phase has ended" or similar

src/autopas/tuning/searchSpace/EvidenceCollection.h
SamNewcome: No I like `EFFECTIVE`. (And I agree `AMORTIZED` sounds too complicated)

src/autopas/LogicHandler.h
thesamriel: Ah okay, I changed it now

src/autopas/tuning/searchSpace/EvidenceCollection.h
thesamriel: Good point, I opened an issue (#1141) and linked it in the logger methods' doxygen.
I also just noticed that the `TuningResultsLogger` is wrong for another reason as well when using multiple interaction types. But I would fix this in another PR if that's ok.

docs/userdoc/AutoTuning.md
```diff
@@ -50,11 +51,12 @@ See [Building](https://github.com/AutoPas/AutoPas/blob/master/docs/userdoc/Build
 
 ## Multiple Interaction Types
 Since PR [751](https://github.com/AutoPas/AutoPas/pull/751), AutoPas also supports algorithms for 3-body interactions.
-Oftentimes, these are used additionally to standard pairwise interactions. For such simulations, AutoPas runs two separate AutoTuners, one for pairwise and one for triwise interaction.
-Both AutoTuners will start their tuning phases at the same iterations by setting the start of the tuning interval when both AutoTuners finish tuning over their search space.  
-_Note: Performance may vary if the AutoTuners end up with configurations that use a different [container option](https://github.com/AutoPas/AutoPas/blob/tuningDoc/src/autopas/options/ContainerOption.h). 
-In this case, AutoPas has to rebuild neighbor lists every time the interaction type changes._
+Oftentimes, these are used additionally to standard pairwise interactions.
+For such simulations, the TuningManager orchestrates between two separate AutoTuners, one for pairwise and one for triwise interaction.
```
SamNewcome:
```suggestion
For such simulations, the TuningManager orchestrates over two separate AutoTuners, one for pairwise and one for triwise interaction.
```

src/autopas/tuning/TuningManager.cpp
```diff
@@ -133,78 +145,108 @@ void TuningManager::setOptimalConfigurations() {
...
+      std::set<ContainerOption> resultContainers;
+      std::ranges::set_intersection(commonContainers, tunerContainers,
+                                    std::inserter(resultContainers, resultContainers.begin()));
+      commonContainers = std::move(resultContainers);
+
+      // Intersect Cell Size Factors
+      std::set<double> resultFactors;
+      std::ranges::set_intersection(commonCellSizeFactors, tunerCellSizeFactors,
+                                    std::inserter(resultFactors, resultFactors.begin()));
+      commonCellSizeFactors = std::move(resultFactors);
```
SamNewcome: I think you can't consider these two independent. E.g. one tuner might have (VCL, CSF1) & (LC, CSF 0.5), the other might have (VLC, CSF 1) & (LC, CSF 1.5). You then get LC as the common container and 1.0 as the common CSF, but neither tuner has (LC, CSF1).

src/autopas/LogicHandler.h
```diff
@@ -167,6 +167,33 @@ class LogicHandler {
...
+      double userProvidedRF = static_cast<double>(_neighborListRebuildFrequency);
+      // The user defined rebuild frequency is considered as the upper bound.
+      // If velocity method estimate exceeds upper bound, set the rebuild frequency to the user defined value.
+      // This is done because we currently use the user defined rebuild frequency as the upper bound to avoid expensive
+      // buffer interactions.
+      if (rebuildFrequencyEstimate > userProvidedRF) {
+        _tuningManager->setRebuildFrequency(userProvidedRF);
+      } else {
+        _tuningManager->setRebuildFrequency(rebuildFrequencyEstimate);
+      }
```
SamNewcome:
```suggestion
      _tuningManager->setRebuildFrequency(std::min(userProvidedRF, rebuildFrequencyEstimate));
```

SamNewcome (changes requested):
Looks nice. There is maybe an issue with getting the common containers + CSFs - check that my reasoning makes sense.

src/autopas/tuning/TuningManager.cpp
thesamriel: You're right, but it would not matter since we will just notice later on that there are no samples for a certain container/csf combination. If a tuning strategy dismisses certain configs for one tuner, this scenario can still happen.
However, since it feels simpler anyway,  I changed the function in [63d1b10](https://github.com/AutoPas/AutoPas/pull/1063/commits/63d1b10983f655d117a5f9c22ce37fe2274a8bb7)

SamNewcome (approved):

PR #1068, opened 2026-02-11
Move logger instance counter to the logger class

thesamriel:
# Description

This removes the awkward `InstanceCounter` struct, which was only needed for taking care of the AutoPas logger.
The logging macro calls a method that performs lazy initialization: if no logger exists, it will first create one.
This avoids the problem where one AutoPas instance deletes the logger while another instance is still trying to log. 

We don't need to explicitly unregister loggers; spdlog handles this internally. (Loggers are stored in `std::shared_ptr<>`)

To use the logger, simply use the logging macro. No setup is needed.
`Logger::create(..)` can be called explicitly to allow logging to another output stream or file. Default is `std::cout`.

### How was this tested

The existing tests are passing and the logging output during the simulation seems as usual.


tests/testAutopas/tests/utils/LoggerTest.cpp
SamNewcome: Can you add simple unit tests for the new functions.

- `get()`: One idea is to test that the logger is fine if called with and without `create()` being called.
- With the others, a simple test that just checks there is no errors if e.g. creating when a logger already exists/doesn't exist and then unregistering is fine if a logger exists/doesn't exist.
- If you have other ideas, feel free to include. I just want a unit test for the new functions.

SamNewcome (changes requested):
I really like this! Just a couple points mostly about unit tests

src/autopas/utils/logging/Logger.h
```diff
@@ -21,44 +21,90 @@
...
    */
-  static void create(std::ostream &oss = std::cout) {
-    // drop an already registered Logger if it exists
-    if (spdlog::get(loggerName())) {
-      unregister();
-    }
+  static void create(std::ostream &logOutputStream = std::cout) {
+    std::lock_guard<std::mutex> lock(_registrationMutex);
+    create_internal(logOutputStream);
+  }
```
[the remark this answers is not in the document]
thesamriel: Yes, becaues it was implicitly dropped in `create_internal()`.
However, I refactored it slightly to explicitly drop and then newly create in [4d5dcfd](https://github.com/AutoPas/AutoPas/pull/1068/commits/4d5dcfd80d4aa430b69cb2f52746bd14225591cd)

tests/testAutopas/tests/utils/LoggerTest.cpp
thesamriel: Added more tests in [c3ef991](https://github.com/AutoPas/AutoPas/pull/1068/commits/c3ef991992a4e4525864910356fd94eb0aadf33b)

SamNewcome (approved):

Issue #658, opened 2021-10-28
`cellSizeFactor`s Different from `1.0` in Octree


johannes-spies:
**Is your feature request related to a problem? Please describe.**

When being constructed, containers receive a `cellSizeFactor` that changes the shape of the cells within the container. At the moment, the octree is not able to handle `cellSizeFactor`s different from `1.0`. This factor would affect the leaf-splitting behavior in the octree.

**Describe the solution you'd like**

Traversals already come with an `isApplicable()` method that evaluates whether the current configuration is safe for the traversal to operate. Since the `cellSizeFactor` is not a property of an (octree) traversal, but of the container itself, it would be wrong to check the `cellSizeFactor` inside of the traversal. It would be best to also add an `isApplicable()` method to the containers. This method could then be checked before trying to change to the container with a given configuration. If this mechanism was in place, the octree could simply refuse configurations where the `cellSizeFactor` is different from `1.0`.

**Describe alternatives you've considered**

Currently, the `cellSizeFactor` is ignored inside the `OctreeLeafNode::insert()` routine. This works if the factor is `1.0`. Erroneous behavior with different factors could not be observed.

PR #1071, opened 2026-02-12
Clearer distinction between functions which check the applicability of a configuration

### LINKED ISSUE #1031: The Logic of `AutoPasInterfaceTest::ConfighasCompatibleValuesVSTraversalIsApplicable` is flawed and `conf.hasCompatibleValues()` does not match `traversal.isApplicable()`.
**Describe the bug**
This test does not actually test different cell size factors (in fact, the cell dimensions used in the test are incorrect even for CSF1, but this doesn't really matter).

In commit 52b8a442, you see a corrected version of this test. This (correctly) fails, because we have some configurations which fail `traversal.isApplicable` but not `conf.hasCompatibleValues()`. I've seen this often with things like traversals not allowing below 1.

There are also a few other cases where `isApplicable` might fail, but the test does not cover these cases. For some of these, I also expect a mismatch with `conf.hasCompatibleValues()`.

We now have to deal with the fact that `conf.hasCompatibleValues()` does not correctly match `traversal.isApplicable()`.

***Things which `traversal.isApplicable` checks but `conf.hasCompatibleValues()` does not***

- Domain is long enough for sliced traversal (All sliced based traversals)
- Domain large enough (LCC04)
- Traversals required cellsizes greater than cutoff (C04, C04_HCP, VLCC08)
- ~Traversals requiring pairwise interactions only (LCC01 w/ SoA + `combineSoA`)~

**Possible Solution**

We want to reduce code distribution because this has IMO this is the fundamental issue. In general, provide a clear separation between issues where the configuration is "internally" incompatible and issues where the configuration is not compatible with the domain. The former is the responsibility of `Configuration::hasCompatibleValues` and the latter is the responsibility of `isApplicable`.

- Modify `Configuration::hasCompatibleValues` to check for cell-size restrictions.
- Modify `Configuration::hasCompatibleValues` to check for interaction type restrictions.
- Rename `isApplicable` to `isApplicableToDomain` to more clearly define its purpose.

SamNewcome:
# Description

This PR cleans up the handling of the applicability of a configuration (see #1031 where I discussed some of the confusing issue).

It does this by providing the three functions that check this, clearer responsibilities:

- `Configuration::hasCompatibleValues` checks the applicability of a configuration independent of functor or traversal.
- `TraversalInterface::isApplicable` checks the applicability of (particularly the traversal) to the domain. It is renamed to `isApplicableToDomain`.
- `LogicHandler::isConfigurationApplicable` calls both functions above and provides an additional check that the configuration is applicable to the functor.

In practice, this means

- [x] in cases where `TraversalInterface::isApplicable` checked the applicability of a configuration independent of the domain, these checks were removed. (This was only ever done inconsistently, and is an unnecessary duplicate check).
- [x] `LogicHandler::isConfigurationApplicable` now calls `Configuration::hasCompatibleValues` instead of its helper function `compatibleTraversals::allCompatibleTraversals` which only checks some things.
- [x] Comments and contributing guidelines make this distinction clearer
- [x] There existed a test that checked that `Configuration::hasCompatibleValues` matches `TraversalInterface::isApplicable`. But these functions have not actually matched since before I started working on AutoPas, and the test only passed because of a mistake. With the above clearer responsibilities, this test is no longer desired and thus removed.

## Related issues regarding CSF

This PR stems from frustrations around CSF<1. During making this PR, I discovered several containers do not work properly with CSF<1.0 but didn't have any handling for this. 

- OT doesn't work with CSF != 1.0 (See #658). This is now enforced.
- VCL doesn't take CSF as an argument (but some traversals do, leading to problems). I enforce CSF == 1.0 now.
- VLC does take CSF < 1.0, but just sets it to CSF 1.0. This isn't broken, but a bit pointless and crowds our search space unnecessarily. I enforce now CSF >= 1.0.
- DS doesn't take a CSF and its somewhat meaningless here. I enforce CSF == 1.0.

-> I assume CSF 1.0 is always allowed in a configuration search space and where it is not allowed, I assume the user is purposely restricting the search space and won't try to force e.g. VCL with CSF 0.5. This kinda makes sense in my mind, and the alternatives would be messy. 

For the containers which do allow CSF < 1.0, some traversals do not. I decided to keep these checks within `TraversalInterface::isApplicableToDomain`. This is because the direct issue is a domain issue (cells smaller than the interation length) and only indirectly from CSF < 1.0. This is a minor preference and you could say the same thing regarding e.g. not allowing VLC with CSF < 1.0. This could be changed, such that we also exclude traversals with CSF < 1 in `Configuration::hasCompatibleValues`.

## Tests

A lot of the issues+bugs regarding CSF came from it being neglected from tests. We have many tests parameterised over the configuration, or the container part, and it is easy for developers to miss things like CSF. In addition, load estimator was also often neglected (although this never led to any bugs).

To avoid this, the generation of sets of (all) configurations or containers are refactored to use only a few helper generator functions, and tests were refactored to only use these. 

## Misc

- Some minor cleaning up of `Configuration::hasCompatibleValues` to improve readability.

## Resolved Issues

- [x] fixes #1031

# How Has This Been Tested?

- [x] CI
- [x] Checked that the same number of configurations are trialled in a full search run (with CSF=1) as in the master


SamNewcome:
@reviewers (@manishmishra6016 @TumBeat ) the failing MacOS GCC-14 CI run appears to be an unrelated issue with some GitHub runners being updated which breaks HWY. But irrelevant for this PR so can be ignored.

src/autopas/containers/verletListsCellBased/VerletListsLinkedBase.h
```diff
@@ -43,14 +43,15 @@ class VerletListsLinkedBase : public ParticleContainerInterface<Particle_T> {
...
    */
   VerletListsLinkedBase(const std::array<double, 3> &boxMin, const std::array<double, 3> &boxMax, const double cutoff,
                         const double skin, const double cellSizeFactor)
       : ParticleContainerInterface<Particle_T>(skin),
         _linkedCells(boxMin, boxMax, cutoff, skin, std::max(1.0, cellSizeFactor)) {
     if (cellSizeFactor < 1.0) {
-      AutoPasLog(DEBUG, "VerletListsLinkedBase: CellSizeFactor smaller 1 detected. Set to 1.");
+      utils::ExceptionHandler::exception(
+          "Trying to construct a VerletListsLinkedBase with CSF < 1.0! Such configurations should not be allowed (they "
+          "should return false with Configuration::hasCompatibleValues).");
```
TumBeat: This feels redundant as such a configuration should be rejected anyways. Or is this more a "This should never happen and we better throw an exception"? If so, then we might want to add these exceptions also to the other containers, e.g. Octree has no exception for cellSizeFactors != 1 even though it takes it as a parameter

src/autopas/tuning/Configuration.cpp
```diff
@@ -51,44 +52,54 @@ std::string autopas::Configuration::getCSVRepresentation(bool returnHeaderOnly)
...
+    if (aosOnlyTraversals.contains(traversal)) {
       return false;
     }
   }
 
+  const auto allContainersSupportingOnlyCSF1 = compatibleCSFs::allContainersOnlySupportingCSF1();
+  if (allContainersSupportingOnlyCSF1.contains(container) and cellSizeFactor != 1.0) {
+    return false;
+  }
+
```
TumBeat, line 97: Just to double check, it is safe to assume that every container that is not filtered out as part of the allContainersOnlySupportingCSF1() check supports CSF > 1, right? Because the allContainersSupportingSub1CSF() could be true for them. So every container that allows CSF < 1 also allows CSF > 1.

src/autopas/LogicHandler.h
```diff
@@ -1323,13 +1334,16 @@ template <typename Particle_T>
...
+  // Check if the configuration is compatible, independent of domain or functor
+  if (not config.hasCompatibleValues()) {
+    AutoPasLog(
+        WARN,
+        "A configuration was rejected by LogicHandler::isConfigurationApplicable, as it was incompatible"
+        " independently of domain or functor. This should not occur and implies illegal configurations are being added"
+        " to the configuration queue during simulation (potentially by a tuning strategy).");
+    AutoPasLog(WARN, "Configuration rejected: {}", config.toString());
+    // Such illegal configurations should be filtered out in the generation of the search space.
+    return {nullptr, /*rejectIndefinitely*/ false};
```
TumBeat: If this is the result of adding configurations illegally, does it not also make sense to signal that this configuration should be deleted from the search space again (by setting rejectIndefinitely to true)? Otherwise, this configuration remains in the search space and triggers this warning again in the next tuning phase which makes identifying if this config is still in the search space or if it was rejected, deleted but added again before the next tuning phase harder.

tests/testAutopas/testingHelpers/GenerateValidConfigurations.h
```diff
@@ -0,0 +1,164 @@
...
+        autopas::LoadEstimatorOption::getAllOptions(),
+    const std::set<autopas::DataLayoutOption> &allowedDataLayoutOptions = autopas::DataLayoutOption::getAllOptions(),
+    const std::set<autopas::Newton3Option> &allowedNewton3Options = autopas::Newton3Option::getAllOptions(),
+    const std::set<double> &allowedCellSizeFactors = {0.5, 1.0, 1.5},
+    const std::set<autopas::VectorizationPatternOption> &allowedVectorPatterns =
+        autopas::VectorizationPatternOption::getAllOptions()) {
+  if (interactionType == autopas::InteractionTypeOption::all) {
+    std::set<autopas::Configuration> allConfigs;
+    for (auto iType : autopas::InteractionTypeOption::getMostOptions()) {
+      const autopas::NumberSetFinite<double> csfs(allowedCellSizeFactors);
```
TumBeat: As this line does not differ depending on the interaction type, it might make sense to have it outside of the if-statement to reduce code duplication.

tests/testAutopas/testingHelpers/GenerateValidConfigurations.h
```diff
@@ -0,0 +1,164 @@
...
+      const auto configs = autopas::SearchSpaceGenerators::cartesianProduct(
+          allowedContainerOptions, allowedTraversalOptions, allowedLoadEstimatorOptions, allowedDataLayoutOptions,
+          allowedNewton3Options, &csfs, allowedVectorPatterns, iType);
+      allConfigs.insert(configs.begin(), configs.end());
+    }
+    return allConfigs;
+  } else {
+    const autopas::NumberSetFinite<double> csfs(allowedCellSizeFactors);
```
TumBeat: see comment above

src/autopas/containers/verletListsCellBased/VerletListsLinkedBase.h
SamNewcome: Personally, I prefer the redundancy as a double check. But I don't feel to strongly about this. In 40b1167f  I rephrased this to make it clearer that is should not happen.

With Octree this is a good catch and I totally agree with you but IMO their are two approaches. Direct Sum and Verlet Cluster Lists simply don't take a cell size factor so nothing is needed. And with these, CSF fundamentally doesn't really make sense, I guess? With Octree, I'm unsure if the best approach is to say CSF doesn't make sense (the container is designed to construct appropriately sized cells anyway) - and then get rid of the constructor variable - or to say it does in which case I put a similar exception.

I asked Manish for his opinion on this because he's a bit closer to this.

src/autopas/tuning/Configuration.cpp
SamNewcome: Yes. We could instead have `allContainersOnlySupportingAbove1CSF` and `allContainersOnlySupportingBelow1CSF` - this way everything becomes opt-in whereas what we currently have is CSF<1 is opt-in and CSF!=1 is opt-out. I see this is maybe a bit confusing as it is.

src/autopas/LogicHandler.h
SamNewcome: Ah yes I agree. Done in 1f80d72b

tests/testAutopas/testingHelpers/GenerateValidConfigurations.h
SamNewcome: Done in c940d3e4

tests/testAutopas/testingHelpers/GenerateValidConfigurations.h
SamNewcome: Done in c940d3e4

src/autopas/containers/linkedCells/traversals/LCC04CombinedSoATraversal.h
```diff
@@ -50,14 +50,13 @@ class LCC04CombinedSoATraversal : public C04BasedTraversal<ParticleCell, Pairwis
 
   /**
    * lc_c04_combined_SoA traversals are only usable with dataLayout SoA.
```
manishmishra6016: should this be removed as we no longer check for data layout here?

src/autopas/tuning/Configuration.cpp
manishmishra6016: If it's not too expensive, what you suggest here might be better!

tests/testAutopas/testingHelpers/GenerateValidConfigurations.h
```diff
@@ -0,0 +1,164 @@
+/**
+ * @file GenerateValidConfigurations.h
+ * @author The AI ghost of F. Gratl
```
manishmishra6016, line 3: was this intentional? :D

manishmishra6016 (commented):
Looks good overall. It might be nice to check a simulation case to see how it compares with the master and make sure we are not missing any configuration. Just a diff on the tuning logger (with runtime values removed), maybe? Ignore if you have already verified this.

src/autopas/tuning/Configuration.cpp
SamNewcome: Done in 89f7a66d

src/autopas/containers/verletListsCellBased/VerletListsLinkedBase.h
SamNewcome: Manish's opinion was that it could be useful to have CSF with Octree still. So I included a similar exception clause in dd9a072f.

tests/testAutopas/testingHelpers/GenerateValidConfigurations.h
SamNewcome: I think the first draft of this file was done by the JetBrains AI which just straight up gave the author as Fabio. Thought it was funny

src/autopas/containers/linkedCells/traversals/LCC04CombinedSoATraversal.h
SamNewcome: Done in c2b16bd1

SamNewcome:
@gemini review

tests/testAutopas/testingHelpers/GenerateValidConfigurations.h
```diff
@@ -0,0 +1,164 @@
...
+      // We also use the first valid traversal for this container.
+      const auto interactionType = autopas::InteractionTypeOption::pairwise;
+      const auto traversals = autopas::compatibleTraversals::allCompatibleTraversals(containerOption, interactionType);
+      if (traversals.empty()) {
+        continue;
+      }
+      const auto &traversalOption = *traversals.begin();
+      const auto loadEstimators = autopas::loadEstimators::getApplicableLoadEstimators(
+          containerOption, traversalOption, autopas::LoadEstimatorOption::getAllOptions());
+      const auto &loadEstimatorOption = *loadEstimators.begin();
```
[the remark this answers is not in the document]
SamNewcome: Done in 7b3250f0

.github/CONTRIBUTING.md
```diff
@@ -129,6 +130,8 @@ Possible log levels are:`trace`, `debug`, `info`, `warn`, `err`, `critical`, `of
   * Add a new enum in `ContainerOption::Value`.
   * Add a new string representation in the `map` of `ContainerOption::getOptionNames()`.
 * Create a new set of compatible traversals in [`CompatibleTraversals`](/src/autopas/containers/CompatibleTraversals.h).
+* If the new container supports cell-size-factors less than 1.0, add it to [`CompatibleCellSizeFactors::allContainersSupportingSub1CSF`](src/autopas/containers/CompatibleCellSizeFactors.h).
+* If the new container supports only cell-size-factor 1.0, or is independent of cell-size-factor, add it to [`CompatibleCellSizeFactors::allContainersOnlySupportingCSF1`](src/autopas/containers/CompatibleCellSizeFactors.h).
```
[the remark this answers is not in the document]
SamNewcome: Done in 7b3250f0

src/autopas/containers/CompatibleCellSizeFactors.h
```diff
@@ -0,0 +1,34 @@
...
+}
+
+/**
+ * Lists all containers that support CSF < 1.0. Note, it is possible that a container supports this but one or more
+ * traversals of that container do not. These should be filtered out in TraversalInterface::isApplicableToDomain().
+ * @return A list of all containers that support CSF < 1.0.
+ */
+[[maybe_unused]] static const std::set<ContainerOption> allContainersSupportingSub1CSF() {
+  return {ContainerOption::linkedCells, ContainerOption::linkedCellsReferences};
+}
```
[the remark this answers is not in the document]
SamNewcome: Done in 7b3250f0

TumBeat (approved):
Looks good

manishmishra6016 (approved):
Looks good to me too!

TumBeat (approved):

manishmishra6016 (approved):

PR #1073, opened 2026-02-12
Add TraceTimer and steady clock

thesamriel:
# Description

Adds a new Timer class `TraceTimer`, which is a wrapper around `Timer`, but only runs at log level `TRACE` or below.
This removes the cluttering `#if #endif` statements from where these timers are used.

Additionally, I added a condition for the Timer to use `steady_clock` if `high_resolution_clock` is in fact not steady.
I occasionally observed negative time measurements, and I assume this is the reason.
(see cppreference [Notes](https://en.cppreference.com/w/cpp/chrono/high_resolution_clock.html))

# How Has This Been Tested?

- [x] added `TraceTimerTest`


SamNewcome (approved):
Looks good. We maybe want a CI run with TRACE active to properly test this. But I feel this is kinda overkill for just this one run. I'd suggest we ignore this for now.

Before you merge, could you confirm that locally this test works correctly with trace enabled and disabled?

manishmishra6016 (approved):
Looks really nice :)

Issue #903, opened 2024-07-05
Dynamically adapt Number of Threads


FG-TUM:
**Is your feature request related to a problem? Please describe.**
Sometimes a domain or scenario is not big enough to fully leverage all available threads. This is especially true in non-mpi simulations on machines with high thread count. 

While this does not give or take time to solution in a significant way it adds unnecessary power consumption. 

**Describe the solution you'd like**
Before every OpenMP parallel loop, the number of threads should be calculated / estimated that have a high parallel efficiency. This is the number of hat should be launched. 

Parameters that might be taken into account for this: 
- Number of parallelizable blocks (most of the time cells per color or slice)
- Particles per cell
  -> particles per thread
- arithmetic intensity of the functor. (can be calculated via neededAttributes and some getFlops())
- Homogeneity (?)

**Describe alternatives you've considered**
It should be possible to dis/enable the feature 

**Additional context**
The number of threads then also should be logged somewhere. Iteration and turning loggers maybe? 

Could be interesting for WindHPC.

Issue #856, opened 2024-05-03
Remove `VarVerletLists`


FG-TUM:
**Is your feature request related to a problem? Please describe.**
`VarVerletLists` were an approach to make the Verlet Container more generic because most different VL approaches only differ in the style of their lists.
However, we only ever implemented one style for this container (`vvl_asBuild`) and implemented the same interface idea over at `VerletListsCells`.

**Describe the solution you'd like**
- [ ] Remove VarVerletLists
- [ ] Port the `asBuild` neighbor list style to the `VerletListsCells`

~**Describe alternatives you've considered**~

**Additional context**
We could also get rid of the regular `VerletLists` container in the same way.

PR #1101, opened 2026-02-20
OMP Thread Count Tuning

rubenhorn:
# Description

Explicitly tune the number of OMP threads in every parallel section (not globally).
Available options are specified in the config.

## Related ~Pull Requests~

- Similar idea to https://github.com/AutoPas/AutoPas/compare/master...AutoPas:AutoPas:903-dynamically-adapt-number-of-threads
- Related to #1138 (also #934)

## Resolved Issues

- [x] #903

# How Has This Been Tested?

- [x] Manual tests show  
   + A) correctness
   + B) difference in performance/energy for different thread counts
   + C) optimal thread count selection
- [x] Unit tests

## Summary by CodeRabbit

* **New Features**
  * Added configurable OpenMP thread counts for AutoPas tuning.
  * Added `--thread-count` and YAML configuration support, including multiple candidate values and automatic maximum-thread fallback.
  * Thread-count choices are now included in tuning search spaces and distributed configurations.
  * Example simulations, traversals, and integrations consistently use the selected tuned thread count.

* **Documentation**
  * Updated iterator examples to use the tuned thread count.

* **Style**
  * Reformatted OpenMP directives for improved readability without changing behavior.


rubenhorn:
@SamNewcome for some reason I still can't request a reviewer, so I guess feel free to review this (anyone) :sweat_smile:

examples/md-flexible/input/AllOptions.yaml
```diff
@@ -24,6 +24,7 @@ output-suffix                        :  usefulNameForLogFiles
 use-LOESS-smoothening                :  false
 tuning-max-evidence                  :  10
 energy-sensor                        :  rapl  # alternatively: likwid
+thread-count                         :  [1,2,4,8] # use [0] to disable thread tuning
```
SamNewcome: Instead of [0], I would have it that providing nothing here defaults to however many threads we would use otherwise.

src/autopas/tuning/tuningStrategy/LiveInfo.h
```diff
@@ -342,7 +342,8 @@ class LiveInfo {
     infos["skin"] = skin;
     infos["rebuildFrequency"] = rebuildFrequency;
     infos["particleSize"] = sizeof(Particle_T);
-    infos["threadCount"] = static_cast<size_t>(autopas_get_max_threads());
+    infos["threadCount"] =
+        static_cast<size_t>(autopas_get_preferred_num_threads());  // NOTE: May not be true everywhere (may use more)
```
SamNewcome: Let's make this "maxThreadCount". I'd like LiveInfo to not include anything relating to the tunable configuration but to relate to the environment for which a configuration should be used.

src/autopas/tuning/Configuration.h
```diff
@@ -140,6 +155,10 @@ class Configuration {
    * Interaction type of the configuration.
    */
   InteractionTypeOption interactionType;
+  /**
+   * OpenMP thread count
```
SamNewcome: Add something like "Preferred OpenMP thread count. In practice, if lower than 1 or higher than the maximum number of threads, will be clamped to these values"

src/autopas/utils/AutoPasConfigurationCommunicator.cpp
```diff
@@ -56,13 +57,14 @@ size_t getSearchSpaceSize(const std::set<ContainerOption> &containerOptions, con
...
  * @param interactionType Handle configurations for this interaction type
+ * @param threadCounts inout
  */
 void generateDistribution(const int numConfigs, const int commSize, const int rank,
                           std::set<ContainerOption> &containerOptions, NumberSet<double> &cellSizeFactors,
                           std::set<TraversalOption> &traversalOptions,
                           std::set<LoadEstimatorOption> &loadEstimatorOptions,
                           std::set<DataLayoutOption> &dataLayoutOptions, std::set<Newton3Option> &newton3Options,
-                          const InteractionTypeOption interactionType) {
+                          const InteractionTypeOption interactionType, NumberSetFinite<int> &threadCounts) {
```
SamNewcome: Add `threadCount` before `interactionType`

src/autopas/tuning/Configuration.h
```diff
@@ -32,26 +38,35 @@ class Configuration {
...
    * @param _cellSizeFactor
    * @param _interactionType
+   * @param _threadCount (optional, for energy tuning, defaults to maximum)
    *
    * @note needs constexpr (hence inline) constructor to be a literal.
    */
   constexpr Configuration(ContainerOption _container, double _cellSizeFactor, TraversalOption _traversal,
                           LoadEstimatorOption _loadEstimator, DataLayoutOption _dataLayout, Newton3Option _newton3,
-                          InteractionTypeOption _interactionType)
+                          InteractionTypeOption _interactionType, int _threadCount = ThreadCountNoTuning)
```
SamNewcome: Put `_threadCount` before `_interactionType`, as that's a bit of a special case.

src/autopas/tuning/Configuration.h
```diff
@@ -32,26 +38,35 @@ class Configuration {
...
       : container(_container),
         traversal(_traversal),
         loadEstimator(_loadEstimator),
         dataLayout(_dataLayout),
         newton3(_newton3),
         cellSizeFactor(_cellSizeFactor),
-        interactionType(_interactionType) {}
+        interactionType(_interactionType),
+        threadCount(_threadCount) {}
```
SamNewcome: Similar

src/autopas/utils/ConfigurationAndRankIteratorHandler.h
```diff
@@ -32,21 +32,26 @@ class ConfigurationAndRankIteratorHandler {
...
-      const std::set<TraversalOption> &traversalOptions, const std::set<LoadEstimatorOption> &loadEstimatorOptions,
-      const std::set<DataLayoutOption> &dataLayoutOptions, const std::set<Newton3Option> &newton3Options,
-      const InteractionTypeOption &interactionType, const int numConfigs, const int commSize)
+  ConfigurationAndRankIteratorHandler(const std::set<ContainerOption> &containerOptions,
+                                      const std::set<double> &cellSizeFactors,
+                                      const std::set<TraversalOption> &traversalOptions,
+                                      const std::set<LoadEstimatorOption> &loadEstimatorOptions,
+                                      const std::set<DataLayoutOption> &dataLayoutOptions,
+                                      const std::set<Newton3Option> &newton3Options,
+                                      const InteractionTypeOption &interactionType, const std::set<int> &threadCounts,
```
SamNewcome: Move `threadCounts` before `interactionType`.

src/autopas/AutoPasDecl.h
```diff
@@ -1025,6 +1026,35 @@ class AutoPas {
     _autoTunerInfo.energySensor = energySensorOption;
   }
 
+  /**
+   * Setter for the thread counts
+   * @param allowedThreadCounts
+   */
+  void setAllowedThreadCounts(const NumberSet<int> &allowedThreadCounts) {
+    if (!allowedThreadCounts.isFinite()) {
```
SamNewcome:
```suggestion
    if (not allowedThreadCounts.isFinite()) {
```
(Typical coding style in AutoPas)

src/autopas/AutoPasDecl.h
```diff
@@ -1025,6 +1026,35 @@ class AutoPas {
...
+      utils::ExceptionHandler::exception("Error: thread count options must be finite!", allowedThreadCounts.getMin());
```
SamNewcome: Not sure if the `allowedThreadCounts.getMin()` is supposed to be here

src/autopas/AutoPasDecl.h
```diff
@@ -1025,6 +1026,35 @@ class AutoPas {
...
+    }
+    if (allowedThreadCounts.getMin() == Configuration::ThreadCountNoTuning and allowedThreadCounts.size() != 1) {
+      utils::ExceptionHandler::exception(
+          "Error: when thread counts contain {} (no thread tuning), no other values may be provided!",
+          Configuration::ThreadCountNoTuning);
+    }
+    if (allowedThreadCounts.getMin() < 0) {
+      utils::ExceptionHandler::exception("Error: minimum thread count must be positive {} < 0!",
+                                         allowedThreadCounts.getMin());
+    }
```
SamNewcome: I guess this can be < 1

tests/testAutopas/tests/containers/TraversalComparison.cpp
SamNewcome, line 1: Don't neglect the `threadCount` in this. It probably makes no difference -> but we should check. (We probably should do this anyway, but I guess this is as good a time as ever).

I think 1, 2, 4 should be fine. You can add a check to skip anything above max threads.

src/autopas/LogicHandler.h
```diff
@@ -1302,6 +1304,9 @@ bool LogicHandler<Particle_T>::computeInteractionsPipeline(Functor *functor,
 
   /// Computing the particle interactions
   AutoPasLog(DEBUG, "Iterating with configuration: {} tuning: {}", configuration.toString(), stillTuning);
+  if (autoTuner.getCurrentConfig().threadCount != 0) {
+    autopas_set_preferred_num_threads(autoTuner.getCurrentConfig().threadCount);
+  }
```
SamNewcome: Move this into `isConfigurationApplicable`. Check that the thread count is not below 1 or above max threads, and if so, reject it (return `nullptr`). If we assume that max threads never changes during a simulation, you can return `true` for `rejectIndefinitely`. I think this is fair, but otherwise return `false`.

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.h
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+#pragma once
+
+#include "AutoPasTestBase.h"
+
+class ThreadCountTuningTest : public AutoPasTestBase {
+ public:
+  /**
+   * Tests if thread count tuning selected the expected number of threads for a given simulation size
```
SamNewcome: "for a given simulation size" is not really clear

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.cpp
```diff
@@ -0,0 +1,94 @@
...
+  autopas::Logger::get()->set_level(autopas::Logger::LogLevel::off);
+  //  autopas::Logger::get()->set_level(autopas::Logger::LogLevel::debug);
+  bool stillTuning = true;
+  autopas::Configuration currentConfig;
+  autopasTools::generators::GridGenerator::fillWithParticles(logicHandler.getContainer(), {boxMax, boxMax, boxMax},
+                                                             Molecule());
+  const size_t numInsertedMolecules = logicHandler.getContainer().size();
+
+  int iterationsAfterTuning = 0;
+  while (stillTuning && iterationsAfterTuning < 1) {
```
SamNewcome:
```suggestion
  while (stillTuning and iterationsAfterTuning < 1) {
```

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.cpp
```diff
@@ -0,0 +1,94 @@
...
+    // Should not have any leaving molecules in this test
+    auto dummyMoleculesVec = logicHandler.updateContainer();
+    stillTuning = logicHandler.computeInteractionsPipeline(&functor, autopas::InteractionTypeOption::pairwise);
+    currentConfig = tunerMap[autopas::InteractionTypeOption::pairwise]->getCurrentConfig();
+    if (!stillTuning) iterationsAfterTuning++;
```
SamNewcome:
```suggestion
    if (not stillTuning) iterationsAfterTuning++;
```

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.cpp
```diff
@@ -0,0 +1,94 @@
...
+            logicHandler.getContainer().size());  // Should not have any leaving molecules in this test
+  // NOTE: currentConfig.threadCount does not return the actual number of threads when thread count tuning is turned off
+  // Instead, check the actual number of threads to be used as set by the current configuration
+  EXPECT_EQ(expectedSelectedThreadCount, autopas::autopas_get_preferred_num_threads());
+}
+
+/**
+ * Tests thread count tuning:
+ * vary small scenario (N=8) -> lowest number of threads
+ * larger scenario (N=33k)   -> highest number of threads
```
SamNewcome:
```suggestion
 * larger scenario (N=33)   -> highest number of threads
```

SamNewcome: Also, where is this 33 coming from?

tests/testAutopas/tests/tuning/AutoTunerTest.cpp
```diff
@@ -33,6 +33,7 @@ using ::testing::_;
 
 TEST_F(AutoTunerTest, testAllConfigurations) {
   const autopas::NumberSetFinite<double> cellSizeFactors({1});
+  const autopas::NumberSetFinite<int> threadCounts{autopas::Configuration::ThreadCountNoTuning};
```
SamNewcome: Give multiple options here.

tests/testAutopas/tests/utils/AutoPasConfigurationCommunicatorTest.h
```diff
@@ -29,6 +29,7 @@ class AutoPasConfigurationCommunicatorTest : public AutoPasTestBase {
   const std::set<autopas::DataLayoutOption> dataLayoutOptions = autopas::DataLayoutOption::getMostOptions();
   const std::set<autopas::Newton3Option> newton3Options = autopas::Newton3Option::getMostOptions();
   const autopas::NumberSetFinite<double> cellSizeFactors = {0.5, 1.0, 1.5};
+  const autopas::NumberSetFinite<int> threadCounts = {autopas::Configuration::ThreadCountNoTuning};
```
SamNewcome: Give multiple options

SamNewcome (changes requested):
Nice


Some more general comments:

- The thread count should be added to all loggers that log the configuration. I'm not sure if this is done automatically, so check this and change these if not. These should be in `utils/logging` (I think)
- I wouldn't explicitly handle thread count tuning disabled. It makes some handling awkward, you will also have to work around it for logging, and I guess its still useful for any learning tuning strategy to have this information, in case it later gets used with thread count tuning enabled. I would simply have AutoPas (and md-flexible) set the `NumberSet` to max threads by default.
- In tests, use multiple thread count options.

examples/md-flexible/input/AllOptions.yaml
rubenhorn: `[0]` is already the implicit default, but you may explicitly use it to show that thread count tuning is disabled.

src/autopas/tuning/Configuration.h
rubenhorn: The value is not clamped or validated in the configuration class, but the Doxygen comment can be enhanced.

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.cpp
rubenhorn: $N$ is the number of particles (hence the k).  
Will rename it to `particles` to be more explicit.

rubenhorn:
> I wouldn't explicitly handle thread count tuning disabled.

This requires the constructors of `Configuration` to not be a `constexpr`. (`[-Winvalid-constexpr]`)

SamNewcome:
> This requires the constructors of `Configuration` to not be a `constexpr`. (`[-Winvalid-constexpr]`)

To be clear, I mean that you always explicitly have the thread count in the configuration. To disable the tuning of thread count, you simply have only one thread count option (similarly to just having AoS to disable data layout tuning) and you make this more seamless by having AutoPas set this to the maximum number of threads only if a thread count setter is never used.

So, from the perspective of the `Configuration` class, thread count is always explicitly given -> it is only from the perspective of AutoPas's interface that the option to ignore thread count tuning exists (and have this then by default set to maximum number of threads).

This is on purpose. We want AutoPas to internally still be aware of the actual thread count. This makes logs clearer, means we don't have to handle tuning disabled any differently, and learning-based tuning strategies can still learn from thread count tuning disabled runs because that information is still included in the Configuration object and the logs.

manishmishra6016 (commented):
Looks good overall. 
Maybe it would be nice to confirm whether the changes here have any overhead compared to master (when threadCount tuning is disabled). Comparison with any example and a comment in the pull request should be enough. I don't expect anything major, but it might be nice to do it for future reference, since we now do a lot of runtime settings.

tests/testAutopas/tests/containers/TraversalComparison.cpp
rubenhorn: Thread count does nothing here. (I don't know how it got there in the first place.)

SamNewcome:
> Comparison with any example and a comment in the pull request should be enough. 

You could go with a very minimal example - a couple of particles. This I guess would be the worst case scenario if there were to be some overhead, I'd guess.

src/autopas/containers/cellTraversals/SlicedBalancedBasedTraversal.h
SamNewcome, line 1: I stumbled across something which I think needs to also be changed. In SlicedBalanced, we specifically create one slice per thread, so in addition to parallelising with the preferred number of threads, you should also change the handling of this in slice creation: I think this is the line

```
    auto numSlices = (size_t)autopas_get_max_threads();
```
in `initSliceThickness`.

docs/userdoc/Iterators.md
```diff
@@ -102,7 +102,7 @@ Iteration of the AutoPas object can also be done in parallel via OpenMP.
 For this, place any of the examples above in a parallel region.
 
 ```c++
-AUTOPAS_OPENMP(parallel)
+AUTOPAS_OPENMP(parallel num_threads(autopas::autopas_get_tuned_num_threads()))
```
SamNewcome, line 105: In your opinion, is there significant performance gain from using the tuned number in iterators. I'm more reluctant about changes here as this affects consumers of AutoPas, as opposed to just an internal mechanism.

src/autopas/tuning/Configuration.cpp
```diff
@@ -23,7 +25,7 @@ std::string autopas::Configuration::getCSVLine() const { return getCSVRepresenta
 bool autopas::Configuration::hasValidValues() const {
   return container != ContainerOption() and cellSizeFactor != -1 and traversal != TraversalOption() and
          loadEstimator != LoadEstimatorOption() and dataLayout != DataLayoutOption() and newton3 != Newton3Option() and
-         interactionType != InteractionTypeOption();
+         interactionType != InteractionTypeOption() and threadCount >= 0 and threadCount <= autopas_get_max_threads();
```
SamNewcome, line 28: Shouldn't this be exclusive "> 0"?

src/autopas/tuning/Configuration.h
```diff
@@ -34,29 +35,38 @@ class Configuration {
    * @param _newton3
    * @param _cellSizeFactor
    * @param _interactionType
+   * @param _threadCount (optional, for energy tuning, defaults to maximum number of threads)
```
SamNewcome: I wouldn't define this as "optional". Every element in the configuration is defined, there is just a clear default option in this case.

src/autopas/utils/WrapOpenMP.h
```diff
@@ -66,6 +66,25 @@ inline int autopas_get_max_threads() { return omp_get_max_threads(); }
  */
 inline void autopas_set_num_threads(int n) { omp_set_num_threads(n); }
 
+inline int &_autopas_get_tuned_num_threads_ref() {
```
SamNewcome, line 75: Add documentation to the function

src/autopas/containers/verletListsCellBased/varVerletLists/neighborLists/asBuild/VerletNeighborListAsBuild.h
SamNewcome, line 1: VarVerletListsAsBuild (i.e. this file and `VVLAsBuildTraversal`) is a bit messy when it comes to threads, as the central idea is to "remember" the thread that built the list and use this to use N3 during the traversal. Due to some weird bug (#721), we don't encourage this traversal/container and issue #856 suggests even removing it. 

Unless you really want, I would rather this traversal doesn't get tuned to the max number of threads, and just uses max threads like before. The cleanest way to implement this would be to modify `Configuration::hasCompatibleValues`, to remove any configuration where this is not the max number of threads.

tests/testAutopas/testingHelpers/GenerateValidConfigurations.h
```diff
@@ -126,51 +129,57 @@ inline std::set<autopas::Configuration> generateAllValidConfigurations(
  * why.
  * @param allowedContainerOptions By default, all options.
  * @param allowedCellSizeFactors By default, {0.5, 1.0, 1.5}
+ * @param allowedThreadCounts By default, {max_threads}
```
SamNewcome, line 132: Include more options -> we need to check that this is correct for other thread counts.

My suggestion: 1, 2, {max_threads}.

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.cpp
```diff
@@ -0,0 +1,129 @@
...
+  autopas::AutoTuner::TuningStrategiesListType tuningStrategies{};
+
+  auto tunerManager = std::make_shared<autopas::TuningManager>(autoTunerInfo);
+  tunerManager->addAutoTuner(
+      std::make_unique<autopas::AutoTuner>(tuningStrategies, searchSpace, autoTunerInfo, verletRebuildFrequency, ""),
+      autopas::InteractionTypeOption::pairwise);
+  autopas::LogicHandler<Molecule> logicHandler(tunerManager, logicHandlerInfo, verletRebuildFrequency, "");
+
+  autopas::Logger::get()->set_level(autopas::Logger::LogLevel::off);
+  //  autopas::Logger::get()->set_level(autopas::Logger::LogLevel::debug);
```
SamNewcome: Remove dead code

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.cpp
```diff
@@ -0,0 +1,129 @@
...
+      &threadCounts, vecPatternOptions, autopas::InteractionTypeOption::pairwise);
+  autopas::AutoTuner::TuningStrategiesListType tuningStrategies{};
+
+  auto tunerManager = std::make_shared<autopas::TuningManager>(autoTunerInfo);
+  tunerManager->addAutoTuner(
+      std::make_unique<autopas::AutoTuner>(tuningStrategies, searchSpace, autoTunerInfo, verletRebuildFrequency, ""),
+      autopas::InteractionTypeOption::pairwise);
+  autopas::LogicHandler<Molecule> logicHandler(tunerManager, logicHandlerInfo, verletRebuildFrequency, "");
+
+  autopas::Logger::get()->set_level(autopas::Logger::LogLevel::off);
```
SamNewcome: Why is this set off? We might miss some warning messages

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.h
```diff
@@ -0,0 +1,24 @@
...
+
+class ThreadCountTuningTest : public AutoPasTestBase {
+ public:
+  /**
+   * Tests if thread count tuning selected a valid option out of a given set and (optionally) the the expected number of
+   * threads (e.g. minimum for a small simulation size, maximum for a large simulation size, or max threads if 0).
+   *
+   * @param boxMax The size of the simulation domain along each side
+   * @param threadCountOptions The possible thread counts in the configuration
+   * @param expectedSelectedThreadCount Optional expected number of threads of the configuration after tuning has
```
SamNewcome, line 19: Use `std::optional`

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.cpp
```diff
@@ -0,0 +1,129 @@
...
+  //  autopas::Logger::get()->set_level(autopas::Logger::LogLevel::debug);
+  bool stillTuning = true;
+  autopasTools::generators::GridGenerator::fillWithParticles(logicHandler.getContainer(), {boxMax, boxMax, boxMax},
+                                                             Molecule());
+  const size_t numInsertedMolecules = logicHandler.getContainer().size();
+
+  int iterationsAfterTuning = 0;
```
SamNewcome, line 62: I'm not sure what the point of this is

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.cpp
```diff
@@ -0,0 +1,129 @@
...
+  while (stillTuning and iterationsAfterTuning < 1) {
```
SamNewcome, line 63: Change this into a `for` loop with a high iteration count. Means that if this fails for whatever reason, we don't get stuck in an infinite loop.

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.cpp
```diff
@@ -0,0 +1,129 @@
...
+#include "autopas/tuning/AutoTuner.h"
+#include "autopas/tuning/Configuration.h"
+#include "autopas/tuning/utils/AutoTunerInfo.h"
+#include "autopas/tuning/utils/SearchSpaceGenerators.h"
+#include "generators/src/GridGenerator.h"
+#include "testingHelpers/commonTypedefs.h"
+
+using ::testing::_;
+
+void ThreadCountTuningTest::testThreadCountTuningWithBoxMax(const size_t boxMax,
```
SamNewcome, line 20: Maybe remove the `WithBoxMax` part - it's a bit confusing because its not really a critical part of the simulation.

src/autopas/utils/WrapOpenMP.h
```diff
@@ -66,6 +66,25 @@ inline int autopas_get_max_threads() { return omp_get_max_threads(); }
  */
 inline void autopas_set_num_threads(int n) { omp_set_num_threads(n); }
```
SamNewcome, line 68: Add documentation that this will set the max number of threads and not the tuned number of threads.

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.cpp
SamNewcome, line 1: There's a couple issues with the tests - some don't get run in the CI is the big issue. 

What do you mean by "Sensitive to shared CPU usage! (Skipped in CI.)"? Is it that the max number of threads ends up being 1? You should be able to set the number of threads to e.g. 4 which is then what is actually used as the max number of threads - even if there aren't 4 hardware threads. This is good enough for testing purposes. (You maybe want to do something similar in the traversal comparison)

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.cpp
```diff
@@ -0,0 +1,129 @@
...
+ * Sensitive to shared CPU usage! (Skipped in CI.)
+ * To run the test explicitly, use:
+ *   ./build/tests/testAutopas/runTests \
+ *     --gtest_filter=ThreadCountTuningTest.DISABLED_testThreadCountTuningLarge \
+ *     --gtest_also_run_disabled_tests
+ */
+TEST_F(ThreadCountTuningTest, DISABLED_testThreadCountTuningLarge) {
+  const int maxThreads = std::min(autopas::autopas_get_max_threads(), 4);
+  testThreadCountTuningWithBoxMax(32, {1, maxThreads}, maxThreads);
+}
```
SamNewcome, line 151: Any reason why we need to test such as large scenario. By my understanding, we don't do an interactions calculations so the number of particles is somewhat arbitrary.

SamNewcome (commented):
I gave it a review from scratch as its been a while and I have a better understanding of all this now.

Some general suggestions:

### Default-max-thread handling

I would handle default-as-max-thread handling at the interface only. This matches some of your error handling. 

So by default, AutoPas comes preset with a default of just a set of max threads only. We might need some option in md-flexible that allows us to either directly set the threads or provide a default -> I don't like using 0 as a placeholder for this as it is rather opaque.

Then internally we just use "72" threads or whatever this is without requiring any special handling. This gets rid of all of these opaque uses of "0" within the code, which are unclear if not explained and require comment explanations constantly (more than what you have currently provided).

This is also useful for debugging and for looking at tuning logs and for learning optimal algorithms.

### Balanced Traversals

IMHO, I would actually do the balancing itself according to the tuned number of threads.

Otherwise, you end up with basically the same per-thread splitting of work as with max threads, but with extra unnecessary overhead to avoid race conditions between partitions A & B when they are handled by the same thread.

I think these balanced methods could be really quite effective with the thread count tuning if you balance according to tuned number of threads.

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.cpp
rubenhorn: The implemented, skipped tests, are more like benchmarks, because they are flaky and depend on parallel CPU load which cannot be controlled for in CI.  

The only similar tests I can think of is providing a configuration and verifying if the actual number of threads matches. (This is already what's being done.)  
I can add another test which tries other thread counts than just the "extremes".

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.cpp
rubenhorn: I did some experiments and with these two scenarios, the tests are most reliable (not flaky)  
...well... except for in the CI.

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.cpp
rubenhorn: Copy paste from other test... I think

tests/testAutopas/testingHelpers/GenerateValidConfigurations.h
rubenhorn: I would say `{1, max_threads}` is better because the tests may run on a single core machine (where 2 is illegal).

src/autopas/tuning/Configuration.cpp
rubenhorn: 0 = max_threads is necessary for portable configurations without thread count tuning.

docs/userdoc/Iterators.md
rubenhorn: I have not tested this, but I thought that this should maybe still fall under the tuning configuration, even if it's on the consumer side.  
(It's also a very small change.)

SamNewcome: >  (It's also a very small change.)

I think its a large change, as this will affect a lot of parts of the code even if the change itself per for loop is small, and will affect consumer code (ls1, MercuryDPM). If there is a good reason for this, I'm happy to keep it, but I want to know there is a good reason for it.

Also, IMHO, I doubt that the optimal number of threads for iterators is the same as for the actually tuned force calculation, and that probably max threads here works a lot better. I may be wrong, but I would like to see that I am wrong before agree to an interface change.

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.cpp
SamNewcome: Can you not mark this as resolved if nothing is done and we've not agreed that doing nothing is the best course of action. Which other test is this? I can have a look

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.cpp
SamNewcome: This doesn't appear to be done - can you not mark it as resolved?

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.cpp
SamNewcome: What do you mean by flaky? Do they fail? Why do they fail?

Design the tests to focus on what actually can be tested. 

If these are benchmarks, then the testing framework is not the appropriate place for them. If you let me know what the intention of these are, I can maybe help suggest something.

Also, some really simply unit tests would be beneficial. E.g. checking that the thread count being set actually does something.

SamNewcome: What do you mean by flaky? Do they fail? Why do they fail?

Design the tests to focus on what actually can be tested.

If these are benchmarks, then the testing framework is not the appropriate place for them. If you let me know what the intention of these are, I can maybe help suggest something.

Also, some really simply unit tests would be beneficial. E.g. checking that the thread count being set actually does something.

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.cpp
SamNewcome: What do you mean by reliable?

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.cpp
rubenhorn: Just forgot to push it

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.cpp
rubenhorn: Related to other comment. After pushing this code does no longer exist.

tests/testAutopas/tests/tuning/utils/ThreadCountTuningTest.cpp
rubenhorn: This is to enure tuning is actually completed (avoid off-by-one error) before evaluating selected configuration.

PR #1129, opened 2026-03-29
Optimal rebuild frequency for minimization of rebuild and remainder traversal time

nata-81:
# Description

This pull request introduces a new dynamic rebuild strategy that decides in each simulation step whether the container should be rebuilt. The decision is based on minimizing the combined rebuild time and remainder traversal time.

For the decision the remainder traversal time of the next simulation iteration is needed. It is estimated using a linear regression (using the [boost library](https://theboostcpplibraries.com/)) over the last five remainder traversal time values, depending on the increasing buffer filling.

A representative of the rebuild time is also required and is estimated as the mean of up to the last 100 rebuild time measurements.

The method significantly reduces rebuild and remainder traversal time in the selected test cases, especially for the Verlet List Cells container type.
However, it may influence container interaction time, which is currently not considered in the implementation.

## Related Pull Requests

- #821 Replaces and was compared against this dynamic rebuild strategy
- #1002 Builds on the idea of placing particles into the buffer if they move more than half the skin
- #1030 The rebuild frequency cannot be estimated in this way for the new type

## Resolved Issues

- None, but the user does not have to choose a rebuild frequency

# How Has This Been Tested?

The estimation of rebuild time and remainder traversal time was tested, as well as the decision making process and the behavior of the new rebuild criterion in the simulation loop.

- testMean: tests rebuild time estimation
- testSimpleLinearRegressionBoost: tests remainder traversal time estimation
- testRebuildDecisionContext: tests rebuild decision logic
- testOneParticleForDynamicRebuild & testParticleInContainerMoveAcrossPeriodicBoundaryForDynamicRebuild:test rebuild decisions within the simulation loop


manishmishra6016:
Feature being developed further in #1170

PR #1135, opened 2026-04-19
HGrid C08

AlexanderGlas:
# Description

This change adds a Fitted  Hierarchical Grids Container and a C08-based Traversal for it.
Adapts HGBlock Traversal to use a traversal that performs just like LC_C08, but respects the larger halo regions to skip unnecessary halo-halo interactions. 

## Related Pull Requests

- #1014 

# How Has This Been Tested?

- These changes were tested with the usual testing pipeline as well as the Hgrid-specific tests from https://github.com/AutoPas/AutoPas/pull/1014. (The Hgrid Traversal Comparison was used for testing, but has since been removed, as it needs to be integrated into Traversal Comparison.)


src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
```diff
@@ -0,0 +1,236 @@
...
+      if (stopCellCenter[1] >= highCornerCell2[1] && stopIndex3D[1] > 0) {
+        stopIndex3D[1]--;
+      }
+      if (stopCellCenter[2] >= highCornerCell2[2] && stopIndex3D[2] > 0) {
+        stopIndex3D[2]--;
+      }
+    }
+    for (size_t z = startIndex3D[2]; z <= stopIndex3D[2]; ++z) {
+      for (size_t y = startIndex3D[1]; y <= stopIndex3D[1]; ++y) {
+        for (size_t x = startIndex3D[0]; x <= stopIndex3D[0]; ++x) {
```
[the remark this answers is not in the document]
AlexanderGlas: This cannot happen: Unfitted grids are considered only if the ratio of cell sizes across levels is at least 2. That means at least one lower-level cell in each direction needs to be inside a higher-level cell.

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -93,6 +95,10 @@ class HierarchicalGrid : public ParticleContainerInterface<Particle_T> {
                  _maxCutoffPerLevel.size());
       _numLevels = _maxCutoffPerLevel.size();
     }
+    if (_fittedGrids) {
+      createFittedGrids(sortingThreshold, loadEstimator);
+      return;
+    }
```
SamNewcome, line 102: Maybe: put the non-fitted case inside a helper function like you did for `createFittedGrids`

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -111,6 +117,86 @@ class HierarchicalGrid : public ParticleContainerInterface<Particle_T> {
...
+   * Create hierarchy levels whose cell grids are fitted to each other.
+   *
+   * @param sortingThreshold Number of particles in two cells from which sorting should be performed.
+   * @param loadEstimator The load estimation algorithm for balanced traversals.
+   */
+  void createFittedGrids(const size_t sortingThreshold, LoadEstimatorOption loadEstimator) {
+    std::vector<double> interactionLengths(_numLevels);
+    interactionLengths[_numLevels - 1] = _maxCutoffPerLevel.back() + _skin;
+
+    // Assure at least 2 levels
```
SamNewcome:
```suggestion
    // Assure at least 2 levels by ...
```
increasing the size of the coarsest grid to double the lowest level.

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -111,6 +117,86 @@ class HierarchicalGrid : public ParticleContainerInterface<Particle_T> {
...
+    interactionLengths[0] = _maxCutoffPerLevel[0] + _skin;
+    if (2.0 * interactionLengths[0] >= interactionLengths.back()) {
+      // Smallest representable value above the exact threshold that still guarantees >=2 lower cells per upper cell.
+      const double minimumSafeRatio = 2.0 * interactionLengths[0] / interactionLengths.back();
+      highLevelRatio = std::nextafter(minimumSafeRatio, std::numeric_limits<double>::infinity());
+    }
+
+    _levels.resize(_numLevels);
+
+    // construct last level first, so we can fit other levels to it
```
SamNewcome: Maybe instead of "last" write "last/smallest"

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -111,6 +117,86 @@ class HierarchicalGrid : public ParticleContainerInterface<Particle_T> {
...
+    _levels[_numLevels - 1] = std::make_unique<autopas::LinkedCells<Particle_T>>(
```
SamNewcome:
```suggestion
    _levels.at(_numLevels - 1) = std::make_unique<autopas::LinkedCells<Particle_T>>(
```

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -111,6 +117,86 @@ class HierarchicalGrid : public ParticleContainerInterface<Particle_T> {
...
+        _boxMin, _boxMax, _maxCutoffPerLevel.back(), _skin, _cellSizeFactor * highLevelRatio, sortingThreshold,
+        loadEstimator);
+
+    auto highestCellLength = _levels[_numLevels - 1]->getCellBlock().getCellLength();
```
SamNewcome:
```suggestion
    auto highestCellLength = _levels.at(_numLevels - 1)->getCellBlock().getCellLength();
```

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -111,6 +117,86 @@ class HierarchicalGrid : public ParticleContainerInterface<Particle_T> {
...
 
+ private:
+  /**
+   * Create hierarchy levels whose cell grids are fitted to each other.
+   *
+   * @param sortingThreshold Number of particles in two cells from which sorting should be performed.
+   * @param loadEstimator The load estimation algorithm for balanced traversals.
+   */
+  void createFittedGrids(const size_t sortingThreshold, LoadEstimatorOption loadEstimator) {
+    std::vector<double> interactionLengths(_numLevels);
```
SamNewcome: I'd suggest a different variable name -> otherwise this gets weird with the idea of `maxCutoffPerLevel`. Maybe `maxInterLenPerLevel`?

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -111,6 +117,86 @@ class HierarchicalGrid : public ParticleContainerInterface<Particle_T> {
...
+    interactionLengths[_numLevels - 1] = _maxCutoffPerLevel.back() + _skin;
```
SamNewcome: I would prefer it if you set all `interactionLengths`/`maxInterLenPerLevel` here already. Otherwise, there is a risk that a future developer tries to access this for one of the intermediate levels before it is set.

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -111,6 +117,86 @@ class HierarchicalGrid : public ParticleContainerInterface<Particle_T> {
...
+
+    // Assure at least 2 levels
+    double highLevelRatio = 1.0;
+    interactionLengths[0] = _maxCutoffPerLevel[0] + _skin;
+    if (2.0 * interactionLengths[0] >= interactionLengths.back()) {
```
SamNewcome: I would add a DEBUG level message if this is true

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -111,6 +117,86 @@ class HierarchicalGrid : public ParticleContainerInterface<Particle_T> {
...
+    // construct last level first, so we can fit other levels to it
+    _levels[_numLevels - 1] = std::make_unique<autopas::LinkedCells<Particle_T>>(
+        _boxMin, _boxMax, _maxCutoffPerLevel.back(), _skin, _cellSizeFactor * highLevelRatio, sortingThreshold,
+        loadEstimator);
+
+    auto highestCellLength = _levels[_numLevels - 1]->getCellBlock().getCellLength();
+    std::array<size_t, 3> cellsPerDimensionHigher = {
+        static_cast<size_t>(std::llround((_boxMax[0] - _boxMin[0]) / highestCellLength[0])),
+        static_cast<size_t>(std::llround((_boxMax[1] - _boxMin[1]) / highestCellLength[1])),
+        static_cast<size_t>(std::llround((_boxMax[2] - _boxMin[2]) / highestCellLength[2]))};
```
SamNewcome: "Higher" -> "Highest"

Also, just get this from cell block. You can use `getCellsPerDimensionWithHalo` and remove the halos (Feel free to just make this a helper function of `CellBlock3D`). This avoids having to mess about with rounding and casting.

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -111,6 +117,86 @@ class HierarchicalGrid : public ParticleContainerInterface<Particle_T> {
...
+    std::array<size_t, 3> cellsPerDimension;
+    size_t lowerCellsPerHigher;
```
SamNewcome:
```suggestion
    size_t numLowerCellsPerHigher;
```

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -111,6 +117,86 @@ class HierarchicalGrid : public ParticleContainerInterface<Particle_T> {
...
+      // increase size of lower level cells to fit the bigger level cell, if it does not fit, remove that level
+      auto cellLength = _levels[i + 1]->getCellBlock().getCellLength();
+      for (size_t d = 0; d < 3; ++d) {
+        lowerCellsPerHigher =
+            static_cast<size_t>(std::floor(cellLength[d] / (interactionLengths[i] * _cellSizeFactor)));
+        // sort out levels too close to be fitted
+        if (lowerCellsPerHigher < 2) {
+          _maxCutoffPerLevel.erase(_maxCutoffPerLevel.begin() + i);
+          interactionLengths.erase(interactionLengths.begin() + i);
+          _levels.erase(_levels.begin() + i);
```
SamNewcome, line 181: Why have the `.begin()`?

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -111,6 +117,86 @@ class HierarchicalGrid : public ParticleContainerInterface<Particle_T> {
...
+      }
+      // if level was removed, create no linked cells and continue with next level
+      if (lowerCellsPerHigher < 2) {
+        continue;
+      }
+      cellsPerDimensionHigher = cellsPerDimension;
+      // @todo: This case is currently dead code and can be removed, I left it for now, so there is some trace of this
+      // existing, in case we might want non-fitted Hgrids using the HGC08 traversal in the future.
+      // Case we dont want to fit the grids perfecty, but just want to make sure they are not too close together, to
+      // keep stride of 3
```
SamNewcome: In my opinion there are two options:

- Remove C08 with non-fitted. 
- Keep it, but implement it nicely i.e. not as deadcode but a genuine option.

I don't really have a strong opinion.

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
```diff
@@ -111,6 +117,86 @@ class HierarchicalGrid : public ParticleContainerInterface<Particle_T> {
...
+      if (!_fittedGrids) {
+        double ratio = interactionLengths[i] / interactionLengths.back();
+        // again the hacky way to get the proper halo size
+        _levels[i] = std::make_unique<autopas::LinkedCells<Particle_T>>(_boxMin, _boxMax, _maxCutoffPerLevel.back(),
+                                                                        _skin, _cellSizeFactor * ratio,
+                                                                        sortingThreshold, loadEstimator);
+      } else {
+        // again the hacky way to get the proper halo size
```
SamNewcome: I would copy the comment from the original or refer to where it is. Otherwise if somebody only reads this they don't know what you mean and where to find out about it

src/autopas/containers/hierarchicalGrid/traversals/HGFitC08Traversal.h
```diff
@@ -0,0 +1,138 @@
...
+        _functor(functor),
+        _dataLayoutConverter(_functor, dataLayout) {}
+
+  void traverseParticles() override;
+
+  [[nodiscard]] TraversalOption getTraversalType() const override { return TraversalOption::hgridFit_c08; };
+
+  /**
+   * C08 traversals are always usable for fitted grids.
+   * @todo: In the future sort out single level Hgrids.
```
SamNewcome, line 80: You can do this already by calling the parent class's `isApplicable` which checks this. I think in the other branch, copilot suggested this for the block traversal, in case you want an example.

src/autopas/containers/hierarchicalGrid/traversals/HGFitC08Traversal.h
```diff
@@ -0,0 +1,138 @@
...
+   */
+  utils::DataLayoutConverter<Functor_T> _dataLayoutConverter;
+
+  /**
+   * HGFitc08Traversal does intra and inter-level traversals combined, so generating seperate intra-level traversals is
+   * not needed.
+   * @param level which HGrid level to generate a traversal for
+   * @return
+   */
+  std::unique_ptr<TraversalInterface> generateNewTraversal(const size_t level) override { return nullptr; }
```
SamNewcome: I think this function was added to `HGTraversalBase` because originally all HG traversals (i.e. block + the others which were removed due to poor performance) all needed intra-level traversals.

Having this function here where it does nothing, and should not be used, is simply a risk.

I would therefore make this function simply a non-virtual member function of `HGBlockTraversal`. You therefore don't need it in the `HGTraversalBase`.

In addition, the `initTraversal` function of `HGTraversalBase` assumes generating intra-level traversals, which is similarly not appropriate for `HGFitC08Traversal`, and is overriden in this class. I would therefore also move that implementation of `initTraversal` to `HGBlockTraversal`.

src/autopas/containers/hierarchicalGrid/traversals/HGFitC08Traversal.h
```diff
@@ -0,0 +1,138 @@
...
+template <class ParticleCell_T, class Functor_T>
+inline void HGFitC08Traversal<ParticleCell_T, Functor_T>::traverseParticles() {
+  const double haloRegionWidth = this->_maxCutoffPerLevel.back() + this->_skin;
+
+  // get a vector of cell blocks for easier access in the traversal
+  std::vector<CellBlock *> cellBlocks;
+  cellBlocks.reserve(this->_levels->size());
+  for (auto &linkedCell : *this->_levels) {
+    cellBlocks.emplace_back(&linkedCell->getCellBlock());
+  }
```
SamNewcome, line 154: @SamNewcome To come back to after finishing review: Do I like this?

src/autopas/containers/hierarchicalGrid/traversals/HGC08SingleLevelTraversal.h
```diff
@@ -0,0 +1,111 @@
...
+#include "autopas/containers/cellTraversals/C08BasedTraversal.h"
+#include "autopas/containers/linkedCells/traversals/LCTraversalInterface.h"
+#include "autopas/utils/WrapOpenMP.h"
+
+namespace autopas {
+
+/**
+ * This class provides the hg_c08 traversal, which is an adaptation of lc_c08 for Hierarchical Grids.
+ * It calculates the intra-level interactions of one level, and the inter-level interactions of this level, with all
+ * lower levels respectively. (e.g. when applied to level 3, it calculates 3-3, 3-2 and 3-1 interactions).
```
SamNewcome:
```suggestion
 * It calculates the inter-level interactions of one level, and the intra-level interactions of this level with all
 * lower levels. (e.g. when applied to level 3, it calculates 3-3, 3-2 and 3-1 interactions).
```

src/autopas/containers/hierarchicalGrid/traversals/HGFitC08Traversal.h
```diff
@@ -0,0 +1,138 @@
...
+ protected:
+  // HGTraversalBase for some reason needs hardcoded FullParticleCell
+  /**
+   * Cell block type used for hierarchical-grid traversal.
+   */
+  using CellBlock = internal::CellBlock3D<FullParticleCell<Particle>>;
+  /**
+   * Pairwise functor used to compute interactions.
+   */
+  Functor_T *_functor;
```
SamNewcome: Make this instead a reference

src/autopas/containers/hierarchicalGrid/traversals/HGC08SingleLevelTraversal.h
```diff
@@ -0,0 +1,111 @@
...
+                                     const std::vector<internal::CellBlock3D<ParticleCell_T> *> &cellBlocks,
+                                     const std::vector<double> &interactionLengthsSquared, const size_t upperLevel,
+                                     double haloRegionWidth, bool fittedGrids = true)
+      : C08BasedTraversal<ParticleCell_T, PairwiseFunctor>(dims, pairwiseFunctor, interactionLength, cellLength,
+                                                           dataLayout, useNewton3),
+        _cellHandler(pairwiseFunctor, this->_cellsPerDimension, interactionLength, cellLength, this->_overlap,
+                     dataLayout, useNewton3, cellBlocks, interactionLengthsSquared, upperLevel, fittedGrids) {
+    for (size_t d = 0; d < 3; ++d) {
+      _haloRegionCellWidth[d] = std::ceil(haloRegionWidth / cellLength[d]);
+    }
```
SamNewcome: You need to be really careful with this kind of thing.

- FP divisions are approximate
- A number divided perfectly by another could result in a number slightly above or slightly below the true solution. E.g. A / B = X.0000001 or A / B = (X-1).9999999
- If you get X.0000001, the ceil operation results in X+1 => incorrect width.

Also, don't do implicit type conversions. Use `static_cast`.


```suggestion
    for (size_t d = 0; d < 3; ++d) {
      _haloRegionCellWidth[d] = static_cast<size_t>(std::llround(haloRegionWidth / cellLength[d]));
    }
```

src/autopas/containers/hierarchicalGrid/traversals/HGC08SingleLevelTraversal.h
```diff
@@ -0,0 +1,111 @@
...
+
+ private:
+  /*
+   * The cell handler for managing the cells in the hierarchical grid.
+   */
+  HGC08CellHandler<ParticleCell_T, PairwiseFunctor> _cellHandler;
+  /**
+   * The width of the halo region in cells.
+   */
+  std::array<size_t, 3> _haloRegionCellWidth;
```
SamNewcome:
```suggestion
  std::array<size_t, 3> _haloRegionCellWidth{};
```

src/autopas/containers/hierarchicalGrid/traversals/HGC08SingleLevelTraversal.h
SamNewcome, line 1: This "Traversal" inherits ultimately from `TraversalInterface` but is not an actual traversal and should not be used like one. Whilst a user can't currently construct and try to use this as a traversal, you run the risk of a misinformed developer from trying to do this, or some other mistake arising from this being the only traversal that should not be used like an actual traversal.

My preferred solution is to simply merge this into `HGFitC08Traversal`. I'm not sure you really need an extra class and it would make things a lot easier. The unique parts of this traversal i.e. `_cellHandler` and `_haloRegionCellWidth` can simply be set/created inside the `HGFitC08Traversal::traverseParticles` function.

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
```diff
@@ -0,0 +1,234 @@
...
+ * This class provides the base for traversals using the hgc08 base step.
+ *
+ * The base step processBaseCell() computes one set of pairwise interactions
+ * between two cells for each spatial direction based on the baseIndex. It also handles top-down the inter-level
+ * interactions in the same spatial region.
+ *
+ * @tparam ParticleCell the type of cells
+ * @tparam PairwiseFunctor The functor that defines the interaction of two particles.
+ */
+template <class ParticleCell, class PairwiseFunctor>
```
SamNewcome:
```suggestion
template <class ParticleCell_T, class PairwiseFunctor_T>
```

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
```diff
@@ -0,0 +1,234 @@
...
+   * @param overlap number of overlapping cells in each direction as result from cutoff and cellLength.
+   * @param dataLayout The data layout with which this traversal should be initialized.
+   * @param useNewton3 Parameter to specify whether the traversal makes use of newton3 or not.
+   * @param cellBlocks The cell blocks of the hierarchical grid, ordered from lowest to highest level.
+   * @param interactionLengthsSquared The squared interaction lengths from each lower level to upperLevel, ordered
+   * from lowest to highest level.
+   * @param upperLevel The upper level of the hierarchical grid
+   * @param fittedGrids Whether the grids of the hierarchical grid are fitted to each other.
+   * @todo Pass cutoff to _cellFunctor instead of interactionLength, unless this functor is used to build verlet-lists,
+   * in that case the interactionLength is needed!
```
SamNewcome: Modify this todo:
Would it be fine to just use `cutoff` instead of `interactionLength`? If not totally sure, I am okay with you just leaving this with `interactionLength` with a todo to check if this would work with `cutoff`.

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
```diff
@@ -0,0 +1,234 @@
...
+   * interactions between cell1 and the lower-level cells. The lower-level cells are determined by the bounding box of
+   * cell2.
+   * @param cell1 The upper-level cell to interact with.
+   * @param cellIndex1 The index of the first upper-level cell.
+   * @param cellIndex2 The index of the second upper-level cell.
+   */
+  void decompose2AndProcessCells(ParticleCell &cell1, const size_t cellIndex1, const size_t cellIndex2);
+};
+
+template <class ParticleCell, class PairwiseFunctor>
```
SamNewcome: Add postfix `_T`

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
```diff
@@ -0,0 +1,234 @@
...
+   * The cell blocks of the hierarchical grid, ordered from lowest to highest level.
+   */
+  std::vector<internal::CellBlock3D<ParticleCell> *> _cellBlocks;
+  /**
+   * The squared interaction lengths from each lower level to upperLevel, ordered from lowest to highest level.
+   */
+  std::vector<double> _interactionLengthsSquared;
+  /**
+   * Shift length, which is smaller than the smallest cell width.
+   */
```
SamNewcome, line 96: Add more to the documentation, as I'm confused by this:

- What is this used for?
- By "smaller than the smallest cell width", what do you mean? Is this expected a close bound? Is this simply any length between 0 and the smallest cell width?

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
```diff
@@ -0,0 +1,234 @@
...
+  if (_fittedGrids) {
+    // Use shifted corners so exact boundary points do not spill into the neighboring cell.
+    highCornerCell2 = {highCornerCell2[0] - _shiftLength[0], highCornerCell2[1] - _shiftLength[1],
+                       highCornerCell2[2] - _shiftLength[2]};
+
+    lowCornerCell2 = {lowCornerCell2[0] + _shiftLength[0], lowCornerCell2[1] + _shiftLength[1],
+                      lowCornerCell2[2] + _shiftLength[2]};
+  } else {
+    // Replace upper corner with lower corner of the diagonally higher corner. Otherwise we might get ambiguous
+    // boundaries, because high and next low corner dont always give the same result, because of floating point errors.
```
SamNewcome:
```suggestion
    // Replace the upper corner with the lower corner of the diagonally higher corner. Otherwise, we might get ambiguous
    // boundaries, because the high and next low corners don't always give the same result, because of floating point errors.
```

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
```diff
@@ -0,0 +1,234 @@
...
+  const auto [lowCornerCell1, highCornerCell1] = _cellBlocks[_upperLevel]->getCellBoundingBox(cellIndex1);
+
+  // decompose for every level below upperLevel
+  for (size_t lowerLevel = 0; lowerLevel < _upperLevel; lowerLevel++) {
+    auto startIndex3D = _cellBlocks[lowerLevel]->get3DIndexOfPosition(lowCornerCell2);
+    auto stopIndex3D = _cellBlocks[lowerLevel]->get3DIndexOfPosition(highCornerCell2);
+
+    // if the grids are not fitted, we count lower cells only to the decomposition of an upper cell, if their center
+    // lies within, to prevent considering the same pair twice or not at all.
+    if (!_fittedGrids) {
```
SamNewcome:
```suggestion
    if (not _fittedGrids) {
```

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
```diff
@@ -0,0 +1,234 @@
...
+      // Calculate the cell centers of the lower-level start and stop index. If they lie outside of the bounds of the
+      // upper cell, we shift the start/stop index by one.
+      const auto [low, high] = _cellBlocks[lowerLevel]->getCellBoundingBox(startIndex3D);
+      std::array<double, 3> startCellCenter = {0.5 * (low[0] + high[0]), 0.5 * (low[1] + high[1]),
+                                               0.5 * (low[2] + high[2])};
```
SamNewcome:
```suggestion
      std::array<double, 3> startCellCenter = 0.5 * (low + high);
```

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
```diff
@@ -0,0 +1,234 @@
...
+  for (size_t lowerLevel = 0; lowerLevel < _upperLevel; lowerLevel++) {
+    auto startIndex3D = _cellBlocks[lowerLevel]->get3DIndexOfPosition(lowCornerCell2);
+    auto stopIndex3D = _cellBlocks[lowerLevel]->get3DIndexOfPosition(highCornerCell2);
+
+    // if the grids are not fitted, we count lower cells only to the decomposition of an upper cell, if their center
+    // lies within, to prevent considering the same pair twice or not at all.
+    if (!_fittedGrids) {
+      // Calculate the cell centers of the lower-level start and stop index. If they lie outside of the bounds of the
+      // upper cell, we shift the start/stop index by one.
+      const auto [low, high] = _cellBlocks[lowerLevel]->getCellBoundingBox(startIndex3D);
```
SamNewcome:
```suggestion
      const auto [startLow, startHigh] = _cellBlocks[lowerLevel]->getCellBoundingBox(startIndex3D);
```
-> to match "stop"

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
```diff
@@ -0,0 +1,234 @@
...
+      std::array<double, 3> startCellCenter = {0.5 * (low[0] + high[0]), 0.5 * (low[1] + high[1]),
+                                               0.5 * (low[2] + high[2])};
+      const auto [stopLow, stopHigh] = _cellBlocks[lowerLevel]->getCellBoundingBox(stopIndex3D);
+      std::array<double, 3> stopCellCenter = {0.5 * (stopLow[0] + stopHigh[0]), 0.5 * (stopLow[1] + stopHigh[1]),
+                                              0.5 * (stopLow[2] + stopHigh[2])};
```
SamNewcome:
```suggestion
      std::array<double, 3> stopCellCenter = 0.5 * (stopLow + stopHigh);
```

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
```diff
@@ -0,0 +1,234 @@
...
+      }
+      if (stopCellCenter[0] >= highCornerCell2[0] && stopIndex3D[0] > 0) {
+        stopIndex3D[0]--;
+      }
+      if (stopCellCenter[1] >= highCornerCell2[1] && stopIndex3D[1] > 0) {
+        stopIndex3D[1]--;
+      }
+      if (stopCellCenter[2] >= highCornerCell2[2] && stopIndex3D[2] > 0) {
+        stopIndex3D[2]--;
+      }
```
SamNewcome: Refactor into one or two `for` loops.
Also: I think this only works if you assume you never need to do this adjustment across periodic boundaries. This is a fair assumption that I would just write it down somewhere.

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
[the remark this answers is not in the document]
SamNewcome: I would write this down in a comment as it will make debugging easier in case somebody breaks this assumption.

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
```diff
@@ -0,0 +1,234 @@
...
+    }
+    for (size_t z = startIndex3D[2]; z <= stopIndex3D[2]; ++z) {
+      for (size_t y = startIndex3D[1]; y <= stopIndex3D[1]; ++y) {
+        for (size_t x = startIndex3D[0]; x <= stopIndex3D[0]; ++x) {
+          // Check if the cells are within range of one another
+          auto [lowCornerLowerCell, highCornerLowerCell] = _cellBlocks[lowerLevel]->getCellBoundingBox({x, y, z});
+          if (lowerLevel < _interactionLengthsSquared.size()) {
```
SamNewcome: I don't quite get this `if`-statement. Surely this should always be true?

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
```diff
@@ -0,0 +1,234 @@
...
+  // get bounds of  cell2
+  auto [lowCornerCell2, highCornerCell2] = _cellBlocks[_upperLevel]->getCellBoundingBox(cellIndex2);
+
+  if (_fittedGrids) {
+    // Use shifted corners so exact boundary points do not spill into the neighboring cell.
+    highCornerCell2 = {highCornerCell2[0] - _shiftLength[0], highCornerCell2[1] - _shiftLength[1],
+                       highCornerCell2[2] - _shiftLength[2]};
+
+    lowCornerCell2 = {lowCornerCell2[0] + _shiftLength[0], lowCornerCell2[1] + _shiftLength[1],
+                      lowCornerCell2[2] + _shiftLength[2]};
```
SamNewcome: I'm not totally sure why you need this logic.

For fitted grids, you use `highCornerCell2` and `lowCornerCell2` only to get the indicies `startIndex3D` and `stopIndex3D` -> but you can just get these directly from the `cellIndex` with `utils::ThreeDimensionalMapping`. `CellBlock3D` also has a `oneToThreeD` member which could do this even more smoothly. This is currently a `private` member but honestly I see no reason why this couldn't be changed to `public`.

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
SamNewcome: May be an irrelevant comment: check https://github.com/AutoPas/AutoPas/pull/1135/changes#r3129671361

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
```diff
@@ -0,0 +1,234 @@
...
+          // @todo: Could potentially calculate and use sorting direction in the future
+          auto &lowerCell = _cellBlocks[lowerLevel]->getCell({x, y, z});
+          this->_cellFunctor.processCellPair(cell1, lowerCell, {0., 0., 0.});
+        }
+      }
+    }
+  }
+}
+
+template <class ParticleCell, class PairwiseFunctor>
```
SamNewcome: Postfix `_T`

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
```diff
@@ -0,0 +1,234 @@
...
+inline void HGC08CellHandler<ParticleCell, PairwiseFunctor>::processBaseCell(size_t baseIndex) {
+  for (auto const &[offset1, offset2, r] : this->_cellPairOffsets) {
+    const unsigned long cellIndex1 = baseIndex + offset1;
+    const unsigned long cellIndex2 = baseIndex + offset2;
+
+    auto &cell1 = _cellBlocks[_upperLevel]->getCell(cellIndex1);
+    auto &cell2 = _cellBlocks[_upperLevel]->getCell(cellIndex2);
+    // if both cells are halo cells, we can skip the interaction,
+    if (cell1.getPossibleParticleOwnerships() == OwnershipState::halo &&
+        cell2.getPossibleParticleOwnerships() == OwnershipState::halo) {
```
SamNewcome:
```suggestion
    if (cell1.getPossibleParticleOwnerships() == OwnershipState::halo and
        cell2.getPossibleParticleOwnerships() == OwnershipState::halo) {
```

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
```diff
@@ -0,0 +1,234 @@
...
+    } else {
+      this->_cellFunctor.processCellPair(cell1, cell2, r);
+    }
+    // Inter-level only top-down
+    if (_upperLevel == 0) {
+      continue;
+    }
+
+    // Inter Level
+    if (!cell1.isEmpty()) {
```
SamNewcome:
```suggestion
    if (not cell1.isEmpty()) {
```

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
```diff
@@ -0,0 +1,234 @@
...
+      decompose2AndProcessCells(cell1, cellIndex1, cellIndex2);
+    }
+    if (cellIndex1 != cellIndex2 && !cell2.isEmpty()) {
```
SamNewcome:
```suggestion
    if (cellIndex1 != cellIndex2 and not cell2.isEmpty()) {
```

src/autopas/containers/hierarchicalGrid/traversals/HGBlockTraversal.h
```diff
@@ -158,10 +163,16 @@ class HGBlockTraversal : public HGTraversalBase<ParticleCell_T>, public HGTraver
    * @return A new traversal that is applicable to a specific LinkedCells level
    */
   std::unique_ptr<TraversalInterface> generateNewTraversal(const size_t level) override {
+    // Instead of LCC08Traversal, we use HGC08SingleLevelTraversal with upperLevel=0 for the intra-level interactions.
+    // This way it works the same, but considers the wider halo region to skip unnecessary halo-halo interactions.
```
SamNewcome: I really like this. I would add to the `HGC08SingleLevelTraversal` and `HGC08CellHandler` documentation about this so this is a clearer use case, including the `upperLevel` parameter. 

What may be cleaner still, is adding a bool flag e.g. `intraLevelOnly` to both `HGC08SingleLevelTraversal` and `HGC08CellHandler`. I'm not too bothered about this but I do think it would be better.

src/autopas/containers/linkedCells/traversals/LCC08CellHandler.h
```diff
@@ -88,7 +88,7 @@ class LCC08CellHandler {
    */
   bool _useNewton3;
 
- private:
+ protected:
```
SamNewcome: I'm happy for the `_cellFunctor` to be protected, but make sure `_interactionLength` and `_cellLength` are still `private`

src/autopas/containers/linkedCells/LinkedCells.h
```diff
@@ -67,6 +67,26 @@ class LinkedCells : public CellBasedParticleContainer<FullParticleCell<Particle_
...
+   * By default all applicable traversals are allowed.
+   */
+  LinkedCells(const std::array<double, 3> &boxMin, const std::array<double, 3> &boxMax, const double cutoff,
+              const double skin, const std::array<unsigned long, 3> &cellsPerDimension,
+              const size_t sortingThreshold = 8,
+              LoadEstimatorOption loadEstimator = LoadEstimatorOption::squaredParticlesPerCell)
+      : CellBasedParticleContainer<ParticleCellType>(boxMin, boxMax, cutoff, skin, sortingThreshold),
+        _cellBlock(this->_cells, boxMin, boxMax, cutoff + skin, cellsPerDimension),
+        _loadEstimator(loadEstimator) {}
+
```
SamNewcome, line 89: Does this ever get used? My CLion can't find any uses of this.

src/autopas/containers/CellBlock3D.h
```diff
@@ -425,6 +457,79 @@ inline void CellBlock3D<ParticleCell>::rebuild(std::vector<ParticleCell> &vec, c
   }
 }
 
+template <class ParticleCell>
+inline void CellBlock3D<ParticleCell>::rebuild(std::vector<ParticleCell> &vec, const std::array<double, 3> &bMin,
```
SamNewcome: As far as I am aware, this should be the same as the other `rebuild`, but instead of `cellSizeFactor`, the number of `cellsPerDim` is given. I would really just make one function, to avoid code duplication. There are also some small differences, which I think are fine for your case but could lead to mistakes in the future. One example is that this function assumes one cell per interaction length, which may not be correct.

My suggestion is to only use the original function, and in the constructor for `CellBlock3D`, determine the relevant cell size factor needed to get the appropriate cells per dimension. A variation of this is to only use this variant, and in the CSF constructor of `CellBlock3D` determine the number of cells per dimension from the cell size factor.

src/autopas/containers/CompatibleTraversals.h
```diff
@@ -144,6 +144,16 @@ std::set<TraversalOption> filterAllOptions(const std::string &prefix, const Inte
   return s;
 }
 
+/**
+ * Lists all traversal options applicable for the HierarchicalGridFitted container.
+ * @return set of all applicable traversal options.
+ */
+[[maybe_unused]] static const std::set<TraversalOption> &allHGFCompatibleTraversals() {
+  // Can use all hgrid traversals and all hgridFit traversals
+  static const auto s = filterAllOptions("hgrid", InteractionTypeOption::pairwise);
```
SamNewcome: I don't love this solution, as I think it's prone to problems in the future.

Two ideas:

- for `allHGCompatibleTraversals`, get two sets: `s1 = filterAllOptions("hgrid_")` and `s2 = filterAllOptions("hgridFit_")` and then merge them.
- Simply manually list the traversals. I don't think automatic is necessary, especially for how few HG traversals there are.

SamNewcome (changes requested):
This generally looks nice.

There's two more major issues:

1. Decide what you want to do about non-fitted C08. My suggestion is to remove this, unless you foresee some case where it might prove significantly beneficial. If you want to keep it, make it actually usable by the user but otherwise there's not really any issues.

2. Add tests. There are basically no new tests and this is the biggest problem with this PR. Particularly relevant are 
- tests for the new helper classes e.g. the Cell Handler. 
- Tests that the new `CellBlock3D` constructor matches the old one for given domain dimensions/interaction length/cell size factor/etc

In general, I would just go over the PR and for anything added, ask yourself what could be broken with it and what could somebody in the future break - and then write a test to check that. In many ways, the main point of tests is to stop somebody from breaking your code in the future.

In my opinion, the tests are the most important thing to be added right now; everything else is much more minor.

src/autopas/containers/linkedCells/LinkedCells.h
AlexanderGlas: Yes, it is used when creating fitted Hierarchical Grids. It is necessary to ensure that the grids are always fully aligned in all dimensions, even when stretching the cells to fit a non-cubic box.

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
AlexanderGlas: The original indices are in the upper-level grid, whereas startIndex3D and stopIndex3D are indices in the lower-level grids, which is why you can't simply do the threeDimensionalMapping. I also considered doing the three-dimensional mapping and using that to calculate the lower-level indices, which should be easy since the grids are fitted. However, handling the halo-cells makes this somewhat messy and error-prone, which is why I thought it would be cleaner to use the get3DIndexOfPosition, etc. functions, as those essentially perform the same arithmetic anyway.

src/autopas/containers/CellBlock3D.h
AlexanderGlas: Sounds good. I would go with the second option because, I think,  there are cases where non-cubic boxes require more than just choosing CSF to guarantee that the cells are fitted in all dimensions.

src/autopas/containers/hierarchicalGrid/traversals/HGC08SingleLevelTraversal.h
AlexanderGlas: I also considered combining the two; however, I was worried about weird inheritance issues when inheriting from both HGTraversalBase and ColorBasedTraversal, since they both inherit from TraversalInterface. Is that fine in this case, or how should I handle this?

src/autopas/containers/linkedCells/LinkedCells.h
SamNewcome: Okay, sounds good!

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
SamNewcome: Yeah this makes sense

src/autopas/containers/hierarchicalGrid/traversals/HGC08SingleLevelTraversal.h
SamNewcome: Yes, that's a good point. What I suggested wouldn't work.

I still want a solution for this. Looking at `HGTraversalBase`, almost all the code is specific to `HGBlockTraversal`. In addition, most of the members inherited from `TraversalInterface` and implemented in this class, have implementations specific to `HGBlockTraversal`.

Therefore, I think you can remove the inheritance of `TraversalInterface` from `HGTraversalBase`, add it to `HGBlockTraversal`. You can then move all of the `HGBlockTraversal` specific code from `HGTraversalBase` to `HGBlockTraversal`. Most of the inherited functions from `TraversalInterface` implemented in `HGTraversalBase` are `HGBlockTraversal`-specific and I think those functions that are not `HGBlockTraversal`-specific are minor/small enough that they can be implemented twice; in `HGBlockTraversal` and `HGFitC08Traversal`. 

I think this is also much neater in general, even ignoring this diamond inheritance issue.

You are then free to merge `HGFitC08Traversal` and `HGC08SingleLevelTraversal`.

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
AlexanderGlas: If I understand it correctly, then periodic boundaries are handled by copying particles into halo cells on the other side. So we don't need to match cells with cells on the other side; interacting with the halo cells is enough. If that is correct, then this logic should work just fine, even with periodic boundaries. However, I can, of course, be mistaken.

SamNewcome: Yes - my point was more write this down as a comment, for future reference

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
AlexanderGlas: Yes, any length between 0 and the smallest cell width should theoretically work. (I extended the documentation now.)

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
AlexanderGlas: Yes, this was originally intended to omit range checks when the levels are too close in size; however, testing showed that these range checks are always worth it, so I have now removed this check.

src/autopas/containers/hierarchicalGrid/traversals/HGC08SingleLevelTraversal.h
AlexanderGlas: I added the static_cast, however ceil should still be the appropriate operation. haloRegionWidth is simply the largest interaction length of the hgrid (I also now renamed it to largestInteractionLength). This is also how the halo width is calculated in CellBlock3D.

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
AlexanderGlas: I think the erase function needs an iterator. Or can integers automatically be cast to iterators?

src/autopas/containers/CompatibleTraversals.h
AlexanderGlas: I went with the second option.

src/autopas/containers/hierarchicalGrid/traversals/HGC08CellHandler.h
AlexanderGlas: I left this with interactionLength and changed the todo to: "Check if passing the cutoff instead of the interaction length to the cell functor works."

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
AlexanderGlas: I added the helper function in CellBlock3D called getCellsPerDimensionWithoutHalo(). I renamed cellsPerDimensionHigher to cellsPerDimensionNextHighest because it is always the value of the next-highest level, which we have to fit to.

src/autopas/containers/hierarchicalGrid/traversals/HGFitC08Traversal.h
AlexanderGlas: I have moved all the intra-level logic from TraversalBase to BlockTraversal, including generateNewTraversal, initTraversal, endTraversal, the _traversals member, and computeIntraLevelInteractions. 
For now, this makes sense, but I could imagine future traversals also using intra-level traversals, which might want to reuse this logic somehow?

src/autopas/containers/hierarchicalGrid/traversals/HGC08SingleLevelTraversal.h
AlexanderGlas: I have now implemented this and moved most of the code only used by HGBlock into the HGBlock file. This required some adaptations to colorTraversal, because we need to adapt the grid sizes for every level of the traversal.

src/autopas/containers/hierarchicalGrid/traversals/HGBlockTraversal.h
AlexanderGlas: I added an intraLevel parameter, which when set indicates to only traverse this one level

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
AlexanderGlas: I removed it along with all the non-fitted HGc08 logic. I kept it in an archived pull request and put the link as a comment in the code.

src/autopas/containers/hierarchicalGrid/traversals/HGFitC08Traversal.h
SamNewcome: Never did come back to this. I think this is good enough, just leave it.

src/autopas/containers/cellTraversals/ColorBasedTraversal.h
```diff
@@ -108,18 +108,35 @@ class ColorBasedTraversal : public CellTraversal<ParticleCell>, public Traversal
...
+   */
+  void changeGrid(const double interactionLength, const std::array<double, 3> &cellLength,
+                  const std::array<unsigned long, 3> &cellsPerDimension) {
+    _interactionLength = interactionLength;
+    _cellLength = cellLength;
+    this->_cellsPerDimension = cellsPerDimension;
+    for (unsigned int d = 0; d < 3; d++) {
+      _overlap[d] = std::ceil(_interactionLength / _cellLength[d]);
+    }
+  }
```
SamNewcome, line 138: I would move this to `HGFitC08Traversal`, as it modifies `protected` members, so should be fine.

src/autopas/containers/hierarchicalGrid/traversals/HGBlockTraversal.h
```diff
@@ -147,6 +149,22 @@ class HGBlockTraversal : public HGTraversalBase<ParticleCell_T>, public HGTraver
...
   };
 
+  void initTraversal() override {
+    this->_traversals.reserve(this->_numLevels);
+    for (size_t level = 0; level < this->_numLevels; ++level) {
+      this->_traversals.emplace_back(std::make_unique<HGFitC08Traversal<ParticleCell_T, Functor_T>>(
+          &_functor, this->_numLevels, this->_dataLayout, this->_useNewton3, level));
+      this->_traversals[level]->setLevels(this->_levels, this->_maxCutoffPerLevel, this->_skin);
+    }
+    // HGFitC08Traversal is an Hgrid traversal which initializes the data layout for all levels
```
SamNewcome, line 159: This comment doesn't make much sense here.

src/autopas/containers/hierarchicalGrid/traversals/HGBlockTraversal.h
```diff
@@ -147,6 +149,22 @@ class HGBlockTraversal : public HGTraversalBase<ParticleCell_T>, public HGTraver
...
+    this->_traversals[0]->initTraversal();
+  }
+
+  void endTraversal() override {
+    // HGFitC08Traversal is an Hgrid traversal which stores the data layout for all levels
```
SamNewcome, line 164: Similar

src/autopas/containers/hierarchicalGrid/traversals/HGFitC08Traversal.h
SamNewcome: Yes I imagine that too. My suggestion in that case (not needed for this PR) would be to keep roughly the same structure, but then one base class for intra-level independent traversals (which would look very similar to Atacan's original `HGTraversalBase`, which was a base for several intra-level independent traversals) and another for combined traversals

src/autopas/containers/hierarchicalGrid/traversals/HGFitC08Traversal.h
```diff
@@ -35,18 +46,33 @@ class HGFitC08Traversal : public HGTraversalBase<ParticleCell_T>, public HGTrave
...
   /**
    * Constructor of the hgridFit_c08 traversal.
-   * @param functor The functor that defines the interaction of two particles.
+   * @param pairwiseFunctor The functor that defines the interaction of two particles.
    * @param numLevels Number of levels in the hierarchical grid.
    * @param dataLayout The data layout with which this traversal should be initialized.
    * @param useNewton3 Parameter to specify whether the traversal makes use of newton3 or not.
+   * @param intraLevel Non-negative integer indicates to only calculate intra-level interactions for that specific
+   * level. Works just like LCC08Traversal, but considers the potentially larger halo regions of HierarchicalGrid levels
+   * to skip unnecessary halo-halo interactions.
```
SamNewcome, line 55: Using numbers to mean other things is often a source of bugs. Here I would recommend something like an `std::optional<size_t>` => If the number is used it is guaranteed by the type to be suitable for indices i.e. `size_t`, and the `optional` makes the optional nature a lot more obvious.

src/autopas/containers/hierarchicalGrid/HierarchicalGrid.h
SamNewcome: Sounds reasonable!

tests/testAutopas/tests/containers/hierarchicalGrid/HGFittedTest.cpp
SamNewcome, line 1: Add class + function documentation. In addition, document each test by explaining briefly what it tests and how.

tests/testAutopas/tests/containers/hierarchicalGrid/HGFittedTest.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+  if (haloLine.rfind(haloPrefix, 0) != 0) {
+    throw std::runtime_error("Unexpected halo line: " + haloLine);
+  }
+  level.cellsPerDimensionWithHalo = parseArray<std::size_t, 3>(extractAfterPrefix(haloLine, haloPrefix));
+
+  return level;
+}
+
+std::array<double, 3> haloWidth(const ParsedLevel &level) {
```
SamNewcome, line 106:
```suggestion
std::array<double, 3> getHaloWidth(const ParsedLevel &level) {
```
Functions do things -> name should have the do-ing word for what it does

tests/testAutopas/tests/containers/hierarchicalGrid/HGFittedTest.cpp
```diff
@@ -0,0 +1,233 @@
...
+      throw std::runtime_error("Unexpected end of hierarchy dump while parsing level information.");
+    }
+
+    levels.push_back(parseLevel(cellLine, interactionLine, haloLine));
+  }
+
+  return levels;
+}
+
+std::size_t cellsPerDimensionWithoutHalo(const ParsedLevel &level, std::size_t dim) {
```
SamNewcome, line 136:
```suggestion
std::size_t getCellsPerDimensionWithoutHalo(const ParsedLevel &level, std::size_t dim) {
```

tests/testAutopas/tests/containers/hierarchicalGrid/HGFittedTest.cpp
```diff
@@ -0,0 +1,233 @@
...
+    std::vector<double> cutoffs;
+    double skin;
+    double cellSizeFactor;
+  };
+
+  const std::vector<TestCase> testCases{
+      {{0., 0., 0.}, {60., 60., 60.}, {0.5, 1.0, 2.0, 4.0}, 0.25, 1.0},
+      {{-3., 2., 5.}, {33., 50., 65.}, {0.75, 1.5, 3.0}, 0.5, 2.0},
+      {{1., 2., 3.}, {25., 32., 41.}, {1.0, 1.4}, 0.1, 0.5},
+  };
```
SamNewcome, line 203: Fyi, it is typically nicer to use GTest's parameterised tests. But I'm okay with this for the PR

SamNewcome: Add explanation for what each test case tests.

tests/testAutopas/tests/containers/hierarchicalGrid/HGFittedTest.h
```diff
@@ -0,0 +1,18 @@
...
+
+#include <array>
+#include <string>
+
+#include "AutoPasTestBase.h"
+
+class HGFittedTest : public AutoPasTestBase {
+ public:
+  static void expectFittedHierarchy(const std::string &hierarchyDump, const std::array<double, 3> &boxMin,
+                                    const std::array<double, 3> &boxMax);
```
SamNewcome, line 17: This function is never used

tests/testAutopas/tests/containers/hierarchicalGrid/HGFittedTest.cpp
```diff
@@ -0,0 +1,233 @@
...
+  return levels;
+}
+
+std::size_t cellsPerDimensionWithoutHalo(const ParsedLevel &level, std::size_t dim) {
+  return level.cellsPerDimensionWithHalo[dim] - 2 * level.cellsPerInteractionLength;
+}
+
+}  // namespace
+
+void HGFittedTest::expectFittedHierarchy(const std::string &hierarchyDump, const std::array<double, 3> &boxMin,
```
SamNewcome, line 142: This function is never used

tests/testAutopas/tests/containers/hierarchicalGrid/HGFittedTest.cpp
```diff
@@ -0,0 +1,233 @@
...
+      SCOPED_TRACE(::testing::Message() << "levels " << levelIndex << " and " << levelIndex + 1);
+
+      const auto &finer = levels[levelIndex];
+      const auto &coarser = levels[levelIndex + 1];
+
+      for (std::size_t dim = 0; dim < 3; ++dim) {
+        const auto finerCellsWithoutHalo = cellsPerDimensionWithoutHalo(finer, dim);
+        const auto coarserCellsWithoutHalo = cellsPerDimensionWithoutHalo(coarser, dim);
+
+        ASSERT_GT(coarserCellsWithoutHalo, 0u);
```
SamNewcome, line 227: Why assert?

tests/testAutopas/tests/containers/hierarchicalGrid/HGFittedTest.cpp
SamNewcome, line 1: I find the parsing logic a bit over complicated. Just add getters.

Avoid adding a direct getter to the `cellBlock`s themselves -> but to the statistics outputted in `HierarchicalGrid::toString()`, I think there are no issues.

tests/testAutopas/tests/containers/linkedCells/LinkedCellsConstructorTest.h
SamNewcome, line 1: Add documentation to classes + functions.

tests/testAutopas/tests/containers/linkedCells/LinkedCellsConstructorTest.cpp
SamNewcome, line 1: Add documentation to classes + functions.

SamNewcome: Add postfix `_T` to all template parameters

tests/testAutopas/tests/containers/linkedCells/LinkedCellsConstructorTest.cpp
```diff
@@ -0,0 +1,236 @@
...
+    for (const auto &particle : cell) {
+      particleData.emplace_back(particle.getID(), particle.getOwnershipState());
+    }
+    std::sort(particleData.begin(), particleData.end(), [](const auto &a, const auto &b) {
+      const auto [idA, stateA] = a;
+      const auto [idB, stateB] = b;
+      if (idA != idB) {
+        return idA < idB;
+      }
+      return toInt64(stateA) < toInt64(stateB);
```
SamNewcome, line 82: I don't really get why this is here. In theory, there should never be two particles with the same ID, so this should never occur.

tests/testAutopas/tests/containers/linkedCells/LinkedCellsConstructorTest.cpp
```diff
@@ -0,0 +1,236 @@
...
+  std::array<unsigned long, 3> cellsPerDim{};
+  for (size_t d = 0; d < 3; ++d) {
+    const auto boxLength = boxMax[d] - boxMin[d];
+    cellsPerDim[d] =
+        std::max(static_cast<unsigned long>(std::floor(boxLength / (interactionLength * cellSizeFactor))), 1ul);
+  }
+  return cellsPerDim;
+}
+
+std::vector<std::array<double, 3>> LinkedCellsConstructorTest::representativePositions(
```
SamNewcome, line 104:
```suggestion
std::vector<std::array<double, 3>> LinkedCellsConstructorTest::getRepresentativePositions(
```

tests/testAutopas/tests/containers/linkedCells/LinkedCellsConstructorTest.cpp
```diff
@@ -0,0 +1,236 @@
...
+      {std::nextafter(boxMin[0], -std::numeric_limits<double>::infinity()), center[1], center[2]},
+      {std::nextafter(boxMax[0], -std::numeric_limits<double>::infinity()), center[1], center[2]},
+      {boxMax[0], boxMax[1], boxMax[2]},
+      {std::nextafter(boxMax[0], std::numeric_limits<double>::infinity()), center[1], center[2]},
+      {center[0], boxMin[1], boxMax[2]},
+      {boxMax[0], center[1], boxMin[2]},
+  };
+}
+
+std::vector<ParticleFP64> LinkedCellsConstructorTest::representativeOwnedParticles(
```
SamNewcome, line 121: -> "get..."

SamNewcome: Also, this function feels really arbitrary. I don't really get it. Maybe with documentation it will be clearer

tests/testAutopas/tests/containers/linkedCells/LinkedCellsConstructorTest.cpp
```diff
@@ -0,0 +1,236 @@
...
+  return {
+      {{boxMin[0] + dx, boxMin[1] + dy, boxMin[2] + dz}, v, 1},
+      {{boxMin[0] + 3. * dx, boxMin[1] + 5. * dy, boxMin[2] + 7. * dz}, v, 2},
+      {{std::nextafter(boxMax[0], -std::numeric_limits<double>::infinity()), boxMin[1] + 2. * dy, boxMin[2] + 8. * dz},
+       v,
+       3},
+  };
+}
+
+std::vector<ParticleFP64> LinkedCellsConstructorTest::representativeHaloParticles(const std::array<double, 3> &boxMin,
```
SamNewcome, line 136: Similar to above

tests/testAutopas/tests/containers/linkedCells/LinkedCellsConstructorTest.cpp
SamNewcome, line 1: The tests in this class test functions independent of the actually used implementations. This means that this test will not fail if somebody breaks what you have done, which largely defeats the value of it.

Refactor this to use the actual functions. You can add getters but, avoid getters which expose more information than necessary e.g. a getter which returns a reference to the actual particle data.

This is also really more a `CellBlock3D` test, so maybe refactor it there. If there is something specific to how Linked Cells uses these functions from `CellBlock3D`, you can add a Linked Cells constructor test that focuses on that.

SamNewcome (changes requested):
Mostly only minor suggestions. 

The most significant issues are with the tests: `HGFittedTest` uses confusing and maybe easily breakable string parsing logic and `LinkedCellsConstructorTest` doesn't test the actual functions used in `src`. In both cases, it seems like these approaches were to get around a lack of getters. Just add getters -> we try to avoid this because it's not really the point of class encapsulation, but ultimately we kinda need this to do meaningful tests in C++. Just avoid being too excessive with the getters.

PR #1142, opened 2026-05-23
Upgrade spd log to v1.17

thesamriel:
# Upgrade spdlog to the newest version

I ran into a bug with logging double values in the globals logger in md-flexible.
Specifically, in DEBUG mode with the Clang compiler.
However, this seems to be an older problem in the `fmt` library that `spdlog` depends on.

I upgraded spdlog, which automatically comes with a newer version of fmt.

The newer versions do not implicitly cast enum values to integer numbers.
Therefore, you either need to cast them specifically or implement a `format_as()` function, which I did for the `Option` class and the `OwnershipState` enum class.


src/autopas/options/Option.h
```diff
@@ -209,5 +209,16 @@ class Option {
...
 };
+
+/**
+ * Function required for modern fmt/spdlog integration (fmt v9+).
+ * @param opt
+ * @return string representation
+ */
+template <typename actualOption>
+inline std::string format_as(const Option<actualOption> &opt) {
+  return opt.to_string();
```
SamNewcome, line 222:
```suggestion
template <typename actualOption_T>
inline std::string format_as(const Option<actualOption_T> &opt) {
```

SamNewcome (changes requested):

src/autopas/options/Option.h
thesamriel: While we're at it, I changed the template parameter name from `actualOption` to `actualOption_T` for the whole `Option` class in [d264479](https://github.com/AutoPas/AutoPas/pull/1142/commits/d264479bace5a6d6aa0a1f757df40c32458f0065)

src/autopas/containers/verletListsCellBased/verletLists/VerletLists.h
```diff
@@ -131,7 +131,7 @@ class VerletLists : public VerletListsLinkedBase<Particle_T> {
       dataLayout = DataLayoutOption::soa;
     } else {
       utils::ExceptionHandler::exception("VerletLists::updateVerletListsAoS(): unsupported BuildVerletListType: {}",
-                                         _buildVerletListType);
+                                         static_cast<int>(_buildVerletListType));
```
manishmishra6016, line 134: Is there a difference in passing an enum directly and converting it to an int?

manishmishra6016 (approved):
Looks good to me!

SamNewcome (approved):

src/autopas/containers/verletListsCellBased/verletLists/VerletLists.h
thesamriel: In the old spdlog/fmt version, the library implicitly cast enums to the underlying integral type and then print that number.
The new library version doesn't do that anymore and would throw a compiler error. So now you either need to provide a `format_as()` method, which I did for our frequently used Option enums. Or alternatively, you need to explicitly cast the enum value to an integral type yourself.

Hope that makes sense :D

PR #1146, opened 2026-05-28
Change functor pointers to references

thesamriel:
# Description

Changes the functor members of all traversals and cell functors to references instead of raw pointers.
These objects require an existing functor to be used. Also, we never included `nullptr` checks anywhere and just assumed an existing functor.
Therefore, it's a logical move imo to change to references instead.


manishmishra6016 (approved):
Looks good to me :)

PR #1148, opened 2026-06-10
Refactor CellFunctor and CellFunctor3B

thesamriel:
# Description

This PR refactors the `CellFunctor` and `CellFunctor3B` to be more consistent and reduce code duplication.
Functionality is the same.
Main changes include
- Update template parameter names to new `_T` convention
- Combine N3 and NoN3 AoS functors to one method.
- *Update: Brought back after review.* Remove the `std::abs()` operation from the sorted cutoff checks. (Our sorting direction is always strictly from cell1 to cell2 and therefore we can save this operation)
- Change the particle ownership check logic slightly.

# How Has This Been Tested?
master commit: https://github.com/AutoPas/AutoPas/commit/6dbbab4b3b6ac32c7e2dcf375134af9e2f8fe356
commit for this PR: https://github.com/AutoPas/AutoPas/pull/1148/commits/ccb4f89befd1203e9815b53d8f5083f54d02eb6a

Build time is slightly faster for the new version, though not really noticeable in the full md-flexible compilation.
(Build Time comparison of a dummy program instantiating only `CellFunctor` and necessary includes. master = 0.58s, this PR = 0.55s)

I ran a constructed performance benchmark for cells with 8 and 64 particles and compared the version in master vs the new version. ( 500000 cells averaged over 3 runs).
The differences are mostly minor and for the low number of particles, the AoSFunctor is better with the new version.

| Configuration | 8 / master | 8 / new | Speedup | 64 / master | 64 / new | Speedup |
|---|---:|---:|---:|---:|---:|---:|
| AoS N3 bidirectional=true pair sorted | 73.3 | 69.4 | 5.2% | 2118.9 | 2060.2 | 2.8% |
| AoS N3 bidirectional=true pair unsorted | 53.6 | 48.5 | 9.5% | 3725.9 | 3736.6 | -0.3% |
| AoS noN3 bidirectional=true pair sorted | 72.9 | 70.8 | 2.9% | 3162.7 | 3174.6 | -0.4% |
| AoS noN3 bidirectional=true pair unsorted | 80.1 | 77.9 | 2.7% | 6503.4 | 6512.0 | -0.1% |
| AoS noN3 bidirectional=false pair sorted | 72.4 | 69.3 | 4.3% | 1903.0 | 1886.5 | 0.9% |
| AoS noN3 bidirectional=false pair unsorted | 53.4 | 50.5 | 5.4% | 3508.4 | 3504.7 | 0.1% |
| AoS N3 single cell | 110.5 | 112.1 | -1.5% | 5771.4 | 5919.3 | -2.6% |
| AoS noN3 single cell | 180.3 | 178.6 | 0.9% | 10482.5 | 10714.5 | -2.2% |
| SoA N3 bidirectional=true pair unsorted | 63.1 | 60.4 | 4.3% | 4061.2 | 4087.4 | -0.6% |
| SoA noN3 bidirectional=true pair unsorted | 113.4 | 109.5 | 3.5% | 7802.1 | 7818.3 | -0.2% |
| SoA noN3 bidirectional=false pair unsorted | 60.3 | 57.2 | 5.1% | 3926.5 | 3931.2 | -0.1% |
| SoA N3 single cell | 88.7 | 89.9 | -1.4% | 5313.4 | 5407.3 | -1.8% |
| SoA noN3 single cell | 88.5 | 89.3 | -0.9% | 5346.1 | 5406.1 | -1.1% |

### DIFF
--- a/src/autopas/baseFunctors/CellFunctor.h
+++ b/src/autopas/baseFunctors/CellFunctor.h
@@ -87,7 +87,7 @@ class CellFunctor {
    */
   [[nodiscard]] bool shouldUseSorting(size_t particleCount, const std::array<double, 3> &sortingDirection) const {
     return particleCount >= _sortingThreshold and
-           (sortingDirection[0] != 0.0 and sortingDirection[1] != 0.0 and sortingDirection[2] != 0.0);
+           (sortingDirection[0] != 0.0 or sortingDirection[1] != 0.0 or sortingDirection[2] != 0.0);
   }
 
   /**
--- a/src/autopas/baseFunctors/CellFunctor.h
+++ b/src/autopas/baseFunctors/CellFunctor.h
@@ -138,60 +138,57 @@ void CellFunctor<ParticleCell_T, ParticleFunctor_T, bidirectional>::setSortingTh
 
 template <class ParticleCell_T, class ParticleFunctor_T, bool bidirectional>
 void CellFunctor<ParticleCell_T, ParticleFunctor_T, bidirectional>::processCell(ParticleCell_T &cell) {
-  // avoid force calculations if the cell contains only halo particles or if the cell is empty (=dummy)
+  const bool isAoS = _dataLayout == DataLayoutOption::aos ? true : false;
+  const bool isSoA = _dataLayout == DataLayoutOption::soa ? true : false;
+
+  // Return early if the cell is empty.
+  if ((isSoA and cell._particleSoABuffer.size() == 0) or (isAoS and cell.isEmpty())) {
+    return;
+  }
+  // Avoid force calculations if the cell contains only halo particles or if the cell is empty (=dummy)
   const bool cellHasOwnedParticles = toInt64(cell.getPossibleParticleOwnerships() & OwnershipState::owned);
   if (not cellHasOwnedParticles) {
     return;
   }
 
-  switch (_dataLayout) {
-    case DataLayoutOption::aos:
-      if (cell.isEmpty()) {
-        return;
-      }
-      processCellAoSImpl(cell);
-      break;
-    case DataLayoutOption::soa:
-      if (cell._particleSoABuffer.size() == 0) {
-        return;
-      }
-      _functor.SoAFunctorSingle(cell._particleSoABuffer, _useNewton3);
-      break;
+  if (isAoS) {
+    processCellAoSImpl(cell);
+  } else if (isSoA) {
+    _functor.SoAFunctorSingle(cell._particleSoABuffer, _useNewton3);
   }
 }
 
 template <class ParticleCell_T, class ParticleFunctor_T, bool bidirectional>
 void CellFunctor<ParticleCell_T, ParticleFunctor_T, bidirectional>::processCellPair(
     ParticleCell_T &cell1, ParticleCell_T &cell2, const std::array<double, 3> &sortingDirection) {
-  const bool cell1HasOwnedParticles = toInt64(cell1.getPossibleParticleOwnerships() & OwnershipState::owned);
-  const bool cell2HasOwnedParticles = toInt64(cell2.getPossibleParticleOwnerships() & OwnershipState::owned);
+  const bool isAoS = _dataLayout == DataLayoutOption::aos ? true : false;
+  const bool isSoA = _dataLayout == DataLayoutOption::soa ? true : false;
 
-  // Avoid force calculations if both cells cannot contain owned particles.
-  if (not cell1HasOwnedParticles and not cell2HasOwnedParticles) {
+  // Return early if a cell is empty.
+  if ((isSoA and (cell1._particleSoABuffer.size() == 0 or cell2._particleSoABuffer.size() == 0)) or
+      (isAoS and (cell1.isEmpty() or cell2.isEmpty()))) {
     return;
   }
 
-  // In Newton3==false and functor is not bidirectional, only interactions from cell1 to cell2 are computed. Therefore,
-  // if cell1 cannot contain owned particles, there is nothing useful to do.
-  if constexpr (not bidirectional) {
-    if (not _useNewton3 and not cell1HasOwnedParticles) {
+  const bool cell1HasOwnedParticles = toInt64(cell1.getPossibleParticleOwnerships() & OwnershipState::owned);
+  if (not cell1HasOwnedParticles) {
+    // Nothing to do if cell1 has no owned particles and we don't write to cell2 particles.
+    if constexpr (not bidirectional) {
+      if (not _useNewton3) {
+        return;
+      }
+    }
+    // Nothing to do if both cells cannot have owned particles.
+    const bool cell2HasOwnedParticles = toInt64(cell2.getPossibleParticleOwnerships() & OwnershipState::owned);
+    if (not cell2HasOwnedParticles) {
       return;
     }
   }
 
-  switch (_dataLayout) {
-    case DataLayoutOption::aos:
-      if (cell1.isEmpty() or cell2.isEmpty()) {
-        return;
-      }
-      processCellPairAoSImpl(cell1, cell2, sortingDirection);
-      break;
-    case DataLayoutOption::soa:
-      if (cell1._particleSoABuffer.size() == 0 or cell2._particleSoABuffer.size() == 0) {
-        return;
-      }
-      processCellPairSoAImpl(cell1, cell2);
-      break;
+  if (isAoS) {
+    processCellPairAoSImpl(cell1, cell2, sortingDirection);
+  } else if (isSoA) {
+    processCellPairSoAImpl(cell1, cell2);
   }
 }
 
--- a/src/autopas/baseFunctors/CellFunctor.h
+++ b/src/autopas/baseFunctors/CellFunctor.h
@@ -147,8 +147,7 @@ void CellFunctor<ParticleCell_T, ParticleFunctor_T, bidirectional>::processCell(
     return;
   }
   // Avoid force calculations if the cell contains only halo particles or if the cell is empty (=dummy)
-  const bool cellHasOwnedParticles = toInt64(cell.getPossibleParticleOwnerships() & OwnershipState::owned);
-  if (not cellHasOwnedParticles) {
+  if (not cell.canHaveOwnedParticles()) {
     return;
   }
 
@@ -171,8 +170,7 @@ void CellFunctor<ParticleCell_T, ParticleFunctor_T, bidirectional>::processCellP
     return;
   }
 
-  const bool cell1HasOwnedParticles = toInt64(cell1.getPossibleParticleOwnerships() & OwnershipState::owned);
-  if (not cell1HasOwnedParticles) {
+  if (not cell1.canHaveOwnedParticles()) {
     // Nothing to do if cell1 has no owned particles and we don't write to cell2 particles.
     if constexpr (not bidirectional) {
       if (not _useNewton3) {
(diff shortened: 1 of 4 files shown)


src/autopas/baseFunctors/CellFunctor.h
```diff
@@ -80,44 +80,49 @@ class CellFunctor {
...
-   * The value of _useNewton3 defines whether or whether not to apply the aos version functor in a newton3 fashion or
-   * not:
-   * - if _useNewton3 is true: the aos functor will be applied once for each pair (only i,j), passing newton3=true.
-   * - if _useNewton3 is false: the aos functor will be applied twice for each pair (i,j and j,i), passing
-   * newton3=false.
+   *
+   * @param particleCount
+   * @param sortingDirection
+   * @return
+   */
```
SamNewcome, line 87: Add documentation

src/autopas/baseFunctors/CellFunctor.h
```diff
@@ -80,44 +80,49 @@ class CellFunctor {
...
+  [[nodiscard]] bool shouldUseSorting(size_t particleCount, const std::array<double, 3> &sortingDirection) const {
+    return particleCount >= _sortingThreshold and
+           (sortingDirection[0] != 0.0 and sortingDirection[1] != 0.0 and sortingDirection[2] != 0.0);
```
SamNewcome: Originally we had: `(sortingDirection != std::array<double, 3>{0., 0., 0.})`. I don't think this is the same. E.g. if the sorting direction is (0,0,1), originally we would still sort, but now we won't.

src/autopas/baseFunctors/CellFunctor.h
```diff
@@ -127,107 +132,113 @@ class CellFunctor {
...
     } else {
       if (not p1.isHalo()) {
-        _functor.AoSFunctor(p1, p2, false);
+        f.AoSFunctor(p1, p2, false);
       }
       if (not p2.isHalo()) {
-        _functor.AoSFunctor(p2, p1, false);
+        f.AoSFunctor(p2, p1, false);
       }
     }
```
SamNewcome: Can you compare the compilation cost vs speed upgrade for making this constexpr. My guess is this doubles the compilation time for AoS + Linked Cells algs, but doesn't result in a speedup because we still use a regular not-constexpr `newton3` check inside the actual functor.

src/autopas/baseFunctors/CellFunctor.h
```diff
@@ -127,107 +132,113 @@ class CellFunctor {
...
+  if (cell.size() >= _sortingThreshold) {
+    SortedCellView<ParticleCell_T> cellSorted(cell, utils::ArrayMath::normalize(cell.getCellLength()));
 
     for (auto cellIter1 = cellSorted._particles.begin(); cellIter1 != cellSorted._particles.end(); ++cellIter1) {
       auto &[p1Projection, p1Ptr] = *cellIter1;
       // start inner loop ahead of the outer loop
       for (auto cellIter2 = std::next(cellIter1); cellIter2 != cellSorted._particles.end(); ++cellIter2) {
         auto &[p2Projection, p2Ptr] = *cellIter2;
-        if (std::abs(p1Projection - p2Projection) > _sortingCutoff) {
+        if (p2Projection - p1Projection > _sortingCutoff) {
```
SamNewcome: Can you check the speedup of just this change? You are correct that it isn't needed right now, but I am concerned that this becomes an easily breakable feature and want to see a speed gain.

src/autopas/baseFunctors/CellFunctor.h
```diff
@@ -244,84 +255,59 @@ void CellFunctor<ParticleCell, ParticleFunctor, bidirectional>::processCellAoS(P
...
-    ParticleCell &cell1, ParticleCell &cell2, const std::array<double, 3> &sortingDirection) {
-  if ((cell1.size() + cell2.size() > _sortingThreshold) and (sortingDirection != std::array<double, 3>{0., 0., 0.})) {
-    SortedCellView<ParticleCell> cell1Sorted(cell1, sortingDirection);
-    SortedCellView<ParticleCell> cell2Sorted(cell2, sortingDirection);
+template <class ParticleCell_T, class ParticleFunctor_T, bool bidirectional>
+template <bool newton3>
+void CellFunctor<ParticleCell_T, ParticleFunctor_T, bidirectional>::processCellPairAoSImpl(
+    ParticleCell_T &cell1, ParticleCell_T &cell2, const std::array<double, 3> &sortingDirection) {
+  const auto interactParticles = [&f = this->_functor](auto &p1, auto &p2) {
+    if constexpr (newton3) {
```
SamNewcome: I feel similarly about this `constexpr (newton3)` - but I think in the cell-pair case we basically had this previously, so I don't care as much if you check how important this is.

src/autopas/baseFunctors/CellFunctor.h
```diff
@@ -244,84 +255,59 @@ void CellFunctor<ParticleCell, ParticleFunctor, bidirectional>::processCellAoS(P
...
+  if (shouldUseSorting(cell1.size() + cell2.size(), sortingDirection)) {
+    // Use sorted cell views
+    SortedCellView<ParticleCell_T> cell1Sorted(cell1, sortingDirection);
+    SortedCellView<ParticleCell_T> cell2Sorted(cell2, sortingDirection);
 
     for (auto &[p1Projection, p1Ptr] : cell1Sorted._particles) {
       for (auto &[p2Projection, p2Ptr] : cell2Sorted._particles) {
-        if (std::abs(p1Projection - p2Projection) > _sortingCutoff) {
+        // p2Projection > p1Projection guaranteed by sorting direction
+        if (p2Projection - p1Projection > _sortingCutoff) {
```
SamNewcome: similar

src/autopas/baseFunctors/CellFunctor3B.h
SamNewcome, line 1: I have similar concerns with this as with the 2B case.

SamNewcome (changes requested):
I think this really nicely cleans up the functor!

I do have some concerns:
- There is one case where you introduce a `constexpr` switch which we didn't have before, and I am a bit unsure if this results in a speedup vs the compilation cost increase.
- I am unsure of the speedup of the `std::abs` removal. This feels like something which would be very easy to break in the future. I think, if there is a noticeable advantage of dropping this, I would add some kind of debug-level message which gets logged if your assumption stops holding in the future.

I would like to get @dmartin13 to review this PR because I think you two together are the two who know the Cell Functor the best.

src/autopas/baseFunctors/CellFunctor.h
thesamriel: Good catch, that should be `or` instead of `and`. I had fixed this before but somehow slipped in again.

src/autopas/baseFunctors/CellFunctor3B.h
```diff
@@ -172,159 +172,174 @@ class CellFunctor3B {
...
+        f.AoSFunctor(p2, p1, p3, false);
       }
       if (not p3.isHalo()) {
-        _functor.AoSFunctor(p3, p1, p2, false);
+        f.AoSFunctor(p3, p1, p2, false);
       }
     }
   };
 
   if (cell.size() > _sortingThreshold) {
```
dmartin13: Small question: The sorting condition still uses `cell.size() > _sortingThreshold`, while `shouldUseSorting()` and the corresponding logic in `CellFunctor` use `>=.`

Should this also be changed to` >=` here?

src/autopas/baseFunctors/CellFunctor.h
thesamriel: added

src/autopas/baseFunctors/CellFunctor.h
thesamriel: I tested this and you are correct that templating `newton3` is not really beneficial.
However, the build time was more or less the same. (Instead of two methods, the compiler compiles one templated method for two variants) Essentially the same work.

Anyways, since there was no performance benefit, I removed the templated `newton3`.
See the new PR description.

src/autopas/baseFunctors/CellFunctor.h
thesamriel: Honestly, I couldn't find a noticeable speedup in any simulations I tried. 
Therefore, I put back the `std::abs()`

src/autopas/baseFunctors/CellFunctor3B.h
thesamriel: Good catch about the `>=`. Changed this.
For consistency, using `shouldUseSorting()` would be nice, but the difference is that for a single cell, we don't check the sorting direction, which would cause a little bit of overhead here.

src/autopas/baseFunctors/CellFunctor.h
```diff
@@ -127,107 +127,92 @@ class CellFunctor {
...
-        _functor.AoSFunctor(p1, p2, false);
-      }
-      if (not p2.isHalo()) {
-        _functor.AoSFunctor(p2, p1, false);
-      }
+  const auto interactParticles = [this](auto &p1, auto &p2) {
+    this->_functor.AoSFunctor(p1, p2, this->_useNewton3);
+
+    if (not this->_useNewton3) {
+      this->_functor.AoSFunctor(p2, p1, false);
```
manishmishra6016, line 200: now we no longer check if p1 or p2 are halo like before, is that not needed anymore or was it already redundant before?

manishmishra6016 (commented):
Looks good, I just have one comment.

src/autopas/baseFunctors/CellFunctor.h
```diff
@@ -144,89 +139,68 @@ void CellFunctor<ParticleCell_T, ParticleFunctor_T, bidirectional>::setSortingTh
...
+  if ((isSoA and (cell1._particleSoABuffer.size() == 0 or cell2._particleSoABuffer.size() == 0)) or
+      (isAoS and (cell1.isEmpty() or cell2.isEmpty()))) {
     return;
   }
 
-  // In Newton3==false and functor is not bidirectional, only interactions from cell1 to cell2 are computed. Therefore,
-  // if cell1 cannot contain owned particles, there is nothing useful to do.
-  if constexpr (not bidirectional) {
-    if (not _useNewton3 and not cell1HasOwnedParticles) {
+  const bool cell1HasOwnedParticles = toInt64(cell1.getPossibleParticleOwnerships() & OwnershipState::owned);
```
SamNewcome: Not a huge fan of this. If this is how it is done elsewhere in the code, I'm happy to leave it. But best would be a helper function (maybe in the cell class?)

src/autopas/baseFunctors/CellFunctor.h
```diff
@@ -144,89 +139,68 @@ void CellFunctor<ParticleCell_T, ParticleFunctor_T, bidirectional>::setSortingTh
...
+  if (not cell1HasOwnedParticles) {
+    // Nothing to do if cell1 has no owned particles and we don't write to cell2 particles.
+    if constexpr (not bidirectional) {
+      if (not _useNewton3) {
+        return;
+      }
+    }
+    // Nothing to do if both cells cannot have owned particles.
+    const bool cell2HasOwnedParticles = toInt64(cell2.getPossibleParticleOwnerships() & OwnershipState::owned);
```
SamNewcome: Similar

SamNewcome (changes requested):
Fixes all my concerns - just a minor point about something which reads a bit unclearly.

src/autopas/baseFunctors/CellFunctor.h
thesamriel: I don't think we have a case where owned and halo particles are mixed in a cell, so I don't think the check is needed.
If this case would somehow still happen, then worst case we do redundant functor calls, but nothing breaks. But since this is a rare exception, I'd rather optimize the regular path.

src/autopas/baseFunctors/CellFunctor.h
thesamriel: To my excuse, that's how it was done before as well :D
But I like your suggestion and changed it in [00f31aa](https://github.com/AutoPas/AutoPas/pull/1148/commits/00f31aa0045ea824e1de0f51e249d5ece8cf0268)

SamNewcome (approved):
Perfect! Thanks

manishmishra6016 (approved):

PR #1149, opened 2026-06-11
Replace ZIP dependencies with Git Subtree

SamNewcome:
# Description

Currently, the dependencies we use are bundled as zips. This has a couple issues:

- Updating a dependency requires manually downloading a particular commit or a package from the browser.
- Updating a dependency shows as a commit where one large file is deleted and another is added. My understanding is that this will bloat our repository, even if only one line is changed.
- You need to manually uncompress the dependency if you view the code. 
- We sometimes need to also work on the dependency's code. Making experimental changes is quite painful with the current system. The main example of this already in master is PMT, but I found this particularly painful when developing with LB4OMP recently.

With this PR, we switch these dependencies to Git subtrees. These are like mini repositories within our repository. Specifically, I used squashed versions of these repositories, so without git history. The main advantages of these are:

- Repositories can be viewed directly.
- Updates are no longer add + remove compressed file. I think in the long run this will reduce git bloat.
- AutoPas specific patches are actual commits in these subtrees. When updating the dependency, these will treated like any other merge from upstream.
- When directly experimenting with changing a dependency's code, we can do so directly.

## Misc

- I added documentation in CONTRIBUTING.md about using git subtrees.
- The PMT patch downgraded PMT to C++17 because at the time it was added, AutoPas was C++17. So I removed any C++17 specific modifications. The PMT patch is now a commit.
- To avoid dependencies polluting our GitHub code statistics, I added a `.gitattributes` file which results in everything added in `/libs/` being ignored. It also by default collapses these files when reviewing, so it is clear which dependency files do not need to be reviewed.
- Harmony was packaged without an `examples/` directory, or this was manually removed by us but not documented. We now have a commit which actively prunes this directory.
- We previously used a dedicated ANTLR4 CPP runtime package, which has no dedicated git repo. This means we cannot use Git subtree for a specific cpp only repository. Instead, I used the regular GitHub repository with runtimes for all the supported languages and pruned everything except the cpp runtime, like above.
- ANTLR4 "required" two dependencies which we also shipped. In this work, I found out that these aren't needed (at least anymore), so I removed them.

As described in the CONTRIBUTING.md, if there are commits modifying dependencies, I should point reviewers to them. I would suggest for these viewing the commit themselves:
- 8abdbca1 The PMT patch applied as a commit.
- e779169e The pruning of `example/` from Harmony
- 1ce8614b The pruning of unneeded runtimes from ANTLR4.

## How to review this PR

Review everything except that which went on in `/libs/` (so the CMake stuff, CONTRIBUTING.md, the `.gitattributes`).

# How Has This Been Tested?

- [ ] CI


thesamriel:
I think that I generally like the subtree approach.
The only open concern I'm having right now is that users always need to download all libraries, right?
For example, when I'm not interested in tests, I still need to download and have diskspace for googletest, right?

I'm not sure how big of a problem this really is, though.

SamNewcome:
> I think that I generally like the subtree approach. The only open concern I'm having right now is that users always need to download all libraries, right? For example, when I'm not interested in tests, I still need to download and have diskspace for googletest, right?
> 
> I'm not sure how big of a problem this really is, though.

In both the master and this branch you need to download all libraries - here it is just uncompressed. For context, googletest in the master takes up about ~1MB, and uncompressed takes up about 4MB. But, when downloading over git clone/pull, compression is used I believe, so I don't think this is a major issue with download speed. And when you build in the master you have to uncompressed the library anyway, so I'm not sure there is a significant on disk space difference. And I believe old versions of the compressed library will also be downloaded when cloning, resulting in an overall larger repo with compressed dependencies that have been updated several times - whereas these git changes from updating dependencies as subtrees will be smaller because each update is just the lines of code changed.

I think it is fair to want a version of AutoPas without all dependencies - I think especially because a lot are not used by a typical user e.g. ALL is for md-flexible only, ANTLR is only used by rule-based tuning. Not sure what is the cleanest way to go about this though.

PR #1155, opened 2026-07-01
Fix macOS CI pipline error by upgrading to gcc 16

schuhmaj:
# Description

macOS 26 SDK is unsupported by gcc 14/15 due to some changes in the`mach/message.h` header. GCC 16 ships the fix.

# How Has This Been Tested?

Test Pipeline runs again


SamNewcome (approved):
Thanks. Wait until CI is done and that this problematic run goes green, then all good to merge

PR #1150, opened 2026-06-18
Feature/soa sorting

hmeyran:
# Descripion
This PR introduces SoAFunctorPairSorted, a sorted variant of SoAFunctorPair that reduces unnecessary distance computations in the SoA path by projecting particles onto the cell-pair direction vector, sorting by projection, and pruning
pairs whose projections exceed the cutoff.

The base class PairwiseFunctor declares SoAFunctorPairSorted. The base implementation, can only throw an error, either if the Sorted Path is used when the Functor did not declare it supports sorting or when the Functor declares it supports sorting but does not overwrite this function (i.e. actually implements it). LJFunctorHWY provides the reference implementation with prepruning. 

A new soaSortingThreshold is added alongside the existing AoS sortingThreshold. CellFunctor selects between SoAFunctorPairSorted and SoAFunctorPair based on this threshold and the cell-pair direction, mirroring the existing AoS sorting dispatch. The threshold is exposed through the full configuration stack: ContainerSelectorInfo, LogicHandler, the AutoPas public API, and the md-flexible CLI and YAML parser.

The CellFunctor does most of the computation, it uses SortedSoAView, similar to the SortedCellView, and from them computes the pruned indeces. These are then passed to the actual Functor in a sortingData struct.

Google Benchmark support is added to the project, along with a benchmark suite for LJFunctorHWY covering single-cell, same-size and ratio-dependent cell pairs, and corner/edge cell configurations. To support controlled benchmarking of the sorted functor, TwoCellsInteractionHitrateGenerator is added to autopasTools: given two adjacent cells and a target hitrate in [0, 1], it fills each cell with n particles such that exactly round(n × hitrate) particles per cell have a guaranteed in-range partner in the other cell, with the remainder placed beyond cutoff. This allows benchmarks to sweep interaction density independently of cell size or particle count.
There were also Tests implemented for this Generator in TwoCellsInteractionHitrateGeneratorTest.h.

## Note
The SoASortingThreshold default is currently set to 50.
# How Has This Been Tested?

- [ ] New Tests in LJFunctorTestHWY, testing all cell arrangements (Face, Edge, Corner), comparing against the preexisting LJFunctor.h.
- [ ] Ran Multiple md-flexible simulations (Exploding Fluid, Falling Drop, Spinodial Decomposition and Rayleigh Taylor)

PR #427, opened 2020-02-20
Remove restrictions on Particle constructors 

SteffenSeckler:
# Description

Removes all restrictions on constructors of the Particle classes.
AutoPas now works with non-default constructible particles and thus will never call any real constructor of a particle.
The copy constructor is, however, required!

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes.

- [x] Adds a test that checks that AutoPas works with Particles that do not provide any real constructors (except for copy, move, ...)

PR #1156, opened 2026-07-02
Feature/soa sorting threshold

hmeyran:
# Description

Introduced and integrated a Benchmark that aims to find the optimum threshold for enabling SoA and AoS sorting. This Benchmark has to run once at the program startup, afterwards it provides the results throughout the whole simulation.

It finds the best thresholds differentiating by n3 and cell layout.
Threshold Benchmark takes about 1 sec to run.

This PR also moves the generators into the AutoPas source as well as refactors how the threshold values are passed to their respective CellFunctors.

## Related Pull Requests

- #1150 

# How Has This Been Tested?

- [] Ran TestSuite to ensure no breaking changes


hmeyran:
Currently there are no standalone unit tests for the benchmark, i ran end to end runs which work as intended. I am honestly unsure what parts of this can properly be unit tested.

SamNewcome:
> Currently there are no standalone unit tests for the benchmark, i ran end to end runs which work as intended. I am honestly unsure what parts of this can properly be unit tested.

I'll see if something comes to me during the review, but right now I agree

examples/md-flexible/src/configuration/MDFlexConfig.h
```diff
@@ -593,6 +593,15 @@ class MDFlexConfig {
       "Threshold for the SoA functor pair path to start sorting. If the sum of the SoA buffer sizes of two cells is "
       "greater or equal to that value, particles are sorted by their projection onto the cell-pair direction vector "
       "before computing interactions."};
+  /**
+   * useSortingThresholdBenchmark
+   * If true, AutoPas runs a micro-benchmark on the first iterate() call to determine per-direction-type
```
SamNewcome: What is meant by "per-direction-type"?

hmeyran:
This can btw also be very easily extended to also cover the aos threshold if this is desired. I might do this locally anyways for the thesis but not sure if this is something that you want to add to AutoPas.

SamNewcome:
> This can btw also be very easily extended to also cover the aos threshold if this is desired. I might do this locally anyways for the thesis but not sure if this is something that you want to add to AutoPas.

This is a good idea and I guess easy to do together. I'd be keen to have this in AutoPas. Can be either in this PR or a separate one.

src/autopas/baseFunctors/CellFunctor.h
```diff
@@ -94,6 +94,24 @@ class CellFunctor {
    */
   void setSoASortingThreshold(size_t soaSortingThreshold);
 
+  /**
+   * Set the per-Newton3-state, per-direction-type AoS pair-sorting thresholds.
+   * First index: 0 = Newton3 disabled, 1 = Newton3 enabled. Second index: number of zero components in
+   * sortingDirection (0=Corner, 1=Edge, 2=Face). Overrides the value set by setAoSSortingThreshold() for all
+   * Newton3 states and direction types. Does not affect the single-cell sorting threshold used by processCell().
```
SamNewcome: Maybe add what threshold single-cell sorting uses.

src/autopas/baseFunctors/CellFunctor.h
```diff
@@ -127,25 +145,21 @@ class CellFunctor {
...
+   * @param useSoA Whether to evaluate against the SoA thresholds (true) or the AoS thresholds (false).
+   * @return Whether the corresponding path should use sorted iteration.
    */
-  [[nodiscard]] bool shouldUseSoASorting(size_t particleCount, const std::array<double, 3> &sortingDirection) const {
-    return particleCount >= _soaSortingThreshold and
-           (sortingDirection[0] != 0.0 or sortingDirection[1] != 0.0 or sortingDirection[2] != 0.0);
+  [[nodiscard]] bool shouldUseSorting(size_t particleCount, const std::array<double, 3> &sortingDirection,
+                                      bool useSoA) const {
+    if (sortingDirection[0] != 0.0 or sortingDirection[1] != 0.0 or sortingDirection[2] != 0.0) {
+      const auto zeroCount = std::count(std::begin(sortingDirection), std::end(sortingDirection), 0.0);
```
SamNewcome: Use ranges

```suggestion
      const auto zeroCount = std::ranges::count(sortingDirection, 0.0);
```

src/autopas/utils/SortingThresholdBenchmark.h
```diff
@@ -0,0 +1,406 @@
...
+ *   [0]   → Newton3 disabled, [1] → Newton3 enabled
+ *   [][0] → Corner (three non-zero components)
+ *   [][1] → Edge   (two non-zero components)
+ *   [][2] → Face   (one non-zero component)
+ *
+ * Newton3 on/off is benchmarked separately because it changes how much work the sorted/unsorted paths do per call
+ * (symmetric force application vs. one-sided passes), which can shift the break-even point.
+ *
+ * The search performed in runSearch() is deliberately biased towards a conservative (higher) threshold, to avoid noisy
+ * measurements influencing the threshold to be too low, causing slowdowns.
```
SamNewcome: Would be worth expanding upon what you mean here

src/autopas/utils/SortingThresholdBenchmark.h
SamNewcome, line 1: Use an `Option` for layout and n3. Already exists for N3. For layout, I don't mind if this is local to this class or not. 

Reasoning: Use `enum` instead of opaque numbers. But as you need to convert this to a string name at points, you might as well use the option class which already exists for N3 and provides an enum and handling for strings.

src/autopas/utils/SortingThresholdBenchmark.h
```diff
@@ -0,0 +1,406 @@
...
+  void runBenchmark(Functor_T &functor) {
+    if constexpr (UseSoA) {
+      for (size_t n3 = 0; n3 < 2; n3++) {
+        for (size_t layout = 0; layout < 3; layout++) {
+          _soaThresholds[n3][layout] = runSearch<Functor_T, Particle_T, true>(functor, layout, static_cast<bool>(n3));
+        }
+      }
+      _hasRunSoA = true;
+    } else {
+      if constexpr (std::constructible_from<Particle_T, std::array<double, 3>, std::array<double, 3>, size_t>) {
```
SamNewcome: I wouldn't do this. It is not reliable/maintainable that `Particle_T` with have such a constructor, and may include non-particle objects.

Just assume the particle type will work. At some point we will do a proper C++-20 update and introduce a particle concept, but should be done differently (e.g. a concept that `Particle_T` is inherited by `ParticleBase` or that it implements all the public functions of `ParticleBase`) and this should be done in a different PR.

src/autopas/utils/SortingThresholdBenchmark.h
```diff
@@ -0,0 +1,406 @@
...
+   *   for functors with Functor_T::supportsSoASorting == true.
+   * - UseSoA == false: benchmarks the AoS path and sets hasRunAoS() to true. Internally
+   *   guarded by whether Particle_T is constructible the way this benchmark needs (position, velocity, id):.
+   * @tparam Functor_T Pairwise functor type.
+   * @tparam Particle_T Particle type.
+   * @tparam UseSoA Whether to benchmark the SoA path (true) or the AoS path (false).
+   * @param functor Functor instance used to drive the benchmark cells.
+   */
+  template <class Functor_T, class Particle_T, bool UseSoA>
+  void runBenchmark(Functor_T &functor) {
```
SamNewcome: As there's no overlap in the AoS and SoA paths, I would just use separate functions.

src/autopas/utils/SortingThresholdBenchmark.h
```diff
@@ -0,0 +1,406 @@
...
+        std::uniform_real_distribution<double>(boxLow[1], boxHigh[1]),
+        std::uniform_real_distribution<double>(boxLow[2], boxHigh[2])};
+    for (size_t i = 0; i < numParticles; ++i) {
+      Particle_T particle(defaultParticle);
+      particle.setR({dist[0](generator), dist[1](generator), dist[2](generator)});
+      particle.setID(i);
+      particle.setOwnershipState(autopas::OwnershipState::owned);
+      cell.addParticle(particle);
+    }
+  }
```
SamNewcome: Why not use the generator?

src/autopas/utils/SortingThresholdBenchmark.h
```diff
@@ -0,0 +1,406 @@
...
+    const double cutoff = functor.getCutoff();
+    const double invSqrt3 = 1. / sqrt(3.);
+    const double invSqrt2 = 1. / sqrt(2.);
+
+    size_t numParticlesEqDistr = static_cast<size_t>(std::ceil(numParticles * (1. - _scatterFactor)));
+
+    size_t numParticlesScatter = numParticles - numParticlesEqDistr;
+
+    size_t numParticlesCell1 = numParticlesEqDistr / 2;
+    size_t numParticlesCell2 = numParticlesEqDistr / 2 + numParticlesEqDistr % 2;
```
SamNewcome: It's maybe better to add the remainder particle to the scatter particles. Probably doesn't matter too much, especially for large particle counts and your 20% scatter, but better safer than sorry

src/autopas/utils/SortingThresholdBenchmark.h
```diff
@@ -0,0 +1,406 @@
...
+      // branch predictor from the first, which would otherwise make it look systematically faster than it is.
+      if (i % 2 == 0) {
+        unsortedDelta = measureUnsorted();
+        sortedDelta = measureSorted();
+      } else {
+        sortedDelta = measureSorted();
+        unsortedDelta = measureUnsorted();
+      }
+
+      AutoPasLog(DEBUG, "SortingThresholdBenchmark rep {}/{} layout={} n={}: unsorted={}ns sorted={}ns", i + 1,
```
SamNewcome:
```suggestion
      AutoPasLog(TRACE, "SortingThresholdBenchmark rep {}/{} layout={} n={}: unsorted={}ns sorted={}ns", i + 1,
```
Just a suggestion, but I think more appropriate

src/autopas/utils/SortingThresholdBenchmark.h
```diff
@@ -0,0 +1,406 @@
...
+
+      // A repetition only counts as a "sorted win" if it clears the margin: see _sortedWinMarginFraction.
+      if (static_cast<double>(sortedDelta) < static_cast<double>(unsortedDelta) * (1. - _sortedWinMarginFraction)) {
+        ++sortedWins;
+      }
+    }
+
+    const long meanSorted = sortedTimer.getTotalTime() / static_cast<long>(_repetitions);
+    const long meanUnsorted = unsortedTimer.getTotalTime() / static_cast<long>(_repetitions);
+    AutoPasLog(INFO, "SortingThresholdBenchmark layout={} n={}: mean unsorted={}ns mean sorted={}ns sortedWins={}/{}",
```
SamNewcome:
```suggestion
    AutoPasLog(TRACE, "SortingThresholdBenchmark layout={} n={}: mean unsorted={}ns mean sorted={}ns sortedWins={}/{}",
```
Trace is a suggestion, but at least DEBUG is a requirement from my side

src/autopas/utils/SortingThresholdBenchmark.h
```diff
@@ -0,0 +1,406 @@
...
+   * @param functor Functor instance used to drive the benchmark.
+   * @param layout Direction-type index (0=Corner, 1=Edge, 2=Face).
+   * @param useNewton3 Whether the benchmarked CellFunctor should apply Newton3.
+   * @return Smallest particle count at which sorted beats unsorted, or the upper search bound if never.
+   */
+  template <class Functor_T, class Particle_T, bool useSoA>
+  size_t runSearch(Functor_T &functor, size_t layout, bool useNewton3) {
+    size_t lowCount = 0;
+    // Use a smaller max particle count for AoS, since the AoS sorted path is expected to pay off at a lower
+    // particle count than the SoA sorted path (see _maxAoSParticles).
```
SamNewcome: I would drop this comment. This is already covered in the description and just means there is one less incorrect comment if this were to change in the future for whatever reason

src/autopas/utils/SortingThresholdBenchmark.h
```diff
@@ -0,0 +1,406 @@
...
+      size_t mid = lowCount + (highCount - lowCount) / 2;
+
+      const auto outcome = executeRun<Functor_T, Particle_T, useSoA>(functor, layout, mid, useNewton3);
+      const double winRatio = static_cast<double>(outcome) / static_cast<double>(_repetitions);
+
+      // Conservative decision rule: only accept "sorted wins" once a clear majority of repetitions
+      // agree by a clear margin (see _sortedWinMarginFraction and _requiredSortedWinRatio).
+      if (winRatio >= _requiredSortedWinRatio) {
+        highCount = mid;
+        AutoPasLog(DEBUG, "SortingThresholdBenchmark search {} layout={} n={}: sorted won {}/{} reps → high={}",
```
SamNewcome:
```suggestion
        AutoPasLog(TRACE, "SortingThresholdBenchmark search {} layout={} n={}: sorted won {}/{} reps → high={}",
```
similar. Suggestion only but I think better

src/autopas/utils/SortingThresholdBenchmark.h
```diff
@@ -0,0 +1,406 @@
...
+                   _newton3Names[useNewton3], _layoutNames[layout], mid, outcome, _repetitions, highCount);
+      } else {
+        lowCount = mid + 1;
+        AutoPasLog(DEBUG, "SortingThresholdBenchmark search {} layout={} n={}: sorted won only {}/{} reps → low={}",
```
SamNewcome:
```suggestion
        AutoPasLog(TRACE, "SortingThresholdBenchmark search {} layout={} n={}: sorted won only {}/{} reps → low={}",
```

src/autopas/utils/SortingThresholdBenchmark.h
```diff
@@ -0,0 +1,406 @@
...
+                   _newton3Names[useNewton3], _layoutNames[layout], mid, outcome, _repetitions, lowCount);
+      }
+    }
+    AutoPasLog(INFO, "SortingThresholdBenchmark {} layout={} threshold={}", _newton3Names[useNewton3],
```
SamNewcome:
```suggestion
    AutoPasLog(DEBUG, "SortingThresholdBenchmark {} layout={} threshold={}", _newton3Names[useNewton3],
```
Should not be `INFO`. I think `DEBUG` makes more sense than `TRACE` for this.

src/autopas/utils/SortingThresholdBenchmark.h
SamNewcome, line 1: I'd be interested in the reasoning for the given parameters. Might be better that we discuss these in our next meeting or here - I have nothing against them but that might change once I know reasoning.

Ultimately, the reasoning should be documented here, but leave the documentation for now and leave this comment unresolved so we don't forget

SamNewcome (changes requested):
Really nice. I think the benchmark itself is well thought through and the code is generally very clean. There are some minor issues listed below as comments, these are mostly documentation or different log levels.

There are two "major" issues, but they are not that bad, just more involved than simple documentation changes:

The more major issue is that you should use a `struct` instead of this 2D array structure, which is opaque. This would also avoid constantly needing comments explaining the array. 

But I would go one step further: 
- Create a base struct `SortingThresholdInfoInterface` that doesn't do anything. 
- You can then create a child struct for 2B, with clear member names for different thresholds.
- Have `setSoASortingThresholds` (similarly, AoS) take the interface, and in the 2B case dynamically checks it was passed a `SortingThresholdInfo2B` and set. We leave the 3B child class to the future.

Reasoning: The 3B equivalent `struct` will almost certainly not look the same as for 2B. So we shouldn't have a 3B function that takes a 2B-only threshold structure, which future developers might be confused by and will ultimately need to modify anyway. 

The second more major issue is that the combination of the original `setSoA/AoSSortingThreshold` and the new setter is a bit confusing. My suggestion is, keep the md-flexible options and the autopas interface, but internally use only one function: `setSoA/AoSSortingThresholds`. With the struct, you can create a single-argument constructor that sets all members to that value, so you don't need to set all values individually. An alternative would be a single-threshold child struct of the above interface, and with some dynamic polymorphism you skip anything needless in the one-threshold case.

src/autopas/utils/SortingThresholdBenchmark.h
hmeyran: Some of the parameters are subject to change and i will need to test them. Maybe a small breakdown here:

- maxAoSParticles = 50 : Wanted a lower value then SoA since AoS seems to break even with much lower numbers (taking from the previous default value and the Gratl dissertation), was not able to test this yet as the cluster is still down (since thursday).
- maxSoAParticles = 250 : Since i thought any value above 250 will probably just be equal to disabling anyways (but can be changed if you think different).
- iterations : Basically the minimum number of times the for loop needs to run to get a stable measurement. In my opinion something in the micro second instead of nanosecond region. 25 worked fine on my machine, same problem with cluster testing.
- repetitions : Adjusted for noise, with 25 i got a Coefficient of Variance of about 10% on my machine, i.e. one standard deviation is roughly 10% of the mean. This might go up or down depending on performance on cluster and i will vary this.
- sortedWinFraction : No real scientific argument for choosing this, for me a speedup of at least 5% sounded like a good metric to measure real speedup instead of measurement noise/difference in particle layout.
- requiredSortedWinRation : Similarly not picked scientifically. I just wanted a supermajority to agree.
- scatterFactor : Similarly i didnt want completely asymmetric  particle loads between cells and 20% randomly distributed felt like a good "jitter".

src/autopas/utils/SortingThresholdBenchmark.h
hmeyran: Would create a circular dependency since the Generators link against autopas: `target_link_libraries(generators PUBLIC autopas)`

src/autopas/utils/SortingThresholdBenchmark.h
hmeyran: Would you prefer a constexpr boolean for "supports Benchmarking", since i need some way of guarding this at compile time otherwise the CI, with for example nonConstructibleParticle, will fail.

examples/md-flexible/src/configuration/MDFlexConfig.h
hmeyran: I meant the same thing as Cell Layout so Face, Edge, Corner. Fixed the comment to keep consistent wording.

src/autopas/utils/SortingThresholdBenchmark.h
SamNewcome: So I did some digging because I wasn't actually sure why we cared about this. The non constructible particle test was added in #427. I'm not 100% sure the original reason, but my guess is that the issue is less that AutoPas can be used without a constructor and more that AutoPas can be used with particles that can be constructed arbitrarily. They do mention explicitly in the PR that the copy constructor is required still. I would keep to their original design ethos - I think it makes a lot of sense. So don't construct any particles, instead use the copy constructor: get a default particle by calling `_currentContainer->begin()` in the logic handler, pass this to `runBenchmark`, and do what you need through copying. You will need to add a protection in the case that `computeInteractions` is called without any particles -> we can simply postpone until the first iteration where there is a particle.

src/autopas/utils/SortingThresholdBenchmark.h
SamNewcome: Okay, sounds good. I think `maxAoS/SoAParticles`, `iterations`, `repetitions` are sensible.

I feel like `sortedWinFraction` and `requiredSortedWinRation` lead to a conservative (high) threshold. That's not nessecarily a bad thing, but I think there should be some justification of what this is for. E.g. if the consequences of going for sorted when it is inappropriate are greater than the opposite (e.g. a higher coefficient of variance), this could be justified.

For `scatterFactor`, I have no idea what is best here, but 20% seems sensible. You could consider a normal distribution instead to determine the number of scatter particles going to either cell, but this is needed for the PR.

src/autopas/utils/SortingThresholdBenchmark.h
SamNewcome: We could simply move generators into AutoPas itself. I have nothing against this but let me ask others first.

SamNewcome: No complaints so go for this. Should fix this whole issue.

hmeyran:
I will just answer to your review comment here since github won't let me answer to it direcctly. I agree with the idea that the array is bad. I have implemented a system that i think implements what you were going for. No guarantees that this completely works yet. I just would like to get feedback if this is the direction you want this to go in. Its a bit of a hassle because of C++ way of handling polymorphism and i haven't found a good solution for the seperate in-cell threshold in the CellFunctor.h. Also the Info class is internally still holding the same array, i could change this very easily though.

src/autopas/utils/SortingThresholdBenchmark.h
hmeyran: Now using the `Option` class, i have slightly modified the newton3Option so the enum also assigns specific numeric values (0 and 1) to disabled enabled that can be used to use the options more elegantly as indeces

SamNewcome:
> I will just answer to your review comment here since github won't let me answer to it direcctly. I agree with the idea that the array is bad. I have implemented a system that i think implements what you were going for. No guarantees that this completely works yet. I just would like to get feedback if this is the direction you want this to go in. Its a bit of a hassle because of C++ way of handling polymorphism and i haven't found a good solution for the seperate in-cell threshold in the CellFunctor.h. Also the Info class is internally still holding the same array, i could change this very easily though.

This is the right direction, yes! A couple points however:

- Use `struct`s. So public members (default for struct), no getters/setters. Does not even need a custom constructor - you could set the members directly (up to you). The purpose is to really just make the meaning of each element transparent, and not needing explaining, and so the struct can be very minimal - as if it were the original `std::array<std::array<size_t, 3>, 2>` but where each element has a name. See `LogicHandlerInfo` as a mostly good example (but I'd use `struct` instead of `class` for this kind of thing, as it signals intent that it just groups data together that can be freely publicly accessed.) Feel free to keep member functions though if they help though - potentially `.getThreshold(newton3, sortingDirection)` is still the cleanest approach.
- Add error handling to dynamic casts which return nullptrs.

Potentially something else will come up when doing the proper review, but these are the main noticeable points.

src/autopas/utils/SortingThresholdBenchmark.h
hmeyran: Should i move them under utils (either flat or utils/generators) or into their own src/generators?

src/autopas/utils/SortingThresholdBenchmark.h
hmeyran: Added an explainer comment, that a too low threshold would actively cause performance regressions while to high threshold would at the worst still perform the same as the baseline. The goal of these thresholds was to never cause a performance regression against baseline.

src/autopas/baseFunctors/CellFunctor.h
hmeyran: This is still a bit of a todo with the new info system. For now it just uses the Face Threshold but i need to think of  a better way to do this.

src/autopas/utils/SortingThresholdBenchmark.h
hmeyran: Implemented this as described. Now takes a particle from the container, postponing the benchmark run until the container has at least one Particle. That particle is then used as the default particle for the UniformGenerator. Which then uses copy construction.

src/autopas/utils/SortingThresholdBenchmark.h
SamNewcome: I think `utils/generators` is best

src/autopas/utils/SortingThresholdBenchmark.h
SamNewcome: Ok, fair.

I think this makes sense, but IMO you don't need to be too conservative. I wouldn't be afraid of performance regression occasionally if overall there is a performance improvement. I'd suggest in the thesis itself testing the performance in a unbiased search (particularly thinking about the win ratio of 70% here, the 5% threshold I'm less concerned about) is better/worse/about the same.

src/autopas/baseFunctors/CellFunctor.h
SamNewcome: I think this is fair. Potentially in the future we can add a separate benchmark but I don't think this is needed in this PR/your thesis.

examples/md-flexible/src/configuration/MDFlexConfig.h
```diff
@@ -604,10 +604,9 @@ class MDFlexConfig {
    */
   MDFlexOption<bool, __LINE__> useSortingThresholdBenchmark{
       false, "use-sorting-threshold-benchmark", true,
-      "If true, AutoPas runs a micro-benchmark to determine the optimal AoS pair-sorting threshold per direction "
-      "type instead of the fixed aos-sorting-threshold, and additionally the optimal SoA sorting threshold per "
-      "direction type instead of the fixed soa-sorting-threshold for functors that "
-      "support SoA sorting."};
+      "If true, AutoPas runs a micro-benchmark to determine the optimal AoS pair-sorting threshold per cell layout "
+      "(Face, Edge, Corner) instead of the fixed aos-sorting-threshold, and additionally the optimal SoA sorting "
+      "threshold per direction type instead of the fixed soa-sorting-threshold for functors that support SoA sorting."};
```
SamNewcome: Rewrite this more concise. You have a separate clause for SoA that says basically the same thing, just with an extra caveat, but written differently. You can just write "AoS/SoA" and similar, and then add the caveat for SoA at the end.

src/autopas/baseFunctors/CellFunctor.h
```diff
@@ -325,8 +302,8 @@ void CellFunctor<ParticleCell_T, ParticleFunctor_T, bidirectional>::processCellA
       this->_functor.AoSFunctor(p2, p1, false);
     }
   };
-
-  if (cell.size() >= _aosSortingThreshold) {
+  // TODO: This is currently arbitrary, think of a good way to handle this
```
SamNewcome: Be clear what "This" is.

src/autopas/containers/cellTraversals/CellTraversal.h
```diff
@@ -40,38 +42,23 @@ class CellTraversal {
...
-   */
-  virtual void setSoASortingThreshold(size_t soaSortingThreshold) = 0;
-
-  /**
-   * Set per-Newton3-state, per-direction AoS pair-sorting thresholds for traversals that use CellFunctor.
    * Traversals that hold a CellFunctor (or LCC08CellHandler) should forward to it. Others should explicitly
    * override with an empty body.
-   * @param thresholds Per-Newton3-state, per-direction thresholds, indexed as [newton3][direction] (direction =
-   * number of zero components in sortingDirection: 0=Corner, 1=Edge, 2=Face).
+   * @param aosSortingThreshold
```
SamNewcome: Add some brief details. E.g. what is accepted

src/autopas/containers/cellTraversals/CellTraversal.h
```diff
@@ -40,38 +42,23 @@ class CellTraversal {
...
 
   /**
-   * Set per-Newton3-state, per-direction SoA sorting thresholds for traversals that use CellFunctor.
+   * Set the SoA sorting-threshold for traversals that use the CellFunctor.
+   * If the sum of the SoA buffer sizes of two cells exceeds this value, the SoA path uses SoAFunctorPairSorted.
    * Traversals that hold a CellFunctor (or LCC08CellHandler) should forward to it. Others should explicitly
    * override with an empty body.
-   * @param thresholds Per-Newton3-state, per-direction thresholds, indexed as [newton3][direction] (direction =
-   * number of zero components in sortingDirection: 0=Corner, 1=Edge, 2=Face).
+   * @param soaSortingThreshold
```
SamNewcome: Similar

src/autopas/containers/directSum/traversals/DSSequentialTraversal.h
```diff
@@ -68,30 +68,18 @@ class DSSequentialTraversal : public CellTraversal<ParticleCell>,
...
-    _cellFunctor.setAoSSortingThresholds(thresholds);
+  void setAoSSortingThresholds(const SortingThresholdInfoInterface &aosSortingThreshold) override {
+    _cellFunctor.setAoSSortingThresholds(aosSortingThreshold);
   }
 
   /**
    * @copydoc autopas::CellTraversal::setSoASortingThresholds()
    */
-  void setSoASortingThresholds(std::array<std::array<size_t, 3>, 2> thresholds) override {
-    _cellFunctor.setSoASortingThresholds(thresholds);
```
SamNewcome: With some brief details of what is expected in the 2B cell functor, you won't be able to copydoc so create your own documentation with a brief comment about what is expected from the soaSortingThreshold

src/autopas/containers/CellBasedParticleContainer.h
```diff
@@ -181,22 +178,16 @@ class CellBasedParticleContainer : public ParticleContainerInterface<typename Pa
...
-   * If the number of particles in a cell exceeds this threshold, single-cell interactions will be sorted.
-   * To be forwarded to cell traversals.
-   */
-  size_t _aosSortingThreshold;
-  /**
-   * Per-Newton3-state, per-direction AoS pair-sorting thresholds, indexed as [newton3][direction] (direction =
-   * number of zero components in sortingDirection: 0=Corner, 1=Edge, 2=Face).
-   * Initialized from the scalar aosSortingThreshold; overridden by setAoSSortingThresholds().
+   * Current AoS pair-sorting threshold, forwarded to freshly generated traversals in prepareTraversal().
+   * Owned as a shared_ptr so the container can hold whichever concrete shape it was given (a uniform
```
SamNewcome:
```suggestion
   * Owned as a shared_ptr to the interface so the container can hold whichever concrete shape it was given (a uniform
```

src/autopas/options/SortingDirectionOption.h
```diff
@@ -0,0 +1,92 @@
+/**
+ * @file CellLayoutOption.h
+ * @date 14.07.2026
```
SamNewcome, line 3: Add author

src/autopas/options/CellLayoutOption.h
```diff
@@ -0,0 +1,92 @@
...
+#include "autopas/options/Option.h"
+
+namespace autopas {
+inline namespace options {
+/**
+ * Class representing the classification of a normalized cell-pair sortingDirection by the number of its zero
+ * components.
+ *
+ * // TODO: name collides conceptually with DataLayoutOption (AoS/SoA); rename if a clearer name is found
+ * // (candidates considered: SortingDirectionOption, DirectionTypeOption).
```
SamNewcome: Do TODO.

Personally, I'm okay with `CellLayoutOption` but others might not be (will add a 2nd reviewer). I quite like `SortingDirectionOption`.

src/autopas/utils/SortingThresholdBenchmark.h
```diff
@@ -25,122 +25,126 @@ namespace autopas {
...
+ * Stores one threshold per Newton3 state and direction type.
  *
  * Newton3 on/off is benchmarked separately because it changes how much work the sorted/unsorted paths do per call
  * (symmetric force application vs. one-sided passes), which can shift the break-even point.
  *
  * The search performed in runSearch() is deliberately biased towards a conservative (higher) threshold, to avoid noisy
- * measurements influencing the threshold to be too low, causing slowdowns.
+ * measurements influencing the threshold to be too low. A too low threshold would actively regress performance while a
+ * too high threshold would at worst still perform the same as baseline. One goal of these thresholds is to avoid any
+ * performance regression.
```
SamNewcome: You need to add performance regression compared to what. I know that this is compared to original no-sorting but somebody in five years time won't.

src/autopas/utils/SortingThresholdBenchmark.h
```diff
@@ -25,122 +25,126 @@ namespace autopas {
...
  */
 class SortingThresholdBenchmark {
  public:
   /**
    * Return all per-Newton3-state, per-direction-type SoA sorting thresholds.
    * Only meaningful once hasRunSoA() is true.
-   * @return Copy of the internal threshold array, indexed as [newton3][direction] (see class documentation).
+   * The returned pointer is stable (the same instance is reused every call) once runSoABenchmark() has completed,
+   * so callers can cheaply re-apply it every iteration (e.g. to a container) without reallocating.
```
SamNewcome: I don't quite get what this comment is intended for. I would either clarify or remove (I'm not sure it is needed)

src/autopas/utils/SortingThresholdBenchmark.h
```diff
@@ -25,122 +25,126 @@ namespace autopas {
...
-   * - UseSoA == false: benchmarks the AoS path and sets hasRunAoS() to true. Internally
-   *   guarded by whether Particle_T is constructible the way this benchmark needs (position, velocity, id):.
+   * Runs the micro-benchmark for the SoA sorting threshold, sweeping both Newton3 states and all three direction
+   * types, and storing the resulting thresholds. Sets hasRunSoA() to true.
+   * Callers should only instantiate this for functors with Functor_T::supportsSoASorting == true.
    * @tparam Functor_T Pairwise functor type.
    * @tparam Particle_T Particle type.
-   * @tparam UseSoA Whether to benchmark the SoA path (true) or the AoS path (false).
    * @param functor Functor instance used to drive the benchmark cells.
+   * @param defaultParticle Template particle (e.g. sampled from the live simulation) whose non-positional
```
SamNewcome:
```suggestion
   * @param defaultParticle Template particle (e.g. sampled from the particle container) whose non-positional
```

src/autopas/utils/SortingThresholdBenchmark.h
```diff
@@ -25,122 +25,126 @@ namespace autopas {
...
-  template <class Functor_T, class Particle_T, bool UseSoA>
-  void runBenchmark(Functor_T &functor) {
-    if constexpr (UseSoA) {
-      for (size_t n3 = 0; n3 < 2; n3++) {
-        for (size_t layout = 0; layout < 3; layout++) {
-          _soaThresholds[n3][layout] = runSearch<Functor_T, Particle_T, true>(functor, layout, static_cast<bool>(n3));
-        }
+  template <class Functor_T, class Particle_T>
+  void runSoABenchmark(Functor_T &functor, const Particle_T &defaultParticle) {
+    std::array<std::array<size_t, 3>, 2> thresholds;
```
SamNewcome: Avoid the array structure here too if possible. You can always use a helper getter/setter that takes the n3, layout options as argument.

src/autopas/utils/SortingThresholdBenchmark.h
```diff
@@ -25,122 +25,126 @@ namespace autopas {
...
+   * types, and storing the resulting thresholds. Sets hasRunAoS() to true.
+   * @tparam Functor_T Pairwise functor type.
+   * @tparam Particle_T Particle type.
+   * @param functor Functor instance used to drive the benchmark cells.
+   * @param defaultParticle Template particle (e.g. sampled from the live simulation) whose non-positional
+   * properties are copied onto every particle generated for the benchmark cells.
+   */
+  template <class Functor_T, class Particle_T>
+  void runAoSBenchmark(Functor_T &functor, const Particle_T &defaultParticle) {
+    std::array<std::array<size_t, 3>, 2> thresholds;
```
SamNewcome: Similar

src/autopas/utils/SortingThresholdInfoSingle.h
```diff
@@ -0,0 +1,33 @@
...
+#include <cstddef>
+
+#include "autopas/utils/SortingThresholdInfoInterface.h"
+
+namespace autopas {
+
+/**
+ * Single threshold value struct.
+ * Stand in for the old single value way of handling the sorting Threshold. Used as a fallback option and the default
+ * value.
```
SamNewcome: Remove this, because we still use this for 3B.

src/autopas/utils/SortingThresholdInfoSingle.h
```diff
@@ -0,0 +1,33 @@
...
+ */
+struct SortingThresholdInfoSingle : SortingThresholdInfoInterface {
+  /**
+   * Constructor for single threshold value.
+   * Intentionally not explicit to make it easier to use this like you would the threshold before.
```
SamNewcome: Avoid comments with temporal comparisons to a "before" because they won't make sense in a year or so

src/autopas/LogicHandler.h
```diff
@@ -842,15 +845,29 @@ class LogicHandler {
...
+   * whenever the container's AoS sorting threshold needs to be reset to a type CellFunctor3B accepts (see
+   * computeInteractionsPipeline()) without reallocating on every triwise call.
+   */
+  std::shared_ptr<const SortingThresholdInfoSingle> _defaultAoSSortingThresholdSingle;
+  /**
+   * Cached SortingThresholdInfoSingle wrapping _soaSortingThreshold, constructed once so that it can be reused
+   * whenever the container's SoA sorting threshold needs to be reset to a type CellFunctor3B accepts (see
+   * computeInteractionsPipeline()) without reallocating on every triwise call.
+   */
+  std::shared_ptr<const SortingThresholdInfoSingle> _defaultSoASortingThresholdSingle;
```
SamNewcome: Just trying to understand this: This is to avoid reallocations from converting to Single threshold in 3B case? I think this is too convoluted. Is the conversion cost that high? This would be done once per iteration.

src/autopas/LogicHandler.h
```diff
@@ -1272,26 +1289,57 @@ bool LogicHandler<Particle_T>::computeInteractionsPipeline(Functor *functor,
   tuningTimer.stop();
   _tuningManager->logTuningResult(tuningTimer.getTotalTime(), _iteration, interactionType);
 
-  // Run SortingThresholdBenchmark lazily on the first iteration and inject into the active container.
+  // Run SortingThresholdBenchmark lazily on the first iteration for which an owned particle is available, and
+  // inject the result into the active container. Until then this is a no-op and hasRunAoS()/hasRunSoA() stay
+  // false, so the benchmark is simply retried on the next call to this function.
```
SamNewcome: Explain why we do this. Emphasise that this is simply required to handle niche no-particle cases.

src/autopas/LogicHandler.h
```diff
@@ -1272,26 +1289,57 @@ bool LogicHandler<Particle_T>::computeInteractionsPipeline(Functor *functor,
...
       auto &sortingThresholdBenchmark = autoTuner.getSortingThresholdBenchmark();
 
-      if (not sortingThresholdBenchmark.hasRunAoS()) {
-        sortingThresholdBenchmark.template runBenchmark<Functor, Particle_T, false>(*functor);
+      const bool needsAoSRun = not sortingThresholdBenchmark.hasRunAoS();
+      const bool needsSoARun = Functor::supportsSoASorting and not sortingThresholdBenchmark.hasRunSoA();
+      if (needsAoSRun or needsSoARun) {
+        // Sample an existing particle from the container to use as the template for the particles generated by the
+        // benchmark, so that particle types needing more than (position, velocity, id) to be valid (e.g. multi-site
+        // molecules) are benchmarked with realistic properties instead of a synthetic default construction.
```
SamNewcome: To be more precise: AutoPas does not assume anything on how Particle_T can be constructed, beyond that it can be copy constructed, so we have to use copied of an existing particle.

SamNewcome (changes requested):
The SortingThreshold struct refactor is really nice, and the code is looking very nice.

What I didn't quite clock before this is that you apply the same threshold regardless of the functor (2B/3B). I don't really want to getting into this topic in this PR/in your thesis, but as the struct polymorphism causes issues with this, you should deal with this, at least enough to deal with to avoid this whole `_defaultSoASortingThresholdSingle` type of thing.

One option is to integrate the benchmark into the AutoTuner. We have one per functor anyway. Alternative, you could create a map of `SortingThresholdInfoInterface`s per interaction type, like we do for AutoTuner.

(These make sense beyond code quality: even if we had two different 2B functors which could take the same struct, I suspect that the thresholds themselves would be different, so it is good to separate the thresholds. Longer term, I think this map between interaction type and AutoTuner(+Thresholds) could become between functor (regardless of interaction type) and AutoTuner(+Thresholds).)

If this turns out to be a lot of work let me know and we'll come up with something simpler.

hmeyran:
> What I didn't quite clock before this is that you apply the same threshold regardless of the functor (2B/3B).  

I think this really only applies to the defaults as the benchmarked values are set via the thresholdBenchmark which is already scoped to interactionType via:
```c++
auto &autoTuner = *_tuningManager->getAutoTuners()[interactionType];
auto &sortingThresholdBenchmark = autoTuner.getSortingThresholdBenchmark();
```
There is of course the constexpr `if(isPairwiseFunctor)` check before that, as the benchmark currently only supports Pairwise functors. For any other i just set the default value. I'm not quite sure what of this needs to be pulled into the AutoTuner, i could let the AutoTuner own the default values.
With that i could also make the SortingThresholdBenchmark a bit more Functor agnostic/easier to extend to other functors, by setting default values and letting the LogicHandler always pull the threshold value from the thresholdBenchmark owned by the correct AutoTuner. I would then only guard the actual benchmark execution behind the isPairswiseFunctor statement.

src/autopas/LogicHandler.h
hmeyran: This was mostly done as in the current state the triwise functor would be reset to the threshold every time and a shared_ptr would have to be allocated every call. I thought this would be the more economical option but this can be of course constructed every iteration if this member clutters the class to much. (This might also just vanish completely depending on how we resolve the main issue you mentioned in your PR comment)

src/autopas/containers/directSum/traversals/DSSequentialTraversal.h
hmeyran: I dont think i understand what you mean by this, why would i not be able to use copydoc here?

SamNewcome: This comment is nonsense, sorry.

src/autopas/LogicHandler.h
SamNewcome: I would assume that the allocation cost is insignificant - this is only need once per computeInteractions right?

SamNewcome:
> > What I didn't quite clock before this is that you apply the same threshold regardless of the functor (2B/3B).
> 
> I think this really only applies to the defaults as the benchmarked values are set via the thresholdBenchmark which is already scoped to interactionType via:
> 
> ```c++
> auto &autoTuner = *_tuningManager->getAutoTuners()[interactionType];
> auto &sortingThresholdBenchmark = autoTuner.getSortingThresholdBenchmark();
> ```
> 
> There is of course the constexpr `if(isPairwiseFunctor)` check before that, as the benchmark currently only supports Pairwise functors. For any other i just set the default value. I'm not quite sure what of this needs to be pulled into the AutoTuner, i could let the AutoTuner own the default values. With that i could also make the SortingThresholdBenchmark a bit more Functor agnostic/easier to extend to other functors, by setting default values and letting the LogicHandler always pull the threshold value from the thresholdBenchmark owned by the correct AutoTuner. I would then only guard the actual benchmark execution behind the isPairswiseFunctor statement.

Ah ok. So if no benchmark is used, we use the default for both 2B/3B and if benchmark is used, we use benchmark for 2B and default for 3B. If I'm correct, I think it is fine to leave as it is.

src/autopas/LogicHandler.h
hmeyran: Yes, i did a small refactoring which removed this anyways so its resolved.

src/autopas/utils/SortingThresholdBenchmark.h
hmeyran, line 1: I refactored the benchmark a little. It now returns the generic `SortingThresholdInfoInterface.` instead of a specific 2B instance. It also now constructs with two default values. All of this enables the Benchmark to always return valid data even if the Benchmark hasn't run. This mean it can more easily be used in the `LogicHandler` and could later be extended to the 3B case. As a result the sortingThreshold defaults now need to live in `AutoTunerInfo` as the AutoTuner constructs the Benchmark objects. Let me know if this is not desired or you find the refactor to not improve code quality.

src/autopas/tuning/selectors/ContainerSelectorInfo.h
```diff
@@ -43,6 +43,8 @@ class ContainerSelectorInfo {
    * @param aosSortingThreshold Number of particles in two cells from which sorting should be performed
    * @param loadEstimator load estimation algorithm for balanced traversals.
    * @param soaSortingThreshold Number of particles in two SoA buffers from which SoA sorting should be performed.
+   * @note This still takes size_t single value thresholds as changing these to structs would be a large refactor,
+   * including all the comparative functions. Struct Threshold Values are only st through the explicit setters.
    */
   explicit ContainerSelectorInfo(const std::array<double, 3> &boxMin, const std::array<double, 3> &boxMax,
```
hmeyran, line 49: This still takes size_t single value thresholds as changing these to structs would be a large refactor, including all the comparative functions. Struct Threshold Values are only set through the explicit setters.

src/autopas/options/CellLayoutOption.h
hmeyran: Should i then also change the doc strings (mainly in the SortingThresholdBenchmark as well as the LJFunctorHWYBench) to something else to keep naming consistent or can i continue to use layout there?

src/autopas/utils/SortingThresholdBenchmark.h
SamNewcome: I'll check this out when reviewing properly but in general this sounds good

src/autopas/options/CellLayoutOption.h
SamNewcome: Yeah best keep naming consistent.

src/autopas/LogicHandler.h
```diff
@@ -55,9 +54,15 @@ class LogicHandler {
    * @param logicHandlerInfo
    * @param rebuildFrequency
    * @param outputSuffix
+   * @param aosSortingThreshold Default AoS sorting threshold used to seed newly generated containers.
+   * Mirrors AutoTunerInfo::aosSortingThreshold, which independently seeds each AutoTuner's
+   * SortingThresholdBenchmark; passed here explicitly since LogicHandler has no other access to AutoTunerInfo.
+   * @param soaSortingThreshold Default SoA sorting threshold used to seed newly generated containers.
```
SamNewcome: By my understanding, this isn't actually needed. It is used to pass to the container selector, which gets passed to the actual traversal, but I _think_ we end up just using the AutoTuner struct one ultimately. Could you check this? I'm somewhat keen to avoid basically having two routes passing the threshold.

src/autopas/LogicHandler.h
```diff
@@ -834,39 +837,26 @@ class LogicHandler {
...
-   * Default comes from the LJFunctorHWY Benchmarks.
+   * Default SoA sorting threshold used to seed newly generated containers. @see _aosSortingThreshold
    */
   size_t _soaSortingThreshold;
 
   /**
-   * Cached SortingThresholdInfoSingle wrapping _aosSortingThreshold, constructed once so that it can be reused
-   * whenever the container's AoS sorting threshold needs to be reset to a type CellFunctor3B accepts (see
-   * computeInteractionsPipeline()) without reallocating on every triwise call.
+   * This is used to store the total number of neighbour lists rebuild.
```
SamNewcome:
```suggestion
   * This is used to store the total number of neighbor lists rebuild.
```
US spelling

src/autopas/LogicHandler.h
```diff
@@ -834,39 +837,26 @@ class LogicHandler {
...
    */
-  std::shared_ptr<const SortingThresholdInfoSingle> _defaultAoSSortingThresholdSingle;
+  size_t _numRebuilds{0};
+
   /**
-   * Cached SortingThresholdInfoSingle wrapping _soaSortingThreshold, constructed once so that it can be reused
-   * whenever the container's SoA sorting threshold needs to be reset to a type CellFunctor3B accepts (see
-   * computeInteractionsPipeline()) without reallocating on every triwise call.
+   * This is used to store the total number of neighbour lists rebuilds in the non-tuning phase.
```
SamNewcome:
```suggestion
   * This is used to store the total number of neighbor lists rebuilds in the non-tuning phase.
```

SamNewcome (changes requested):
Nice. There was this one point about this other route this `size_t` sorting threshold can take to the functor and I _think_ we can just remove this but I might be mistaken.

src/autopas/LogicHandler.h
hmeyran: As far as i can tell this value is only used to construct the Containers with default values, nowhere else. I would just construct the thresholds 0 initialized since they get set anyway by the setter. Removing all the cases where the thresholds get passed in the constructor.

SamNewcome: I think my point is more, we probably don't need the containers to have these default values, right?

hmeyran: I think we are talking about the same thing. 
Currently it flows like this: `LogicHandler` constructs the Container with `size_t` defaults -> The Container converts the `size_t` values to `SortingThresholdInfoSingle`s and stores them -> LogicHandler overwrites the stored values via `setAoS/SoAThresholds`.
By removing the size_t input to Containers these defaults would fall away. I guess the question now is, should there still be a 0 initialized `shared_ptr` threshold be constructed or should they be left unset? 
The only problem with the second option is that now the containers would be non-functional/would crash if the thresholds are not set before. This would make the system a bit more fragile and depend on a specific execution order in LogicHandler. Containers could not be used directly like they are in some tests currently. So should the containers be always safe to use or is it ok if they require more setup after the constructor?

hmeyran: To add to this currently these `size_t` defaults in LogicHandler provide the sortingThresold data if the benchmarking is disabled. I could find away around this (By pulling `_currentContainer->setAoSSortingThresholds(sortingThresholdBenchmark.getAoSThreshold());` Out of the `if (_logicHandlerInfo.useSortingThresholdBenchmark)` block. This would mean some benchmark code (the default values) is still accessed even if the benchmark is disabled, but would make completely removing the `size_t` possible/easy.

SamNewcome: Hmmm, okay. Then then I would change the variable name + documentation to make it clear that this is really a fallback default value.

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -791,8 +795,9 @@ class LJFunctorHWY
           fy2Ptr, fz2Ptr, typeID1Ptr, typeID2Ptr, virialSumX, virialSumY, virialSumZ, uPotSum, 0, jVecStart, jVecEnd);
     }
     if constexpr (vecPattern != VectorizationPattern::p1xVec) {
-      // Rest I can't occur in 1xVec case
-      const size_t restI = n1 - i;
+      // Rest I can't occur in 1xVec case. Bounded by endI, not n1: i-particles from endI onwards cannot
+      // interact with any j-particle and are skipped entirely rather than processed as a no-op remainder.
+      const size_t restI = static_cast<size_t>(endI - i);
```
SamNewcome: The potential underflow here makes me a little uncomfortable. I think you can just keep this all as the same `std::ptrdiff_t` as used for everything else array-index related, then there's no static cast and if this were to go negative, it would correctly skip the `if`-statement.

src/autopas/baseFunctors/PairwiseFunctor.h
```diff
@@ -21,16 +21,23 @@ template <class Particle>
...
-  const std::vector<size_t> &maxIndex;  ///< Per-particle upper bound index into soa2 (exclusive).
-  const std::vector<size_t> &minIndex;  ///< Per-particle lower bound index into soa2 (inclusive).
+  size_t startI;  ///< First index into soa1 that can interact with soa2, skipping soa1 particles too far below.
+  size_t endI;    ///< Exclusive upper bound index into soa1 that can interact with soa2, skipping soa1 particles
+                  ///< too far above. Whether this or startI is the no-op bound depends on the call, see struct doc.
+  const std::vector<size_t> &maxIndex;  ///< Per-i (soa1) exclusive upper bound index into soa2, skipping soa2
+                                        ///< particles too far above particle i.
+  const std::vector<size_t> &minIndex;  ///< Per-i (soa1) inclusive lower bound index into soa2, skipping soa2
+                                        ///< particles too far below particle i. Whether this or maxIndex is the
+                                        ///< no-op bound depends on the call, see struct doc.
```
SamNewcome: Use AutoPas-style documentation, especially for multi-line docs:

```
/**
 *
 */
```

SamNewcome (changes requested):
Looks very nice. Two minor points only.

SamNewcome (approved):
From my side, everything is perfect. But I want a second reviewer to approve before merging. I've added now a third reviewer in case they are able to get around to reviewing this first.

src/autopas/utils/SortingThresholdBenchmark.h
```diff
@@ -0,0 +1,381 @@
...
+   * cell and 20% being randomly distributed between the cells.
+   * @param newton3 Whether the benchmarked CellFunctor should apply Newton3.
+   * @return The number of repetitions (out of _repetitions) in which the sorted path was measured as faster than
+   * the unsorted path by at least _sortedWinMarginFraction.
+   */
+  template <class Functor_T, class Particle_T, bool useSoA>
+  size_t executeRun(Functor_T &functor, const Particle_T &defaultParticle, SortingDirectionOption cellDirection,
+                    size_t numParticles, Newton3Option newton3) {
+    using BenchCell = FullParticleCell<Particle_T>;
+    using BenchCF = internal::CellFunctor<BenchCell, Functor_T, true>;
```
dmartin13: Is there a specific reason why the `CellFunctor` is created only with `bidirectional=true`, and `bidirectional=false` is not benchmarked?

src/autopas/utils/SortingThresholdBenchmark.h
```diff
@@ -0,0 +1,381 @@
...
+      } else {
+        lowCount = mid + 1;
+        AutoPasLog(TRACE,
+                   "SortingThresholdBenchmark search {} cell direction={} n={}: sorted won only {}/{} reps → low={}",
+                   newton3, cellDirection, mid, outcome, _repetitions, lowCount);
+      }
+    }
+    AutoPasLog(DEBUG, "SortingThresholdBenchmark {} cell direction={} threshold={}", newton3, cellDirection, lowCount);
+    return lowCount;
+  }
```
dmartin13: If sorting does not win for any tested particle count, the binary search still returns the initial upper bound. The upper bound itself is never evaluated, since mid is always smaller than `highCount`.

The returned value is later treated as a valid crossover point via `particleCount >= threshold`. Consequently, sorting is enabled at and above the upper bound even though no winning measurement was observed.

Could we explicitly evaluate the upper bound and use a dedicated "never sort" value if sorting still loses there? Or is there a reason why enabling sorting beyond the benchmarked range is considered safe here?

dmartin13 (changes requested):
Very nice!

I only found two points. But it's also possible that a comment in the code (or here) is enough to clarify them.

thesamriel (changes requested):
Generally, really nice code and I love the addition of this pre-simulation micro benchmark.

My main concern when going through the changes was that "sortingThreshold" seemed to be everywhere. I think this can be simplified to also make it clearer where the eventually used sorting values are actually coming from.

For example:
The `LogicHandler` always calls `_currentContainer->setAoSSortingThresholds(sortingThresholdBenchmark.getAoSThreshold());` (same for SoA) before the interactions. Therefore, I believe the "fallbacks" from the container constructors can be removed, as the true fallback values are stored centrally and forwarded in above setSortingThreshold call.
Potentially, these values can be removed completely from the containers and passed directly to the traversals. The containers don't seem to actually do anything with the sorting thresholds. But this might be more involved refactoring.

src/autopas/utils/SortingThresholdBenchmark.h
hmeyran: 250 Particles was a value i assumed would be never reached in molecular dynamics simulations and basically was supposed to function as a never use sorting value. I think this was set to low and should be increased since it doesn't make the benchmark take that much longer (maybe 3 or for more binary search steps for the soa path). The question would be what upper bound would actually guarantee the sorted path to never be taken. I was thinking of 1000?

dmartin13: Ok I see.  I do not think there is a finite upper bound that can generally guarantee that the sorted path is never taken. Even 1000 particles per cell could potentially occur for unusual densities, cell sizes, ...

Could we instead separate the benchmark upper bound from the "never sort" value? The benchmark could still stop at 250, 500 or 1000 (not sure which one makes most sense), so its runtime remains bounded. If sorting never wins within that range, we could return `std::numeric_limits<size_t>::max()` without actually benchmarking up to that value.

hmeyran: Implemented now, i kept the upper search bounds the same, but when the search never finds a lower value it now returns `std::numeric_limits<size_t>::max()`.

src/autopas/utils/SortingThresholdBenchmark.h
hmeyran: I was unsure on how to approach `bidirectional` here. Since for my thesis i am using lj and linked cells it made more sense to default to `bidirectional=true`. 
However `bidirectional` changes the performance of the SoA version heavily so it 100% makes sense to add. On the same note it Makes sense to add VecPatterns here since they also change this. The thing is that this would increase the number of stored configurations/variables drastically and it maybe worth to consider storing these differently then just as individual variables. Would you like me to just add bidirectional for now?

hmeyran:
@thesamriel i think pushing the thresholds directly to the traversal, skipping the container would be a very large refactor, since one would need to adjust all containers and traversals. Dropping the fallback was tried once before the one thing that came up was: there are quite a few places in the code where the container is used without the LogicHandler running first, meaning they would never have any threshold values set through the LogicHandler (mostly in tests but also in `TraversalComparison.cpp` and `Newton3OnOffTest.cpp`). There is also the `CellBasedParticleContainer` which needs some value for the thresholds at construction, so the value could not be dropped there. This refactor would be smaller but still some work.

src/autopas/utils/SortingThresholdBenchmark.h
dmartin13: Yeah, I think adding bidirectional for now would make sense, especially since it seems to have quite a big impact on the SoA performance. But if this turns out to require too many changes, I wouldn’t make it a requirement for this PR either.

For the VecPatterns, I think it would be good to create an issue describing the idea and the current limitation. If bidirectional also turns out to be too much work for this PR, you could just add that to the same issue as well.

thesamriel:
> @thesamriel i think pushing the thresholds directly to the traversal, skipping the container would be a very large refactor, since one would need to adjust all containers and traversals. Dropping the fallback was tried once before the one thing that came up was: there are quite a few places in the code where the container is used without the LogicHandler running first, meaning they would never have any threshold values set through the LogicHandler (mostly in tests but also in `TraversalComparison.cpp` and `Newton3OnOffTest.cpp`). There is also the `CellBasedParticleContainer` which needs some value for the thresholds at construction, so the value could not be dropped there. This refactor would be smaller but still some work.

That's fair. Could you then create an issue and link it here / describe the problems with the tests? It's fine not to do this in this PR, but I would like to refactor it in the future.

src/autopas/utils/SortingThresholdBenchmark.h
hmeyran: Okay added bidirectional as a swept parameter and the VecPatterns issue was created as #1174

hmeyran:
> That's fair. Could you then create an issue and link it here / describe the problems with the tests? It's fine not to do this in this PR, but I would like to refactor it in the future.

Okay, added as issue #1173

dmartin13 (approved):

thesamriel (approved):

PR #1158, opened 2026-07-21
Interaction detection functor

SamNewcome:
# Description

Many codes handle force calculations with two passes: one pass gets a list of neighbours in contact, the next pass applies forces to those in the list. This PR introduces* a functor which can be used to generate the list of neighbours in contact, with which the user can independently calculate the relevant forces.

Note: This is probably inefficient compared to our "integrated" force calculation functors where the force is calculated in the functor, integrated together with the cutoff check, and for future GPU developments, would probably require CPU-GPU transfers. So I do discourage its use but it is good to support as it allows for much easier use of AutoPas in such codes, rather than writing dedicated functors.

## Details

We actually already had a functor which could do this, used for creating Verlet Lists. So this functor is really just an expansion of that. In particular, I

- Created the distinction between AutoPas using newton3 to create the lists (i.e. like using it with the LJFunctor) and whether the lists created are newton3 lists. This allows the user to request newton3 lists or not, regardless of the algorithmic configuration AutoPas uses. We restrict AutoPas to use only newton3 enabled when creating newton3 lists, because supporting newton3 disabled was getting ugly and I think ultimately unneeded.
- Created an extensive testing suite, which didn't exist before.

## Open Question

This functor is kept in `baseFunctors`, but I am unsure about this. `baseFunctors` is otherwise entirely internal. Typically such public-facing functors are kept in `applicationLibrary`, but we can't do this as we need this functor internally.

Three options:
- Keep as it is, in `baseFunctors`
- Move the functor in line with the `AutoPas.h` file i.e. in line with the other public-facing files in the repo.
- Create a file in `applicationLibrary`, that serves really just as an alias for this functor (which could be kept somewhere like `utils`). This gives us something public-facing where we'd expect it to be, but gives something internal for Verlet Lists to use.

# How Has This Been Tested?

- [ ] CI
- [ ] New tests


src/autopas/baseFunctors/InteractionListGeneratorFunctor.h
```diff
@@ -0,0 +1,323 @@
...
+   * @param newton3 Whether AutoPas is using N3 or not for list generation. This is regardless of whether the user
+   * requests N3 lists or not. If this and gatherNewton3Lists match, the handling is trivial. If newton3=true and
+   * gatherNewton3Lists=false, we just add each particle to each other's list. We do not allow newton3=false with
+   * gatherNewton3Lists=true - there are no fundamental issues with this but messy to implement and probably not too
+   * needed.
+   */
+  void AoSFunctor(Particle_T &i, Particle_T &j, bool newton3) override {
+    using namespace autopas::utils::ArrayMath::literals;
+
+    [[unlikely]] if (_gatherNewton3Lists and not newton3) {
```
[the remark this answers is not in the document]
SamNewcome: Done in 62582e17

src/autopas/baseFunctors/InteractionListGeneratorFunctor.h
```diff
@@ -0,0 +1,323 @@
...
+   * @param soa1 soa of first cell
+   * @param soa2 soa of second cell
+   * @param newton3 Whether AutoPas is using N3 or not for list generation. This is regardless of whether the user
+   * requests N3 lists or not. If this and gatherNewton3Lists match, the handling is trivial. If newton3=true and
+   * gatherNewton3Lists=false, we just add each particle to each other's list. We do not allow newton3=false with
+   * gatherNewton3Lists=true - there are no fundamental issues with this but messy to implement and probably not too
+   * needed.
+   */
+  void SoAFunctorPair(SoAView<SoAArraysType> soa1, SoAView<SoAArraysType> soa2, bool newton3) override {
+    [[unlikely]] if (_gatherNewton3Lists and not newton3) {
```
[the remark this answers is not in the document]
SamNewcome: Done in 62582e17

src/autopas/baseFunctors/InteractionListGeneratorFunctor.h
```diff
@@ -0,0 +1,323 @@
...
+   * @param verletList indices of the potential neighbors of indexFirst
+   * @param newton3 Whether AutoPas is using N3 or not for list generation. This is regardless of whether the user
+   * requests N3 lists or not. If this and gatherNewton3Lists match, the handling is trivial. If newton3=true and
+   * gatherNewton3Lists=false, we just add each particle to each other's list. We do not allow newton3=false with
+   * gatherNewton3Lists=true - there are no fundamental issues with this but messy to implement and probably not too
+   * needed.
+   */
+  void SoAFunctorVerlet(SoAView<SoAArraysType> soa, const size_t indexFirst,
+                        const std::vector<size_t, AlignedAllocator<size_t>> &verletList, bool newton3) override {
+    [[unlikely]] if (_gatherNewton3Lists and not newton3) {
```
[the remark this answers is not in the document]
SamNewcome: Done in 62582e17

src/autopas/baseFunctors/InteractionListGeneratorFunctor.h
```diff
@@ -0,0 +1,323 @@
...
+   * @return
+   */
+  bool allowsNewton3() override { return true; }
+
+  /**
+   * Whether InteractionListGeneratorFunctor allows non-newton3. This is not allowed if not gathering N3 lists. (Not
+   * a fundamental issue but messy implementation and not really needed).
+   * @return
+   */
+  bool allowsNonNewton3() override { return not _gatherNewton3Lists; }
```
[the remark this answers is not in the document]
SamNewcome: Done in 62582e17

src/autopas/baseFunctors/InteractionListGeneratorFunctor.h
```diff
@@ -0,0 +1,323 @@
...
+
+        if (dr2 < _interactionLengthSquared) {
+          // This SoAFunctorSingle implementation always used newton3 in practice, regardless of the option.
+          _neighborListsAoS.at(ptrptr[i]).push_back(ptrptr[j]);
+          if (not _gatherNewton3Lists) {
+            _neighborListsAoS.at(ptrptr[j]).push_back(ptrptr[i]);
+          }
+        }
+      }
+    }
```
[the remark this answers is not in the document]
SamNewcome: Done in 62582e17

src/autopas/baseFunctors/InteractionListGeneratorFunctor.h
```diff
@@ -0,0 +1,323 @@
...
+        const double dr2 = drx2 + dry2 + drz2;
+
+        if (dr2 < _interactionLengthSquared) {
+          _neighborListsAoS.at(ptr1ptr[i]).push_back(ptr2ptr[j]);
+          if (newton3 and not _gatherNewton3Lists) {
+            _neighborListsAoS.at(ptr2ptr[j]).push_back(ptr1ptr[i]);
+          }
+        }
+      }
+    }
```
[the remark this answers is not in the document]
SamNewcome: Done in 62582e17

src/autopas/baseFunctors/InteractionListGeneratorFunctor.h
```diff
@@ -0,0 +1,321 @@
+/**
+ * @file InteractionListGeneratorFunctor.h
+ * @author seckler
+ * @date 27.04.18
```
TumBeat, line 4: This should be updated accordingly

src/autopas/baseFunctors/InteractionListGeneratorFunctor.h
```diff
@@ -0,0 +1,321 @@
...
+
+        const double drx2 = drx * drx;
+        const double dry2 = dry * dry;
+        const double drz2 = drz * drz;
+
+        const double dr2 = drx2 + dry2 + drz2;
+
+        if (dr2 < _interactionLengthSquared) {
+          // This SoAFunctorSingle implementation always used newton3 in practice, regardless of the option.
+          iList.push_back(ptrptr[j]);
```
TumBeat, line 185: I would add a comment here something like "thread safe because see comment for AoSFunctor"

src/autopas/baseFunctors/InteractionListGeneratorFunctor.h
```diff
@@ -0,0 +1,321 @@
...
+        const double drz = z1ptr[i] - z2ptr[j];
+
+        const double drx2 = drx * drx;
+        const double dry2 = dry * dry;
+        const double drz2 = drz * drz;
+
+        const double dr2 = drx2 + dry2 + drz2;
+
+        if (dr2 < _interactionLengthSquared) {
+          iList.push_back(ptr2ptr[j]);
```
TumBeat, line 244: Same comment about thread safety here

manishmishra6016 (commented):
Just some quick points/ questions: 

1. As far as I see, this does not affect how AutoPas works existing cases (md-flexible, ls1). right? So, that's good! 
2. I agree it might be useful to make the Neigh. List accessible to the simulator for integrating with existing ones. What I didn't get here is how the simulator accesses the generated NL directly for collision detection + force computation? Is there a public function to simply iterate over the neighbors (without rebuilding the list, or is that a todo)? `computeInteraction` will rebuild the NL again, right?
3. I think generating list separately is also nice when you want to reuse the list more than once. Currently, in the case of calling the `computeInteraction`, we will generate the list twice. Am I right?
4. `applicationLibrary` is application-specific, and functors stored there rely on Pairwise and Triwise functors in `baseFunctors`, and this one fits here better as it is not application-specific. Similarly, ParticleBase is in AutoPas and an application-specific particle class in `applicationLibrary`. So, imo, this is fine here. Does the argument make sense?

Changes look fine to me overall, although I haven't looked in detail.

src/autopas/baseFunctors/InteractionListGeneratorFunctor.h
```diff
@@ -0,0 +1,321 @@
...
+      const double drz = zFirst - zptr[j];
+
+      const double drx2 = drx * drx;
+      const double dry2 = dry * dry;
+      const double drz2 = drz * drz;
+
+      const double dr2 = drx2 + dry2 + drz2;
+
+      if (dr2 < _interactionLengthSquared) {
+        firstList.push_back(ptrptr[j]);
```
TumBeat, line 304: Same comment about thread safety

SamNewcome:
> Just some quick points/ questions:
> 
>     1. As far as I see, this does not affect how AutoPas works existing cases (md-flexible, ls1). right? So, that's good!

Yes. I thought about an md-flexible implementation to show off how it is, but decided overall my time is better spent on other things. Can always do this in a future PR.

> 
>     2. I agree it might be useful to make the Neigh. List accessible to the simulator for integrating with existing ones. What I didn't get here is how the simulator accesses the generated NL directly for collision detection + force computation? Is there a public function to simply iterate over the neighbors (without rebuilding the list, or is that a todo)? `computeInteraction` will rebuild the NL again, right?

So they would provide the empty, initialised map when constructing the functor, which would get filled. It is then up to the user what they do with this. E.g. MercuryDPM already has a function to produce such lists, which we would replace with AutoPas compute interactions with this functor. They then loop through the particles in a separate pass for force calculation, using the list (very similar to how our Verlet List List Iteration works, but with thread buffers to use newton3). I think ms2 is similar, although I don't remember exact details. 

I don't want to provide a public function to iterate over the neighbours, as I believe that ultimately this 2-pass system is inefficient on CPU and infeasible on GPU so don't want to encourage it, and for MercuryDPM they already have that functionality. But we could do this in a future PR/provide a simple example.

> 
>     3. I think generating list separately is also nice when you want to reuse the list more than once. Currently, in the case of calling the `computeInteraction`, we will generate the list twice. Am I right?

Do you mean if we wanted to call `computeInteraction` twice for different functors versus once with this and then handling the two forces separately? This could be worthwhile, but I think ultimately somehow integrating the two force calculations into one functor call is still better.

> 
>     4. `applicationLibrary` is application-specific, and functors stored there rely on Pairwise and Triwise functors in `baseFunctors`, and this one fits here better as it is not application-specific. Similarly, ParticleBase is in AutoPas and an application-specific particle class in `applicationLibrary`. So, imo, this is fine here. Does the argument make sense?

Yes, it makes sense.

> 
> 
> Changes look fine to me overall, although I haven't looked in detail.


Thanks for the review!

src/autopas/baseFunctors/InteractionListGeneratorFunctor.h
SamNewcome: This is actually correct, but the git change is a bit messed up so shows as a new file.

TumBeat:
About the CPU-GPU-communication, yes, this will require CPU-GPU communication in every step if the returned neighbor lists are not a data structure that can be stored on the GPU (std::map<..., std::vector<>> does definitely not work on the GPU because they both manage their own memory). I am currently thinking about implementing something like an AutoPasVector oriented on Kokkos::DualViews (with the main idea of hiding how and where data is stored from the user) but with similar functionality like an std::vector (push_backs etc.).
This can come in handy for avoiding data transfers as part of the updateContainer() interface which so far returns a std::vector of the migrating particles in every iteration. Here, we could provide the possibility to loop over the neighbor lists as a feature (similar to AutoPas' forEach interface which I am also currently porting to Kokkos).

But this will definitely require some more thoughts also in cooperation with David as this will be closely linked to domain decomposition and stuff.

So my honest suggestion would be to not care about possible incompatibilities with GPU-code for now as this will have to change anyway.

thesamriel (changes requested):
I like the code changes, but conceptually I have similar concerns to what you mentioned in the open questions.

Having the comments on contact detection in this internal file are a bit confusing because the context is missing.

I would prefer the internal Functor to be used for (internal) neighbor list generation only and then have an extra functor for contact detection in the `applicationLibrary`, with a separate explanation on how this functor  can be used for that application.
There, you could also override the `isRelevantForTuning()` function and would no longer need the `isInternal` template parameter.

SamNewcome:
> I like the code changes, but conceptually I have similar concerns to what you mentioned in the open questions.
> 
> Having the comments on contact detection in this internal file are a bit confusing because the context is missing.
> 
> I would prefer the internal Functor to be used for (internal) neighbor list generation only and then have an extra functor for contact detection in the `applicationLibrary`, with a separate explanation on how this functor can be used for that application. There, you could also override the `isRelevantForTuning()` function and would no longer need the `isInternal` template parameter.

Thanks! This internal/external confusion was somewhat my concern as well. But I'd be reluctant in creating an entirely separate new functor when the two do the same thing.

"contact detection" is a bit confusing as this is really DEM terminology. I just mean neighbour identification i.e. checking what's within cutoff range. I shouldn't have put this term in for a functor designed to be used for all of MD/DEM/SPH/etc.... I will replace this with "Neighbor Identification" as this is, as far as I am aware, more universally understood across MD/DEM/SPH.

One possibility is to create in `applicationLibrary` an alias to the internal functor, which automatically sets the `isInternal` parameter to false so this is hidden completely from the user. This way, we basically expose a public-facing functor alias that uses the internal functor, without reimplementing anything. The alias definition would include any comments on how to use the functor. The internal functor then includes no comments about how it is used externally. Do you think this would resolve the confusion?

src/autopas/baseFunctors/InteractionListGeneratorFunctor.h
SamNewcome: Done in 60950928

src/autopas/baseFunctors/InteractionListGeneratorFunctor.h
SamNewcome: Done in 60950928

src/autopas/baseFunctors/InteractionListGeneratorFunctor.h
SamNewcome: Done in 60950928

thesamriel:
> "contact detection" is a bit confusing as this is really DEM terminology. I just mean neighbour identification i.e. checking what's within cutoff range. I shouldn't have put this term in for a functor designed to be used for all of MD/DEM/SPH/etc.... I will replace this with "Neighbor Identification" as this is, as far as I am aware, more universally understood across MD/DEM/SPH.
> 
> One possibility is to create in `applicationLibrary` an alias to the internal functor, which automatically sets the `isInternal` parameter to false so this is hidden completely from the user. This way, we basically expose a public-facing functor alias that uses the internal functor, without reimplementing anything. The alias definition would include any comments on how to use the functor. The internal functor then includes no comments about how it is used externally. Do you think this would resolve the confusion?

Yeah, I think that's a very valid solution!

SamNewcome:
> About the CPU-GPU-communication, yes, this will require CPU-GPU communication in every step if the returned neighbor lists are not a data structure that can be stored on the GPU (std::map<..., std::vector<>> does definitely not work on the GPU because they both manage their own memory). I am currently thinking about implementing something like an AutoPasVector oriented on Kokkos::DualViews (with the main idea of hiding how and where data is stored from the user) but with similar functionality like an std::vector (push_backs etc.). This can come in handy for avoiding data transfers as part of the updateContainer() interface which so far returns a std::vector of the migrating particles in every iteration. Here, we could provide the possibility to loop over the neighbor lists as a feature (similar to AutoPas' forEach interface which I am also currently porting to Kokkos).
> 
> But this will definitely require some more thoughts also in cooperation with David as this will be closely linked to domain decomposition and stuff.
> 
> So my honest suggestion would be to not care about possible incompatibilities with GPU-code for now as this will have to change anyway.

I removed the GPU discussion in 60950928, and replaced with a very general statement about the use of the functor not taking full advantage of AutoPas.

SamNewcome:
@thesamriel I created a public-facing version of the functor `NeighborIdentificationFunctor` in the application library. I went for a child class over an alias, to handle the functor name and I think it's a little more obvious for users if they can see the constructor in their public-facing class. Done in 24512f40. The contact detection -> neighbor identification terminology change was done in 60950928

thesamriel (approved):

TumBeat (approved):

PR #1160, opened 2026-07-23
Changes to the CMake for inclusion in MercuryDPM

SamNewcome:
# Description

Changes AutoPas's CMake handling primarily to make it suitable as a dependency for MercuryDPM, but also for other projects.

In particular, this PR:

- Makes dependencies by default not force-bundled, and we use parent-project dependency targets instead where they exist
- Wraps `CMAKE_BUILD_TYPE` force-setting to avoid this leaking into parent projects. 

### Dependency handling

Makes dependencies by default not force-bundled. This is important because if the consumer (i.e. the parent project e.g. MercuryDPM, ls1-MarDyn) provides their own version of that dependency, we will end up with two versions of one dependency, violating the One Definition Rule (in the MercuryDPM case, there would be two versions of Eigen, i.e. two definitions of all the functions, classes, etc... shared between the two versions).

We instead:

- By default use not force-bundled
- We check if there is already a target defined for that dependency, provided by the parent project. This then gets used.
- If not, we check for a system version with `find_package()`. We check this meets a minimum version required (i.e. like we have before for the non-bundled case).
- If we don't find the target above, or we force bundled, then we install the bundled version.

The exception is PMT. Because we need a custom version, we check if the parent project has provided a target, and provide a warning if this is the case. Otherwise, we use the bundled version. We remove the force bundled option because this is unnecessary: we want to use our own PMT, and another version is problematic.

An alternative to the above:

We keep force-bundled enabled by default, but regardless of if this is set or not, we check if there is a target defined by the parent project. If we force bundled and the target is defined already, we give a warning like for PMT.

Based on discussion with @thesamriel , I added a master switch to force bundle everything.

### Build type leak protection

- AutoPas no longer force sets a build type if not building a standalone project.
- As PMT makes the same mistake, we provide a little wrapper around its set up to stop it setting a build type by default

### Fixes Warning Suppression

Spotted during work for the review, I found that warnings were being suppressed in AutoPas through a `PUBLIC -w` compile option on PMT. In 1c9c84b0, I removed this warning suppression, corrected the labelling of PMT as third-party, suppressed PMT warnings, and corrected some warnings in our `EnergySensor`. In d1746bd4, I corrected some warnings which were suppressed, which arose from SoA Sorting Benchmarking #1156 .

## Related Pull Requests

- #PR 1158 - implements a functor needed for MercuryDPM. We base this branch off of that so there is one branch with everything we need for the mercury integration.

# How Has This Been Tested?

- [x] CI
- [x] Testing the AutoPas still compiles with force-bundled turned off or on for the dependencies.

## Summary by CodeRabbit

- **New Features**
  - Added optional standalone tools, including `ruleChecker` and `tuningLogToSQL`.
  - Added configurable example builds and clearer validation when required libraries are unavailable.
  - Added support for reusing parent-project or system-installed dependencies, with options to force bundled copies.
  - Added the `AutoPas::autopas` CMake target for consumer projects.
  - Added optional ALL load-balancer support for MD-Flexible.

- **Documentation**
  - Updated build requirements, dependency guidance, tool prerequisites, and load-balancer configuration instructions.


cmake/modules/autopas_dependency_helpers.cmake
```diff
@@ -0,0 +1,44 @@
+# Helpers for handling third-party dependency modules
+
+# Umbrella switch for the per-dependency <dep>_ForceBundled options
+option(AUTOPAS_FORCE_ALL_BUNDLED "Ignore all provided/installed dependencies and always use the bundled copies." OFF)
```
dmartin13, line 4: `AUTOPAS_FORCE_ALL_BUNDLED` sounds like it should apply to all dependencies, but ANTLR, GoogleTest, Harmony, and ALL still reuse parent-provided targets, if I see this correctly. 
(only PMT is documented as exception)

Could we either handle the option there as well or adjust the documentation to make the scope clearer?

cmake/modules/autopas_spdlog.cmake
```diff
@@ -1,63 +1,63 @@
...
-    target_include_directories(spdlog SYSTEM PUBLIC "${propval}")
-endif()
\ No newline at end of file
+if (NOT spdlog_ForceBundled AND NOT AUTOPAS_FORCE_ALL_BUNDLED)
+    # Path 1: reuse a spdlog target a parent project already defined. AUTOPAS_MIN_LOG_LVL is not applied
+    # here, as the target belongs to the parent project, which decides how its spdlog is compiled.
+    if (TARGET spdlog::spdlog OR TARGET spdlog)
+        message(STATUS "AutoPas: Reusing spdlog provided by parent project")
+        autopas_alias_dependency(spdlog spdlog::spdlog)
+        return()
```
dmartin13, line 19: If we find an spdlog in the parent project and then return immediately here, won't that skip the setting of `AUTOPAS_MIN_LOG_LVL`, which happens in the if statement further down?

cmake/modules/autopas_gtest.cmake
```diff
@@ -1,3 +1,9 @@
+# Reuse a googletest a parent project already provides (avoids a duplicate `gtest` target collision).
+if (TARGET gtest)
```
dmartin13: If the parent provides gtest but not gmock, we return here and gmock remains unavailable, even though the AutoPas tests link against it. Should we check for both targets before reusing the parent-provided GoogleTest?

dmartin13 (changes requested):
Overall nice!

I found only a few cases that could potentially lead to problems.

One of the larger things I noticed:
The bundled PMT requires CMake 3.17.5, while AutoPas still allows CMake 3.14. I can reproduce this with CMake 3.14.7, where configuration fails directly in `libs/pmt/CMakeLists.txt`:

```
-- pmt - using bundled version (commit 7a56fa3a) and patch
CMake Error at libs/pmt/CMakeLists.txt:1 (cmake_minimum_required):
  CMake 3.17.5 or higher is required.  You are running version 3.14.7
```

Maybe we should bump the minimum CMake version for AutoPas to a newer one?

SamNewcome:
> Maybe we should bump the minimum CMake version for AutoPas to a newer one?

Yeah. I would bump then to 3.18. This is still comfortably low - HSUper has 3.27 and CoolMUC4 has 3.30, and ls1 requires 3.24 and MercuryDPM requires 3.24.

cmake/modules/autopas_dependency_helpers.cmake
SamNewcome: You're right, that was sloppy of me. I've handled ANTLR and GoogleTest similarly in 64a43130, but I am leaving Harmony and ALL for now and will leave this convo unresolved as I want to see if we want to just remove Harmony and make ALL non-bundled.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.14 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.18 FATAL_ERROR)
```
thesamriel, line 1: Can you remove the redundant cmake policies below this?

Also, in line 32 following there's a snippet for cmake version < 3.17, which can probably be removed now

examples/CMakeLists.txt
```diff
@@ -1,14 +1,17 @@
...
 if (NOT AUTOPAS_BUILD_EXAMPLES)
     return()
 endif ()
 
+# md-flexible links molecularDynamicsLibrary and the sph examples link SPHLibrary. Without them the
+# examples fail to compile with missing headers, so demand them here rather than at compile time.
+if (NOT AUTOPAS_BUILD_MD_LIBRARY OR NOT AUTOPAS_BUILD_SPH_LIBRARY)
+    message(
+        FATAL_ERROR
+            "AUTOPAS_BUILD_EXAMPLES=ON requires AUTOPAS_BUILD_MD_LIBRARY=ON and AUTOPAS_BUILD_SPH_LIBRARY=ON."
```
thesamriel, line 11:
```suggestion
            "AUTOPAS_BUILD_EXAMPLES=ON requires AUTOPAS_BUILD_MD_LIBRARY=ON or AUTOPAS_BUILD_SPH_LIBRARY=ON."
```

cmake/modules/autopas_all.cmake
```diff
@@ -6,20 +6,25 @@ option(MD_FLEXIBLE_ENABLE_ALLLBL "Enable load balancing via ALL for MD-Flex" OFF
...
-    URL_HASH MD5=aa30549b3decce8df1ee82e90b198b04
-  )
-  
-  # Get ALL source and binary directories from CMake project
-  FetchContent_GetProperties(allfetch)
-  
-  if (NOT allfetch)
-    FetchContent_MakeAvailable(allfetch)
+  # Reuse an ALL target a parent project already provides; otherwise fetch the bundled copy.
+  if (TARGET ALL)
```
thesamriel, line 10: Just reminds me how annoying of a name "ALL"  is :D

CMakeLists.txt
thesamriel: One more reference also in `autopas_other-compileroptions.cmake` - line 18

cmake/modules/autopas_dependency_helpers.cmake
SamNewcome: In ff275bce, ALL was made non-bundled. Harmony is still a TODO and I'll leave this unresolved.

SamNewcome: In #1180 I remove Harmony.

cmake/modules/autopas_spdlog.cmake
SamNewcome: Good point. There is no nice way to handle this, as far as I am aware. One main issue is that AutoPas headers included in the consumer would compile log levels at the consumer's spd active level, so setting AutoPas's log level to something different would be largely unsuitable.

In 8f4f2221 (and minor correction in 0cf70ea1), I basically changed from using spd log's compile gate system to us providing our own. It's not ideal for us, but as AutoPas is designed to be consumed and I would want the consumer to have freedom with their spd log level, I think this is the best solution.

cmake/modules/autopas_gtest.cmake
SamNewcome: Handled in 64a43130. We check against gmock, which pulls in gtest. If a parent provides gtest only, it doesn't seem like there is a nice solution so we fatal error.

cmake/modules/autopas_dependency_helpers.cmake
```diff
@@ -0,0 +1,44 @@
...
+# becomes a forwarding INTERFACE target, which (unlike an ALIAS) also works when the existing target is a
+# non-global imported target from an outer scope (e.g. provided by a parent project). No-op if both, or neither, already exist.
+function(autopas_alias_dependency plain namespaced)
+    if (TARGET ${plain} AND NOT TARGET ${namespaced})
+        add_library(${namespaced} ALIAS ${plain})
+    elseif (TARGET ${namespaced} AND NOT TARGET ${plain})
+        add_library(${plain} INTERFACE)
+        target_link_libraries(${plain} INTERFACE ${namespaced})
+    endif ()
+endfunction()
```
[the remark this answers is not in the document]
SamNewcome: Dealt with in 088b2f64

cmake/modules/autopas_spdlog.cmake
SamNewcome: Plus minor correction in f70ac5a4

cmake/modules/autopas_spdlog.cmake
```diff
@@ -1,63 +1,63 @@
...
+    # Path 2: installed version
+    set(expectedVersion 1.17.0)
+    find_package(spdlog ${expectedVersion} QUIET)
+    if (spdlog_FOUND)
+        message(STATUS "spdlog - using installed system version ${spdlog_VERSION}")
+        autopas_promote_global(spdlog)
+        autopas_promote_global(spdlog::spdlog)
+        autopas_alias_dependency(spdlog spdlog::spdlog)
+        # System targets can only take INTERFACE options, and are already treated as SYSTEM headers by CMake
+        target_compile_options(spdlog INTERFACE -DSPDLOG_ACTIVE_LEVEL=SPDLOG_LEVEL_${AUTOPAS_MIN_LOG_LVL})
```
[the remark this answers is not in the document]
SamNewcome: No longer relevant as of f70ac5a4

src/autopas/CMakeLists.txt
```diff
@@ -23,6 +23,10 @@ endif ()
 
 add_library(autopas STATIC ${MY_SRC})
 
+# Namespaced alias so consumers (e.g. via FetchContent / add_subdirectory) can link a stable,
+# collision-safe target name and use `if (NOT TARGET AutoPas::autopas)` guards
+add_library(AutoPas::autopas ALIAS autopas)
```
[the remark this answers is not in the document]
SamNewcome: Done in 01b3a44a

cmake/modules/autopas_yaml-cpp.cmake
```diff
@@ -1,22 +1,28 @@
...
+# version via find_package, or the bundled copy in libs/yaml-cpp. Set yaml-cpp_ForceBundled=ON to force bundled.
+option(yaml-cpp_ForceBundled "Ignore any provided/installed yaml-cpp and always use the bundled copy." OFF)
+mark_as_advanced(yaml-cpp_ForceBundled)
 
-if (NOT ${yaml-cpp_ForceBundled})
+if (NOT yaml-cpp_ForceBundled AND NOT AUTOPAS_FORCE_ALL_BUNDLED)
+    # Path 1: reuse a yaml-cpp target a parent project already defined.
+    if (TARGET yaml-cpp OR TARGET yaml-cpp::yaml-cpp)
+        message(STATUS "AutoPas: Reusing yaml-cpp provided by parent project")
+        autopas_alias_dependency(yaml-cpp yaml-cpp::yaml-cpp)
```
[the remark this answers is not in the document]
SamNewcome: Handled in 43dc9b2f. We always error if force bundled and parent provided a dependency.

cmake/modules/autopas_highway.cmake
```diff
@@ -1,4 +1,33 @@
...
+        autopas_promote_global(hwy::hwy)
+        autopas_alias_dependency(hwy hwy::hwy)
+        return()
+    endif ()
+    message(STATUS "Highway - no installed version >= ${expectedVersion} found; using bundled copy")
+endif ()
+
+# Path 3 + fallback: bundled version.
+message(STATUS "Highway - using bundled version 1.4.0")
+
```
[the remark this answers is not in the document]
SamNewcome: Handled in 43dc9b2f.

CMakeLists.txt
SamNewcome: Done in f257db0f

applicationLibrary/CMakeLists.txt
```diff
@@ -2,189 +2,219 @@
...
 # --- sph ---
 
-file(
-        GLOB_RECURSE
-        LIB_SPH_SRC
-        "sph/SPHLibrary/*.h"
-        "sph/SPHLibrary/*.cpp"
+option_conditional(AUTOPAS_BUILD_SPH_LIBRARY
+        "Build the SPH application library. Opt-in when AutoPas is used as a subproject."
+        "AUTOPAS_STANDALONE_BUILD OR AUTOPAS_BUILD_EXAMPLES"
```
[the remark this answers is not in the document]
SamNewcome: Leaving as is - we only build the mdLib as its needed for internal AutoPas tests. For non-standalone builds, which do not want to use the sphLib or neighborIndentificationLib, building these and testing these is irrelevant.

cmake/modules/autopas_pmt.cmake
```diff
@@ -5,25 +5,23 @@ option(AUTOPAS_ENABLE_ENERGY_MEASUREMENTS
...
-    endif()
+    # AutoPas requires its own patched PMT (see the patch note below), so it never looks for an
+    # installed one. If a parent project already provides a `pmt` target, warn and reuse it: the parent
+    # must either disable AutoPas's energy measurements or drop its own `pmt` so ours is used.
+    if (TARGET pmt)
+        message(
+            WARNING
+                "AutoPas needs its own patched PMT, but a parent project already provides a 'pmt' target. Set AUTOPAS_ENABLE_ENERGY_MEASUREMENTS=OFF, or remove your 'pmt' target so AutoPas can build its patched version. Reusing the provided PMT for now; energy measurements may be incorrect."
+        )
+        return()
```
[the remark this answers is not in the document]
SamNewcome: No. This is maybe still valid so WARN don't kill

docs/userdoc/Building.md
```diff
@@ -14,7 +14,12 @@ Optional:
 There are a few more dependencies, however you don't need to install them because they come bundled with AutoPas.
 See [libs/](/libs) for a complete list.
 
-If you insist on using your locally installed version of any of them, set the CMake variable `<LIBNAME>_ForceBundled=OFF`
+By default, AutoPas prefers a version provided by a parent project or installed on your system over the bundled copy.
+To always use the bundled copy of one library, set `<LIBNAME>_ForceBundled=ON`, or set `AUTOPAS_FORCE_ALL_BUNDLED=ON` to do so for all of them.
+
+The one exception is [ALL](https://gitlab.jsc.fz-juelich.de/SLMS/loadbalancing), an optional load balancer for
+md-flexible that is not bundled. If you enable it with `MD_FLEXIBLE_ENABLE_ALLLBL=ON` and do not have it
+installed, CMake downloads it at configure time, which requires network access.
```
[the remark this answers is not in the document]
SamNewcome: No. md-flexible is an end consumer of AutoPas and therefore independent of a parent project even if it were to be built alongside.

examples/CMakeLists.txt
SamNewcome: Done in ccf0bd0c

dmartin13 (approved):

PR #1161, opened 2026-07-24
Prepare for transitioning dependencies to git subtrees.

thesamriel:
# Description

From #1149

> Currently, the dependencies we use are bundled as zips. This has a couple issues:
> 
> Updating a dependency requires manually downloading a particular commit or a package from the browser.
> Updating a dependency shows as a commit where one large file is deleted and another is added. My understanding is that this will bloat our repository, even if only one line is changed.
> You need to manually uncompress the dependency if you view the code.
> We sometimes need to also work on the dependency's code. Making experimental changes is quite painful with the current system. The main example of this already in master is PMT, but I found this particularly painful when developing with LB4OMP recently.
> With this PR, we switch these dependencies to Git subtrees. These are like mini repositories within our repository. Specifically, I used squashed versions of these repositories, so without git history. The main advantages of these are:
> 
> Repositories can be viewed directly.
> Updates are no longer add + remove compressed file. I think in the long run this will reduce git bloat.
> AutoPas specific patches are actual commits in these subtrees. When updating the dependency, these will treated like any other merge from upstream.
> When directly experimenting with changing a dependency's code, we can do so directly.

- The `CONTRIBUTING.md` file has been updated to reflect how we want to handle dependencies
- `.gitattributes` has already been configured to ignore the dependencies we plan to transition to git subtrees. This should make the reviewing of the next PRs easier.


.github/CONTRIBUTING.md
```diff
@@ -196,3 +196,69 @@ To update the grammar, simply run [`generateRuleLanguage.sh`](/src/autopas/tunin
...
 **WARNING** The CLion plugin and Antlr version must match!. For example for Antlr version 4.9.1 the plugin version 1.16 is needed, otherwise incompatible parser code is generated. See [Antlr plugin's GitHub page](https://github.com/antlr/intellij-plugin-v4/releases) for what is compatible and get the plugin's binary from the [jetbrains plugin webpage](https://plugins.jetbrains.com/plugin/7358-antlr-v4/versions).
+
+## Dependencies
+
+> [!NOTE]
+> AutoPas is currently transitioning its dependency management from bundled `.zip` archives to **git subtrees**.
+> The guidelines below describe the **new** git subtree workflow. Until all dependencies are migrated, they remain as `.zip` archives in [`libs/`](/libs) which are extracted at configure time.
+
+* Dependencies are kept in [`libs/`](/libs) as **git subtrees** of their upstream repositories whenever possible. 
+* If a dependency lacks a clean, standalone upstream git repository (for example, `libuuid` originates from the massive `util-linux` project, and pulling that entire tree would unnecessarily bloat the repository), it should be kept as a bundled `.zip` archive.
```
SamNewcome: So in #1149 , we dropped the `libuuid` dependency entirely, so I wouldn't include this as an example.

`antlr4cpp` is a very similar example though, but there in #1149 we handled this by our git subtree pruning (i.e. a git commit local to our git subtree which removed) the unnecessary parts of the larger repository. (`harmony` is similar, but only the examples directory is pruned away)

I would personally much prefer this over zips for all the reasons we originally did the change + it makes maintenance much easier as otherwise every time we update the dependency, we have to also prune away everything unnecessary again.

IMO, if there is something with genuinely no upstream, I'm a bit reluctant about that from a maintenance point of view.

SamNewcome (changes requested):
Nice. thanks! Just one comment

.github/CONTRIBUTING.md
thesamriel: Ah okay, I see. It somehow doesn't show that you dropped libuuid in #1149 for me. But it makes sense.
Not sure if we might ever have a non-git/upstream dependency, but if so we can discuss this then I suppose.

Adjusted the description in [54699a3](https://github.com/AutoPas/AutoPas/pull/1161/commits/54699a3f3a2ddad6ef3292f9670c7ca03bec648b)

SamNewcome: > It somehow doesn't show that you dropped libuuid in #1149 for me. 

Ah that's probably then a mistake of mine, sorry!

Thanks

SamNewcome (approved):
Looks good!

PR #1162, opened 2026-07-26
Improve Performance of Verlet Lists Traversal

### LINKED ISSUE #771: Too much memory movement during neighbor lists for verlet-list based containers
**Describe the bug**
Whenever we rebuild neighbor lists we call `clear()` on them first. This invokes the destructor on all elements which implies the deletion of many `std::vector` that are part of a `std::unordered_map`. --> This is expensive! 

Example: https://github.com/AutoPas/AutoPas/blob/8ae7ba5d754e124d5c8b81311eb81f1909e65509/src/autopas/containers/verletListsCellBased/varVerletLists/neighborLists/asBuild/VerletNeighborListAsBuild.h#L91-L96

**Expected behavior**
Do not free and reallocate the memory but keep the allocation and only extend the structure when necessary.

**Additional context**
Described issue affects probably all verlet-list based containers.
Fixed for:
- [x] VerletClusterLists #781
- [ ] VarVerletLists
- [ ] VerletLists
- [x] VerletListsCells

**Suggested Solution**
A structure that holds memory for a given number of neighbor lists and maps them to arbitrary particles.

thesamriel:
# Description

This PR improves the performance for our VerletLists container.
The main change includes a different data format for storing the neighbor lists.
Instead of a `std::unordered_map` we use a CRS-based data layout.
This improves neighbor list memory access.

Verlet Lists with newton3=true has been enabled and it is implemented via a basic 27-cell coloring scheme to avoid race conditions.

## Related Pull Requests

#1158 moved the NeighborListGeneratorFunctor to the base functors directory to also share the functionality with the `NeighborIdentificationFunctor`.
I reverted some of these changes here to better separate internal neighbor list functionality and external functors

- [] fixes #771 for VL

## Comparison

As an example, I ran `fallingDrop.yaml` with only VerletList options:
<details>
<summary> Config File </summary>

``` yaml
# This yaml file is for single-site molecular simulation. Uncomment the Molecules option to run this experiment using
## md-flexible compiled for multi-site molecules.
container                        :  [VerletLists]
# verlet-rebuild-frequency         :  10
verlet-skin-radius               :  1.0
data-layout                      :  [AoS, SoA]
traversal                        :  [vl_list_iteration] # Please see AllOptions.yaml for a comprehensive list of traversals
tuning-strategies                :  []
tuning-interval                  :  2500
tuning-samples                   :  3
tuning-max-evidence              :  10
functor                          :  lj-hwy
newton3                          :  [disabled, enabled]
cutoff                           :  3
box-max                          :  [7.25, 7.25, 7.25]
deltaT                           :  0.0005
iterations                       :  15000
energy-sensor                    :  rapl
boundary-type                    :  [reflective,reflective,reflective]
globalForce                      :  [0,0,-12]
Sites:
  0:
# Uncomment below to run a multi-site simulation.
#Molecules:
#  0:
#    site-types                   :  [ 0 ]
#    relative-site-positions      :  [ [0, 0, 0 ] ]
#    moment-of-inertia            :  [ 1., 1., 1. ]
Objects:
  # "water"
  CubeClosestPacked:
    0:  
      particle-spacing           :  1.122462048
      bottomLeftCorner           :  [1, 1, 1]
      box-length                 :  [48, 28, 10]
  Sphere:
    0:  
      center                     :  [18, 15, 30]
      radius                     :  6
      particle-spacing           :  1.122462048
# thermostat:
#   initialTemperature             :  1
#   targetTemperature              :  4
#   deltaTemperature               :  0.1
#   thermostatInterval             :  10
#   addBrownianMotion              :  false
# vtk-filename                     :  fallingDrop
# vtk-write-frequency              :  1000
# vtk-output-folder                :  vtkOutputFolder
log-level                        :  info
no-progress-bar                    :  true
```
</details>

On master: 45.870s
This PR:      29.130s

Below are some runtime comparisons from CoolMUC4, scaled up to only 56 threads, to keep all threads on one socket.
The graphs below compare a homogeneous scenario for different densities for aos and soa. The rebuild frequency is a fixed interval of 10.

image
image
image

image
image
image

## Summary by CodeRabbit

* **New Features**
  * Improved Verlet-list processing with a unified neighbor-list representation for AoS and SoA workflows.
  * Added policy-based neighbor-list handling for more flexible neighbor identification.
  * Added parallel neighbor-list construction and validation.
  * Improved Newton3 traversal handling and parallel execution.

* **Bug Fixes**
  * Corrected Verlet-list traversal setup to account for halo-cell dimensions.
  * Expanded coverage for Newton3, particle ownership, filtering, and multithreaded neighbor-list scenarios.


TumBeat:
> I reverted some of these changes here to better separate internal neighbor list functionality and external functors

To be honest, I am not too sure if this is really required, but I do not have an ultimate opinion on this.

As far as I can see, the VerletListGeneratorFunctor and the NeighborIdentificationFunctor only differ in a few lines of codes that do not necessarily have to be different. E.g. in the AoSFunctor in the IdentificationFunctor, we have

`_neighborListsAoS.at(&i).push_back(&j);`

and in the VerletListGeneratorFunctor, we have

`_neighborLists[_particleToIndex.at(&i)].push_back(_particleToIndex.at(&j));`

Since neighborListsAoS is templated anyway in the IdentificationFunctor, the datatype of the lists does not matter. The only main difference is the key of the neighbor lists (Particle* vs size_t). Maybe we can have another templated mapping in form of a lambda, something like this:

(GeneratorFunctor) `[](Particle* ptr) { return _particleToIndex.at(ptr); }`
(IdentificationFunctor) `[](Particle* ptr) { return ptr; }`

This also should keep the additional overhead for both versions low but results in the end in (hopefully) less code duplication if we keep the InteractionListGeneratorFunctor and instantiate this one from both routine (NeighborIdentificationFunctor and neighbor list rebuilding).
Additionally, this way we can separate preparing the neighbor list data structure (in Markus' case with the CSR format) from the way we identify the neighbors and fill the lists (traversal + functor) which "forces" us to provide a generic interface.

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctor.h
```diff
@@ -562,12 +562,8 @@ class LJFunctor
   // clang-format off
   /**
    * @copydoc autopas::PairwiseFunctor::SoAFunctorVerlet()
-   * @note If you want to parallelize this by openmp, please ensure that there
-   * are no dependencies, i.e. introduce colors!
    */
-  // clang-format on
-  void SoAFunctorVerlet(autopas::SoAView<SoAArraysType> soa, const size_t indexFirst,
-                        const std::vector<size_t, autopas::AlignedAllocator<size_t>> &neighborList,
+  void SoAFunctorVerlet(autopas::SoAView<SoAArraysType> soa, const size_t indexFirst, std::span<const size_t> neighborList,
```
SamNewcome: Any reason why the `clang-format on` was removed?

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
```diff
@@ -1106,16 +1106,136 @@ class LJFunctorHWY
...
-  * @copydoc autopas::PairwiseFunctor::SoAFunctorVerlet()
-  * @note If you want to parallelize this by openmp, please ensure that there
-  * are no dependencies, i.e. introduce colors and specify iFrom and iTo accordingly.
-  */
-  // clang-format on
-  inline void SoAFunctorVerlet(autopas::SoAView<SoAArraysType> soa, const size_t indexFirst,
-                               const std::vector<size_t, autopas::AlignedAllocator<size_t>> &neighborList,
+   * @copydoc autopas::PairwiseFunctor::SoAFunctorVerlet()
+   */
+  inline void SoAFunctorVerlet(autopas::SoAView<SoAArraysType> soa, const size_t indexFirst, std::span<const size_t> neighborList,
```
SamNewcome: Similar about clang format not turned back on

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorAVX.h
```diff
@@ -604,13 +604,8 @@ class LJFunctorAVX
...
   /**
    * @copydoc autopas::PairwiseFunctor::SoAFunctorVerlet()
-   * @note If you want to parallelize this by openmp, please ensure that there
-   * are no dependencies, i.e. introduce colors and specify iFrom and iTo accordingly.
    */
-  // clang-format on
-  inline void SoAFunctorVerlet(autopas::SoAView<SoAArraysType> soa, const size_t indexFirst,
-                               const std::vector<size_t, autopas::AlignedAllocator<size_t>> &neighborList,
-                               bool newton3) final {
+  inline void SoAFunctorVerlet(autopas::SoAView<SoAArraysType> soa, const size_t indexFirst, std::span<const size_t> neighborList, bool newton3) final {
```
SamNewcome: Similar clang-fromat

SamNewcome (changes requested):
Looks very nice and some very good performance gains. The code looks very clean. I haven't got super far into reviewing this properly yet but won't have more time until later this week and there is one more major structure feedback that I have that I think is necessary and will also speedup the code review:

Can you reverse the migration of the old InteractionListGeneratorFunctor -> NeighborListFunctor, to keep the structure we currently have? This way we reduce the code duplication, we can use anything you've developed here for that functor, we can use both your and my tests on the same functor. The CRS data format is probably better for the neighbour lists used externally to AutoPas, and if not I can write an adaptor or something in a later PR.

thesamriel:
> Looks very nice and some very good performance gains. The code looks very clean. I haven't got super far into reviewing this properly yet but won't have more time until later this week and there is one more major structure feedback that I have that I think is necessary and will also speedup the code review:
> 
> Can you reverse the migration of the old InteractionListGeneratorFunctor -> NeighborListFunctor, to keep the structure we currently have? This way we reduce the code duplication, we can use anything you've developed here for that functor, we can use both your and my tests on the same functor. The CRS data format is probably better for the neighbour lists used externally to AutoPas, and if not I can write an adaptor or something in a later PR.

I understand that you and also @TumBeat want to avoid code duplication, but to me it feels a bit awkward to mix internal and external usage for the functor.

If I revert the structure to what we have in master, could you check whether you are okay with the data format from the `VerletListGeneratorFunctor`? It requires a map mapping from particle pointers to their index (`_particleToIndex`) and neighbor lists in the format of `std::vector<std::vector<size_t>>`, which is not yet the CRS format.

SamNewcome:
> > Looks very nice and some very good performance gains. The code looks very clean. I haven't got super far into reviewing this properly yet but won't have more time until later this week and there is one more major structure feedback that I have that I think is necessary and will also speedup the code review:
> > Can you reverse the migration of the old InteractionListGeneratorFunctor -> NeighborListFunctor, to keep the structure we currently have? This way we reduce the code duplication, we can use anything you've developed here for that functor, we can use both your and my tests on the same functor. The CRS data format is probably better for the neighbour lists used externally to AutoPas, and if not I can write an adaptor or something in a later PR.
> 
> I understand that you and also @TumBeat want to avoid code duplication, but to me it feels a bit awkward to mix internal and external usage for the functor.
> 
> If I revert the structure to what we have in master, could you check whether you are okay with the data format from the `VerletListGeneratorFunctor`? It requires a map mapping from particle pointers to their index (`_particleToIndex`) and neighbor lists in the format of `std::vector<std::vector<size_t>>`, which is not yet the CRS format.

I'm a bit reluctant with the split from a maintainability perspective. If you ignore the exact data format, the two do identical things. Even in this PR you can see the issue - the internal functor had a unit test suite and now doesn't appear to have any unit tests of its own. There is also the possiblity this index-based format could be useful in the public-facing functor. Tbh, this wouldn't be the case with MercuryDPM as that is a AoS-based code, but for a hypothetical SoA-based code that wanted to do their neighbour-list generation with AutoPas, they would actively want this functionality.

My ideal solution would therefore be something either like what @TumBeat  suggested or alternatively, you could add have both neighbour list types as members, the original AoS/ptr-lists constructor in addition to the SoA-lists constructor, a simple template parameter bool with constexpr if statements to determine which list is actually added to. 

These solutions would have the added benefit that you wouldn't need the map you construct at the start of `initTraversal` to convert your CRF-index lists to a pointer-based list, you could simply just use the functor in AoS-list-generation mode.

If the CRF-stuff would be ugly for a public-facing functor, I'd be happy with the public-facing functor being limited to pointer-lists for now, as this would at least make a possible extension relatively easy.


If you still want to keep the split, then I'm okay, but I think you should then add the neighbour identification functor tests to test this functor (adapting list format accordingly + removing irrelevant tests).

TumBeat:
> If I revert the structure to what we have in master, could you check whether you are okay with the data format from the `VerletListGeneratorFunctor`? It requires a map mapping from particle pointers to their index (`_particleToIndex`) and neighbor lists in the format of `std::vector<std::vector<size_t>>`, which is not yet the CRS format.

I think this should also not be a problem since the neighbor lists are templated. The Functor could then have the templated neighbor list (can be of type `std::vector<std::vector<size_t>>` or `std::vector<std::vector<Particle*>>` and an additional templated function in the form of

`template <typename Result, typename Lambda>
Result getMappingKey (Particle* ptr, Lambda mapping) {
   return mapping(ptr);
}`

The mapping lambda can then either be `[](Particle* ptr) { return _particleToIndex.at(ptr); }` or `[](Particle* ptr) { return ptr; }` (here, the _particleToIndex mapping can be just an empty list that has to exist for compatibility).

Disclaimer: I haven't checked the exact syntax of this, I just encountered that C++ has some (weird/annoying/...) constraints regarding lambdas. Just an idea, would need proper testing :P

Additionally, with this approach we might even get rid of the VerletListFillerFunctor since the structure is also the same. However, I have to admit that replacing `_neighborList.indices[_fillPos[iIdx].value.fetch_add(1, std::memory_order_relaxed)] = jIdx;` with some lambda could be more complicated.

thesamriel:
@SamNewcome 
I added a `NeighborListPolicy` template parameter to the `InteractionListGeneratorFunctor`, which defines how a particle is added to another particle's neighbor list.
The Verlet Lists policy for the CRS lists would be here: https://github.com/AutoPas/AutoPas/blob/6149b242e6b3578e0fcf81eb5bafb51196534a2c/src/autopas/containers/verletListsCellBased/verletLists/VerletListHelpers.h#L69
The aos  version with unordered map here: https://github.com/AutoPas/AutoPas/blob/6149b242e6b3578e0fcf81eb5bafb51196534a2c/applicationLibrary/neighborIdentification/neighborIdentificationLibrary/NeighborIdentificationFunctor.h#L23

Would this solution be okay? It goes in the direction of what @TumBeat suggested.

SamNewcome:
> Would this solution be okay?

Yes, this is great! Thanks!

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctor.h
thesamriel: This was on accident, but I removed both off and on now, since the CI seems to no longer have issues with it.

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
SamNewcome: I would try to refactor the rest-part into the verlet kernel. I think this will be cleaner and more performant. But I would leave this up to you and this can just be added as an issue.

applicationLibrary/neighborIdentification/tests/NeighborIdentificationFunctorTest.cpp
SamNewcome, line 1: The Neighbor List Initialization and End-to-End tests use an actual compilation of AutoPas with the functor. For this reason, originally we had an explicit template initialisation - add this to the application library tests or move these tests back to the internal tests

src/autopas/baseFunctors/InteractionListGeneratorFunctor.h
```diff
@@ -6,44 +6,47 @@
...
 #include "autopas/utils/ArrayMath.h"
 #include "autopas/utils/SoA.h"
 namespace autopas {
 
+/**
+ * Concept defining the requirements for a NeighborListPolicy used by the InteractionListGeneratorFunctor.
+ * @tparam Policy The policy type to check.
+ * @tparam Particle_T The particle type the policy operates on.
+ */
+template <class Policy, class Particle_T>
```
SamNewcome:
```suggestion
template <class Policy_T, class Particle_T>
```

src/autopas/baseFunctors/InteractionListGeneratorFunctor.h
```diff
@@ -6,44 +6,47 @@
...
 /**
  * This functor generates lists of particles within interactionLength of each other: can be used internally for Verlet
  * list generation and provides a base for a public-facing externally used version, provided in the applicationLibrary.
  *
- * @details After applying AutoPas's computeInteractions function with this functor, a
- * std::unordered_map<Particle_T *, std::vector<Particle_T *>> is filled, mapping from each particle pointer to a vector
- * of pointers to all particles within interactionLength of the first.
+ * @details After applying AutoPas's computeInteractions function with this functor, a std::vector<std::vector<size_t>>
+ * is filled, mapping from each particle index to a vector of indices belonging to all particles within
+ * interactionLength of the first.
```
SamNewcome: Make this generalised for different policies.

src/autopas/baseFunctors/InteractionListGeneratorFunctor.h
```diff
@@ -126,15 +111,14 @@ class InteractionListGeneratorFunctor
...
 
-    double distsquare = utils::ArrayMath::dot(dist, dist);
-    if (distsquare < _interactionLengthSquared) {
-      // Assuming this functor is used like any other functor, this is thread safe: _neighborListsAoS is an
-      // unordered_map, meaning we can push_back to particle i's list with the same thread safety as for writing to its
-      // force buffer in e.g. a LJ functor.
+    double distSquared = utils::ArrayMath::dot(dist, dist);
+    if (distSquared < _interactionLengthSquared) {
+      // Assuming this functor is used like any other functor, this is thread safe: _neighborLists is a vector of
+      // vectors, meaning we can push_back to particle i's list with the same thread safety as for writing to it force
```
SamNewcome: Generalise this comment for different policies

src/autopas/baseFunctors/InteractionListGeneratorFunctor.h
```diff
@@ -126,15 +111,14 @@ class InteractionListGeneratorFunctor
...
+      // buffer in e.g. a LJ functor.
 
-      // This is only thread-safe if all keys are inserted prior to applying this functor, as a rehash might lead to
-      // dangling references, but at() protects against this: if the key exists, we can push_back safely; if a key
-      // doesn't, an exception is thrown anyway.
+      // This is only thread-safe if a vector (neighbor list) exists for all particles and their corresponding indices.
+      // These indices as stored in _particleToIndex must not exceed the total number of particles.
```
SamNewcome: Sim.

src/autopas/containers/verletListsCellBased/verletLists/traversals/VLListIterationTraversal.h
```diff
@@ -43,14 +50,31 @@ class VLListIterationTraversal : public TraversalInterface, public VLTraversalIn
...
+    if (_useNewton3) {
+      // Initialize/clear color cells lists
+      for (auto &colorGroup : _colorCells) {
+        colorGroup.clear();
+      }
+
+      for (size_t c = 0; c < cells.size(); ++c) {
+        if (offsets[c] < offsets[c + 1]) {
+          auto c3D = utils::ThreeDimensionalMapping::oneToThreeD(c, _cellsPerDim);
+          const size_t color = (c3D[0] % 3) + 3 * (c3D[1] % 3) + 9 * (c3D[2] % 3);
```
SamNewcome: This is basically a C27 colouring, right? Might be good to write a brief comment.

src/autopas/containers/verletListsCellBased/verletLists/traversals/VLListIterationTraversal.h
```diff
@@ -72,31 +96,38 @@ class VLListIterationTraversal : public TraversalInterface, public VLTraversalIn
...
-          for (size_t bucketId = 0; bucketId < buckets; bucketId++) {
-            auto endIter = aosNeighborLists.end(bucketId);
-            for (auto bucketIter = aosNeighborLists.begin(bucketId); bucketIter != endIter; ++bucketIter) {
-              ParticleType &particle = *(bucketIter->first);
-              for (auto neighborPtr : bucketIter->second) {
-                ParticleType &neighbor = *neighborPtr;
-                _functor.AoSFunctor(particle, neighbor, false);
-              }
+          // Each particle i owns its own list slice — no write conflict between iterations.
+          AUTOPAS_OPENMP(parallel for schedule(static))
```
SamNewcome: I'm not against this shift from `dynamic`, but it might be good to justify it with some data. I am concerned that this will be slower for inhomogeneous scenarios. But I don't care too much if this changes.

I will probably look at this when I look at my OpenMP stuff again, so I don't actually mind what you set this to and feel free to not justify this. If you just want a vaguely sensible guess based very roughly on my preliminary findings, my recommendation would be something like `schedule(static, 8)` - will be better load balanced but some degree of data reuse still.

And to match Manish's points, whatever you do I'd keep things consistent.

src/autopas/containers/verletListsCellBased/verletLists/VerletLists.h
```diff
@@ -91,141 +90,200 @@ class VerletLists : public VerletListsLinkedBase<Particle_T> {
...
-   * Fills SoA neighbor list with particle indices.
+   * Single-threaded rebuild: One traversal with VerletListGeneratorFunctor writing into per-particle
+   * std::vector<size_t>, followed by a serial prefix-sum + copy into the flat CRS.
    */
-  void generateSoAListFromAoSVerletLists() {
-    // resize the list to the size of the aos neighborlist
-    _soaNeighborLists.resize(_aosNeighborLists.size());
-    // clear the aos 2 soa map
-    _particlePtr2indexMap.clear();
+  void updateNeighborListsSingleThread(size_t N, double interactionLength, DataLayoutOption dataLayout,
```
manishmishra6016, line 172: `updateNeighborListsMultiThread()` is under private scope, but this is not; this could also be moved there.

src/autopas/containers/verletListsCellBased/verletLists/VerletLists.h
```diff
@@ -91,141 +90,200 @@ class VerletLists : public VerletListsLinkedBase<Particle_T> {
...
+   *   Compute CRS offsets; allocate the flat indices array.
+   *
+   * Pass 2 (parallel, VerletListFillerFunctor):
+   *   Write neighbor indices directly into the pre-allocated CRS slice
+   *   via per-particle atomic fetch-add fill cursors (also padded).
+   *
+   * Optimal when autopas_get_max_threads() > 1.
    */
-  InteractionListGeneratorFunctor<Particle_T, true>::NeighborListAoSType _aosNeighborLists;
+  void updateNeighborListsMultiThread(size_t N, double interactionLength, DataLayoutOption dataLayout,
```
manishmishra6016, line 220: is this significantly worse than `updateNeighborListsSingleThread()` for serial case? If not, we could just keep this one function! $\rightarrow$ Very rarely someone would run a serial case, even on personal laptops.

src/autopas/containers/verletListsCellBased/verletLists/traversals/VLListIterationTraversal.h
manishmishra6016:
```suggestion
          AUTOPAS_OPENMP(parallel for schedule(dynamic))
```
We have `dynamic` in the old code as well as everywhere else in the changes, so maybe we also change this for consistency.

tests/testAutopas/tests/containers/verletListsCellBased/verletLists/VerletListHelpersTest.cpp
```diff
@@ -0,0 +1,190 @@
...
+ */
+TEST_F(VerletListHelpersTest, FillerFunctorThreadSafety) {
+  constexpr size_t numParticles = 1000;
+  generateDenseParticles(numParticles);
+
+  typename Helpers::NeighborListCRS neighborList;
+
+  // Simulate Prefix Sum
+  neighborList.offsets.resize(numParticles + 1, 0);
+  for (size_t i = 0; i < numParticles; ++i) {
```
manishmishra6016, line 64:
```suggestion
  neighborList.offsets.resize(numParticles - 1, 0);
  for (size_t i = 0; i < numParticles; ++i) {
```
Looking at the loop, it seems like we only need `numParticles -1`

manishmishra6016 (approved):
Looks pretty good! I just had some very minor comments, so you can have a look and see what you think.

src/autopas/containers/verletListsCellBased/verletLists/traversals/VLListIterationTraversal.h
SamNewcome: I have a very annoying comment (sorry!). The VLListIterationTraversal is now effectively two traversals: whether you use N3 or not, this traversal behaves now very differently! With N3 off, this is an iteration over every list; and with N3 on, this is a C27 traversal. 

This causes a few problems ranging from pedantic to actually awkward: 

- the name VLListIterationTraversal is then not quite true for newton3 enabled -> you are not iterating in parallel over the lists
- if we were to in a paper describe the traversal, we would have to give different explanations for N3 on or off which we don't have to do for any other traversal
- if we were disabled by default N3-off versions of N3-capable traversals (which makes sense for all other true N3 capable traversals), we would then loose the true VL list iteration traversal which is very useful in scenarios where we want particle level parallelisation (e.g. high load imbalance), where a cell-based traversal performs poorly

I would strongly suggest a separate traversal for this, to separate the two fundamentally different concepts for applying shared-memory parallelisation. 

Also, it feels to me like a VLC traversal implemented in the standard Verlet List container, and I would generally expect C27 to be outperformed by C08. Does this traversal actually outperform VLC traversals?

src/autopas/containers/verletListsCellBased/verletLists/VerletLists.h
thesamriel: The multi-threaded version requires two C08 traversals vs. just one in the single-threaded case.
So, it is quite a difference, but it does kind of "bloat" the class a bit :/

tests/testAutopas/tests/containers/verletListsCellBased/verletLists/VerletListHelpersTest.cpp
thesamriel: We write to `neighborList.offsets[i+1]`, that's what we need the + 1 for.

thesamriel:
@coderabbitai review

thesamriel:
@SamNewcome I refactored the newton3 version to be its own traversal (`VLListIterationC27Traversal`, because just `VLC27Traversal` looks like a verlet list cells traversal).
I also noticed that the graphs in the PR description already depicts both versions (noN3 and N3 with C27) in comparison to master.

src/autopas/containers/verletListsCellBased/verletLists/traversals/VLListIterationTraversal.h
```diff
@@ -72,31 +72,31 @@ class VLListIterationTraversal : public TraversalInterface, public VLTraversalIn
...
-              ParticleType &neighbor = *neighborPtr;
-              _functor.AoSFunctor(particle, neighbor, _useNewton3);
+          // Newton3 cannot be parallelized here
+          for (size_t i = 0; i < numParticles; ++i) {
+            ParticleType &particleI = *indexToParticle[i];
+            const size_t numNeighbors = neighborList.count(i);
+            const size_t *neighborsIPtr = neighborList.begin(i);
+            for (size_t j = 0; j < numNeighbors; ++j) {
+              _functor.AoSFunctor(particleI, *indexToParticle[neighborsIPtr[j]], true);
             }
```
SamNewcome, line 107: This is not really part of your PR as the issue existed beforehand, so feel free to ignore. We should never access this branch, as this is not a valid configuration. It might be cleaner to switch this to an exception.

src/autopas/containers/verletListsCellBased/verletLists/VerletListHelpers.h
```diff
@@ -0,0 +1,432 @@
...
+    bool allowsNewton3() override { return true; }
+    bool allowsNonNewton3() override { return true; }
+
+    /**
+     * @copydoc autopas::PairwiseFunctor::AoSFunctor()
+     */
+    void AoSFunctor(Particle_T &i, Particle_T &j, bool newton3) override {
+      using namespace autopas::utils::ArrayMath::literals;
+      if (i.isDummy() or j.isDummy()) return;
+      auto dist = i.getR() - j.getR();
```
SamNewcome:
```suggestion
      const auto dist = i.getR() - j.getR();
```

Also, my personal preference is for `displacement` or `disp` for the 3D vector.

src/autopas/containers/verletListsCellBased/verletLists/VerletListHelpers.h
```diff
@@ -0,0 +1,432 @@
...
+   * Pass-1 functor for the two-pass CRS build.
+   *
+   * For every interacting pair (i, j) it increments a per-particle neighbor counter. The counters are then
+   * prefix-summed to produce the CRS offsets before pass 2 runs.
+   *
+   * Thread safety: each counter is a std::atomic<size_t> padded to a full cache line (64 bytes) so that adjacent
+   * particles never share a cache line. This eliminates false-sharing stalls that would otherwise dominate on
+   * multi-socket / many-core systems.
+   */
+  class VerletListCounterFunctor : public PairwiseFunctor<Particle_T, VerletListCounterFunctor> {
```
SamNewcome, line 108: I would prefer if you handled newton3 similarly to `gatherNewton3Lists` and functors using `newton3` in `InteractionListGeneratorFunctor`: i.e. a class parameter (e.g. `countNewton3Lists`) for whether `_counts` is for N3 lists or not; independent of if the functor itself is being used with N3. This means that the functor gives the same result regardless of AoS/SoA, N3 on/off, container, etc...

If you don't want to do this, I would accept simply a comment that the functor produces different results whether N3 on or off is used.

src/autopas/containers/verletListsCellBased/verletLists/VerletListHelpers.h
```diff
@@ -0,0 +1,432 @@
...
+   * Pass-2 functor for the two-pass CRS build.
+   *
+   * The CRS offsets are already set. For each interacting pair (i, j) this functor writes j's index directly into the
+   * pre-allocated _neighborList.indices slice for particle i, advancing a per-particle fill cursor with a relaxed
+   * atomic fetch-add.
+   *
+   * The fill cursors reuse the same cache-line-padded PaddedAtomic array as the counter pass (reset to the CRS start
+   * offset before this pass runs).
+   */
+  class VerletListFillerFunctor : public PairwiseFunctor<Particle_T, VerletListFillerFunctor> {
```
SamNewcome, line 251: Same N3 on/off comment as above, plus the `const` and the potential rename of `dist`.

src/autopas/containers/verletListsCellBased/verletLists/VerletListHelpers.h
```diff
@@ -0,0 +1,432 @@
...
+  };
+
+  /**
+   * This functor checks the validity of neighborhood lists.
+   * If a pair of particles has a distance of less than the cutoff radius it
+   * checks whether the pair is represented in the CRS neighbor list.
+   * If the pair is not present in the list the neighborhood lists are invalid
+   * and neighborlistsAreValid() will return false.
+   */
+  class VerletListValidityCheckerFunctor : public PairwiseFunctor<Particle_T, VerletListValidityCheckerFunctor> {
```
SamNewcome: This never appears to get used. Correct me if I'm wrong

src/autopas/containers/verletListsCellBased/verletLists/VerletLists.h
```diff
@@ -91,141 +90,200 @@ class VerletLists : public VerletListsLinkedBase<Particle_T> {
...
   }
 
- protected:
   /**
-   * Update the verlet lists for AoS usage
-   * @param useNewton3
+   * Rebuilds _particleToIndex and _neighborList from scratch.
+   *
+   * Dispatches to the single-pass path when only one thread is available (lower overhead, no atomics) and to the
+   * two-pass lock-free path when multiple threads are active (eliminates false sharing and malloc contention).
```
SamNewcome: Maybe briefly explain the passes

src/autopas/containers/verletListsCellBased/verletLists/VerletLists.h
```diff
@@ -91,141 +90,200 @@ class VerletLists : public VerletListsLinkedBase<Particle_T> {
...
+      this->_linkedCells.computeInteractions(&traversal);
+    }
+  }
 
   /**
-   * Mapping of every particle, represented by its pointer, to an index.
-   * The index indexes all particles in the container.
+   * Mapping of every particle pointer to its dense SoA index.
+   * Built once per rebuild in buildParticleIndex() before the traversal.
+   * Shared with VerletListGeneratorFunctor during list construction and with the traversal for the AoS force
```
SamNewcome:
```suggestion
   * Shared with InteractionListGeneratorFunctor during list construction and with the traversal for the AoS force
```

src/autopas/containers/verletListsCellBased/verletLists/VerletLists.h
```diff
@@ -91,141 +90,200 @@ class VerletLists : public VerletListsLinkedBase<Particle_T> {
...
+    } else {
+      updateNeighborListsMultiThread(N, interactionLength, dataLayout, useNewton3);
     }
-
-    return numParticles;
   }
 
   /**
-   * Fills SoA neighbor list with particle indices.
+   * Single-threaded rebuild: One traversal with VerletListGeneratorFunctor writing into per-particle
```
SamNewcome:
```suggestion
   * Single-threaded rebuild: One traversal with InteractionListGeneratorFunctor writing into per-particle
```

SamNewcome (changes requested):
Nice. Only some minor points.

src/autopas/containers/verletListsCellBased/verletLists/traversals/VLListIterationTraversal.h
SamNewcome: I think this is no longer applicable

src/autopas/containers/verletListsCellBased/verletLists/traversals/VLListIterationTraversal.h
SamNewcome: I think this is no longer applicable

src/autopas/containers/verletListsCellBased/verletLists/VerletListHelpers.h
thesamriel: Changed to 'displacement'.

src/autopas/baseFunctors/InteractionListGeneratorFunctor.h
thesamriel: I changed the description in [12f0d95](https://github.com/AutoPas/AutoPas/pull/1162/commits/12f0d9527eb6873466fbc23b6e63d2d6ca13ed33)

applicationLibrary/neighborIdentification/tests/NeighborIdentificationFunctorTest.cpp
thesamriel: I added the template initialization to the library tests in [12f0d95](https://github.com/AutoPas/AutoPas/pull/1162/commits/12f0d9527eb6873466fbc23b6e63d2d6ca13ed33).
I don't want to move them back to the internal tests because they rely on the `NeighborIdentificationFunctor` instead of the internal `InteractionListGeneratorFunctor`.

src/autopas/containers/verletListsCellBased/verletLists/VerletListHelpers.h
thesamriel: Since this is a performance critical part of the rebuilding code, I would prefer to not add additional if-statements if not necessary.
Added a note instead in [56a04d1](https://github.com/AutoPas/AutoPas/pull/1162/commits/56a04d1c8654207cac498cd5bba32f6351c798a7).

src/autopas/containers/verletListsCellBased/verletLists/traversals/VLListIterationTraversal.h
thesamriel: I think it could be kept, but can only run single-threaded.
For now, I added a warning in [ed5da05](https://github.com/AutoPas/AutoPas/pull/1162/commits/ed5da053c246a9ffadedc9c0090bdb8f81da02da)

applicationLibrary/molecularDynamics/molecularDynamicsLibrary/LJFunctorHWY.h
thesamriel: Good idea! I refactored that part in [4a203be](https://github.com/AutoPas/AutoPas/pull/1162/commits/4a203be1a17162f90687605066bb08d8d0807911)

src/autopas/containers/verletListsCellBased/verletLists/traversals/VLListIterationTraversal.h
```diff
@@ -43,14 +50,14 @@ class VLListIterationTraversal : public TraversalInterface, public VLTraversalIn
...
-      std::vector<size_t> offsets(cells.size() + 1);
-      std::inclusive_scan(
-          cells.begin(), cells.end(), offsets.begin() + 1,
-          [](const size_t &partialSum, const auto &cell) { return partialSum + cell.size(); }, 0);
+    // First, resize the SoA to the required number of elements to store. This avoids resizing successively the SoA in
+    // SoALoader.
+    std::vector<size_t> offsets(cells.size() + 1);
+    std::inclusive_scan(
+        cells.begin(), cells.end(), offsets.begin() + 1,
+        [](const size_t &partialSum, const auto &cell) { return partialSum + cell.size(); }, 0);
```
SamNewcome: Any reason why this is no longer wrapped in the `if (_dataLayout == DataLayoutOption::soa)`?

thesamriel: No, that's a left over from the C27 part which we moved to another traversal. Change the scope back here

SamNewcome (approved):
Looks good!

PR #1163, opened 2026-07-28
Switch GoogleTest from ZIP file to git subtree

thesamriel:
# Description

This changes the GoogleTest dependency from a ZIP file to a git subtree.

It was added via:
`git subtree add --prefix libs/googletest https://github.com/google/googletest.git v1.17.0 --squash`

- Version bumped to the newest release version of gtest: v1.17.0 (from 1.15.2)
- Changed `Fetch_Content` to `add_subdirectory` in the CMake module

## Related Pull Requests

- #1149 

### Misc
Somehow, the PR still shows all new files from `libs/googletest`. I thought that our `.gitattributes` file would avoid this. Not sure why it doesn't.


SamNewcome (approved):
Nice. I went through the CMake module differences, this is definitely better than what I had in #1149.

> Somehow, the PR still shows all new files from libs/googletest. I thought that our .gitattributes file would avoid this. Not sure why it doesn't.

Yeah so when I looked at this in #1149, there was no way to actually hide this completely, the `.gitattributes` settings just makes the code not appear in repo statistics and that by default these vendored files are collapsed when viewing the files changed.

PR #1165, opened 2026-07-28
Switch spdlog from ZIP file to git subtree

thesamriel:
# Description

This changes the spdlog dependency from a ZIP file to a git subtree.

It was added via:
`git subtree add --prefix libs/spdlog https://github.com/gabime/spdlog.git v1.17.0 --squash`

- Changed `Fetch_Content` to `add_subdirectory` in the CMake module
- Updated the CMake file fixing an early return bug in case of using system installed spdlog

## Related Pull Requests

- #1149


SamNewcome (approved):
All good, just fix that merge conflict

PR #1169, opened 2026-07-30
Adding a ParticleType concept

SamNewcome:
# Description

Adds a particle type concept.

## Potentially awkward

I can't add the `markAsDeleted` function to the concept as it is private, and that `internal::markParticleAsDeleted` is a friend of the class. So this is left out of the concept.

# How Has This Been Tested?

- [x] CI


src/autopas/utils/ParticleConcept.h
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * @note Particles are deliberately not required to be default constructible, only copyable, to avoid imposing any
+ * requirement upon the constructor.
+ * @note Only the SoA attributes that every particle defines (ptr, posX, posY, posZ and ownershipState) are required
+ * here. Attributes such as id or the force components are not universal, e.g. sphLib::SPHParticle defines neither.
+ *
+ * @warning In addition to this concept, we require a markAsDeleted() function on the particle. As this is a private
+ * function, accessed only through a friend function, we cannot check it in this concept. If developing a custom
+ * particle derived from ParticleBase, this function will already be included, but a purely custom particle must
+ * implement it and befriend internal::markParticleAsDeleted():
```
manishmishra6016, line 33: This would only be useful if `markAsDeleted()` makes the particle a dummy; not sure how it will work if the user implements it differently(!). If I understand correctly, the main goal of the PR is to allow users to implement custom particles that do not derive from `ParticleBase`, is that correct? Is there a scenario where creating everything from scratch would be more beneficial? (Additionally, the user can always overload these functions)

I am generally fine with the changes; it doesn't conflict with how AutoPas works right now. But in my opinion, I prefer to restrict users to deriving from `ParticleBase`. And if the issue is that our base class has additional parameters (like force?), we can move these to derived classes and make our base class leaner.

manishmishra6016: ~~It also seems like the whole idea of using a friend class and making it private in the base class was that the user cannot call this function in the simulator (and instead should use `AutoPas::deleteParticle(iterator)`), but if the custom particle class is implemented directly with these, then the user can call these functions in the simulator code outside AutoPas (am I right?~~ edit: this should not happen if they are private. Although, I am a bit unclear about why it was implemented this way before. Just to avoid calling this in the derived particle class?).

src/autopas/utils/ParticleConcept.h
```diff
@@ -0,0 +1,106 @@
...
+
+  // Force.
+  { cp.getF() } -> std::same_as<const std::array<double, 3> &>;
+  { p.setF(vec) } -> std::same_as<void>;
+  { p.addF(vec) } -> std::same_as<void>;
+  { p.subF(vec) } -> std::same_as<void>;
+
+  // Id. Phrased so that the id type does not have to be named.
+  { cp.getID() } -> std::integral;
+  { p.setID(cp.getID()) } -> std::same_as<void>;
```
manishmishra6016, line 71: In the comments, we mention that these are not universal, so shouldn't we skip them in the concept?

tests/testAutopas/testingHelpers/NonConstructibleParticle.h
```diff
@@ -69,4 +69,32 @@ class NonConstructibleParticle : public autopas::ParticleBaseFP64 {
       autopas::utils::ExceptionHandler::exception("ParticleBase::get() unknown attribute {}", attribute);
     }
   }
+
+  /**
+   * Setter, which allows set an attribute using the corresponding attribute name (defined in AttributeNames).
```
manishmishra6016, line 74:
```suggestion
   * Setter, which allows setting an attribute using the corresponding attribute name (defined in AttributeNames).
```

tests/testAutopas/tests/particles/ParticleConceptTest.cpp
```diff
@@ -0,0 +1,39 @@
...
+              "NonConstructibleParticle does not satisfy the ParticleType concept!");
+
+namespace {
+/**
+ * A particle that hides one of the required members.
+ */
+struct ParticleWithoutGetF : public autopas::ParticleBaseFP64 {
+ private:
+  using autopas::ParticleBaseFP64::getF;
+};
```
manishmishra6016, line 27: If we decide to remove Force from the concept, this must change. Perhaps the cases without position or ownership would be better.

manishmishra6016 (commented):
The changes seem fine to me. I just had some thoughts about it. Let me know what you think.

src/autopas/utils/ParticleConcept.h
SamNewcome: > If I understand correctly, the main goal of the PR is to allow users to implement custom particles that do not derive from ParticleBase, is that correct?

The main goal is really that we get cleaner code development and clearer error messages, because it is clear what functions the particle class actually has. This is independent of custom particles/inherited particles and will hold for both from the perspective of us developing internal AutoPas logic, but I think it will also aid development in projects using custom particles. 


> Is there a scenario where creating everything from scratch would be more beneficial? (Additionally, the user can always overload these functions)

Yes. If AutoPas is used in an existing code-base, they would need to have their particle classes inherit from our particle base. This gets messy if they wish for AutoPas to be a togglable feature or if they already implement their own position, velocity, force and don't want duplicates or to rename their variables. E.g. in MercuryDPM, they spell out position as `position` instead of just `r`, they use a suffix `_` to indicate member variable and we use a prefix, they have custom `Vec3D` class and we use `std::array<double, 3>`, so the actual data itself is not compatible - but we don't need the data to be comptable, just the getters/setters. Requiring only the public interface functions means we can have a `getR()` function, that we only add if AutoPas is enabled, that use their underlying `Vec3D position_`.


> edit: this should not happen if they are private. Although, I am a bit unclear about why it was implemented this way before. Just to avoid calling this in the derived particle class?).

I think the idea is that we need to be able to mark a particle as deleted but we don't want users using this as it breaks AutoPas internal owned/halo/dummy statistics. We maybe could rework this so its not needed but it would need some rework.

manishmishra6016: > > If I understand correctly, the main goal of the PR is to allow users to implement custom particles that do not derive from ParticleBase, is that correct?
> 
> The main goal is really that we get cleaner code development and clearer error messages, because it is clear what functions the particle class actually has. This is independent of custom particles/inherited particles and will hold for both from the perspective of us developing internal AutoPas logic, but I think it will also aid development in projects using custom particles.

That is a good point.

> > Is there a scenario where creating everything from scratch would be more beneficial? (Additionally, the user can always overload these functions)
> 
> Yes. If AutoPas is used in an existing code-base, they would need to have their particle classes inherit from our particle base. This gets messy if they wish for AutoPas to be a togglable feature or if they already implement their own position, velocity, force and don't want duplicates or to rename their variables. E.g. in MercuryDPM, they spell out position as `position` instead of just `r`, they use a suffix `_` to indicate member variable and we use a prefix, they have custom `Vec3D` class and we use `std::array<double, 3>`, so the actual data itself is not compatible - but we don't need the data to be comptable, just the getters/setters. Requiring only the public interface functions means we can have a `getR()` function, that we only add if AutoPas is enabled, that use their underlying `Vec3D position_`.

Okay, fair point. They will still need to adjust function names, but that's less hassle than changing all internal variables!

> I think the idea is that we need to be able to mark a particle as deleted but we don't want users using this as it breaks AutoPas internal owned/halo/dummy statistics. We maybe could rework this so its not needed but it would need some rework.

I would add somewhere that if a custom particle class is used, the `markAsDeleted()` should still be implemented exactly the same as in the `ParticleBase()` (and not just any implementation).

src/autopas/utils/ParticleConcept.h
```diff
@@ -0,0 +1,106 @@
...
+
+    requires(Particle_T p, const Particle_T cp, const std::array<double, 3> &vec, double maxDistSquared,
+             std::ostream &os) {
+  // Types needed to build and address the SoAs.
+  typename Particle_T::SoAArraysType;
+  typename Particle_T::AttributeNames;
+  requires std::is_enum_v<typename Particle_T::AttributeNames>;
+
+  // Position.
+  { cp.getR() } -> std::same_as<const std::array<double, 3> &>;
```
dmartin13, line 52: Just one thing I was thinking about:

Is the strict `std::same_as<const std::array<double, 3> &>` requirement mainly for performance reasons, i.e. to guarantee that `getR()` returns a reference instead of potentially creating a temporary or conversion?

For custom particle types with a different internal vector representation, a less restrictive requirement would make adapting `getR()` easier (maybe `std::convertible_to`), but I assume that could introduce additional copies in a performance-critical path.

(same for velocity and force)

dmartin13 (changes requested):
Nice idea in this PR! I just had one thought, see my comment.

PR #1171, opened 2026-07-31
Switch Eigen from ZIP file to git subtree (and upgrade to 5.0.1)

thesamriel:
# Description

This changes the Eigen dependency from a ZIP file to a git subtree.

It was added via:
`git subtree add --prefix libs/eigen https://gitlab.com/libeigen/eigen.git 3.4.0 --squash`
and then updated to version 5.0.1

- Removed `Fetch_Content` in the CMake module
- eigen 5.0.1 is actually eigen 3.5.0.1, they changed their versioning

## Related Pull Requests

- #1149


SamNewcome (approved):
> eigen 5.0.1 is actually eigen 3.5.0.1, they changed their versioning

This makes a lot of sense - I was getting kinda confused why everybody was on v5, I didn't think we were so far behind.

Looks good, thank you!

PR #1179, opened 2026-09-13
Update md-flexible particle objects initialization

### LINKED ISSUE #977: Don't add padding to md-flexible objects
In md-flexible, we have multiple types particle objects to create e.g. cubes of equally spaced particles.
`CubeUniform` and `CubeClosestPacked` specify a box-length. To prevent issues with periodic boundaries, the actual domain length is padded around these created objects. This leads to the actual domain being larger than the specified box-length of the cube objects. 
This regularly leads to confusion when setting up scenarios with a specific density.

## Solution ideas
- As a minimum, there should be a warning, when the actual domain size is being increased due to padding.
- This problem is arising often when trying to create initial states of a specific density. A cube generator for a specified density would be a great feature.

thesamriel:
# Description

This PR updates some of md-flexible's particle object generators.

- `CubeUniform`, `CubeGrid` and `CubeClosestPacked` can be created with a given `particle-density` instead of num-particles, spacing or spacing respectively.
- `CubeGrid` and `CubeClosestPacked` have an additional parameter `centered`, which determines whether the first particle is place at the cube origin (`centered = false`) or at the center of the first unit cell. The default behavior has been changed to set this to true. This avoids the need to add padding around cubes for periodic boundaries (see #977). 
- The simulation box padding was removed and users are encouraged to simply use grid objects and keep `centered = true`.
- `CubeClosestPacked` has been updated to support generating grids as either an fcc or hcp lattice. fcc lattice is more deterministic in the number of particles that will fit in the domain, because the unit cell has the same orientation as the object cube. (See on [wikipedia](https://en.wikipedia.org/wiki/Close-packing_of_equal_spheres))

## Resolved Issues

- [x] fixes #977

# How Has This Been Tested?

- [x] Added new md-flexible tests
- [x] Ran old example simulations
  - Falling Drop: Now uses a FCC lattice and slightly different box dimensions than before
  - Exploding Liquid: Also uses a FCC lattice now and therefore particles are at different starting positions than before
  - Exploding Liquid Big: Moved the box to start at [0, 0, 0] as the bottom left corner.
  - Other Scenarios are kept the same. (Achieved by using HCP lattice and non-centering alignment)

## Summary by CodeRabbit

* **New Features**
  * Added particle-density configuration for grid, uniform, and closest-packed objects.
  * Added FCC and HCP lattice selection and centered-alignment options.
  * Added command-line support for particle density and closest-packing structure.
  * Configuration output now reports density, lattice structure, and alignment.
* **Bug Fixes**
  * Improved simulation-box calculation while preserving explicitly configured bounds.
  * Added validation for conflicting or missing spacing and density settings, plus invalid lattice values.
* **Tests**
  * Expanded coverage for density-based generation, lattice options, parsing, and box-bound preservation.


examples/md-flexible/src/configuration/objects/CubeClosestPacked.h
```diff
@@ -10,62 +10,90 @@
...
-        _bottomLeftCorner(bottomLeftCorner),
-        _topRightCorner(autopas::utils::ArrayMath::add(bottomLeftCorner, boxLength)),
-        _xOffset(particleSpacing * 1. / 2.),
-        _yOffset(particleSpacing * sqrt(1. / 12.)) {}
+  CubeClosestPacked(const std::array<double, 3> &velocity, const size_t typeId, const double particleSpacing,
+                    const std::array<double, 3> &boxLength, const std::array<double, 3> &bottomLeftCorner,
+                    const LatticeStructure structure = FCC, const bool centered = true)
+      : CubeClosestPacked(velocity, typeId, boxLength, bottomLeftCorner, structure, centered) {
+    _particleSpacing = particleSpacing;
+    _density =
```
manishmishra6016: Maybe using `_particleDensity` or something different would be better? The `density` of the simulation would also include the factor `mass`, which may differ between species.

examples/md-flexible/src/configuration/objects/CubeClosestPacked.h
```diff
@@ -10,62 +10,90 @@
...
-    const double numLayers = std::ceil(_boxLength[2] / spacingLayer);
-    // Add up all even and odd layers.
-    return evenLayer * std::ceil(numLayers / 2.) + oddLayer * std::floor(numLayers / 2.);
+    if (_structure == FCC) {
+      return autopasTools::generators::FCCGenerator::getNumberOfParticles(_bottomLeftCorner, _topRightCorner,
+                                                                          _particleSpacing, _centered);
+    }
+
+    return autopasTools::generators::HCPGenerator::getNumberOfParticles(_bottomLeftCorner, _topRightCorner,
+                                                                        _particleSpacing, _centered);
```
SamNewcome, line 96: Do this as an if, else-if, else error.

examples/md-flexible/src/configuration/objects/CubeClosestPacked.h
```diff
@@ -103,15 +137,39 @@ class CubeClosestPacked : public Object {
     // dummy particle used as a template with id of the first newly generated one
     const ParticleType dummyParticle = getDummyParticle(particles.size());
 
-    autopasTools::generators::ClosestPackingGenerator::fillWithParticles(
-        particlesWrapper, _bottomLeftCorner, _topRightCorner, dummyParticle, _particleSpacing);
+    if (_structure == FCC) {
+      autopasTools::generators::FCCGenerator::fillWithParticles(particlesWrapper, _bottomLeftCorner, _topRightCorner,
+                                                                dummyParticle, _particleSpacing, _centered);
+    } else {
```
SamNewcome, line 143: Similar

examples/md-flexible/src/configuration/objects/CubeGrid.h
```diff
@@ -66,17 +89,13 @@ class CubeGrid : public Object {
     using namespace autopas::utils::ArrayMath::literals;
 
     const auto particlesPerDimDouble = autopas::utils::ArrayUtils::static_cast_copy_array<double>(_particlesPerDim);
-    // subtract one because the first particle is at bottomLeftCorner
-    const auto particlesPerDimSubOne = particlesPerDimDouble - 1.;
-    const auto lastParticleRelative = particlesPerDimSubOne * _particleSpacing;
-    auto lastParticleAbsolute = _bottomLeftCorner + lastParticleRelative;
-
-    return lastParticleAbsolute;
+    const auto totalLengthRelative = particlesPerDimDouble * _particleSpacing;
```
SamNewcome, line 92: I'm not totally sure why we dropped the subtract 1

examples/md-flexible/src/configuration/MDFlexConfig.cpp
```diff
@@ -464,22 +461,35 @@ std::string MDFlexConfig::to_string() const {
 
 void MDFlexConfig::calcSimulationBox() {
   const double interactionLength = cutoff.value + verletSkinRadius.value;
-  const auto preBoxMin = boxMin.value;
-  const auto preBoxMax = boxMax.value;
 
-  // helper function so that we can do the same for every object collection
-  // resizes the domain to the maximal extents of all objects
+  const bool userDefinedBox =
+      (boxMax.value[0] > boxMin.value[0] and boxMax.value[1] > boxMin.value[1] and boxMax.value[2] > boxMin.value[2]);
```
SamNewcome, line 466: Not too keen on this, as an incorrect user defined box (e.g. with mins greater than maxes) gets treated as an undefined user box. My suggestion is to only assume an undefined box if all these values really are all a nonsensical default value, and I would probably use NaN instead of max and lowest.

examples/md-flexible/src/configuration/MDFlexConfig.h
```diff
@@ -630,6 +637,22 @@ class MDFlexConfig {
    */
   MDFlexOption<double, __LINE__> particleSpacing{1.1225 * 1, "particle-spacing", true,
                                                  "Space between two particles for the grid generator."};
+  /**
+   * particleDensity
+   */
+  MDFlexOption<double, __LINE__> particleDensity{
+      0.0, "particle-density", true, "Density of particles. This is the number of particles per unit volume."};
```
SamNewcome, line 644: Add a brief comment about either spacing or density in the description.

examples/md-flexible/tests/configuration/GeneratorsTest.cpp
SamNewcome, line 1: Add a brief description to all the tests you added.

examples/md-flexible/tests/configuration/MDFlexConfigTest.cpp
```diff
@@ -225,4 +225,20 @@ TEST_F(MDFlexConfigTest, correctMolParsing) {
   for (int i = 0; i < 3; i++) {
     EXPECT_EQ(expectedMoI1[i], momentOfInertia1[i]);
   }
+}
+
+TEST_F(MDFlexConfigTest, UserSpecifiedBoxMinMaxPreserved) {
```
SamNewcome, line 230: Add a brief description.

examples/md-flexible/tests/configuration/YamlParserTest.cpp
```diff
@@ -67,4 +68,311 @@ TEST_F(YamlParserTest, loadBalancerAllParsedAsALL) {
...
+    obj.generate(particles);
+    ASSERT_GT(particles.size(), 0);
+    const double a = std::sqrt(2.0) * 1.0;
+    EXPECT_NEAR(particles[0].getR()[0], 1.0 + a / 4.0, 1e-10);
+    EXPECT_NEAR(particles[0].getR()[1], 2.0 + a / 4.0, 1e-10);
+    EXPECT_NEAR(particles[0].getR()[2], 3.0 + a / 4.0, 1e-10);
+  }
+}
+
+TEST_F(YamlParserTest, parseCubeGridAlignment) {
```
SamNewcome, line 336: Sim.

tools/generators/src/FCCGenerator.h
```diff
@@ -0,0 +1,149 @@
...
+inline size_t getNumberOfParticles(const std::array<double, 3> &boxMin, const std::array<double, 3> &boxMax,
+                                   const double spacing = 1.0, const bool centeredAlignment = true) {
+  if (spacing <= 0.0) {
+    return 0;
+  }
+  for (size_t d = 0; d < 3; ++d) {
+    if (boxMax[d] <= boxMin[d]) {
+      return 0;
+    }
+  }
```
SamNewcome, line 46: Maybe errors/warnings are better

tools/generators/src/FCCGenerator.h
```diff
@@ -0,0 +1,149 @@
...
+
+  const double a = std::sqrt(2.0) * spacing;
```
SamNewcome, line 48: Is `a` a common notation used in FCC literature? Otherwise use descriptive name.

tools/generators/src/FCCGenerator.h
```diff
@@ -0,0 +1,149 @@
...
+                           typename autopas::utils::ParticleTypeTrait<Container>::value(),
+                       const double spacing = 1.0, const bool centeredAlignment = true) {
+  if (spacing <= 0.0) {
+    return;
+  }
+  for (size_t d = 0; d < 3; ++d) {
+    if (boxMax[d] <= boxMin[d]) {
+      return;
+    }
+  }
```
SamNewcome, line 108: Similar

tools/generators/src/FCCGenerator.h
```diff
@@ -0,0 +1,149 @@
...
+
+  const double a = std::sqrt(2.0) * spacing;
```
SamNewcome, line 110: Similar

tools/generators/src/FCCGenerator.h
```diff
@@ -0,0 +1,149 @@
...
+            boxMin[2] + static_cast<double>(z) * a + offset[2],
+        };
+        for (const auto &b : fccBasis) {
+          const std::array<double, 3> pos = {
+              cellOrigin[0] + b[0] * a,
+              cellOrigin[1] + b[1] * a,
+              cellOrigin[2] + b[2] * a,
+          };
+          if (pos[0] >= boxMin[0] and pos[0] < boxMax[0] - offset[0] and pos[1] >= boxMin[1] and
+              pos[1] < boxMax[1] - offset[1] and pos[2] >= boxMin[2] and pos[2] < boxMax[2] - offset[2]) {
```
SamNewcome, line 75: How immune is this from FP errors? Maybe worth a brief comment explaining why this isn't an issue, or correcting if it is.

tools/generators/src/FCCGenerator.h
```diff
@@ -0,0 +1,149 @@
...
+            boxMin[2] + static_cast<double>(z) * a + offset[2],
+        };
+        for (const auto &basis : fccBasis) {
+          const std::array<double, 3> pos = {
+              cellOrigin[0] + basis[0] * a,
+              cellOrigin[1] + basis[1] * a,
+              cellOrigin[2] + basis[2] * a,
+          };
+          if (pos[0] >= boxMin[0] and pos[0] < boxMax[0] - offset[0] and pos[1] >= boxMin[1] and
+              pos[1] < boxMax[1] - offset[1] and pos[2] >= boxMin[2] and pos[2] < boxMax[2] - offset[2]) {
```
SamNewcome, line 137: Similar

tools/generators/src/HCPGenerator.h
```diff
@@ -0,0 +1,125 @@
...
+inline size_t getNumberOfParticles(const std::array<double, 3> &boxMin, const std::array<double, 3> &boxMax,
+                                   const double spacing = 1.0, const bool centeredAlignment = true) {
+  if (spacing <= 0.0) {
+    return 0;
+  }
+  for (size_t d = 0; d < 3; ++d) {
+    if (boxMax[d] <= boxMin[d]) {
+      return 0;
+    }
+  }
```
SamNewcome, line 36: Similar

tools/generators/src/HCPGenerator.h
```diff
@@ -0,0 +1,125 @@
...
+                           typename autopas::utils::ParticleTypeTrait<Container>::value(),
+                       const double spacing = 1., const bool centeredAlignment = true) {
+  if (spacing <= 0.0) {
+    return;
+  }
+  for (size_t d = 0; d < 3; ++d) {
+    if (boxMax[d] <= boxMin[d]) {
+      return;
+    }
+  }
```
SamNewcome, line 88: Similar

SamNewcome (changes requested):
Looks really nice and adds some really useful features and will stop my little python sessions where I work out what particle spacing and number and offset from boundary I need to achieve certain densities ;) 


Some points about `centered`:

- Does centering also squeeze in the top right corner? If so, add this to the relevant places in the documentation. In the yaml examples it sounds a bit like it just offsets the bottom right, but not sure this is the case.
- Does this change the particle spacing, where this is provided.
- How does this work with density? Do we get the specified density over the box, just that it begins the simulation squashed in slightly?

About `particle-density`:

- Really useful!

About `structure`:

- I would like to try to make md-flex input files slightly backwards compatible, although I care less about this than the actual AutoPas interface. I think you have set it up such that `CubeClosestPacked` has a default where there is no `structure` entry (if not, I would like this), but I think that this defaults to whichever of FCC and not the previous HCP. I would like the default option to remain the same as before. If there is a case that FCC is better, I am happy for you to add a warning message e.g. if no structure is provided, you say that the existing default HCP structure will change to FCC by September 2027, or something like this.

--------

When you say

> Ran old example simulations

Did you check that the starting position of particles match? There are some input files where this will definitely not be the case -> that's totally fine but I would like to know that the input files match where they did before in the cases where they should.

Happy to take your word for it but would like a little discussion on which input files match what they did before and that where they don't match, that this is intentional.

tools/generators/src/FCCGenerator.h
thesamriel: Yes, but changed it to `latticeConstant` anyways.

tools/generators/src/FCCGenerator.h
thesamriel: This is a bit messy with how the yaml parser handles errors. 
I would keep the spacing check here to return 0 as number of particles (which is not wrong).
The actual error handling is happening in the yaml parser though, as it has been for other objects as well.

SamNewcome: Sounds okay!

examples/md-flexible/src/configuration/MDFlexConfig.cpp
thesamriel: I updated this method and initialize with NaN as suggested. Done in [caa49d8](https://github.com/AutoPas/AutoPas/pull/1179/commits/caa49d872aa0b0a60e7e37ad4a2e0b046b3817d5)

examples/md-flexible/src/configuration/objects/CubeGrid.h
thesamriel: When using centered no subtraction is needed, but when not centered, it's still needed. Fixed now.

tools/generators/src/FCCGenerator.h
thesamriel: I removed the ` - offset[d]`. I believe that's not needed actually and then there should not be any rounding dangers

thesamriel:
> * Does centering also squeeze in the top right corner? If so, add this to the relevant places in the documentation. In the yaml examples it sounds a bit like it just offsets the bottom right, but not sure this is the case.

=> I added more information in different places of the documentation. Essentially also 'squeezes' in the top right corner, yes.


> * Does this change the particle spacing, where this is provided.

=> No, particle spacing is only adjusted when initializing with density.


> * How does this work with density? Do we get the specified density over the box, just that it begins the simulation squashed in slightly?

=> Essentially yes. But I would think different about it. The padding we added around objects to work for boundaries is now part of the object.


> * I would like to try to make md-flex input files slightly backwards compatible, although I care less about this than the actual AutoPas interface. I think you have set it up such that `CubeClosestPacked` has a default where there is no `structure` entry (if not, I would like this), but I think that this defaults to whichever of FCC and not the previous HCP. I would like the default option to remain the same as before. If there is a case that FCC is better, I am happy for you to add a warning message e.g. if no structure is provided, you say that the existing default HCP structure will change to FCC by September 2027, or something like this.

=> Personally, I prefer FCC most of the times. It's easier to set up, looks better and leads to more predictable number of particles. But I don't mind adding it explicitly in the future. Reverted the default back to HCP.
Old yaml files might still not work out of the box though, because the domain is no longer implicitly increased.

> Did you check that the starting position of particles match? There are some input files where this will definitely not be the case -> that's totally fine but I would like to know that the input files match where they did before in the cases where they should.

=> Added this also to the PR description. Essentially, "Falling Drop" and "Exploding Liquid" are slightly different now. All other scenarios are kept as before. (Old behavior usually is achieved by using HCP with `centered=false`)

examples/md-flexible/tests/configuration/GeneratorsTest.cpp
```diff
@@ -104,3 +112,371 @@ TEST_F(GeneratorsTest, MultipleObjectGeneration) {
...
+  const CubeClosestPacked cube(velocity, typeId, boxLength, bottomLeft, density,
+                               CubeClosestPacked::LatticeStructure::HCP);
+  EXPECT_DOUBLE_EQ(cube.getParticleDensity(), static_cast<double>(cube.getParticlesTotal()) / (4.0 * 4.0 * 4.0));
+  EXPECT_DOUBLE_EQ(cube.getParticleSpacing(), std::cbrt(std::sqrt(2.0) / density));
+
+  std::vector<ParticleType> particles;
+  cube.generate(particles);
+  EXPECT_EQ(particles.size(), cube.getParticlesTotal());
+  EXPECT_GT(particles.size(), 0);
+}
```
SamNewcome, line 533: Why is there no comparison of `cube.getParticleDensity` with `density` like for FCC?

Btw, I'm fine if these tests to not produce exact equalities, that's always going to be difficult. But maybe tests that `cube.getParticleDensity` is near to `density` for a couple different densities and box sizes might be good. (Same point applies to FCC and I guess CubeUniform and similar)

SamNewcome (changes requested):
One minor point plus what we discussed about a notice for changing to FCC in v3.

examples/md-flexible/tests/configuration/GeneratorsTest.cpp
thesamriel: While trying to do so, I noticed that the density was further off than I wanted in some cases.
Reworked the `CubeClosestPacked` constructor with density to find a better particle spacing and end up with a density closer to the desired one.
Also refactored and extended the tests on this.
Done in [b14feb4](https://github.com/AutoPas/AutoPas/pull/1179/commits/b14feb4bcfd7e0d145d2fc9988557fb56683b229)

thesamriel:
> One minor point plus what we discussed about a notice for changing to FCC in v3.

I added a warning to the yaml parser. Also added a warning and reverted the default behavior of the `centered` lattice alignment to be false. Aka. the first particle is placed directly on the given box origin, as it was before.

src/autopas/utils/generators/HCPGenerator.h
SamNewcome, line 27: Could you add a little regression test somewhere that this does equal the original? Its a little hard for me to check that the logic matches. A simple test which compares this to the original methods for a couple of values would suffice. If already covered somewhere, lmk.

SamNewcome (changes requested):

src/autopas/utils/generators/HCPGenerator.h
thesamriel: I added a test in [a9fa76a](https://github.com/AutoPas/AutoPas/pull/1179/commits/a9fa76a60005cbb203a4ac5ea8d31d6433311fc8). Hope that is what you had in mind.

